PR #3, opened 2016-11-18
Support for flame graph building

incubos:
Now you can dynamically attach `async-profiler` and pass it the following options:
1. `frameBufferSize:<size>` to set the size of the frame buffer
2. `start` to start profiling
3. `stop` to stop profiling
4. `dumpRawTraces:<file>` to dump collected stack traces into `<file>` in `lightweight-java-profiler` format

Then you can process the stack traces and build flame graphs:
```sh
echo "Collapsing stacks into ${STACKS}..."
cat $RAW | ${FLAMEGRAPH_DIR}/stackcollapse-ljp.awk >$STACKS
echo "Collapsing recursion into ${NON_RECURSIVE}..."
cat $STACKS | sort | ${FLAMEGRAPH_DIR}/stackcollapse-recursive.pl >$NON_RECURSIVE
echo "Building SVG into ${SVG}..."
cat $NON_RECURSIVE | ${FLAMEGRAPH_DIR}/flamegraph.pl >$SVG
```

Changes:
  * `Agent_OnAttach` resolves `AsyncGetCallTrace` each time to support dynamic attach
  * Dumping collected traces in `lightweight-java-profiler` format to be able to build flame graphs
  * For the history (to be able to revert to legacy behavior):
    * Properly stripping frames if the trace is too long
    * Dynamically tuneable stack trace depth
  * Replaced preallocated frames with shared frame buffer
  * Ignoring `bci` when storing frames

### DIFF
--- a/src/asyncProfiler.cpp
+++ b/src/asyncProfiler.cpp
@@ -25,6 +25,7 @@
 #include "vmEntry.h"
 
 ASGCTType asgct;
+int maxFrames = DEFAULT_MAX_FRAMES;
 Profiler Profiler::_instance;
 
 static void sigprofHandler(int signo, siginfo_t* siginfo, void* ucontext) {
@@ -56,7 +57,7 @@ MethodName::~MethodName() {
 
 void CallTraceSample::assign(ASGCT_CallTrace* trace) {
     _call_count = 1;
-    _num_frames = MIN(trace->num_frames, MAX_FRAMES);
+    _num_frames = MIN(trace->num_frames, maxFrames);
     
     const int offset = trace->num_frames - _num_frames;    
     for (int i = 0; i < _num_frames; i++) {
@@ -221,7 +222,13 @@ void Profiler::start(long interval) {
     _calls_safepoint = 0;
     _calls_unknown = 0;
     memset(_hashes, 0, sizeof(_hashes));
+    for (int i = 0; i < MAX_CALLTRACES; i++) {
+        free(_traces[i]._frames);
+        _traces[i]._frames = NULL;
+    }
     memset(_traces, 0, sizeof(_traces));
+    for (int i = 0; i < MAX_CALLTRACES; i++)
+        _traces[i]._frames = (ASGCT_CallFrame *) malloc(maxFrames * sizeof(ASGCT_CallFrame));
     memset(_methods, 0, sizeof(_methods));
 
     setTimer(interval / 1000, (interval % 1000) * 1000);
--- a/src/asyncProfiler.cpp
+++ b/src/asyncProfiler.cpp
@@ -25,7 +25,6 @@
 #include "vmEntry.h"
 
 ASGCTType asgct;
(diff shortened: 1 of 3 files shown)


karianna:
LGTM

src/vmEntry.cpp
```diff
@@ -81,20 +83,52 @@ Agent_OnLoad(JavaVM* vm, char* options, void* reserved) {
...
+                std::cerr << "Couldn't open: " << fileName << std::endl;
+                return -1;
+            }
+
+            std::cout << "Dumping raw traces to " << fileName << std::endl;
+            Profiler::_instance.dumpRawTraces(dump);
+            dump.close();
+        }
+        
+        token = strtok(NULL, OPTION_DELIMITER);
```
apangin: strtok modifies source string inplace, which is not good

src/vmEntry.cpp
```diff
@@ -81,20 +83,52 @@ Agent_OnLoad(JavaVM* vm, char* options, void* reserved) {
...
+const char OPTION_DELIMITER[] = ",";
+const char FRAME_BUFFER_SIZE[] = "frameBufferSize:";
+const char START[] = "start";
+const char STOP[] = "stop";
+const char DUMP_RAW_TRACES[] = "dumpRawTraces:";
+
 extern "C" JNIEXPORT jint JNICALL
 Agent_OnAttach(JavaVM* vm, char* options, void* reserved) {
     VM::attach(vm);
+    asgct = getJvmFunction<ASGCTType>("AsyncGetCallTrace");
```
apangin: Can be moved to init()

src/asyncProfiler.cpp
```diff
@@ -98,9 +97,24 @@ void Profiler::storeCallTrace(ASGCT_CallTrace* trace) {
         }
         if (++i == MAX_CALLTRACES) i = 0;
     } while (i != bucket);
+    
+    if (_frameBufferSize - _freeFrame < trace->num_frames) {
+        // Don't have enough space to store call trace
+        _frameBufferOverflow = true;
+        return;
+    }
```
apangin: Check it before the loop

incubos: BTW, we might not be able to increment the counter for a huge trace, which is **already** stored, but there is not enough space for another instance of this trace.

incubos:
Fixed.

src/asyncProfiler.cpp
apangin: Makes sense. Sorry for wrong suggestion then.

kkydesnikk:
How set frameBufferSize and dumpRawTraces
Can I see example

incubos:
You can use https://github.com/apangin/jattach:

```sh
jattach $PID load $ASYNC_PROFILER_SO true interval:$INTERVAL,frameBufferSize:$SIZE,start
jattach $PID load $ASYNC_PROFILER_SO true dumpRawTraces:$PATH
```

PR #4, opened 2016-11-23
Support for custom profiling frequency through agent option

incubos:
A user can pass `freq:<freq>,start` options to the agent to start profiling with custom frequency.

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -86,6 +86,7 @@ Agent_OnLoad(JavaVM* vm, char* options, void* reserved) {
 
 const char OPTION_DELIMITER[] = ",";
 const char FRAME_BUFFER_SIZE[] = "frameBufferSize:";
+const char FREQ[] = "freq:";
 const char START[] = "start";
 const char STOP[] = "stop";
 const char DUMP_RAW_TRACES[] = "dumpRawTraces:";
@@ -101,6 +102,8 @@ Agent_OnAttach(JavaVM* vm, char* options, void* reserved) {
     }
     strncpy(args, options, sizeof(args));
     
+    int freq = 1000 / DEFAULT_INTERVAL;
+    
     char *token = strtok(args, OPTION_DELIMITER);
     while (token) {
         if (strncmp(token, FRAME_BUFFER_SIZE, strlen(FRAME_BUFFER_SIZE)) == 0) {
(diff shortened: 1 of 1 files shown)


karianna:
I think this needs to be clearer to the user what unit of frequency is being applied here.

src/vmEntry.cpp
```diff
@@ -101,22 +102,47 @@ Agent_OnAttach(JavaVM* vm, char* options, void* reserved) {
     }
     strncpy(args, options, sizeof(args));
     
+    int freq = 1000 / DEFAULT_INTERVAL;
```
karianna: 1000 is a magic number

src/vmEntry.cpp
```diff
@@ -101,22 +102,47 @@ Agent_OnAttach(JavaVM* vm, char* options, void* reserved) {
...
         if (strncmp(token, FRAME_BUFFER_SIZE, strlen(FRAME_BUFFER_SIZE)) == 0) {
             const char *text = token + strlen(FRAME_BUFFER_SIZE);
             const int value = atoi(text);
             std::cout << "Setting frame buffer size to " << value << std::endl;
             Profiler::_instance.frameBufferSize(value);
+        } else if (strncmp(token, FREQ, strlen(FREQ)) == 0) {
+            const char *text = token + strlen(FREQ);
+            const int value = atoi(text);
+            if (value <= 0 || value > 1000) {
+                std::cerr << "Frequency must be in (0, 1000]: " << value << std::endl;
```
karianna: Is the (0, 1000] format correct?

src/vmEntry.cpp
```diff
@@ -101,22 +102,47 @@ Agent_OnAttach(JavaVM* vm, char* options, void* reserved) {
...
-            std::cout << "Profiling started" << std::endl;
-            Profiler::_instance.start(DEFAULT_INTERVAL);
+            if (Profiler::_instance.is_running()) {
+                std::cerr << "Profiler is already running" << std::endl;
+                return -1;
+            }
+            
+            std::cout << "Profiling started with frequency " << freq << "Hz" << std::endl;
+            Profiler::_instance.start(1000 / freq);
         } else if (strcmp(token, STOP) == 0) {
```
karianna: 1000 is a magic number

karianna:
Also feels like the logic of whether the profiler can or cannot execute should be extracted into it's own little state machine / grouping, will defer to better minds than mine though :-)

incubos:
A user can pass `interval:<ms>,start` options to the agent to start profiling with custom frequency.
Intervals are more consistent with preexisting Java API. As a result magic constants disappeared.
Error reporting is fixed.
Profiler state check logic is quite tricky. At the moment `Profiler` silently ignores duplicate `start`/`stop` commands. We don't want to lie to users by reporting ineffective intervals. So `start` check is left.
Thank you for your comments!

PR #6, opened 2016-11-25
Initialise JVMTI once and forever

incubos:
Fixed allocating multiple `jvmtiEnv` per each `On_Attach()` call.


apangin:
Merged manually

PR #10, opened 2016-12-02
Obligatory profiling deadline

incubos:
* Profiling duration is obligatory
  * Default value is 1h
  * We use `time()`, because seconds resolution is enough

### DIFF
--- a/src/asyncProfiler.cpp
+++ b/src/asyncProfiler.cpp
@@ -150,6 +150,16 @@ void Profiler::storeMethod(jmethodID method) {
 }
 
 void Profiler::recordSample(void* ucontext) {
+    // Check deadline
+    if (time(NULL) > _deadline) {
+        char error[] = "Disabling profiler due to deadline\n";
+        write(STDERR_FILENO, error, sizeof(error));
+
+        _running = false;
+
+        setTimer(0, 0);
+    }
+    
     _calls_total++;
 
     JNIEnv* jni = VM::jni();
@@ -209,7 +219,10 @@ void Profiler::setTimer(long sec, long usec) {
         perror("couldn't start timer");
 }
 
-void Profiler::start(long interval) {
+void Profiler::start(int interval, int duration) {
+    if (interval <= 0) return;
+    if (duration <= 0) return;
+    
     if (_running) return;
     _running = true;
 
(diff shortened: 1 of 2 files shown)


src/asyncProfiler.cpp
```diff
@@ -150,6 +150,16 @@ void Profiler::storeMethod(jmethodID method) {
 }
 
 void Profiler::recordSample(void* ucontext) {
+    // Check deadline
+    if (time(NULL) > _deadline) {
+        char error[] = "Disabling profiler due to deadline\n";
+        write(STDERR_FILENO, error, sizeof(error));
```
apangin: This will issue a warning #8

src/asyncProfiler.cpp
```diff
@@ -150,6 +150,16 @@ void Profiler::storeMethod(jmethodID method) {
...
+
+        _running = false;
+
+        setTimer(0, 0);
```
apangin: Are you sure calling setitimer inside signal handler is safe?

src/asyncProfiler.cpp
incubos: What can I do with the return value in this case? It is an error reporting in signal handler.
BTW, no warning in case of Mac OS:
```
$ make
mkdir -p build
g++ -O2 -I/Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/include -I/Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/include/darwin -fPIC -shared -o build/libasyncProfiler.so src/*.cpp
```
Or Linux:
```
$ make
mkdir -p build
g++ -O2 -I/usr/java/jdk180_64_60/include -I/usr/java/jdk180_64_60/include/linux -fPIC -shared -o build/libasyncProfiler.so src/*.cpp
```

src/asyncProfiler.cpp
incubos: It seems to be true, because `alarm()` [can be safely called inside a signal handler](http://man7.org/linux/man-pages/man7/signal.7.html) and [it is implemented](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/alarm.c;h=f7d97f1f330d3c698e27514616826a149b666315;hb=bd2260a2069993cdb14cffcbb2b0dc04b9888aee) using `setitimer()`.

apangin:
Merged manually

Issue #1, opened 2016-06-02
Update ReadMe.md with Example about "How to use this profile on remote JVM"


DmitriyLuckyman:
Steps I follow:
1) download
2) compile
3) Write Wrap class which call throw JNI start and stop profiler method.
4) Run java application
5) invoke wrap class methods
6) See in output stream summary of profilling

How can I do this for remote jvm?

apangin:
You can dynamically inject agent library into the running JVM using HotSpot Attach API.
For instance, you can do this with my [jattach](https://github.com/apangin/jattach) utility:

`jattach PID load /path/to/libasyncProfiler.so true start`
`jattach PID load /path/to/libasyncProfiler.so true stop`

leventov:
@apangin I've followed this and it just prints
```
Connected to remote JVM
Response Code = -1
```

apangin:
This means the command was recognized, but the library could not be loaded.
Make sure you specify the **absolute** path to the library, and the file has proper permissions.

tartemov:
I'm trying to attach and dump summary info:
`
$ ./jattach $PID load /tmp/libasyncProfiler.so true start 1 10
Connected to remote JVM
Response code = 0
0
$ ./jattach $PID load /tmp/libasyncProfiler.so true dumpRawTraces
Connected to remote JVM
Response code = 0
0
$ ./jattach $PID load /tmp/libasyncProfiler.so true stop
Connected to remote JVM
Response code = 0
0
$ ./jattach $PID load /tmp/libasyncProfiler.so true dumpRawTraces
Connected to remote JVM
Response code = 0
0
`
but I can't get any methods info. What am I doing wrong?

apangin:
Note that profiling stats are printed on `stdout` of Java process (not on the console of jattach).

apangin:
Alternatively you may output raw traces in the specified file:

`./jattach $PID load /tmp/libasyncProfiler.so true dumpRawTraces:/path/to/profile.txt`

tartemov:
Thanks. With dumpRawTraces method it's working fine. But with other methods - summary, dumpMethods and others it doesn't work - no file created.

tartemov:
Hi again! Could you please get any advice how to get summary/dumpMethods info in file?

apangin:
This is not supported right now, but will get fixed soon.

balmaster:
@tartemov you can use Brendan Gregg scripts to fold stacks and make svg

[...]

PR #13, opened 2017-04-08
Loader support and initial documentation

goldshtn:
This patchset introduces a Java loader tool that helps inject the agent into
a running Java process, and send the start, stop, and dump commands to it.
Additionally, initial documentation is introduced to the README, which
explains how to build and use the tool.

Finally, some minor issues are fixed: proper handling of sampling interval
(in cycles), proper handling of trace duration (in seconds), and printing
stack and method summaries in ascending order for user convenience.


README.md
```diff
@@ -1,4 +1,151 @@
...
+post-process it. The following example uses Brendan Gregg's
+[FlameGraph](https://github.com/BrendanGregg/FlameGraph) scripts, but a similar
+solution can be tailored to other visualization tools.
+
+```
+$ jps
+9234 Jps
+8983 Computey
+$ ./run-command.sh 8983 interval:1,start
+$ ./run-command.sh dumpRawTraces:/tmp/traces.txt
```
brendangregg: missing PID

brendangregg: missing PID (sorry, kept clicking the wrong github button).

README.md
```diff
@@ -1,4 +1,151 @@
...
+		Primes$1::run
+		java.lang.Thread::run
+```
+
+This stack appeared 1056 times in the trace. This is the same stack trace format
+used by the [Lightweight Java Profiler](https://code.google.com/archive/p/lightweight-java-profiler/),
+and can be easily used for generating flame graphs.
+
+## Restrictions/Limitations
+
```
brendangregg: Should mention overhead of this approach. What is the overhead of AsyncGetCallTrace?

One of our major microservices has a java call stack that gets to over 3000 frames deep, 1000 frames after inlining. That's the outlier where -XX:+PreserveFramePointer can add as much as 10% overhead at times (most microservices that overhead is < 0.3%, and close to 0.0%).

Even if the overhead of AsyncGetCallTrace was high, it may still be preferable to always-on frame pointer overhead. But it's important to give some guidance so the end user can make informed decisions...

brendangregg: PS. It's ok to say the overhead is not yet quantified. :)

README.md
```diff
@@ -1,4 +1,151 @@
...
+* Accessing `perf_events` requires root privileges. Because the agent is loaded
+into the Java process, **the Java process must run as root**.
```
apangin: Not really. PERF_COUNT_SW_CPU_CLOCK counter (used by default) does not require root privileges. async-profiler opens per-thread events that are also available for unprivileged users. Before Linux 4.6 it was even possible to get kernel callchains. Since 4.6 only user-mode profiling is available by default, but it is possible to override this by setting /proc/sys/kernel/perf_event_paranoid to 1 or below. Also /proc/sys/kernel/kptr_restrict should be 0 to allow resolving kernel symbols by non-root.
In fact, I could successfully get full-stack profiles under non-root account on all systems I've tested, e.g. CentOS 7 / 3.10 kernel or Ubuntu 16.04 / 4.4 kernel.

run-command.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo java -cp $JAVA_HOME/lib/tools.jar:./build Attach ./build/libasyncProfiler.so "$@"
```
apangin: As mentioned above, sudo is not required

src/Attach.java
```diff
@@ -0,0 +1,50 @@
...
+                System.out.printf("The agent at %s could not be found",
+                                  fullPath);
+                System.exit(1);
+            }
+            else vm.loadAgentPath(fullPath, options);
+        } finally {
+            vm.detach();
+        }
+    }
+}
```
apangin: I was actually thinking on including [jattach](https://github.com/apangin/jattach) launcher for this purpose, as it has faster startup and does not require tools.jar available in JDK but not JRE.

goldshtn: Oh cool, would you like me to just copy the Linux implementation in and build a standalone tool out of it? Or do you want to tell people they need to download jattach separately (I think it's not as nice in terms of friction)?

README.md
goldshtn: Great! Will fix the instructions.

README.md
goldshtn: Makes sense. Will measure and report back.

run-command.sh
goldshtn: Fixed.

README.md
goldshtn: Fixed.

README.md
goldshtn: Fixed.

src/Attach.java
apangin: The tool is rather small - it's OK to copy it here.

goldshtn: Sounds good - I will copy it from there.

goldshtn:
@apangin @incubos @brendangregg Added jattach, reworked the helper script (profiler.sh) to be a little more useful, and rebased to clean up the commits. I think this is good to go, and I will file an issue for the overhead measurement work, which I think should be done separately.

goldshtn:
Also, I believe this resolves #1.

README.md
```diff
@@ -168,3 +168,11 @@ of 127 frames. On recent Linux kernels, this can be configured using
...
 which means that in some rare cases, the captured Java stack might not match
 the captured native (user+kernel) stack.
+
+* When the JVM option `-XX:+PreserveFramePointer` is not enabled (available
+since 1.8u60), `perf_events` might not be able to walk the full stack through
+the Java frames. This should not prevent you from getting a good stack trace,
+but you will not see the non-Java frames _preceding_ the Java frames on the
+stack. For example, if `start_thread` called `JavaMain` and then your Java
+code started running, you will not see the first two frames in the resulting
+stack if `perf_events` is unable to unwind the whole stack.
```
apangin: This is general limitation of the approach, even with PreserveFramePointer enabled. The native stack ends at the first Java frame; everything below it is obtained via AsyncGetCallTrace.

src/profiler.cpp
```diff
@@ -149,8 +149,7 @@ void Profiler::checkDeadline() {
         ssize_t w = write(STDERR_FILENO, error, sizeof(error) - 1);
         (void) w;
 
-        _running = false;
-        setTimer(0, 0);
+        stop();
```
apangin: It is not safe to call Profiler::stop() from within signal handler.
Automatic stop mechanism should be either re-implemented  (e.g. by sleeping in a separate thread) or removed altogether. It's OK to remove it for now and fix later in a separate issue.

README.md
goldshtn: Fixing, thanks.

src/profiler.cpp
goldshtn: Right. I'll get rid of the whole deadline thing for now, then, and create a separate issue to track it, as you suggested.

goldshtn:
@apangin Addressed the last two comments, opened an issue about the automatic stop in #15.

apangin:
Great! Thanks for all your efforts.

PR #17, opened 2017-04-11
Dump traces in FlameGraph format

apangin:
- `dumpRawTraces` now prints stack traces in the FlameGraph format, so the intermediate step of running `stackcollapse-ljp` is no longer needed;
 - Java classes are printed in the slashed format (e.g. java/lang/Thread), and `_[k]` is appended to the kernel symbols to match `--colors=java` scheme of the FlameGraph script.

### DIFF
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -35,17 +35,8 @@ static inline jmethodID asJavaMethod(const char* name) {
     return (jmethodID)(-(intptr_t)name);
 }
 
-static inline void escape(char* name) {
-    char *pos = name;
-    while (*pos) {
-        if (*pos == ' ') {
-            *pos = '_';
-        }
-        pos++;
-    }
-}
 
-MethodName::MethodName(jmethodID method, const char* sep) {
+MethodName::MethodName(jmethodID method, bool dotted) {
     if (method == NULL) {
         _str = "[unknown]";
     } else if (isString(method)) {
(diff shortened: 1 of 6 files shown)


goldshtn:
@apangin Which brings me to a question -- why in this particular example are we seeing so many samples in `JVM_FindSignal` and our signal handler that calls `AsyncGetCallTrace`? I haven't seen anything like it with a sample app that spends 100% spinning on-CPU. Here's the sample application I used: https://github.com/goldshtn/linux-tracing-workshop/blob/master/buggy/Allocy.java

goldshtn:
I'm seeing similar results with an even simpler application, that just calls `System.gc()` in a loop. >90% of the samples in `JVM_handle_linux_signal`, `AsyncGetCallTrace`, and `JVM_FindSignal`. Sample code here:

```java
class OnlyGC {
  public static void main(String[] args) {
    while (true) {
      System.gc();
    }
  }
}
```

apangin:
Looks like a bug. I'll investigate this.

goldshtn:
Cool. But that shouldn't detract from this PR or block it - I've reproduced the same behavior on the master branch.

apangin:
Seems that you run OpenJDK where libjvm.so has no symbol table. Function names were shown incorrectly in this case. Now it will display just library name instead of wrong symbols.

goldshtn:
Tested, that helps. I have external debuginfo installed for my JVM (on Fedora), and symbols are not resolved. I noticed you opened a separate issue #18 to track this, will comment there.

src/codeCache.cpp
```diff
@@ -60,22 +51,22 @@ MethodName::MethodName(jmethodID method, const char* sep) {
...
-        if (*s == '/') *s = '.';
+char* MethodName::fixClassName(char* name, bool dotted) {
+    if (dotted) {
+        for (char* s = name + 1; *s; s++) {
+            if (*s == '/') *s = '.';
+        }
     }
-    s[-1] = 0;
+
+    name[strlen(name) - 1] = 0;
```
incubos: Is it necessary?

src/codeCache.cpp
```diff
@@ -60,22 +51,22 @@ MethodName::MethodName(jmethodID method, const char* sep) {
...
     return name + 1;
```
incubos: Why do we skip the first character?

apangin: [Class signature](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.9.1) is a string of form `Ljava/lang/Thread;`
So we have to remove the first `L` and the last `;`

incubos: Pushed as a comment.

PR #23, opened 2017-04-13
Parse external debuginfo

apangin:
If shared library misses .strtab section, look for debug symbols in the external sources:

1. /usr/lib/debug/.build-id/ab/cdef1234.debug, where abcdef1234 - is ID obtained from `.note.gnu.build-id` section;
2. /usr/lib/debug/path/to/lib.so
3. /usr/lib/debug/path/to/lib.so.debug
4. /path/to/lib.so.debug

Note: to keep things simple we do not parse .gnu_debuglink contents, but rather search through a list of common locations.

### DIFF
--- a/src/symbols.cpp
+++ b/src/symbols.cpp
@@ -202,7 +202,7 @@ void Symbols::parseElf(CodeCache* cc, const char* addr, const char* base, BuildI
 int Symbols::parseMaps(CodeCache** array, int size) {
     int count = 0;
     if (count < size) {
-        CodeCache* cc = new CodeCache("[kernel]", true);
+        CodeCache* cc = new CodeCache("[kernel]");
         parseKernelSymbols(cc);
         cc->sort();
         array[count++] = cc;
@@ -214,7 +214,7 @@ int Symbols::parseMaps(CodeCache** array, int size) {
     while (count < size && std::getline(maps, str)) {
         MemoryMap map(str.c_str());
         if (map.isExecutable() && map.file()[0] != 0) {
-            CodeCache* cc = new CodeCache(map.file(), false, map.addr(), map.end());
+            CodeCache* cc = new CodeCache(map.file(), map.addr(), map.end());
             const char* base = map.addr() - map.offs();
 
             if (map.inode() != 0) {
--- a/src/symbols.cpp
+++ b/src/symbols.cpp
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-#include <elf.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
@@ -29,13 +28,13 @@
 #include "symbols.h"
 
 
-class SymbolMap {
+class SymbolDesc {
   private:
     const char* _addr;
     const char* _type;
 
   public:
-      SymbolMap(const char* s) {
+      SymbolDesc(const char* s) {
           _addr = s;
           _type = strchr(_addr, ' ') + 1;
       }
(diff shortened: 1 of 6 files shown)


apangin:
Running OpenJDK with extra debuginfo package installed:

fg

goldshtn:
Hmm, I am still getting stacks with no symbols, despite installing the debuginfo package. Going to check where exactly the debuginfo files ended up. By the way, why not parse the debuglink section rather than guessing?

goldshtn:
So that's weird, the symbols for libjvm.so are in the obvious place:

```
$ ls /usr/lib/debug/lib/jvm/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/jre/lib/amd64/server/*
/usr/lib/debug/lib/jvm/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/jre/lib/amd64/server/libjsig.so.debug
/usr/lib/debug/lib/jvm/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/jre/lib/amd64/server/libjvm.so.debug
```

But as you see in the flame graph above, async-profiler doesn't find them. This is FC25, OpenJDK 1.8.

src/symbols.cpp
```diff
@@ -91,11 +100,61 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
             }
         }
     }
+}
+
+void Symbols::parseLibrarySymbols(CodeCache* cc, const char* lib_name, const char* base) {
+    // Parse symbols from the original .so
+    // If there is no .strtab section, try loading symbols from an external debug library
```
goldshtn: I think it is actually preferable to load symbols from external debuginfo first, if it is present, than from the actual binary. The reason is that the binary might have stripped symbols, or partial symbols, while the external debuginfo is very likely to have more complete symbol information.

apangin:
OK, I'll take care of .gnu_debuglink

apangin:
Updated PR to handle debuglink section

src/symbols.cpp
```diff
@@ -76,89 +76,193 @@ class MemoryMap {
...
+}
+
+// Look for debuginfo file specified in .gnu_debuglink section
+bool ElfParser::loadSymbolsUsingDebugLink() {
+    ElfSection* section = findSection(SHT_PROGBITS, ".gnu_debuglink");
+    if (section == NULL || section->sh_size <= 4) {
+        return false;
+    }
+
+    const char* basename = strrchr(_file_name, '/');
```
goldshtn: Perhaps use basename(3)?

src/symbols.cpp
```diff
@@ -76,89 +76,193 @@ class MemoryMap {
...
+    if (basename == NULL) {
+        return false;
+    }
+
+    char* dirname = strndup(_file_name, basename - _file_name);
```
goldshtn: Perhaps use dirname(3)?

apangin: Yes, I know about these library functions, but I avoided them intentionally.
dirname(3) yields extra '/' for the root directory. Also man page is not quite clear whether it is safe to call dirname+basename on the same string. Since these library calls do not make the code substantially shorter or clearer, I would stay with the given implementation.

goldshtn:
Cool! Quick question: do you like merging with merge commits (like you did so far), or would you prefer rebasing (so there are no merge commits) or squashing (so the whole PR becomes one commit)? I know it's mostly a personal preference, I personally prefer rebasing, but what do you think?

apangin:
I don't have any good reason to choose one over another; rebasing will be OK. For the short-lived branches squash commits will do as well.

Issue #25, opened 2017-04-17
Hard crash (core dump) while dumping traces


goldshtn:
I have a fairly reproducible hard crash when dumping traces. This happened as I was working on an overhead testing script. To try and reproduce, clone the [goldshtn:overhead](https://github.com/goldshtn/async-profiler/tree/overhead) branch from my repo, build the agent, and then run `overhead-test.sh` from the test directory. After a couple of lines, I get a hard crash. Some basic analysis:

- SIGSEGV in "Attach Listener" thread
- Crash address is in `strlen` called by `vsnprintf`
- After building with `-g` and looking at the core file in gdb, here's the full stack trace:

```
$ gdb java -c core.25942
GNU gdb (GDB) Fedora 7.12.1-47.fc25
...
Core was generated by `java Busy 1 1100 7'.
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
58	}
[Current thread is 1 (Thread 0x7f5e0c7f7700 (LWP 25958))]
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
#1  0x00007f5e2333951a in __GI_abort () at abort.c:89
#2  0x00007f5e22be74d9 in os::abort (dump_core=<optimized out>) at /usr/src/debug/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/openjdk/hotspot/src/os/linux/vm/os_linux.cpp:1500
#3  0x00007f5e22e12968 in VMError::report_and_die (this=this@entry=0x7f5e0c7f3bd0) at /usr/src/debug/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/openjdk/hotspot/src/share/vm/utilities/vmError.cpp:1060
#4  0x00007f5e22bf09b5 in JVM_handle_linux_signal (sig=sig@entry=11, info=info@entry=0x7f5e0c7f3e70, ucVoid=ucVoid@entry=0x7f5e0c7f3d40, abort_if_unrecognized=abort_if_unrecognized@entry=1)
    at /usr/src/debug/java-1.8.0-openjdk-1.8.0.121-10.b14.fc25.x86_64/openjdk/hotspot/src/os_cpu/linux_x86/vm/os_linux_x86.cpp:556
[...]

PR #26, opened 2017-04-17
#25: Guard state transitions by a mutex

apangin:
Change profiler state between IDLE/RUNNING/TERMINATED  under state_lock.
This will delay VM termination until all pending profiler output is completed.


goldshtn:
LGTM, hope to test this later today.

PR #30, opened 2017-04-21
Suppot linux-arm config

sergey-nbl:


src/perfEvent.cpp
```diff
@@ -25,7 +25,14 @@
 #include "perfEvent.h"
 #include "vmEntry.h"
 
+// from aarch32_port
+#if (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_7__)
+#define rmb() __asm__ __volatile__ ("dmb ish"   : : : "memory")
+#elif defined (__arm__) || (__thumb__)
```
apangin: `(__thumb__)` --> `defined(__thumb__)` ?

src/spinLock.h
```diff
@@ -17,8 +17,12 @@
 #ifndef _SPINLOCK_H
 #define _SPINLOCK_H
 
+#if defined(__arm__) || defined(__thumb__)
+#include <sched.h>
+#define pause() sched_yield()
+#else
 #define x86_pause()  asm volatile("pause")
```
apangin: In this case `x86_pause()` should be also renamed. But `pause()` is a library function defined in <unistd.h>

src/spinLock.h
```diff
@@ -17,8 +17,12 @@
 #ifndef _SPINLOCK_H
 #define _SPINLOCK_H
 
+#if defined(__arm__) || defined(__thumb__)
+#include <sched.h>
+#define pause() sched_yield()
```
apangin: ARMv6K+ has `YIELD` instruction that is similar to x86 `PAUSE`

src/stackFrame.cpp
```diff
@@ -17,7 +17,19 @@
 #include <ucontext.h>
 #include "stackFrame.h"
 
+#if defined(__arm__) || defined(__thimb__)
```
apangin: typo: thumb

sergey-nbl:
Ouch! So many error, will fix later without haste.

And yes, it works on RPi2

sergey-nbl:
OK, I've updated the files. Now I limited a number of supported CPU by ARMv7 and higher due to DMB and YIELD instructions need. Verified on RPI2, build check only for x86

PR #31, opened 2017-04-22
Support linux-arm platfrom. ARMv7 and higher only

sergey-nbl:
Recreated request to remove redundant commits. 
OK, I've updated the files. Now I limite a number of supported CPU by ARMv7 and higher due to DMB and YIELD instructions need. Verified on RPI2, build check only for x86

### DIFF
--- a/src/perfEvent.cpp
+++ b/src/perfEvent.cpp
@@ -25,7 +25,12 @@
 #include "perfEvent.h"
 #include "vmEntry.h"
 
+// from aarch32_port
+#if defined (__arm__) || defined(__thumb__)
+#define rmb() __asm__ __volatile__ ("dmb ish"   : : : "memory")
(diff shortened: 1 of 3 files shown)


goldshtn:
Perhaps this would be a good opportunity to update the README, specifically the "Building" section, to explain which platforms are supported and tested? The CI server will not test the ARM build, so for now we'd have to verify manually once in a while that we haven't broken ARM.

apangin:
I've verified that async-profilers compiles and runs on my Galaxy smartphone. However, it sometimes crashes when printing profiler results. The problem is known, I will address it in a separate issue - #32.

apangin:
I think this PR is a good starting point for ARM support. README updates and minor cosmetic changes will follow shortly.

PR #34, opened 2017-04-27
Ignore SIGPIPE during socket operations to avoid silent exits

qwwdfsad:
Currently if one has problems with permissions, jattach will print
`Connected to remote JVM` to stdout and silently exits due to unhandled `SIGPIPE` signal.

Before patch:
```
./profiler.sh -p ... -a start 
echo $?
141
```

After patch:
```
./profiler.sh -p ... -a start
Error writing to socket: Broken pipe
echo $?
1
```

### DIFF
--- a/src/jattach.c
+++ b/src/jattach.c
@@ -128,6 +128,7 @@ int main(int argc, char** argv) {
         return 1;
     }
 
+    signal(SIGPIPE, SIG_IGN);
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for the fix. It would be nice to have it in https://github.com/apangin/jattach as well.

PR #36, opened 2017-06-08
Expose method for raw traces to Java

WIgor:
I want to integrate this profiler to an application. To make a more convenient way of getting traces for flamegraph. And to do this I need to expose dumpRawTraces to Java.


apangin:
Makes sense. Thank you for contribution.

PR #37, opened 2017-06-29
Resolves #32: Revise usage of jmethodID to encode strings

### LINKED ISSUE #32: Revise usage of jmethodID to distinguish between Java methods and native symbols
`jmethodID` has ambiguous meaning in the profiler: it can denote either Java method or a native symbol. Currently this type is encoded in the highest bit of `jmethodID` pointer. This works well on 64-bit Linux, but may not be appropriate for 32-bit platforms (e.g. linux-arm).

apangin:
Call traces are now stored as an array of `ASGCT_CallFrame` (instead of an array of `jmethodID`).

`bci` field is used to distringuish Java methods from native functions. The special value -2 means that jmethodID should be treated as a string holding the native function name.

Related changes:

- StringTable is completely removed; all function names are strdup'ed.
- CodeCache is split into Java CodeCache and NativeCodeCache: the former holds `jmethodID` while the latter holds `const char*`.
- JITted methods and VM generated stubs are now in two separate CodeCaches.
- CodeCache is updated under lock; when it grows, the previously allocated memory is reclaimed.

Linux/ARM builds do not crash anymore during profile dumping.

### DIFF
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -14,158 +14,29 @@
  * limitations under the License.
  */
 
-#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
-#include <stdio.h>
-#include <cxxabi.h>
 #include "codeCache.h"
-#include "vmEntry.h"
 
 
-static inline bool isString(jmethodID method) {
-    return (intptr_t)method < 0;
-}
-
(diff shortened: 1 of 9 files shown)


goldshtn:
LGTM, tests seem to pass and the changes make sense.

Issue #27, opened 2017-04-20
Add "info" command


alekseytomin:
Please to add "jattach.sh -p XXXX -i" - print info (profile was (not)attached and any other information).

goldshtn:
Sorry, not sure what you mean by that. What would the -i flag do? What is jattach.sh?

alekseytomin:
Sorry, profiler.sh :)

I want to see status:
"profiler was not attached to this process"
"profiler attached at 123434 seconds"

goldshtn:
So would you want the -i switch to print this information _after_ you already attached the profiler? Or is that something you want each command to print as you execute it? It would be nice if you could mock the terminal output you would expect, to understand this better.

goldshtn:
@alekseytomin Ping.

alekseytomin:
I will create merge request in this week.

apangin:
Added `status` command.

Issue #19, opened 2017-04-12
Allow redirecting all output to a file


goldshtn:
Currently, only the raw stack dumps can be redirected to a file. The rest of the output goes to the Java program's stdout, which can be undesired in case that output is parsed or redirected in the first place.

I propose another agent parameter: `file:<path-to-file>`, which can be appended to the start command. For example: `jattach ... file:/tmp/profile.log,interval:1000000,start`.

goldshtn:
@apangin Does the `file:<path-to-file>` parameter make sense?

apangin:
Yes, it is fine. I suppose this parameter will also replace an optional file name in `dumpRawTraces` command.

PR #38, opened 2017-07-04
Better argument processing. Resolves #15, #19, #27.

### LINKED ISSUE #15: Automatically stop profiling after a specified duration
Reintroduce the "duration:N" parameter support in the profiler by having it detect that the specified number of seconds has elapsed and stop automatically. This requires stopping the profiler from a separate thread, or after returning from the signal handler, because it is not safe to stop the profiler from the signal handler that detects the timeout.

Perhaps we should consider leaving this up to the user, and removing the "duration" parameter altogether? Alternatively, the helper script could accept a duration parameter and do this automatically:

```
$ ./profiler.sh -p 8983 -a start -d 10      # duration: 10 seconds, script stops automatically
```

apangin:
Here is a number of improvements both to parsing of Agent arguments and to profiler.sh command-line interface.

1. Moved arguments processing to the separate `arguments.cpp` file.
2. Removed the timeout check from the agent code (it did not work anyway). Now it is responsibility of the wrapper script to stop the profiling after the specified period of time.
3. Added `-d` (duration) option to profiler.sh. It allows to execute start-wait-stop sequence with just one command.
4. Allowed to dump any combination of profiling data (summary / flat profile / call stacks / flamegraph) also with just one command.
5. Filename option now works not only for the flamegraph format, but for any combination of output formats above.
6. Added --status option to print whether profiler is active and for how long.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -21,6 +21,16 @@ usage() {
     exit 1
 }
 
+show_agent_output() {
+    if [[ $USE_TMP ]]; then
+        if [[ -f $FILE ]]; then
+            cat $FILE
+            rm $FILE
+        fi
+    fi
+}
+
+
 OPTIND=1
 SCRIPT_DIR=$(dirname $0)
 JATTACH=$SCRIPT_DIR/build/jattach
@@ -29,6 +39,7 @@ PROFILER=$(readlink -f $SCRIPT_DIR/build/libasyncProfiler.so)
 ACTION="collect"
 DURATION="60"
 FILE=""
+USE_TMP="true"
 INTERVAL=""
 FRAMEBUF=""
 OUTPUT="summary,traces=200,methods=200"
@@ -46,7 +57,8 @@ while [[ $# -gt 0 ]]; do
             shift
             ;;
         -f)
-            FILE=",file=$2"
+            FILE="$2"
+            unset USE_TMP
             shift
             ;;
         -i)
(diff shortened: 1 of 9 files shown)


apangin:
Agree. I didn't feel comfortable with -- options either. Replaced them with actions. If action is omitted, timed profiling will take place.

apangin:
One more addition:

- Any profiler output can now be redirected to a file, including messages printed by "start", "stop" and "status" commands.
- If no `-f` option is given, `profiler.sh` makes agent dump to a temporary file, which is then printed to a caller console. It is no longer necessary to switch between terminals; profiling output is printed to the script's console by default.

profiler.sh
```diff
@@ -74,22 +86,30 @@ done
 
 [[ "$PID" == "" ]] && usage
 
+# if no -f argument is given, use temporary file to transfer output to caller terminal
+if [[ $USE_TMP ]]; then
+    FILE="/tmp/async-profiler-${PID}.tmp";
```
goldshtn: Wouldn't mktemp be safer?

goldshtn:
LGTM except the temp file name comment. I've written a quick blog post about async-profiler which I plan to publish tomorrow, just FYI.

profiler.sh
apangin: All right; changed to mktemp.

apangin:
Greate, thank you!
Please note that I've moved the repository to jvm-profiling-tools as @nitsanw suggested, so make sure you've updated the links in your post.

PR #41, opened 2017-08-07
Heap allocation profiler

apangin:
Implemented heap allocation profiler.

It reuses all infrustructure for attaching, handling signals, collecting samples and dumping results, though the event source is different.

The idea of this allocation profiler is to intercept HotSpot allocation tracing callbacks. There are basically two of them:

- `AllocTracer::send_allocation_in_new_tlab_event` - called when an object is allocated in newly created TLAB (thread local allocation buffer);
- `AllocTracer::send_allocation_outside_tlab_event` - called when an object is allocated in slow path outside TLAB.

It seems those functions are designed specially for Mission Control / Flight Recorder. However, they *do exist* in OpenJDK, and what is even better, they *are called unconditionally*. So, this allocation profiler works both with OpenJDK and Oracle JDK and it does not rely on any commercial feature or closed source. At the same time it provides nearly the same grade of allocation profiling as JMC.

The technique to intercept allocation callbacks works as follows:

1. Use either built-in or external ELF symbol table to find the address of the interesting function.
2. Remap the memory page with those function in writeable mode.
3. Insert a breakpoint trap at the entry of the function.
4. Install signal handler to process SIGTRAP.
5. Whenever trap is hit, read the function arguments from `ucontext` to find the allocated object class and the object size.
6. Use `VMStructs` from Serviceability Agent to find the offset of the class name.
7. Use `AsyncGetCallTrace` to collect the stack trace.
8. Record allocation sample,

Tested on Linux amd64/i386/ARM.


ceeaspb:
Hi Andrei,
Looks really useful, it may be of interest to cross reference with other work/discussions:
There was some discussion including Alloc Tracer and tlab sampling in this thread:
http://mail.openjdk.java.net/pipermail/serviceability-dev/2015-June/017543.html
which led to http://openjdk.java.net/jeps/8171119
with follow up emails: http://mail.openjdk.java.net/pipermail/serviceability-dev/2017-April/021212.html (not all the thread is linked)
It going into the jvm won't be ready for a while...

Alloc tracer functions with uprobes/perf works fine also, but clearly with this tool you get much more information, and it could be lower overhead.

apangin:
@ceeaspb Thank you for the pointers and for your interest in the topic. I've seen the discussion and I believe this JEP is the right way to go. The support for low overhead profiling should be in JVM in the first place.

The motivation for implementing this feature right now in async-profiler was the following.
1. async-profiler already has the required infrastructure for profiling, including attach-on-demand capability, stack traces collector, different ways of presenting output etc.
2. The approach just works with out-of-the-box Java distributions ranging from JDK 7 to modern JDK 9 at no additional cost. It does not even require special command line flags, preloaded agents and so on.
3. Performance analysis often involves both CPU and memory profiling. It is nice to have one tool for both. We already use async-profiler in production for CPU profiling; now it will allow us to collect heap profiles for free.

PR #53, opened 2017-09-27
Update jattach with changes from upstream

goldshtn:
* macOS support for temporary directory discovery

* Container support by entering mount namespace and considering
  pid namespace differences between container and host

### DIFF
--- a/src/jattach.c
+++ b/src/jattach.c
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


goldshtn:
Would be nice to have an automatic way of keeping the repos in sync. Could add a step to the Makefile that fetches `jattach_linux.c`. Or use a submodule, which I personally don't like. Thoughts?

PR #57, opened 2017-10-12
Perf events support

nitsanw:
Add support for different perf events:
 * HW cycles
 * branch-misses
 * context-switches
 * LLC/L1D load misses
 * cache-misses

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -17,6 +17,7 @@
 #include <limits.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cstdio>
 #include "arguments.h"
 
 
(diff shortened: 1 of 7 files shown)


src/arguments.cpp
```diff
@@ -55,8 +56,31 @@ const char* Arguments::parse(char* args) {
...
+                } else if (strcmp(event, "cache-misses") == 0) {
+                    _event_type = EVENT_TYPE_CACHE_MISSES;
+                } else if (strcmp(event, "L1-dcache-load-misses") == 0) {
+                    _event_type = EVENT_TYPE_L1D_LOAD_MISSES;
+                } else if (strcmp(event, "LLC-load-misses") == 0) {
+                    _event_type = EVENT_TYPE_LLC_LOAD_MISSES;
+                }
+                else {
+                    const char *pattern = "unknown event type :%s";
+                    char *err_msg = static_cast<char *>(malloc(strlen(pattern) + strlen(event)));
```
qwwdfsad: Shouldn't it be freed on exit?
Or copied to _buf

apangin: Agree. I'd personally get rid of malloc here altogether. Returning "Unknown event type" message is enough.

src/perfEvents_linux.cpp
```diff
@@ -101,16 +106,56 @@ int PerfEvents::tid() {
...
+                                           PERF_COUNT_HW_CACHE_RESULT_MISS);
+            break;
+        case EVENT_TYPE_LLC_LOAD_MISSES:
+            precise_ip = 0;
+            type = PERF_TYPE_HW_CACHE;
+            config = hw_cache_event_config(PERF_COUNT_HW_CACHE_LL,
+                                           PERF_COUNT_HW_CACHE_OP_READ,
+                                           PERF_COUNT_HW_CACHE_RESULT_MISS);
+            break;
+    }
```
apangin: It's already long enough switch, and it'll be even longer when new events added. Probably array-backed mapping will be good.

src/arguments.cpp
```diff
@@ -55,8 +56,31 @@ const char* Arguments::parse(char* args) {
             _action = ACTION_STOP;
         } else if (strcmp(arg, "status") == 0) {
             _action = ACTION_STATUS;
-        } else if (strcmp(arg, "cpu") == 0) {
+        } else if (strncmp(arg, "cpu", 3) == 0) {
```
apangin: Does `-m cpu:cache-misses` syntax have any advantages over just `-m cache-misses`? Any objects against extending `enum Mode` with all these events?

nitsanw: I don't mind, I was treating 'cpu' as the gateway to perf profiling, so in that sense 'perf:cpu' and 'perf:cache-misses' made sense to me. But modes work equally well, or following the perf notation we could add an option: '-e cache-misses'

src/perfEvents_linux.cpp
nitsanw: Sounds good

src/arguments.cpp
nitsanw: sounds good

nitsanw:
I'm going to be quite busy in the near future, so I will fix the allocation but would not have time to play around with reworking the modes, or converting the switch to a lookup table. I hope the PR offers enough value as is.

apangin:
OK, no problem. I'll merge it as is and leave the final decorations to myself.

PR #58, opened 2017-10-16
Alloc details

### LINKED ISSUE #46: Report for number of allocations, in addition to aggregate size
Ideally, both would be recorded and available to the output formatter.

A new command line option `-a size|count` could control which counter the current formatters choose to report.

More advanced formatters/visualization tools could display both.

apangin:
- Collect counter values in addition to number of samples.
- Added option to output either counter total value or number of samples in collapsed format. In particular, heap profiles can now display either object count or their total size.
- Distinguish between allocations in new TLAB vs. allocations outside TLAB. They are displayed in different color in flame graph.
- Pretty class names in allocation profile, e.g. `java.lang.Integer[]` instead of `[Ljava/lang/Integer;`

Resolves #46

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -193,6 +193,7 @@ void PerfEvents::signalHandler(int signo, siginfo_t* siginfo, void* ucontext) {
     }
 
     Profiler::_instance.recordSample(ucontext, counter, 0, NULL);
+    ioctl(siginfo->si_fd, PERF_EVENT_IOC_RESET, 0);
     ioctl(siginfo->si_fd, PERF_EVENT_IOC_REFRESH, 1);
 }
 
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -187,7 +187,12 @@ void PerfEvents::installSignalHandler() {
 }
 
 void PerfEvents::signalHandler(int signo, siginfo_t* siginfo, void* ucontext) {
-    Profiler::_instance.recordSample(ucontext, 1, 0, NULL);
+    u64 counter;
+    if (read(siginfo->si_fd, &counter, sizeof(counter)) != sizeof(counter)) {
+        counter = 1;
+    }
+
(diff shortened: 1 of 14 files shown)


src/profiler.h
```diff
@@ -40,8 +39,15 @@ static inline int cmp64(u64 a, u64 b) {
 }
 
 
+union CallTraceBuffer {
+    ASGCT_CallFrame _asgct_frames[MAX_STACK_FRAMES];
+    jvmtiFrameInfo _jvmti_frames[MAX_STACK_FRAMES];
```
qwwdfsad: It's not used.
Is it stub  for https://github.com/jvm-profiling-tools/async-profiler/issues/51?

src/perfEvents_linux.cpp
```diff
@@ -187,7 +187,13 @@ void PerfEvents::installSignalHandler() {
 }
 
 void PerfEvents::signalHandler(int signo, siginfo_t* siginfo, void* ucontext) {
-    Profiler::_instance.recordSample(ucontext, 1, NULL);
+    u64 counter;
+    if (read(siginfo->si_fd, &counter, sizeof(counter)) != sizeof(counter)) {
```
qwwdfsad: Could you please elaborate why is it needed here?

src/profiler.h
apangin: Exactly

src/perfEvents_linux.cpp
apangin: read(siginfo->si_fd) reads the value of perf counter. I assume this read() should always succeed. `if` is left here just in case. Do you think it is better to ignore return value or handle it some other way?

qwwdfsad: I think it's less confusing to handle it as `if (read(siginfo->si_fd, &counter, sizeof(counter)) == -1`

apangin: Not sure this is better. I expect to read the counter variable fully - that's what I've written.

PR #63, opened 2017-10-31
Load AsyncGetCallTrace from libjvm.so

vsarunas:
Attempting to use async-profiler on Java code that has been launched by C++ code which dynamically loads libjvm.so and then starts embedded JVM via [JNI_CreateJavaVM()](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html) results in:

> Could not find AsyncGetCallTrace function

That is because current attempt to load this function is only via RTLD_DEFAULT. In case of failure, we can try loading AsyncGetCallTrace from libjvm.so explicitly to handle this edge case.

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -70,6 +70,14 @@ void VM::init(JavaVM* vm, bool attach) {
     PerfEvents::init();
 
     _asyncGetCallTrace = (AsyncGetCallTrace)dlsym(RTLD_DEFAULT, "AsyncGetCallTrace");
+    if (_asyncGetCallTrace == NULL) {
+        void* libjvm_handle = dlopen("libjvm.so", RTLD_NOW);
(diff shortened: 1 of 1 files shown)


src/vmEntry.cpp
```diff
@@ -70,6 +70,14 @@ void VM::init(JavaVM* vm, bool attach) {
     PerfEvents::init();
 
     _asyncGetCallTrace = (AsyncGetCallTrace)dlsym(RTLD_DEFAULT, "AsyncGetCallTrace");
+    if (_asyncGetCallTrace == NULL) {
```
apangin: Looks good, but I would leave a comment here when this is needed.
Something like in PR description would be sufficient.

vsarunas: Added a comment.

PR #71, opened 2017-11-27
Add ability to perform a continuous profiling

stIncMale:
async-profiler implements cool ideas on how to make profiling better, but it can't do continuous profiling, unlike Java Flight Recorder. One can't, however, use Java Flight Recorder in production for free, so it would be cool to have a similar functionality in async-profiler. I don't know C++ :) and hence I can't implement anything like this in the profiler itself, but I have written a `monitoring.sh` script that allows running async-profiler's profiling iterations (sessions) infinitely while the target process is available. It does this by periodically running `profiler.sh` script with the given duration and options. `monitoring.sh` mimics all the suitable options of `profiler.sh` and deals with the constantly growing gathered data by removing outdated files.

I would like to suggest either adding this script to async-profiler, or implementing such a functionality in the source code of the profiler (which would be better).

### DIFF
--- a/monitor.sh
+++ b/monitor.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+usage() {
+    echo "This utility allows to run async-profiler's (https://github.com/jvm-profiling-tools/async-profiler)"
+    echo "  profiling iterations (sessions) infinitely while the target process is available."
+    echo ""
+    echo "Usage: ${0} [options] <pid>"
+    echo "Options:"
+    echo "  --profiler profilerScript   Path to 'profiler.sh'."
+    echo "  --history preserveN         Number of files with gathered data to preserve."
+    echo "                              <preserveN> * <duration> is the length of the history in seconds."
+    echo "  -e event                    Profiling event: cpu|alloc|lock|cache-misses etc."
+    echo "  -d duration                 Duration of a single profiling session in seconds."
+    echo "  -f filenamePrefix           Output filename prefix."
+    echo "  -i interval                 Sampling interval in nanoseconds."
+    echo "  -b bufsize                  Frame buffer size in bytes."
+    echo "  -t                          Profile different threads separately."
+    echo "  -o fmt[,fmt...]             Output format: summary|traces|flat|collapsed."
+    echo ""
+    echo "<pid> is a numeric process ID of the target JVM,"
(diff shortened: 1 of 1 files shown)


rpulle:
@apangin the following is the result.

-j 10
image

-j 5
image

toktarev:
Is it possible to execute continuous profiling directly inside your Java application using Async profiler Java API ?

krzysztofslusarski:
I've create a spring starter for that purpose:
https://github.com/krzysztofslusarski/continuous-async-profiler

All you need to do is to do stop/start/wait in a loop.

toktarev:
@krzysztofslusarski, thanks.
Good work !
Valentin(@stIncMale ), we could also Krzysztof 's code as JavaAgent.

I am sure that it is possible to write the same on C++ leve.

stIncMale:
@toktarev 
Sure, you may reuse the code from https://github.com/krzysztofslusarski/continuous-async-profiler, or may use async-profiler's Java API (https://github.com/jvm-profiling-tools/async-profiler/tree/master/src/api) directly when depending on Spring artifacts is not desirable. If you want to have a Java agent that does in-process profiling, you want as little dependencies as possible, as each dependency has to be repacked in the agent's JAR, i.e., using async-profiler's Java API directly seems preferable.

toktarev:
@stIncMale 
Sure, I was not able to use Async Profiler as a java agent in my real scenario.
Only direct java-api helped.

apangin:
After 4 years, I've finally committed the alternative implementation of continuous profiling, natively supported by the agent :)

Since version 2.0, async-profiler could already run day-long profiling sessions and record all events in the JFR format.
Now I added two more options:
  - `timeout=N` - to automatically stop profiling after N seconds;
  - `loop=N` - to restart profiler every N seconds (minutes/hours/days).

The corresponding profiler.sh arguments are `--timeout N` and `--loop N`. They are compatible not only with JFR, but with other formats, too.

If the output file name contains `%t` placeholder, it will be substituted by the current timestamp.
Example: `./profiler.sh --loop 1h -f /var/log/profile-%t.jfr <pid>`

apangin:
`loop/timeout` options work only when attaching profiler as an agent, they are not supported in Java API.
For continuous profiling inside a Java application, Krzysztof's solution is just fine.

toktarev:
I've checked Andrei's commit and think that someone from the community could do the same :)

PR #79, opened 2018-01-24
Built-in generator of Flame Graphs in SVG format

apangin:
- New output format: `-o svg`
- Automatically choose output format depending on file extension
- Removed references to `flamegraph.pl`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Features
+ - Produce SVG files out of the box; flamegraph.pl is no longer needed
+
 ## [1.1] - 2017-12-03
 
 ### Features
--- a/README.md
+++ b/README.md
@@ -153,20 +153,19 @@ Samples: 679 (98.84%)
 This indicates that the hottest method was `Primes.isPrime`, and the hottest
 call stack leading to it comes from `Primes.primesThread`.
 
-## Generating Flame Graphs
+## Flame Graph visualization
 
-To generate flame graphs or other visualizations from the collected profiling
-information, you will need to dump the raw collected traces to a file and then
-post-process it. The following example uses Brendan Gregg's
-[FlameGraph](https://github.com/BrendanGregg/FlameGraph) scripts, but a similar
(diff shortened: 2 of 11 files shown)


apangin:
Thank you for review!

rockiee281:
With -f flame.svg ,no output int flame.svg  but something like 'Stopped profiling after 30 seconds'. 

Without '-f' option, I can get the load trace normally. 

env: Centos 7.2, jdk8 with hotspot, 32G memory with G1,  build the async-profiler from master

apangin:
@rockiee281 I guess an older version of `libasyncProfiler.so` had been already loaded into Java process before you started profiling with a new version. Can it be the case?

rockiee281:
Hi, I build a fresh aysnc-profiler from source, and never use this tools before so I think there won't be any   libasyncProfiler.so load by java process before

rockiee281:
Anyway , now I use the old way to generate colloapsed tarce file and make flame graph by flamegraph.pl . This is really great tool, thanks

apangin:
@rockiee281 Hmm. `-f flame.svg` is also supposed to work. If it does not - it is a bug that I'd like to fix. If you can easily reproduce the problem, could you please add the following `echo` line to see the full jattach command issued by `profiler.sh`? Thanks.

```
diff --git a/profiler.sh b/profiler.sh
index 644a25b..d389950 100755
--- a/profiler.sh
+++ b/profiler.sh
@@ -45,6 +45,7 @@ check_if_terminated() {
 }

 jattach() {
+    echo $JATTACH $PID load "$PROFILER" true $1
     $JATTACH $PID load "$PROFILER" true $1 > /dev/null
     RET=$?
```

rockiee281:
@apangin  this output 2 lines：
./build/jattach 134836 load /<dir_where_build>/async-profiler/build/libasyncProfiler.so true start,event=cpu,file=flame1.svg,svg
./build/jattach 134836 load /<dir_where_build>/async-profiler/build/libasyncProfiler.so true stop,file=flame1.svg,svg

And this time it works well. I think its my mistake. Thanks you  👍

PR #85, opened 2018-02-19
Check if PAGE_SIZE is already defined.

jeroentervoorde:
Fixes compile on alpine linux where PAGE_SIZE is already defined in /usr/include/bits/limits.h (and included via /usr/include/limits.h)


apangin:
Thank you for the fix!

PR #115, opened 2018-05-09
Option to limit java stack depth to reduce CPU overhead

rpulle:
We have experimented with reduced java stack and observed less CPU overhead. This could be a useful option for others too.

Added a new option "-j jstackdepth" to control the java stack depth that AsyncGetCallTrace fetches.

Conditions are added to make sure that the total stack frames (native+java) are not exceeding MAX_STACK_FRAMES.

This option is applied only when event type is "cpu"

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -40,6 +40,7 @@ const Error Error::OK(NULL);
 //     traces[=N]    - dump top N call traces
 //     flat[=N]      - dump top N methods (aka flat profile)
 //     interval=N    - sampling interval in ns (default: 1'000'000, i.e. 1 ms)
+//     jstackdepth=N - java stack depth (default: MAX_STACK_FRAMES, will be ignored if the option is greater than default)
 //     framebuf=N    - size of the buffer for stack frames (default: 1'000'000)
 //     threads       - profile different threads separately
 //     simple        - simple class names instead of FQN
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -19,7 +19,6 @@
 #include <string.h>
 #include "arguments.h"
 
-
 // Predefined value that denotes successful operation
 const Error Error::OK(NULL);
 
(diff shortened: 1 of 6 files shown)


src/arguments.cpp
```diff
@@ -41,6 +40,7 @@ const Error Error::OK(NULL);
 //     traces[=N]    - dump top N call traces
 //     flat[=N]      - dump top N methods (aka flat profile)
 //     interval=N    - sampling interval in ns (default: 1'000'000, i.e. 1 ms)
+//     jstackdepth=N - java stack depth (default: MAX_STACK_FRAMES, will be ignored if the option is greater than default)
```
apangin: Would you like to add a paragraph in README.md as well?

src/profiler.cpp
```diff
@@ -614,6 +617,7 @@ void Profiler::dumpFlat(std::ostream& out, int max_methods) {
 void Profiler::runInternal(Arguments& args, std::ostream& out) {
     switch (args._action) {
         case ACTION_START: {
+            if(strcmp(args._event,"cpu") == 0) { _jstackdepth = args._jstackdepth; }
```
apangin: I think it's a kind of general argument, not specific to CPU profiling.

src/arguments.h
```diff
@@ -71,6 +71,7 @@ class Arguments {
     Counter _counter;
     const char* _event;
     long _interval;
+    long _jstackdepth;
```
apangin: `int` is definitely enough for this. I wonder how this compiles without a warning when you later assign it to `int max_depth`.

rpulle:
@apangin  Thank you for a quick response on the PR. I have made changes as per review comments.

src/profiler.cpp
```diff
@@ -338,7 +338,11 @@ void Profiler::recordSample(void* ucontext, u64 counter, jint event_type, jmetho
     }
 
     if (event == NULL || _JvmtiEnv_GetStackTrace == NULL) {
-        num_frames += getJavaTraceAsync(ucontext, frames + num_frames, MAX_STACK_FRAMES - 1 - num_frames);
+        int max_depth = MAX_STACK_FRAMES - 1 - num_frames;
+        if( _jstackdepth != 0 && _jstackdepth < max_depth) {
+		max_depth =  _jstackdepth + 1 + num_frames;
```
apangin: `+ 1 + num_frames` seems wrong, doesn't it?

rpulle:
@apangin the following is the result.

-j 10
image

-j 5
image

rpulle:
default:
image

apangin:
Thanks! Your change have been integrated into master.

PR #117, opened 2018-05-14
Call tree and Backtrace tree option

rpulle:
@apangin 

I have implemented call tree and backtrace tree output format.  I still need to do some code cleanup but wanted to get a review on functionality. Can you please try the options "-o calltree" and "--reverse -o calltree" and let me know any thing more useful. Any early feedback is really helpful.

Output is as follows:

Calltree:
image

Backtrace tree:
image


apangin:
Wow, that's impressive!
I was thinking about a different approach to tree view: exporting the profile to VisualVM and/or Java Mission Control format, since these tools being a part of JDK already have decent UI for browsing call trees. In fact, I'm already working on this.
But since you've already implemented the feature, and the changes are not very big, I think we may give it a chance. Do you think HTML tree generator would be useful if we had the support for above tools?

rpulle:
The html view can be integrated into any web view. For example we have integrated svg into a html panel of our UI.  The idea is to have call tree and back trace in the UI itself. Since we are targeting this for  production profiling, html view can be accessible from remote. 

I will create a clean PR for the change. I have done color codings to identify areas of the code similar to flame graph. 

You are right, Visual VM and MC can also be used to manually analyze the content.

apangin:
After playing with call trees for a while I have the following comments:

1. The name of a method is a part of a `<a>` link with handpoint cursor, but it does not expand/collapse subtrees. This is OK, but in this case I suppose it should have normal text cursor and allow selecting the text.
2. Colors are too light for the white background. The palette should be not random as in flame graphs. I guess just having one css class for each method type would be enough.
3. An option to expand the whole subtree at once will be helpful.

As to the source code, I noticed that `Profiler::dumpTree` is almost the same as `Profiler::dumpFlameGraph`. There is definitely a place for some cleanup.

BTW, thank you very much for your interest and contributions to async-profiler.

rpulle:
I have fixed comments 1, 2 and 3.
Also did code cleanup to optimize. To get the ability of generating both calltree and backtrace, I have provided two options and not depending on reverse option.

new output:
image

rpulle:
Added time spent in self:
image

rpulle:
image

rpulle:
@apangin  Based on my team feedback I have added %self and frame depth in the view.
Please let me know if you have any other input for this PR.

rpulle:
image
image

rpulle:
image

image

rpulle:
With search feature:
image

image

rpulle:
@apangin  I have added frame search feature in UI.

apangin:
Impressive changes! I'll take a deeper look a bit later.

README.md
```diff
@@ -250,6 +250,8 @@ This is a comma-separated list of the following options:
   [FlameGraph](https://github.com/brendangregg/FlameGraph) script. This is
   a collection of call stacks, where each line is a semicolon separated list
   of frames followed by a counter.
+  - `calltree[=C]`  - produce call tree in HTML format. C is counter type: 'samples' or 'total'.  
+  - `backtrace[=C]` - produce backtrace tree HTML format. C is counter type: 'samples' or 'total'.
```
apangin: Why two different options? One "tree" with optional "--reverse" (as was before) looks cleaner and more logical.
It does not work anyway if two options specified at the same time.
Since there is already too many dump options, I'm thinking about changing the set of booleans to one enum.

src/flameGraph.cpp
```diff
@@ -445,13 +558,30 @@ void FlameGraph::printFooter(std::ostream& out) {
...
+    out << TREE_HEADER;
+    if(type == BACK_TRACE){
+      out << "<ul>Backtrace view, total [sample/counter]: " << total << "</ul>\n";        
+    } else {
+      out << "<ul>Call tree view, total [sample/counter]: " << total << "</ul>\n";
+    }
+    out << "<ul class=\"tree\">\n";
+    out << "<button type='button' onclick='treeView(0)'>++</button><button type='button' onclick='treeView(1)'>--</button>\n";
+    out << "<input type='text' id='search' value='' size='35' onkeypress=\"if(event.keyCode == 13) document.getElementById('searchBtn').click()\">\n";
+    out << "<button type='button' id='searchBtn' onclick='search()'>search</button>\n";
```
apangin: This can be probably included to TREE_HEADER constant, not to pollute the code too much. If required, the header can be customized with sprinf, like with SVG_HEADER.

src/flameGraph.cpp
```diff
@@ -479,7 +609,58 @@ double FlameGraph::printFrame(std::ostream& out, const std::string& name, const
...
+            {
+                childtotal +=  itr->second._total;
+            }
+	    if(type == CALL_TREE) {
+                snprintf(_buf, sizeof(_buf),
+                "<li><a href=\"#\">[%d] %.2f%% %lld self: %.2f%% %lld</a><span style=\"color: #%06x\"> %s</span>",
+                depth, pairs[i].second._total * _pct,pairs[i].second._total,(pairs[i].second._total-childtotal)* _pct,(pairs[i].second._total-childtotal),color,full_title.c_str()); 
+            }else {
+               snprintf(_buf, sizeof(_buf), 
+               "<li><a href=\"#\">[%d] %.2f%% %lld</a><span style=\"color: #%06x\"> %s</span>", depth, pairs[i].second._total * _pct,pairs[i].second._total,color,full_title.c_str());
```
apangin: I'd suggest changing `style="color: xxx"` to `class="paletteName"`, so that
1) the output will be shorted;
2) colors will be customizable in css.
Color are better now, but still a little bit acid.

apangin:
As to `++` button - I was actually thinking about a different way to expand subtrees.
Probably there is no need to expand the whole tree at once, but rather a given subtree.
E.g. when clicking on `+` while holding Ctrl (or may be Alt or Shift) - the subtree will expand starting from the current node.
Alternatively you may choose to automatically expand subtree until the first node with 2 or more children - like VisualVM and JMC do.

apangin:
One more tiny UI improvement would be to decrease indent on each tree level, otherwise deep stack will go too far to the right.

README.md
rpulle: Can please revisit this. We want to have both calltree and backtrace in every run.  Users can select what they want.

Backtrace is important for analysis and then if we give reverse option, flame graph is not generated in normal. So I feel better to have both for this feature.

Now both "calltree" and "backtrace" both work in same UI. I have fixed this.

src/flameGraph.cpp
rpulle: Yes, I have moved as suggested.

src/flameGraph.cpp
rpulle: Yes,  I am using class="paletteName" now.

src/flameGraph.cpp
```diff
@@ -489,23 +712,51 @@ int FlameGraph::selectFrameColor(std::string& name) {
...
-        return yellow.getColor();
+        return palette ? yellow.getColor() : YELLOW;
     } else if ((int)name.find('/') > 0 || ((int)name.find('.') > 0 && name[0] >= 'A' && name[0] <= 'Z')) {
         // Java regular method
-        return green.getColor();
+        return palette ? green.getColor() : GREEN;
     } else {
         // Other native code
-        return red.getColor();
+        return palette ? red.getColor() : RED;
```
apangin: It looks confusing when the function returns either a color or enum value.
I've pushed the refactoring that will help to use palette names without ambiguous arguments and without extra `switch`.

rpulle:
@apangin  I have submitted changes as per review comments. I have removed backtrace option now. As long as I can generate it with a stop command.

src/flameGraph.cpp
```diff
@@ -484,6 +636,73 @@ double FlameGraph::printFrame(std::ostream& out, const std::string& name, const
     return framewidth;
 }
 
+void FlameGraph::printTreeHeader(std::ostream& out, long total) {
+    char buf[sizeof(TREE_HEADER) + 256];
+    std::string treeName = "Calltree";
+    if(_reverse){
+        treeName = "Backtrace";
+    } 
+    sprintf(buf, TREE_HEADER, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse,_reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, treeName.c_str(), total, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse, _reverse);
```
apangin: This looks odd :)
Are all these `_reverse` parameters still needed? Even if you plan to merge multiple tree views later, I'm sure JS functions can be tuned to work only with elements from the right context.

src/arguments.cpp
```diff
@@ -86,6 +87,9 @@ Error Arguments::parse(const char* args) {
         } else if (strcmp(arg, "flamegraph") == 0 || strcmp(arg, "svg") == 0) {
             _dump_flamegraph = true;
             _counter = value == NULL || strcmp(value, "samples") == 0 ? COUNTER_SAMPLES : COUNTER_TOTAL;
+        } else if (strcmp(arg, "calltree") == 0) {
```
apangin: Do you mind if we rename the option to "tree"? Just to type fewer letters in the command line :)
Also profiler.sh can automatically select this option if the filename ends with .html, similarly to how we select flamegraph output if the filename ends with .svg.

rpulle: I have changed calltree to tree and submitted changes.

rpulle:
@apangin  I have done changes for _reverse from cout and calltree to tree.

rpulle:
@apangin  sure, can you please check the latest changes?
I have done both 1 and 2.

apangin:
Exactly what I meant. Thanks!

apangin:
It has finally gone into master. Thank you for the contribution!

rpulle:
You are welcome and thank you for quick action on this.

PR #126, opened 2018-05-31
Fix svg command line flags in example

dsyer:
The text says "Specify `-o svg`" but the example command neglects to do that. This change fixes that ombission.


dsyer:
I see that now. It didn't seem to work for me. Probably user error.

PR #130, opened 2018-06-06
self % does not make sense for backtrace

rpulle:
@apangin  sorry for another small change.  Its a one line change but let me know if you want to merge or add it by your self. It got slipped when I removed explicit backtrace option.


apangin:
Why do you count self time manually at all instead of using Trie._self field?

rpulle:
you are right, I am using _self now.

apangin:
I've committed the fix.

PR #135, opened 2018-06-27
Pass JAVA_HOME variable to test scripts

jchipmunk:
Hi,

I found the issue:
```
$ make test
test/smoke-test.sh
+ '[' -z '' ']'
+ echo 'JAVA_HOME is not set'
JAVA_HOME is not set
```
when dynamically calculated `JAVA_HOME` variable is not passed to test scripts, like `smoke-test.sh` and `alloc-smoke-test.sh`.

This PR adds export of `JAVA_HOME` variable to the environment of subsequently executed commands.


apangin:
Thank you for the fix!

PR #137, opened 2018-06-30
macOS native symbols

apangin:
- Parse all Mach-O symbols, including Fat binaries
- Get native call trace by walking stack manually


vlsi:
Just letting you know: it builds and works in macos 10.13.5 / Java 1.8.0u152
This is great!

apangin:
Thanks for confirming it works! Given positive feedback from early adopters, I'm merging this into master.

PR #142, opened 2018-07-13
Improve JVM thread names support

KirillTim:
This PR brings better JVM thread names reporting.
- VMStructs symbols are loaded at the same time when AGCT symbols do: it fix bug when no thread names was reported by perf engine.
- Thread names storage is updated on every thread start and thread end: we can get names not only for active at the moment of dump threads , but also for all threads that had already stopped.
- I slightly changed thread output format to `[<thread Name> tid=<thread id>]`, because we need to distinguish different threads with the same name and to understand if dump contains thread labels or not.

### DIFF
--- a/src/concurrencyUtil.cpp
+++ b/src/concurrencyUtil.cpp
@@ -1,0 +1,8 @@
+#include "concurrencyUtil.h"
+
+void initLock(pthread_mutex_t* lock) {
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+    pthread_mutex_init(lock, &attr);
+}
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -21,6 +21,7 @@
 #include "frameName.h"
 #include "vmStructs.h"
 #include "threadNames.h"
+#include "profiler.h"
 
 
 FrameName::FrameName(bool simple, bool dotted, bool use_thread_names) :
@@ -32,8 +33,9 @@ FrameName::FrameName(bool simple, bool dotted, bool use_thread_names) :
     _saved_locale = uselocale(newlocale(LC_NUMERIC_MASK, "C", (locale_t)0));
 
     if (use_thread_names) {
-        updateAllKnownThreads();
-        _thread_names = ThreadNames::_instance.getNames();
+        ThreadNames storage = Profiler::_instance.threadNames();
+        storage.updateAllKnownThreads();
+        _thread_names = storage.getNames();
     }
 }
 
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -20,82 +20,30 @@
 #include <string.h>
 #include "frameName.h"
 #include "vmStructs.h"
+#include "threadNames.h"
 
 
 FrameName::FrameName(bool simple, bool dotted, bool use_thread_names) :
     _cache(),
     _simple(simple),
-    _dotted(dotted),
-    _thread_count(0),
-    _threads(NULL) {
+    _dotted(dotted) {
 
     // Require printf to use standard C format regardless of system locale
     _saved_locale = uselocale(newlocale(LC_NUMERIC_MASK, "C", (locale_t)0));
 
-    if (use_thread_names && VMThread::available()) {
-        initThreadMap();
+    if (use_thread_names) {
+        updateAllKnownThreads();
+        _thread_names = ThreadNames::_instance.getNames();
     }
 }
 
 FrameName::~FrameName() {
-    jvmtiEnv* jvmti = VM::jvmti();
-    for (int i = 0; i < _thread_count; i++) {
-        jvmti->Deallocate((unsigned char*)_threads[i]._name);
-    }
-    free(_threads);
-
     freelocale(uselocale(_saved_locale));
 }
 
-void FrameName::initThreadMap() {
-    JNIEnv* env = VM::jni();
-    jclass threadClass = env->FindClass("java/lang/Thread");
-    if (threadClass == NULL) {
-        return;
-    }
-    jfieldID eetop = env->GetFieldID(threadClass, "eetop", "J");
-    if (eetop == NULL) {
-        return;
-    }
-
-    jvmtiEnv* jvmti = VM::jvmti();
-    jthread* thread_objects;
-    if (jvmti->GetAllThreads(&_thread_count, &thread_objects) != 0) {
-        return;
-    }
-
-    _threads = (ThreadId*)calloc(_thread_count, sizeof(ThreadId));
-
-    // Create a map [OS thread ID] -> [Java thread name] backed by a sorted array
-    for (int i = 0; i < _thread_count; i++) {
-        VMThread* vm_thread = (VMThread*)(uintptr_t)env->GetLongField(thread_objects[i], eetop);
-        jvmtiThreadInfo thread_info;
-        if (vm_thread != NULL && jvmti->GetThreadInfo(thread_objects[i], &thread_info) == 0) {
-            _threads[i]._id = vm_thread->osThreadId();
-            _threads[i]._name = thread_info.name;
-        }
-    }
-
-    qsort(_threads, _thread_count, sizeof(ThreadId), ThreadId::comparator);
-    jvmti->Deallocate((unsigned char*)thread_objects);
-}
-
 const char* FrameName::findThreadName(int tid) {
-    int low = 0;
-    int high = _thread_count - 1;
-
-    while (low <= high) {
-        int mid = (unsigned int)(low + high) >> 1;
-        if (_threads[mid]._id < tid) {
-            low = mid + 1;
-        } else if (_threads[mid]._id > tid) {
-            high = mid - 1;
-        } else {
-            return _threads[mid]._name;
-        }
-    }
-
-    return NULL;
+    NamesMap::iterator it = _thread_names.find(tid);
+    return it == _thread_names.end() ? NULL : it->second.c_str();
 }
 
 const char* FrameName::cppDemangle(const char* name) {
@@ -211,10 +159,10 @@ const char* FrameName::name(ASGCT_CallFrame& frame) {
             int tid = (int)(uintptr_t)frame.method_id;
             const char* name = findThreadName(tid);
             if (name != NULL) {
-                return name;
+                snprintf(_buf, sizeof(_buf), "[%s tid=%d]", name, tid);
+            } else {
+                snprintf(_buf, sizeof(_buf), "[tid=%d]", tid);
             }
-
-            snprintf(_buf, sizeof(_buf), "[thread %d]", tid);
             return _buf;
         }
 
--- a/src/perfEvents.h
+++ b/src/perfEvents.h
@@ -32,7 +32,6 @@ class PerfEvents : public Engine {
     static PerfEvent* _events;
     static PerfEventType* _event_type;
     static long _interval;
-    static bool _is_active;
 
     static bool createForThread(int tid);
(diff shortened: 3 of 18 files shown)


apangin:
Thank you! The changes sound good. I'll review them shortly.

src/perfEvents.h
```diff
@@ -39,6 +41,10 @@ class PerfEvents : public Engine {
     static void installSignalHandler();
     static void signalHandler(int signo, siginfo_t* siginfo, void* ucontext);
 
+    static void setActive(bool value) {
+        __atomic_store_n(&_is_active, value, __ATOMIC_SEQ_CST);
```
apangin:
```
src/perfEvents.h:45: error: ‘__ATOMIC_SEQ_CST’ was not declared in this scope
src/perfEvents.h:45: error: ‘__atomic_store_n’ was not declared in this scope
```

src/threadNames.cpp
```diff
@@ -0,0 +1,48 @@
+#include "threadNames.h"
+#include "profiler.h"
+
+ThreadNames ThreadNames::_instance;
+
+void ThreadNames::update(jthread thread) {
+    MutexLocker ml(_lock);
```
apangin: Mutex is only needed to protect `_storage`; there is no need to perform other possibly heavy operations like `GetThreadInfo` under lock. I'd suggest to keep critical section as short as possible.

src/threadNames.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return;
+    }
+    _storage[vm_thread->osThreadId()] = std::string(info.name);
+    jvmti->Deallocate((unsigned char *)info.name);
+}
+
+NamesMap ThreadNames::getNames() {
+    MutexLocker ml(_lock);
+    std::map<int, std::string> result = _storage;
+    return result;
```
apangin: Is it really required to copy the whole map? We need to update the map only while profiler is running, but read it only after profiler is stopped.

src/threadNames.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+void updateAllKnownThreads() {
```
apangin: Please avoid global functions. You can make it a static member of a class instead.

src/vmEntry.cpp
```diff
@@ -68,6 +69,8 @@ void VM::init(JavaVM* vm, bool attach) {
     _jvmti->SetEventNotificationMode(JVMTI_ENABLE, JVMTI_EVENT_COMPILED_METHOD_LOAD, NULL);
     _jvmti->SetEventNotificationMode(JVMTI_ENABLE, JVMTI_EVENT_COMPILED_METHOD_UNLOAD, NULL);
     _jvmti->SetEventNotificationMode(JVMTI_ENABLE, JVMTI_EVENT_DYNAMIC_CODE_GENERATED, NULL);
+    _jvmti->SetEventNotificationMode(JVMTI_ENABLE, JVMTI_EVENT_THREAD_START, NULL);
+    _jvmti->SetEventNotificationMode(JVMTI_ENABLE, JVMTI_EVENT_THREAD_END, NULL);
```
apangin: Though thread start/end are not performance critical, I'd suggest not to generate events while profiler is idle anyway.

src/threadNames.h
```diff
@@ -0,0 +1,33 @@
...
+typedef std::map<int, std::string> NamesMap;
+
+class ThreadNames {
+  private:
+    pthread_mutex_t _lock;
+    NamesMap _storage;
+    jclass _threadClass;
+    jfieldID _eetop;
+  public:
+    static ThreadNames _instance;
```
apangin: I'm afraid keeping the same map throughout the lifetime of the application may have some undesired side effects. Java thread name will be tied to the given PID forever, even if PID reused weeks later for some native thread. Also this may lead to excessive memory usage if threads keep borning and dying.

KirillTim: I agree. I will create new map and register callbacks every time when profiler is started and unregister callbacks when it stops.

src/threadNames.cpp
KirillTim: It is still possible that thread related events will be called after profiler is stopped, so I think it is better to copy the whole map and be safe.

src/threadNames.cpp
KirillTim: OK.
The idea was to use one lock for storage and for `_eetop` updates.

src/threadNames.cpp
KirillTim: ОК

src/vmEntry.cpp
KirillTim: ОК

src/perfEvents.h
apangin: gcc 4.6.4
And it is not even the oldest one - I know a customer who runs async-profiler on Linux 2.6.39 with ten years old glibc :) Enterprise IT is tough (sigh)... so I'd prefer to keep requirements and dependencies as low as possible.

KirillTim: Oh...
OK, I will think about it.
Is it possible to downgrade gcc to target (4.6.4) version on CI? Travis use `g++ (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4` and it give us a false confidence when check is passed.

apangin: I switched travis configuration to `precise` dist that comes with gcc 4.6.3.

KirillTim: Thanks.
One more question: can we compile with `-std=gnu++0x` and use features listed as available here: https://gcc.gnu.org/gcc-4.6/cxx0x_status.html ?

apangin: Why? I don't see what problem this will solve, but this may definitely introduce compatibility issues. Even if you have `__atomic` intrinsics, `onThreadStart/onThreadEnd` functions are still not thread-safe.

KirillTim:
Could you please take a look at updated version.
Is it also possible to install java8 on CI server? :)

src/concurrencyUtil.cpp
```diff
@@ -0,0 +1,8 @@
+#include "concurrencyUtil.h"
+
+void initLock(pthread_mutex_t* lock) {
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+    pthread_mutex_init(lock, &attr);
+}
```
apangin: Global function again.
Why not encapsulate `pthread_mutex_t` and related operations in a Mutex class, and place it together with MutexLocker in `mutex.h/cpp`?

src/frameName.cpp
```diff
@@ -20,82 +20,32 @@
...
-    _threads(NULL) {
+    _dotted(dotted) {
 
     // Require printf to use standard C format regardless of system locale
     _saved_locale = uselocale(newlocale(LC_NUMERIC_MASK, "C", (locale_t)0));
 
-    if (use_thread_names && VMThread::available()) {
-        initThreadMap();
+    if (use_thread_names) {
+        ThreadNames storage = Profiler::_instance.threadNames();
```
apangin: Doesn't this create a redundant copy of the whole map?

src/profiler.h
```diff
@@ -234,6 +224,16 @@ class Profiler {
         _instance.addRuntimeStub(address, length, name);
     }
 
+    static void JNICALL ThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
+        _instance._thread_names.update(thread);
+        _instance._engine->onThreadStart();
```
apangin: `_engine` can be concurrently deleted, right?

src/threadNames.cpp
```diff
@@ -0,0 +1,52 @@
+#include "threadNames.h"
```
apangin: New source files should include standard license header.

src/threadNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+jfieldID ThreadNames::_get_eetop(JNIEnv *env) {
+    MutexLocker ml(_cached_eetop_lock);
+    if (_cached_eetop == NULL) {
+        jclass threadClass = env->FindClass("java/lang/Thread");
+        if (threadClass != NULL) {
+            _cached_eetop = env->GetFieldID(threadClass, "eetop", "J");
+        }
+    }
+    return _cached_eetop;
+}
```
apangin: Looks like overcomplication. `jfieldID` needs to be initialized just once, and it will be valid throughout the JVM lifetime. It is safe to initialize `jfieldID` at the agent attach time or at the profiler start time without extra lock.

src/threadNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (eetop == NULL) {
+        return;
+    }
+    VMThread *vm_thread = (VMThread *)(uintptr_t)env->GetLongField(thread, eetop);
+    jvmtiThreadInfo info;
+    jvmtiEnv *jvmti = VM::jvmti();
+    if (jvmti->GetThreadInfo(thread, &info) != 0 || info.name == NULL) {
+        return;
+    }
+    MutexLocker ml(_storage_lock);
```
apangin: Narrowing the scope of the mutex:
```
{
    MutexLocker ml(_storage_lock);
    _storage[vm_thread->osThreadId()] = std::string(info.name);
}
```

src/concurrencyUtil.cpp
KirillTim: My bad. I personally prefer top level function (being a Kotlin fan). Will fix.
The question is: should I just put two classes `Mutex` (with one static method) and `MutexLocker` in one `mutex.h/cpp` ?

src/profiler.h
KirillTim: Hm, yes. Is it ok to add yet another mutex to give thread safe access to `_engine`? Or can we just ignore possible data race and check that `_engine != NULL` ?

src/threadNames.cpp
KirillTim: OK. Who should be the copyright owner for new files? 
While it is told 
> You _must_ cause any modified files to carry prominent notices stating that You changed the files; and ...

> You _may_ add Your own copyright statement to Your modifications and ...

It seems strange for me to be an owner for two files in you project, though.

src/threadNames.cpp
KirillTim: OK

src/threadNames.cpp
KirillTim: Forget about this language feature, will fix.

src/frameName.cpp
KirillTim: The idea was to be sure, that `ThreadNames` instance won't be cleared. 
In fact, if `start` for recently attached agent will be called just after `stop` and before `dump*` we have problem much bigger than loss of `ThreadNames`. Should it be fixed somehow?
However, I think we should copy the whole map in `ThreadNames::getNames()`.

apangin: I understand why you create a copy, but you create two copies in a row (`_instance.threadNames()` + `storage.getNames()`), which seems redundant.

I didn't get what problem you are talking about. start/stop/dump commands are executed under profiler's lock.

src/profiler.h
apangin: Data race is not a problem. But unfortunately `_engine != NULL` does not help, since `_engine` can be deleted by another thread right after this check,

src/threadNames.cpp
apangin: The above statements are about redistribution.
When contributing changes, you typically grant non-exclusive copyright to the project owner. I'll formulate contribution instructions in a document a bit later.

So, please leave ` * Copyright 2018 Andrei Pangin` line in the header. If you wish, you *may* add an extra copyright line with your name.

apangin:
I've finally managed to revise and integrate this pull request while leaving your original idea.
Feedback is welcome!

PR #159, opened 2018-10-03
Use absolute path for output file

SylvainJuge:
If we parameters as described in documentation, we don't get the expected output file.
Using an absolute path seems to be required to behave as expected.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -204,6 +204,10 @@ if [[ "$OUTPUT" == "" ]]; then
     fi
 fi
 
+# using absolute path name is required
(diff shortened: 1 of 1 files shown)


apangin:
Makes sense. Thank you!

apangin:
Unfortunately I have to revert this commit since it breaks profiler on Alpine Linux, see #162.
Busybox version of `readlink` does not work for non-existent files.

apangin:
@SylvainJuge Yes, I think this will serve as a workaround.

PR #160, opened 2018-10-03
improve profiler.sh for work with symlink

QIvan:
Hi! I use your profiler with `update-alternatives` and now when I try to connect to process I get something like 
```
/my/path/to/symlink/build/jattach: No such file
```

this fix works for me and work with usual usage as well.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -86,7 +86,7 @@ function abspath() {
 
 
 OPTIND=1
-SCRIPT_DIR=$(dirname $0)
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
 JATTACH=$SCRIPT_DIR/build/jattach
 PROFILER=$(abspath $SCRIPT_DIR/build/libasyncProfiler.so)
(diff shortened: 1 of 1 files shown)


profiler.sh
```diff
@@ -86,7 +86,7 @@ function abspath() {
 
 
 OPTIND=1
-SCRIPT_DIR=$(dirname $0)
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
```
apangin: Have you checked this on macOS? AFAIK, `readlink -f` does not work there.

QIvan: damn... you're right the flag  `-f` not in the POSIX standard... sorry

QIvan:
what do you think about to add a function `readlink` to `profiler.sh`  with code from here https://stackoverflow.com/a/1116890 ?

apangin:
A few lines above is [abspath](https://github.com/jvm-profiling-tools/async-profiler/blob/master/profiler.sh#L78) function.

PR #165, opened 2018-10-14
fix thread names format

KirillTim:
In our inner fork we use '[' and ']' to understand that thread dump has named threads(if first frame starts with '[' and ends with ']' and it isn't  `[unknown]` it should be thread name)
And when users decide to use upstream version they have funny bugs, like this: https://youtrack.jetbrains.com/issue/IDEA-200453
I also add thread id information for JVM threads do distinguish different threads with same name and to match profiler info to other collected thread related data, if any.

### DIFF
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -211,10 +211,10 @@ const char* FrameName::name(ASGCT_CallFrame& frame) {
             int tid = (int)(uintptr_t)frame.method_id;
             const char* name = findThreadName(tid);
             if (name != NULL) {
-                return name;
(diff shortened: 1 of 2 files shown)


apangin:
It's OK to change the format now, but please note that relying on such peculiarity in the plugin is not a good idea anyway. There are other cases of brackets in a frame name, e.g. `[frame_buffer_overflow]` or  `[jvmtiError NN]`. I plan to add more signals shown as pseudo-frames.

Why not just check whether argument line contains `threads`?

apangin:
I would say metadata is a must if you plan to export/import profiles.

But I have no objections to this particular PR. Merged.

PR #167, opened 2018-10-19
fix compilation for macos 10.8.5

KirillTim:
IntelliJ IDEA supports macs 10.8.5+ so we need to compile agent library on very old build agents.
This small change fix compilation for us.

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -33,7 +33,17 @@
 #define ARRAY_SIZE(arr)  (sizeof(arr) / sizeof(arr[0]))
 
 #ifndef htonll
+
+#ifdef __APPLE__
+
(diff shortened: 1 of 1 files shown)


KirillTim:
Is it better to just use 
```
#include <libkern/OSByteOrder.h>
#define bswap_64(x) OSSwapInt64(x)
```
instead of 
```
#include <byteswap.h>
``` 
than?

I'm not a C++ man so could you, please, suggest better solution, if any?

apangin:
Yes, given that we already have `#ifdef __APPLE__`, it's possible to include `<libkern/OSByteOrder.h>` there and implement platform-specific `htonll` under this ifdef.

I've committed the fix - please check if it helps.

PR #177, opened 2018-11-13
Allow to run without changing perf_event_paranoid

jpbempel:
If we get a EACCES when opening perf event because we are not a root user and perf_event_paranoid is 
set to 2, try to exclude kernel from perf event. It allows to profile without changing globally the setting
(security issue)


apangin:
It's a good thing to allow profiler work even with restricted perf_events. However, I'm a bit concerned that the absense of kernel stack will now be silently hidden from a user. It'll be hard to understand that something needs to be configured in order to get full profile.

jpbempel:
Agreed. We can print a Warning for the degraded mode.
Would you want an explicit option as well?

nitsanw:
This is a cool enhancement, I’m with @apangin that if people can fix this we should be notifying rather than silently degrading. How about going for a suggestive error on failure + a user only mode(can also add kernel only while you are at it) and follow the perf convention for these options?

apangin:
@nitsanw Yes, I like this approach.

nitsanw:
This will be enable getting some interesting profiling overhead baselines :-)

jpbempel:
still have an issue with JDK 7 which is weird. Works fine with JDK8, still investigating

jpbempel:
Done! Since a while I have been c++ 🙄

apangin:
I got this finally merged. Thanks again!

jpbempel:
Thanks for the review and merge 👍

PR #194, opened 2019-01-21
Added additional case for Troubleshooting section

doom369:
Extended troubleshooting section for https://github.com/jvm-profiling-tools/async-profiler/issues/193


apangin:
Thanks for updating the documentation.

PR #209, opened 2019-03-19
Distinguish inlined frames

ghaug:
Async profiler doesn't distinguish frame types (compiled/inlined/interpreted), which is relevant information for developers. With this change the frame type is tracked at least for top frames.

The basic principle used to gather the frame type is simple: If the pc of topmost java frame is within the interpreter, the top frame is interpreted. If not, the `methodID` of the topmost java frame is looked up in the code cache model. Then, the `vframe` stack trace obtained from the VM is walked till the `methodID` is found. This is the topmost physical frame. The ones possibly found above this are virtual frames of inlined methods.

I had to change the code in several places in order to obtain the necessary information.

The most profound change was necessary to the code cache model. The Hotspot VM has only one code cache for nmethods as well as for other blobs, e.g. the interpreter. Even with the segmented code cache introduced with Java 9 IIRC, it is still one continuous memory region. I've changed async profiler's model accordingly. There is now only one VM code cache where each blob can either be an `nmethod` which has a `methodID` or a non-nmethod which has a `cstring` name. The VM code cache is now ordered to allow for binary searching. Each new blob has to be inserted in the appropriate location. The new class is called `VmCodeCache.` I've renamed the model for libraries to `NativeLib` to be a little bit more clear. Both inherit from a base class called `CodeUnit.`

`recordSample` has been extended in the way described above. Frame type information is stored as an offset to the bci. Therefore, I had to slightly change the hashing mechanism to take the bci into account. This is done for all frames, even though it is not absolutely required right now, but I think it might be the basis for a worthwhile future extension.

Frames are now color coded in the flame graph (svg) output. Interpreted top frames appear in magenta, while inlined frames are blue (aqua). Frames of the same method may occur in different types (e.g. interpreted first and inlined later on), what is represented by a gradient. In order to transfer the information from the traces to the flame graph generator, the `Trie` class has been extended. The type of a frame is coded by a sigle character extension to its name. This now works independently from the annotation mechanism.

There are a few extensions to the simple output functions (summary, flat and traces) which are self-explanatory. jfr output still works but doesn't take advantage of the additional information, yet.

Finally, I've changed the color of "all" and threads dummy frames in the flamer graph (svg) output to white (gray) and salmon respectively, while the new ERROR frames continue to appear in red. This, however, is actually unrelated to the rest of the change and it can be easily be reverted or modified to any other color scheme.

Please see the attachment for an example flame graph:
[out.svg.zip](https://github.com/jvm-profiling-tools/async-profiler/files/2982879/out.svg.zip)

### DIFF
--- a/src/allocTracer.cpp
+++ b/src/allocTracer.cpp
@@ -35,7 +35,7 @@ volatile u64 AllocTracer::_allocated_bytes;
 
 
 // Resolve the address of the intercepted function
-bool Trap::resolve(NativeCodeCache* libjvm) {
+bool Trap::resolve(NativeLib* libjvm) {
     if (_entry != NULL) {
         return true;
     }
@@ -128,7 +128,7 @@ Error AllocTracer::start(Arguments& args) {
         return Error("VMStructs unavailable. Unsupported JVM?");
     }
 
-    NativeCodeCache* libjvm = Profiler::_instance.jvmLibrary();
+    NativeLib* libjvm = Profiler::_instance.jvmLibrary();
     if (!(_in_new_tlab.resolve(libjvm) || _in_new_tlab2.resolve(libjvm)) ||
         !(_outside_tlab.resolve(libjvm) || _outside_tlab2.resolve(libjvm))) {
         return Error("No AllocTracer symbols found. Are JDK debug symbols installed?");
--- a/src/allocTracer.h
+++ b/src/allocTracer.h
@@ -35,7 +35,7 @@ class Trap {
     Trap(const char* func_name) : _func_name(func_name), _entry(NULL) {
     }
 
-    bool resolve(NativeCodeCache* libjvm);
+    bool resolve(NativeLib* libjvm);
     void install();
     void uninstall();
 
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -17,9 +17,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include "codeCache.h"
+#include "vmEntry.h"
 
-
-void CodeCache::expand() {
+void CodeUnit::expand() {
     CodeBlob* old_blobs = _blobs;
     CodeBlob* new_blobs = new CodeBlob[_capacity * 2];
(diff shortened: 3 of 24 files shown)


ghaug:
Sorry, I can't reproduce the failed test. The tests complete successfully on Linux as well as macOS with me. I tried with several different JDKs, including JDK7.

ghaug:
Finally found the problem. The last change fixes the failing test.

simonis:
Hi Andrei,
did you had a chance meanwhile to look at this enhancement?
Is there anything we can do in order to support you pushing this change?
Please let us know if you have any questions or see any problems.
Thank you and best regards,
Volker

apangin:
Excuse me for a delay. I haven't looked deeply at the code yet, but even the basic experiments demonstrate some oddities in the generated flamegraphs.

Here is a fragment of the flamegraph I've got when profiling a GUI application on macOS.
flamegraph
It's OK that some frames are "unknown", since there are no symbols for them, but the strange things are:

- Some unknown frames are red, and the other are brown. This is even more strange since brown means kernel frames, and there is no kernel profiling on macOS at all. I suppose this might me caused by the CodeCache refactoring, and I would prefer not to mix the new features with the refactoring that is not directly related to the proposed feature.
- The frame names are trimmed incorrectly, the number of samples should appear only in a popup hint.

Unfortunately, there is no a dedicated quality engineer in the project, and I don't really wish to become one :) so, only the thoroughly tested code with no visible regressions could be accepted. Thank you for your understanding.

ghaug:
Hi Andrei,
sorry for being late as well! I've updated and rebased the change. There were in fact issues with the color of unknown frames and formatting of method names which are both fixed now.
Would you mind to try it and take a look at the code? We still believe that that identifying inlined and interpreted frames is a useful feature.
Best,
Gunter

apangin:
Thank you for fixing the issues. I'll give this another try.

parttimenerd:
This PR is superseded by the PR #545.

PR #230, opened 2019-08-01
keep all method displays consistent

huangchen007:
Keep all method displays consistent.
Parameter _dotted should be passed all through.
Hard code need be removed like below:
'javaClassName(symbol->body(), symbol->length(), _simple, true)

### DIFF
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -139,13 +139,13 @@ const char* FrameName::name(ASGCT_CallFrame& frame) {
 
         case BCI_SYMBOL: {
             VMSymbol* symbol = (VMSymbol*)frame.method_id;
-            char* class_name = javaClassName(symbol->body(), symbol->length(), _simple, true);
+            char* class_name = javaClassName(symbol->body(), symbol->length(), _simple, _dotted);
             return strcat(class_name, _dotted ? "" : "_[i]");
         }
 
         case BCI_SYMBOL_OUTSIDE_TLAB: {
             VMSymbol* symbol = (VMSymbol*)((uintptr_t)frame.method_id ^ 1);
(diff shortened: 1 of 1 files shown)


apangin:
Any comments about the reasons of this PR?

huangchen007:
I recommented it.

apangin:
Pull Requests without a valid justification (even so small and simple as this one) cannot be accepted, sorry. A change should either solve some problem or introduce an objective improvement. I don't see what problem the proposed change addresses, but it can definitely break some existing configurations that rely on the particular output format.

huangchen007:
I felt so sorry to hear that. Hard code "true" definitely should be changed to member variable '_dotted'. I think I had explant it very clearly. You 'll see if you took a look at the code. It's small but it absolutely should be corrected.

apangin:
Surely I've seen this code, since I wrote it. `true` was put there intentionally to force Java language notation for the pseudo-frames that denote allocated objects or locked instances.
Sorry if I didn't get your point, but there was no explanation from your side except "I recommend it".

OK, let me show why your proposed change is broken.

1. The title says the purpose of this change is to "keep all method displays consistent" (whatever it means). But the modified code has nothing to do with **method** names at all. It changes the way how the **class** names are printed. This relates only to allocated objects in `alloc` profiling mode
and the locked objects in `lock` mode.

2. There are two meaningful notations for Java classes: either Java language notation, e.g. `java.lang.Object[]` or a JVM descriptor, e.g. `[Ljava/lang/Object;`. Your PR changes the notation to `java/lang/Object[]` which does not make sense in either context.

3. Furthermore, this change irreversibly breaks existing deployments that already rely on the current format. This didn't break alloc-smoke-test only by a lucky chance, since the test checks for `.` which stands for any symbol in a regexp context.

PR #238, opened 2019-08-29
Adding simple Dockerfile and docker-compose.yml examples

gustavotemple:


apangin:
Contributions to the project are generally welcome, but in this case I don't see how the users of async-profiler will benefit from the proposed change. This looks like a specific setup for a particular environment, and it can't be treated as a recommended way to use async-profiler or docker.

PR #243, opened 2019-08-31
Update README.md

gustavotemple:


apangin:
Sorry for closing this again, but IMO this does not add much value to the existing documentation.
README already explicitly says that the profiler uses HotSpot-specific API and works on OpenJDK, Oracle JDK and other Java runtimes based on HotSpot JVM. It is not intended to work with OpenJ9 or JVMs that do not support Dynamic Attach (not sure what JVMs exactly you were referring to).

Issue #237, opened 2019-08-26
Wall clock profiler optimizations


guser21:
Hi,

I have been using your wall clock profiler recently and it has proved itself really well in my humble tests. The only problem is when I try to use it for a process which has more than 2000 threads the frequency of the sampling per thread drops dramatically and to reach reasonable results one needs to lower the sampling interval so much that the profiler becomes a huge overhead. Unfortunately we have many processes on our production which have an order of 10k threads.  

To solve this problem I suggest 2 approaches. 

1) One idea would be to have a regex which would enable the users filter threads that needs to be profiled. Usually after doing a broad profiling we needed to focus on a single threadpool, which might be particularly interesting. 

The way I imagine how this could be implemented is to filter out the pids of the threads, which thread names match to a specific regex and to send the signals only to them. The major problem that I see in this approach is that usually threadpools are dynamic and during the profiling there might be new threads created or deleted. So one might need to go through all those threads periodically during sampling to ensure that we have all the pids of the interesting threads. 

What do you think about this approach ?  How can we escape the additional overhead of periodically matching the threads to the regex (or do we really need to do that)? 

2) The second idea is a little bit more complicated.  Most of the threads, most of the time do nothing, so it is kind of useless to walk through their stack every time. The idea is to count the number of context switches per thread and before sending the the new signal, check if it has changed dramatically (dramatically because I believe that the signal handling itself is a context switch) . This way we can avoid useless stack walking for idle threads and increase the sampling frequency for working threads. 

There are a couple of problems with this approach. 
Firstly, it is not crystal clear to me how this could be implemented. I don't really know if the approach with the context switches would work. The other part is how we would get the stacks for the idle threads efficiently (concurrency issues).  
Secondly the overhead of the profiler will become dynamic as the less threads you have the more data you generate as the time spent on useless stackwalking will be spent on the stackwalking of working threads. 
Third the distribution will get skewed as the working threads will get more sampled than they would have been sampled. 

What are your ideas about this approach? Would this work and may be you could give clues how this could be implemented efficiently? 

So does any of these ideas make any sense and would they be useful? 
I would be love to hear your feedback on these ideas before I actually start to implement them. 

P.S. 
I am a student and still a newbie in the JVM world, so if something is complete nonsense correct me ))).  

@apangin

apangin:
Thank you for the detailed feedback and for the constructive ideas!
[...]

PR #246, opened 2019-09-11
Filter threads

guser21:
@apangin This is a prototype implementation of the feature that I was talking about in the #237 issue. The implementation is immature and has a lot of place for improvement but I still wanted to get your opinion.  

I have tested the following scenarios and it behaved as it was supposed to
1) Changing thread name
2) Exiting from running threads
3) New thread creation
4) High load of threads ~ 10k threads 

I decided to divide the changes into small pr-s so if all goes well I will implement it for the CPU profiler as well.

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -383,7 +383,7 @@ bool Profiler::addressInCode(const void* pc) {
             return true;
         }
     }
-    
+
     // This can be some other dynamically generated code, but we don't know it. Better stay safe.
     return false;
 }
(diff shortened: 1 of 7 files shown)


guser21:
It seems to me that the CI is using  g++ version 4.3 and there were no regular expressions implemented back then. Could you change the CI g++ version or should I somehow change my implementation ?

guser21:
Here is a simpler version where it basically checks if the given word is a substring of the thread name. Still a simple and premature implementation but I would appreciate your feedback ))

gsafaryan:
@apangin Did you have a chance to look at the new changes )))?

guser21:
@apangin Did you have chance to take a look at the changes )))?

apangin:
> Did you have chance to take a look at the changes )))?

Not yet, sorry. I'm a bit busy this week, but I'll definitely take a look later.

src/profiler.cpp
```diff
@@ -519,11 +520,49 @@ void Profiler::updateThreadName(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
...
     }
 }
 
+void Profiler::addThreadToFilteredList(int osThreadId, const char *name) {
+    MutexLocker lock(_filtered_tids_mutex);
+
+    if (_filter_threads == NULL) return;
+
+    if (strstr(name, _filter_threads)) {
+        _filtered_tids.push_back(osThreadId);
```
apangin: I'd say `vector` is not a proper structure for holding filtered tids.
One thread id may appear in `vector` multiple times, and the `vector` may grow infinitely.

src/profiler.cpp
```diff
@@ -519,11 +520,49 @@ void Profiler::updateThreadName(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
...
+        if (vm_thread != NULL && jvmti->GetThreadInfo(thread, &thread_info) == 0) {
+            _filtered_tids.erase(std::remove(_filtered_tids.begin(), _filtered_tids.end(), vm_thread->osThreadId()),
+                                 _filtered_tids.end());
+            jvmti->Deallocate((unsigned char *) thread_info.name);
+        }
+    }
+}
+
+std::vector<int> Profiler::getFilteredTidsRange(int from, int count) {
+    std::vector<int> threads_for_this_tick;
```
apangin: Why create a new vector each time?
Can't we walk over _filtered_tids?

Furthermore, with this approach the order of traversed tids will break if `_filtered_tids.size()` changes between iterations.

src/profiler.cpp
```diff
@@ -660,6 +705,11 @@ Error Profiler::stop() {
     }
 
     _engine->stop();
+    if (_filter_threads != NULL) {
+        MutexLocker lock(_filtered_tids_mutex);
+        _filter_threads = NULL;
+        _filtered_tids.clear();
```
apangin: Resetting profiler structures is typically done in `start()` rather than in `stop()`, since profiling can be later resumed (i.e. stop == pause).

src/wallClock.cpp
```diff
@@ -94,3 +109,44 @@ void WallClock::timerLoop() {
...
+        }
+    }
+}
+
+void JNICALL WallClock::threadSetNameTrap(JNIEnv* env, jobject obj, jstring name) {
+    VMThread* vm_thread = VMThread::fromJavaThread(env, obj);
+    int threadId = vm_thread->osThreadId();
+
+    const char* threadName = env->GetStringUTFChars(name, NULL);
+    Profiler::_instance.addThreadToFilteredList(threadId, threadName);
```
apangin: Need to make sure that threadName != NULL

guser21:
Sorry, I have been busy lately, in the upcoming week will submit the new version.

llasram:
What's the status on this?  If @guser21 is no longer interested in this feature, I'm happy to take a swing at getting the PR into an acceptable state.

felixbarny:
The work on filtering continues in https://github.com/jvm-profiling-tools/async-profiler/tree/thread-filter

apangin:
Closing the PR as obsolete.

PR #249, opened 2019-09-27
Fix allocation profiler `alloc_size` for JDK 7-9

pingtimeout:
According to the signatures of both `send_allocation_in_new_tlab` and
`send_allocation_in_new_tlab_event`, the previous code was taking the
tlab size as the requested allocation size.

This commit fixes this by using the correct parameter.

Tested with https://github.com/pingtimeout/humongous-allocator.  Without
the fix, 8 instances of `java.lang.Thread_[i]` were reported with a `-o
collapsed=total` above 12M, which cannot be correct.  The same happened
with other `java.lang` basic data types.

### DIFF
--- a/src/allocTracer.cpp
+++ b/src/allocTracer.cpp
@@ -77,13 +77,13 @@ void AllocTracer::signalHandler(int signo, siginfo_t* siginfo, void* ucontext) {
     // PC points either to BREAKPOINT instruction or to the next one
     if (frame.pc() - (uintptr_t)_in_new_tlab._entry <= sizeof(instruction_t)) {
         // send_allocation_in_new_tlab_event(KlassHandle klass, size_t tlab_size, size_t alloc_size)
-        recordAllocation(ucontext, frame.arg0(), frame.arg1(), false);
+        recordAllocation(ucontext, frame.arg0(), frame.arg2(), false);
     } else if (frame.pc() - (uintptr_t)_outside_tlab._entry <= sizeof(instruction_t)) {
         // send_allocation_outside_tlab_event(KlassHandle klass, size_t alloc_size);
         recordAllocation(ucontext, frame.arg0(), frame.arg1(), true);
     } else if (frame.pc() - (uintptr_t)_in_new_tlab2._entry <= sizeof(instruction_t)) {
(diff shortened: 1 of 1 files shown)


apangin:
Measuring total TLAB size is intended.

The allocation sampling mechanism provides no means to count the exact number of allocations or the exact amount of allocated bytes.
Current async-profiler behavior is aligned with JFR which measures so called "allocation pressure". If I record the size of individual objects instead, the relative allocation pressure would be misleading. E.g. if there are three types of objects: A (100 bytes), B (200 bytes) and C (300 bytes), and the number of allocations of each object is the same, the allocation ratio shown by the profiler should be 1:2:3, which is the case with the current approach, but not with your suggested approach.

apangin:
> Without the fix, 8 instances of java.lang.Thread_[i] were reported with a -o collapsed=total above 12M, which cannot be correct. The same happened with other java.lang basic data types.

One more clarification: async-profiler shows the estimated total size of *all* allocated objects of the given class, not only the size of the *sampled* objects. On the long distance, TLAB size would give [statistically] more precise results. However, I admit that with a small number of allocated objects (like `java.lang.Thread` in your case) the collected results may seem incorrect.

pingtimeout:
Ah I see, thanks @apangin for the explanation.  I have a related question for you then.  I have tried to make async-profiler record humongous object allocations to precisely know where they are created, so that a codebase can become more G1GC friendly.  To do that, I needed some `if(alloc_size > threshold)` code and this this.

The code is still very hacky at this point.  But eventually I was thinking about submitting it here.  Do you have any recommendation for how I should implement this feature?  In my current version, that `if` logic is in `AllocTracer::recordAllocation`.  But that cannot work very accurately if the tlab size is used instead of the actual object size.

apangin:
Humongous allocations are not related to TLAB at all. Async-profiler makes difference between allocations in new TLAB and outside TLAB. Humongous allocations are always outside.

Alternatively, you can profile humongous allocations directly, using async-profiler's feature to intercept arbitrary native functions. In this case, you'll need to intercept `G1CollectedHeap::humongous_obj_allocate`:

```
./profiler.sh -e _ZN15G1CollectedHeap22humongous_obj_allocateEmh --sync-walk ...
```

I've just added `--sync-walk` feature to make Java stack traces more accurate when profiling VM runtime functions. It's not yet tested well, use carefully.

Issue #250, opened 2019-09-27
v1.6 - SIGSEGV on libasyncProfiler.so ElfParser::parseFile


nobeh:
I am trying to run async-profiler from a host Linux on a docker container running a Java application:

Host:

```
$ uname -a
Linux nobeh13 5.0.0-29-generic #31-Ubuntu SMP Thu Sep 12 13:05:32 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ perf version
perf version 5.0.21

```

Example Dockerfile:

```
FROM ubuntu:18.04

SHELL [ "/bin/bash", "-c" ]

RUN \
    apt-get update && \
    apt-get install -y openjdk-8-jdk-headless && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

RUN \
    groupadd --gid 1000 bnobakht && \
    useradd --uid 1000 --gid 1000 --shell /bin/bash bnobakht && \
    chown -R bnobakht:root /opt && \
    mkdir /app && \
    chown -R bnobakht:root /app

ENV APP_ROOT /app

COPY ./highcpu/target/highcpu.jar ${APP_ROOT}/highcpu.jar
COPY ./run-highcpu.sh ${APP_ROOT}
COPY ./run.sh ${APP_ROOT}
COPY ./profile.sh ${APP_ROOT}

USER bnobakht
WORKDIR ${APP_ROOT}
CMD [ "./run-highcpu.sh" ]

```

For the docker container, I've tried docker base images `openjdk:8-jdk`, `ubuntu:19.04` and `ubuntu:18.04` (just to rule out potential issue on `perf_event_open` versions).

The following is how async-profiler is run from the host:

```bash
$ docker top $(docker ps | grep highcpu |cut -d ' ' -f 1)
$ sudo /opt/async-profiler/profiler.sh -d 10 -e cpu -o collapsed -f ./perf.txt JPID
```
and how the docker container is run:

```bash
$ docker run -it --rm --security-opt=seccomp:unconfined --volume $(pwd)/perf:/perf:rw --volume /opt/async-profiler:/opt/async-profiler:rw
```

The following is the header of the core dump:

```
# JRE version: OpenJDK Runtime Environment (8.0_222-b10) (build 1.8.0_222-8u222-b10-1ubuntu1~18.04.1-b10)
[...]

PR #251, opened 2019-10-04
Update README.md

gustavotemple:


apangin:
It wouldn't be quite correct to say that the bug occurs only on 5.0.0-2x.
E.g. in #250 the same problem was seen on Linux 5.3.0-13-generic.
I'd prefer to leave only the link to the original issue, where all the details are available firsthand.

PR #254, opened 2019-10-11
Document the placeholders in the file name

dmitry-timofeev:
Document the placeholders for the PID and the timestamp
in a filename.

See https://github.com/jvm-profiling-tools/async-profiler/blob/master/src/arguments.cpp#L189-L190

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,8 +324,10 @@ Example: `./profiler.sh -t 8983`
 * `--title TITLE`, `--width PX`, `--height PX`, `--minwidth PX`, `--reverse` - FlameGraph parameters.  
 Example: `./profiler.sh -f profile.svg --title "Sample CPU profile" --minwidth 0.5 8983`
 
-* `-f FILENAME` - the file name to dump the profile information to.  
-Example: `./profiler.sh -o collapsed -f /tmp/traces.txt 8983`
+* `-f FILENAME` - the file name to dump the profile information to.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -324,8 +324,10 @@ Example: `./profiler.sh -t 8983`
 * `--title TITLE`, `--width PX`, `--height PX`, `--minwidth PX`, `--reverse` - FlameGraph parameters.  
 Example: `./profiler.sh -f profile.svg --title "Sample CPU profile" --minwidth 0.5 8983`
 
-* `-f FILENAME` - the file name to dump the profile information to.  
-Example: `./profiler.sh -o collapsed -f /tmp/traces.txt 8983`
+* `-f FILENAME` - the file name to dump the profile information to.
+`%p` in the file name is expanded to the PID of the target JVM;
+`%t` — to the timestamp.
```
dmitry-timofeev: It might be worth clarifying at which time the time stamp is recorded. Is that correct that the timestamp corresponds to the time the command is received (i.e., if file option is passed with the start command, then it corresponds to the start time, if with the stop — then to the stop time)? If that's correct, how'd you put it in a concise way?

apangin: You are right, '%t' is expanded when the profiler parses arguments, i.e. at the time of command execution.

apangin:
Thank you for updating the documentation! I merged the PR.

dmitry-timofeev:
Great, thank you for the tool and the review!

PR #265, opened 2019-11-12
Update README.md

RoySunnySean007:
Provide solution to install debug info on yum based linux distribution, refer to https://github.com/jvm-profiling-tools/async-profiler/issues/264.


apangin:
Thanks, I've merged the README updates.

PR #270, opened 2019-11-27
Fix syntax for Docker seccomp, and add note about --cap-add

tomgoren:
When attempting to run the async-profiler in a Docker context, I encountered issues with the syntax for `seccomp` as specified in the README.

From https://docs.docker.com/engine/reference/run/:

> use `--security-opt seccomp=unconfined` when adding capabilities


apangin:
Makes sense. Thank you!

PR #276, opened 2019-12-09
port async profiler to Power Linux little endian (linuxppc64le)

ghaug:
Hi Andrei,
here is one more pull request. This one is a port of async profiler to LinuxPPC64 little endian. I've tested it extensively and found no issues. The code is mostly straight forward. There is an additional section in arch.h and the new stackFrame_ppc64.cpp file. With these additions the perf event based traces and the wall clock trace as well as the lock trace work perfectly.

Unfortunately the alloc trace requires changes in shared code. The problem is that hotspot itself makes use of SIGTRAP on PPC. Therefore the signal handlers have to be chained, i.e.  the agent's handler has to call the VM's handler once it has been determined that the SIGTRAP is not caused by a trap installed by the agent. There now is an additional function declared in os.h and implemented in the respective os_*.cpp files to get the address of a signal handler currently installed. This is called from allocTracer.cpp and the handler's address is stored in a static variable.

Moreover, the trap instruction may not be installed at the very beginning of a function on PPCLE (see arch.h for an explanation why). I had to introduce the macro `BREAKPOINT_OFFSET` which is 0 on all existing platforms and 2 on PPCLE. `Trap::install()` implemented in allocTracer.cpp now installs the trap with the appropriate offset.

### DIFF
--- a/src/allocTracer.cpp
+++ b/src/allocTracer.cpp
@@ -29,7 +29,7 @@ Trap AllocTracer::_outside_tlab("_ZN11AllocTracer34send_allocation_outside_tlab_
 // JDK 10+
 Trap AllocTracer::_in_new_tlab2("_ZN11AllocTracer27send_allocation_in_new_tlab");
 Trap AllocTracer::_outside_tlab2("_ZN11AllocTracer28send_allocation_outside_tlab");
-void (*AllocTracer::_next_handler)(int, siginfo_t *, void *) = 0;
+void (*AllocTracer::_next_handler)(int, siginfo_t *, void *) = NULL;
 
 u64 AllocTracer::_interval;
 volatile u64 AllocTracer::_allocated_bytes;
@@ -53,7 +53,8 @@ bool Trap::resolve(NativeCodeCache* libjvm) {
     return false;
 }
 
-// Insert breakpoint at the very first instruction
+// Insert breakpoint - usually at the very first instruction
+// but some architectures require an offset (see arch.h)
 void Trap::install() {
     if (_entry != NULL) {
         _saved_insn = *(_entry + BREAKPOINT_OFFSET);
@@ -143,11 +144,11 @@ Error AllocTracer::start(Arguments& args) {
     _interval = args._interval;
     _allocated_bytes = 0;
 
+    // remember the previously installed handler to call it next (only once)
     if (_next_handler == NULL) {
         _next_handler =  (void (*)(int, siginfo_t*, void*)) OS::getSignalHandler(SIGTRAP);
     }
     OS::installSignalHandler(SIGTRAP, signalHandler);
-
 
     _in_new_tlab.install();
     _outside_tlab.install();
--- a/src/allocTracer.cpp
+++ b/src/allocTracer.cpp
@@ -29,6 +29,7 @@ Trap AllocTracer::_outside_tlab("_ZN11AllocTracer34send_allocation_outside_tlab_
 // JDK 10+
 Trap AllocTracer::_in_new_tlab2("_ZN11AllocTracer27send_allocation_in_new_tlab");
 Trap AllocTracer::_outside_tlab2("_ZN11AllocTracer28send_allocation_outside_tlab");
+void (*AllocTracer::_next_handler)(int, siginfo_t *, void *) = 0;
 
 u64 AllocTracer::_interval;
 volatile u64 AllocTracer::_allocated_bytes;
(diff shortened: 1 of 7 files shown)


src/stackFrame_ppc64.cpp
```diff
@@ -0,0 +1,104 @@
...
+    // C1 epilogue and critical section (posX)
+    //        3821**** add     r1,r1,xx
+    // pos3   xxxxxxxx
+    // pos2   1000e1eb ld      r31,16(r1)
+    // pos1   a603e87f mtlr    r31
+    //        xxxxxxxx
+    //        2000804e blr
+    if (*((unsigned int *)pc + 1) == 0xebe10010 && *((unsigned int *)pc + 2) == 0x7fe803a6 ||
+        *((unsigned int *)pc + 0) == 0xebe10010 && *((unsigned int *)pc + 1) == 0x7fe803a6 ||
+        *((unsigned int *)pc - 1) == 0xebe10010 && *((unsigned int *)pc + 0) == 0x7fe803a6) {
```
apangin: Is accessing memory before current PC always safe?
I can't tell about ppc, but on x86 and ARM there were situations where the address [pc-1] was not mapped.

apangin:
Hi Gunter,
I appreciate your work. The changes overall look fine. However, my biggest concern is how to deal with ppc port further.

I don't have the appropriate hardware and have no resources to test and maintain the port. At the same time I don't want this to become a bottleneck in delivering new features and bug fixes.

For instance, I'm about to commit a fix that detects more frames with non-standard layout in order to reduce the amount of `unknown_Java` samples. This change affects pop-frame logic, and will probably break your code.

I *may* accept PRs related to other architectures, but only with an inherent disclaimer regarding no guarantees to support, release or even keep those ports compilable.

ghaug:
Hi Andrei,
I fully understand your point. We would be fine with a disclaimer along the lines you describe above. We just discussed this within our team and came up with the idea of a "friendly fork" in parallel to our [SapMachine](sapmachine.io) project. We would pull automatically from the async profiler repository and build on PPC (and possibly other platforms) nightly. We would of course provide necessary fixes upstream or - in case you're interested - you could take a look at the build results. Moreover, we could provide a "last known good" version for the convenience of our users.

src/stackFrame_ppc64.cpp
ghaug: Not always, but as long as PC is within the VM's code cache it is. However, it it's a good idea to simply check if (PC-1) is within the code cache to be really robust. I'll add this check.

src/allocTracer.cpp
```diff
@@ -137,6 +144,10 @@ Error AllocTracer::start(Arguments& args) {
     _interval = args._interval;
     _allocated_bytes = 0;
 
+    // remember the previously installed handler to call it next (only once)
+    if (_next_handler == NULL) {
```
toktarev, line 148: sorry, just curiosity

why don't you use nullptr here ?

shipilev:
I used patch from this PR without a problem on ppc64le. Is there any blockers to get this integrated? I think we can collectively support it, and Andrei can add some caveat language about "community support" for ppc64le, so it does not block the releases?

ghaug:
I still would be happy to update the pull request if there is any interest to integrate it.

apangin:
Hi Gunter and Alexey,
As I've told before, I don't mind integrating the change, with the caveat that I will not maintain the port in future. I can't promise that I will even keep the sources buildable, sorry.

If you are OK with that, please rebase the PR on top of v2.0 branch - it will become the main development branch in the nearest future.

I have also a couple of comments regarding the code.
1. Make sure there is no access to the memory which is not absolutely known to be accessible.  
    I mean, `*((unsigned int *)pc - 1)`. A simple workaround is to check that `pc - sizeof(int)` belongs to the same page. In the worst case, the sample will be just rejected, but this is not a problem, I suppose.
2. I would warn against modifying the stack / registers in the signal handler, as it is very dangerous and may potentially cause hard-to-debug crashes. Stack walking code must be read-only from the application's point of view. Any changes made to the thread's state to facilitate stack walking, should be rolled back before return from the signal handler.

shipilev:
> As I've told before, I don't mind integrating the change, with the caveat that I will not maintain the port in future. I can't promise that I will even keep the sources buildable, sorry.

That is absolutely fine with me! (I meant to set up some CI jobs for it anyway).

ghaug:
Closing this PR,  there is a new one now.

Issue #297, opened 2020-02-04
Json output format of tree view


rpulle:
The HTML output is very useful, however we are also looking to implement a profile comparison feature on our UI. We can convert html tree to Json and then do a diff,  however, JSON output from async_profiler is the right way we believe.

rpulle:
@apangin  Can I create a pull request for the json output format?

apangin:
Thank you for your interest in async-profiler and for enhancement proposal.

async-profiler already supports a number of output options, including
 - `collapsed` stacktraces - a text format which is easy to parse and to convert into other views;
 - `jfr` - the comprehensive binary format.

I'm not a fan of adding more application-specific dump formats into the profiler core. Actually, I'd like to do the opposite: make the core agent generate the single comprehensive binary output, and provide the number of external plugins that will convert this output to more formats. I'll be happy to accept contributions of new plugins, once the converters logic is moved out of the C++ agent.

See #134 for the related discussions.

rpulle:
@apangin  Thank you for your quick response, by the time I saw this message, I already created a PR.
It will be really helpful for our current profile comparison feature if you can consider this. We are using async profiler to analyze production incidents and is very useful, and a comparison is very well needed for our use case.

apangin:
I appreciate pull requests in general. But the proposed changes confirmed my concerns that the new format is application-specific. I mean, other users will hardly get an immediate benefit from it, and there is no available software to parse and visualize it. I wouldn't like to complicate the existing code base or add an overhead of maintaining the feature in future.

As I've told before, it's fine to have various output generators as plugins, but I believe this is not a job of the profiler core.

> Writing a converter may be an alternative, however it requires additional compute resources to convert to a different format

I would argue that it's rather a benefit to convert the output to any desired format *offline* at any convenient time in a standalone process. In contranst, built-in generators run in the context of the JVM that executes application code.

apangin:
Considering the above reasons, I'm inclined to refuse the pull request. Sorry.
I hope, standalone coverters will solve the problem.

PR #298, opened 2020-02-05
Call tree json output 

rpulle:
@apangin This PR is to get json output for call tree and back trace view. https://github.com/jvm-profiling-tools/async-profiler/issues/297

image


apangin:
See the discussion in #297

PR #303, opened 2020-02-24
 profiler.sh: work on minimal systems

cfstras:
- use POSIX sh instead of bash (not available on most embedded linux, alpine, etc)
- replace some bash-isms by sh alternatives
- some shellcheck warnings
- replace usage of readlink by cd & pwd (not everybody has coreutils installed)

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -74,7 +74,7 @@ jattach() {
     if [ $RET -ne 0 ]; then
         if [ $RET -eq 255 ]; then
             echo "Failed to inject profiler into $PID"
-            if [ "$UNAME_S" = "Darwin" ]; then
+            if [ "$(uname -s)" = "Darwin" ]; then
                 otool -L "$PROFILER"
             else
                 ldd "$PROFILER"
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for your interest in async-profiler. The goal of the changes makes sense, but unfortunately the proposed implementation doesn't seem to work well enough.

I gave it a quick try, and found a number of issues:

- `profiler.sh -v` does not work;
- if the agent library fails to load, the script silently terminates with no error message;
- `--title` option does not accept special characters, e.g. `<`

Please bear in mind that the project has no a dedicated QA engineer, so authors are required to thoroughly test their changes before submitting, ideally both on Linux and macOS. Thanks.

cfstras:
Yes, I didn't take the time to test it thoroughly. Sorry.
I'll see whether I can find the time to update this. In the meantime: Your profiler is awesome, and is the only thing that works reliably on the slow hardware I'm currently working with. Thank you!

ceeaspb:
Hi, resolving this would be useful, how to proceed (apart from it needing more testing)? Thanks

`profiler.sh -v` replace `$PID` with `${PID:-}`

Agent library fails to load - remove the `set -e`

We need to escape the `&`s in sed processing of `--title`.

apangin:
@ceeaspb I'll be glad to review the updated pull request. Feel free to finish the work, if you wish.

As a mid-term goal, I plan to get rid of the shell script entirely, replacing it with a binary executable to be installed to `/usr/bin`.

cfstras:
@apangin got around to making the fixes. let me know if there's anything else amiss.

cfstras:
thanks 🙏

apangin:
Seems to work fine now. Thank you for the contribution 👍

Issue #162, opened 2018-10-05
Problems running in Alpine Docker container 'file must not be empty'


StephenFox1995:
Docker Image: `openjdk:8u171-jdk-alpine3.8`

It seems when I run the profiler script it doesn't output anything and within the target process the following message is logged 'file must not be empty'.

When I run `make test` this is the output, it also seems to output 'file must not be empty':

```
bash-4.4# make test
test/smoke-test.sh
+ '[' -z /usr/lib/jvm/java-1.8-openjdk ']'
++ dirname test/smoke-test.sh
+ cd test
+ '[' Target.class -ot Target.java ']'
+ FILENAME=/tmp/java.trace
+ JAVAPID=283
+ + sleep /usr/lib/jvm/java-1.8-openjdk/bin/java 1Target

+ ../profiler.sh -f /tmp/java.trace -o collapsed -d 5 283
file must not be empty
file must not be empty
+ kill 283
+ assert_string 'Target.main;Target.method1 '
+ grep -q 'Target.main;Target.method1 ' /tmp/java.trace
grep: /tmp/java.trace: No such file or directory
+ exit 1
make: *** [Makefile:57: test] Error 1
```

I'm not sure if this is a limitation with running async-profiler in the docker container / alpine or async-profiler itself.

Any help would be appreciated.

PR #307, opened 2020-03-14
resolve relative output file name

vision57:
Output file path is not resolved correctly if relative path is specified.


apangin:
What problem does the proposed patch solve?
The issue with relative output paths has been already fixed in #159.
Also note that `abspath` does not work here for the reasons described in #162.

vision57:
It should be ok in your way.

PR #316, opened 2020-04-26
Ensure code blob exists properly

KirillTim:
Check that code blob is not removed to avoid returning NULL
when another code blob loaded at the similar address range
that was used by removed one

Contributed by @semoro


apangin:
Makes sense.
Thanks, the fix went into master.

PR #317, opened 2020-04-26
Report how many samples were dropped because of 'CallTraceSample' overflow

KirillTim:
On heavy workloads, e.g when profiling IntelliJ or Kotlin compiler many samples can be dropped. We need to understand the impact of the loss.

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -937,11 +937,13 @@ void Profiler::dumpCollapsed(std::ostream& out, Arguments& args) {
 
     FrameName fn(args, args._style, _thread_names_lock, _thread_names);
     u64 unknown = 0;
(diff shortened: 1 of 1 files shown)


apangin:
There are some things in the proposed PR that I find questionable:
1. The changes affect only `dumpCollapsed`. In general, the output should be consistent between all available formats.
2. Duplicated frames seem like a workaround for the specific parser. Even in a threaded output, an error frame does not belong to any particular thread.
3. I believe the best way to report dropped stack traces is to not drop the traces in the first place. It's better to fix the original problem. Will address this in the next major release.

apangin:
As of v2.0, async-profiler does not limit the number of call traces.

PR #323, opened 2020-05-18
Add an argument to collect C stack from ucontext

QuantumBear:
Related to this issue https://github.com/jvm-profiling-tools/async-profiler/issues/321

For some old Linux kernel (CentOS 6.5 Final), native stack can't get retrieved correctly from perf event. So I add an argument to swtich to walk stack using ucontext.

Tested on CentOS 6.5 Final, kernel version 2.6.32-696.3.2.el6.x86_64.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -87,6 +87,7 @@ const size_t EXTRA_BUF_SIZE = 512;
 //     height=PX       - FlameGraph frame height
 //     minwidth=PX     - FlameGraph minimum frame width
 //     reverse         - generate stack-reversed FlameGraph / Call tree
+//     ucontext        - collect C stack frames using ucontext
 //
 // It is possible to specify multiple dump options at the same time
 
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -162,7 +162,9 @@ Error Arguments::parse(const char* args) {
                 if (value == NULL || value[0] == 0) {
                     return Error("event must not be empty");
                 }
-                _event = value;
+                _walk_stack_frame = (strcmp(value, "stackframe") == 0);
+                if (!_walk_stack_frame)
+                    _event = value;
 
             CASE("interval")
                 if (value == NULL || (_interval = parseUnits(value)) <= 0) {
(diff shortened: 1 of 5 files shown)


apangin:
There is already `cstack` option for choosing how to walk native stack. I think it's better to reuse this one. One more similar option would be confusing.

Meanwhile I'll get an image of the above OS version for testing.

QuantumBear:
Antother choice is check the value of "u64 nr = ring.next();" in PerfEvents::getNativeTrace, if nr <= 1 then switch to use ucontext, then we don't need modify the arguments.

apangin:
I installed exactly the same kernel version and could reproduce the problem.

However, what I found interesting, is that the problem happens only with `PERF_COUNT_SW_CPU_CLOCK` counter (`event=cpu`). With `event=cycles`, call stacks are fine, including the kernel stack.

So, the suggested alternative is to use `event=cycles`. It is semantically the same as `event=cpu`, but provides higher accuracy and also shows kernel stacks.

QuantumBear:
Thanks for your suggestion, very helpful!

But after to use event=cycles, I still find some weird stack trace (using my HeapOOM example)

```
SpinPause;apic_timer_interrupt_[k];smp_apic_timer_interrupt_[k];local_apic_timer_interrupt_[k];hrtimer_interrupt_[k];__run_hrtimer_[k];tick_sched_timer_[k];update_process_times_[k];run_local_timers_[k];hrtimer_run_queues_[k] 1
```
Most stack trace is correct, so I think is ok to swtich to cycles for daily use.

Another phenomenon is that when using event=cycles, the result is quite different. When I use async-profiler to profile my Hadoop job, it shows jit complier consumes lots of CPU time when profiling using event=cpu or event=itimer. After swtich to event=cycles, almost no compiler consumption reports. 

In my understanding, event=cycles and event=cpu are sematically the same, both of them are reporting CPU consumption. But why the huge difference?

apangin:
With the default interval, `cycles` produces much more samples than `cpu` or `itimer`. This might explain the difference. Try larger interval (`-i 10ms`) and increase the frame buffer size (`-f 10m`).

Also, if you run the application under unprivileged user, make sure `kernel.perf_event_mlock_kb` sysctl option is large enough. 8KB per thread is required. E.g. if your application has 500 threads, `perf_event_mlock_kb` should be at least 4000 (the default value is 516). This is not needed when the application runs under `root`.

QuantumBear:
Thanks for your help, I think stack walking using ucontext is useless now.

PR #338, opened 2020-08-04
Allow thread filter to be enabled dynamically via Java API

retronym:
- Set the `_enabled` flag after adding a thread to the filter.
  - Check the `_enabled` flag on each sample, rather than just
    when commencing the sampling loop.

### DIFF
--- a/src/threadFilter.cpp
+++ b/src/threadFilter.cpp
@@ -94,6 +94,7 @@ void ThreadFilter::add(int thread_id) {
     if (!(__sync_fetch_and_or(&word(b, thread_id), bit) & bit)) {
         atomicInc(_size);
     }
+    _enabled = true;
 }
 
 void ThreadFilter::remove(int thread_id) {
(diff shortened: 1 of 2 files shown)


retronym:
Should we also set `_enabled = false` in `ThreadFilter::remove` when the last thread is removed?

retronym:
My use case is having the new JMH integration start profiling, then having the workload-under-profile itself add one or more threads to the thread filter.

retronym:
This might be the wrong approach -- perhaps we should be passing `filter=0` when starting the profiler (which we can do even without knowing which threads will run the workload).

apangin:
Right, changing `_enabled` in runtime would change the semantics.
If no threads added to the filter, then the profiler should sample nothing, while with the filter disabled it samples everything.

Javadoc do `addThread` / `removeThread` mentions that
```
     * 'filter' option must be enabled to use this method.
```

To use filters, just add `filter` option to the agent arguments (without `=0`).

PR #345, opened 2020-09-01
Clean up debug symbols instructions and troubleshooting

yuzawa-san:
I ran into an odd case where I tried to alloc profile a running process, the debug symbols were missing, I installed the debug symbols, the profiling continued to fail. This was because the JDK was upgraded when I installed the debug symbols. I added to the troubleshooting section to spare others the investigative work.

I broke the debug symbol information into a separate subsection for anchoring/linking. I also added note in the CPU section that the debug symbols are needed if one wanted resolve frames within libjvm.so. "Wall-clock profiling" section was promoted since it is not a subsection of "ALLOCATION profiling".

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,7 +499,7 @@ stack traces.
 ```
 No AllocTracer symbols found. Are JDK debug symbols installed?
 ```
-The OpenJDK debug symbols are an hard requirement for allocation profiling.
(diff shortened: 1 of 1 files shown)


apangin:
Thanks! I very much appreciate additions to the documentation.

PR #346, opened 2020-09-01
Specify JVM process by 'jps' application name

yuzawa-san:
I found the `jps` parameter helpful, so I extended it to use the application name as it would appear in the `jps` list.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -40,10 +40,12 @@ usage() {
     echo ""
     echo "<pid> is a numeric process ID of the target JVM"
     echo "      or 'jps' keyword to find running JVM automatically"
+    echo "      or the application's name as it would appear in the jps tool"
     echo ""
     echo "Example: $0 -d 30 -f profile.svg 3456"
     echo "         $0 start -i 999000 jps"
     echo "         $0 stop -o summary,flat jps"
+    echo "         $0 -d 5 -e alloc MyAppName"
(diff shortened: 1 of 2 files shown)


profiler.sh
```diff
@@ -199,10 +201,21 @@ while [ $# -gt 0 ]; do
...
+            if [ $# -eq 1 ]; then
+                # the last argument is the application name as it would appear in the jps tool
+                PID=$(jps -J-XX:+PerfDisableSharedMem | grep " $1$" | head -n 1 | cut -d ' ' -f 1)
+                if [ "$PID" = "" ]; then
+                    echo "No Java process '$1' could be found!"
+                fi
+            else
+                echo "Unrecognized option: $1"
+                usage
+            fi
```
apangin, line 218: Interesting suggestion! Looks useful.
I'd only exclude options starting with `-`
If there is an unrecognized option starting with `-`, immediately fail without looking for a process

yuzawa-san: I just added a `-*)` case. I would have normally done something bash-y with the `&& [[ ]]` but I saw https://github.com/jvm-profiling-tools/async-profiler/pull/303 was done. I believe what I have is still POSIX compliant since it passes on shellcheck.net

Since this was done

apangin:
Looks great! Thank you for the contribution.

Issue #93, opened 2018-03-09
Publish Jar (with embedded native agent) to Maven central


advancedxy:
Some ideas about the Java API:
1. published to maven central, so it can be easily added as dependency in other Maven based projects.
2. Jar contains libAsynProfiler.so(dylib) so that the jar can be used independently 
3. Java API execute() method supports 'file' argument (already committed)

apangin:
OK, it would be nice, if someone volunteered to help with this.

advancedxy:
I may help with jar packing and API in my spare time. But I have never published jars to central personally, that needs an experienced one

incubos:
I can do the publishing part.

advancedxy:
Some other issues popped during my experience with async-profiler, one feature I hope Java API has:
ability to specify dumped file path in API.
The rationale is that sometimes we want to defer the start of profiler: after loading other native library by calling `System.load` or similar, so the native library's symbol can be identified correctly. Then the profiler output can be dumped at the JVM exit. However it's impossible for Java API to specify an output file.

@apangin what do you think? If this seems ok. I will try to send a PR. And Let's keep this issue open for a while

apangin:
Java API returns profiling data as a String, then you do whatever you want with this String: write to a file etc.
Shutdown Hook can be used to stop profiler before VM exit.

advancedxy:
> Java API returns profiling data as a String, then you do whatever you want with this String: write to a file etc.

Yeah, I know, but that requires wrapper code to flush the output into file. If we can flush output to file, the execute API can be used as same as Java Agent(or Attach)'s options.

apangin:
OK. Now if `execute(command)` contains `file=` option, the output will be dumped directly to the given file.

advancedxy:
>OK. Now if execute(command) contains file= option, the output will be dumped directly to the given file.

Thanks, it works as expected after I check the master branch. 

However my old approach was persisting execute's argument into vmEntry's internal _agent_args: so it's dumped automatically at JVM exit.

apangin:
I would argue that Java API should not modify global arguments designed primarily for the agent loaded at start-up. At Java level an application can still use Java mechanisms such as Shutdown Hook.

advancedxy:
> I would argue that Java API should not modify global arguments designed primarily for the agent loaded at start-up. At Java level an application can still use Java mechanisms such as Shutdown Hook.

[...]

PR #365, opened 2020-11-03
Mavenizing async-profiler Java library

JigarJoshi:
#93 

Step 1: Mavenize the Java library <-- this PR takes care for this
Step 2: Create friendly Java API (take care of extracting and loading library .so)
Step 3: publish it to maven central


incubos:
Could you please temporarily publish the artifact from [your fork using JitPack](https://jitpack.io/#JigarJoshi/async-profiler) and test it (i.e. the library is loaded successfully)?

pom.xml
```diff
@@ -0,0 +1,70 @@
...
+                        <phase>package</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>/usr/bin/zip</executable>
+                            <arguments>
+                                <argument>-u</argument>
+                                <argument>target/async-profiler-${project.version}.jar</argument>
+                                <argument>build/libasyncProfiler.so</argument>
```
incubos, line 62: `async-profiler` supports [many platforms](https://github.com/jvm-profiling-tools/async-profiler#download) having prebuilt libraries and Maven artifact we are going to publish is expected to be used from any supported platform, so we have to find a way to pack all the binaries.

incubos (changes requested):
Multiplatform support is needed if we are going to publish a single artifact.

apangin:
Hi,
Sorry for the delay - that was a rather busy month for me.

First, please create PRs against v2.0 branch - it will become mainline soon. No new features are planned for 1.8 - it will remain only for bugfixes.

My main concern is how to package async-profiler agent binaries. As Vadim noticed, async-profiler supports many platforms, each requiring a separate binary. In future, the number of platforms will likely increase (think of Apple M1, for example). I don't like an idea of packaging all possible binaries into a single JAR, otherwise it'll get enormously large. Furthermore, I expect that some users will prefer to install async-profiler package with the OS-specific package manager like RPM, and some will build the profiler on their own. Therefore, I think it makes sense to create a separate Maven artifact with the API only, and separate artifacts with `.so` for each individual platform.

Side note: by default, `make` builds the profiler agent, `jattach` helper, API jar, and converters jar. When preparing `libasyncProfiler.so`, we may build just the agent library and skip other steps.

apangin:
Closing PR, since it has been superseded by the ongoing work within #93.

PR #386, opened 2021-01-22
Added a note about IntelliJ IDEA

flounder4130:


apangin:
Added the note with a minor edition. Thanks.

PR #402, opened 2021-03-22
port async profiler 2.0 to Power Linux little endian (linuxppc64le)

ghaug:
Hi Andrei,
here is a new pull request to port the async profiler 2.0 to ppc64le as discussed in the original pull request. It requires less changes in shared code, as chaining of signals is now already done.
However, two changes in shared code are still required. First, the trap instruction may not be installed at the very beginning of a function on PPCLE (see arch.h for an explanation why). I had to introduce the constant BREAKPOINT_OFFSET which is 0 on all existing platforms and 2 on PPCLE. Trap::assign() implemented in trap.cpp now installs the trap with the appropriate offset.
Second, we may never move SP on PPC. I have therefore introduced a bool constant CAN_MOVE_SP which is only false on PPC and prevents the respective attempts in profiler.cpp.


apangin:
Thank you for updating the PR.

I don't have an experience with PPC64, but I am uncertain if BREAKPOINT_OFFSET is added at the right place.

1. Traps are used not only for allocation profiling, but also for `--begin/--end` feature, where a user may supply an arbitrary function, or maybe even an arbitrary address.
2. What about hardware breakpoints? (I think PPC64 has them, too). If I specify `-e local_func`, won't the hardware breakpoint be set at the global entry point?

ghaug:
The reason for that BREAKPOINT_OFFSET trouble is that on PPC64le the first two instructions of a function set up the TOC, i.e. load a constant to a register. As the TOC is the same for a compilation unit, these two instructions are skipped when a function is invoked from within the same compilation unit. Therefore setting the TRAP at the very beginning of a function will only work if the function is invoked from a different compilation unit. Just to avoid confusion: interestingly calling conventions on PPC big endian are completely different.

To address your questions:
1. I'm aware of that and it is not an issue for arbitrary functions. Arbitrary addresses are fine, as long as they are function entries. I guess it's rare that a user supplies an arbitrary address within a function but we could add some documentation of the behaviour.
2. I'm not sure about HW BPs. However, I've tested this implementation and it works without issues. Setting BREAKPOINT_OFFSET to zero breaks the allocation trace as the respective function is called from within the same compilation unit.
3. That's a really good catch, I didn't think of that. It is probably rare, as you say and I would have to check the documentation to see if it can happen at all. But it's certainly better to make sure we unprotect the correct page. I'll update the change.

apangin:
Sorry if I've missed the point. Can you please clarify - have you checked that HW breakpoints work on ppc64le?
For example, does `-e G1CollectedHeap::humongous_obj_allocate` correctly show stack traces of humongous allocations? I ask, because this VM function seems to be called from the same compilation unit, but BREAKPOINT_OFFSET is not added to the address in `getBreakpoint`.

IMO, it would make sense to add the offset close to the symbol resolution rather than in `Trap.assign`:

- this should work both for hardware and software breakpoints;
- this will make possible to use raw addresses and symbols with explicit offsets (in these cases BREAKPOINT_OFFSET should not be added).

ghaug:
Ah, sorry! Now I understand. I'll check and update the PR accordingly.

ghaug:
Sorry for being late, I literally tried for weeks to get my hands on a PPC host where HW breakpoints work, but in vain. At least I can't get execution HW BPs to work and they don't in gdb or perf either. It may be a problem with the virtualization but that's just a guess. We've been in contact with IBM but to no avail so far.
The approach to add the offset during symbol resolution appears fine to me. The only little drawback would be the tiny little bit of accuracy we would loose in case a sample really occurred in the first two instruction when called from outside the compile unit. In my opinion that's acceptable, do you agree?

ghaug:
I've updated the PR according to your suggestion. I wasn't able to test it with WW BPs as described above, sorry. However, the the begin/end mechanism works, as well as the alloc trace, of course. Besides linux/ppc64le I've tested it on linux/x86_64 and MacOS/x86_64 and it doesn't break anything. I don't think it breaks anything on arm either.

apangin:
Hi @ghaug,
I'm going to publish a new release soon, and I thought it's a good chance to include ppc64le port there, too.
If PR is still relevant, can you please rebase it on top of up-to-date master? I've already pushed the required changes in shared code.
Thanks.

ghaug:
Thanks for the heads-up! Yes, it is still relevant to us. I'll try to rebase it tomorrow.

ghaug:
Hi Andrei,
I've updated the pull request. Unfortunately I messed up completely last time. When I tried to implement your suggestion to move the thing with the BP offset closer to symbol resolution I somehow managed to push a completely broken version without the actual change.
Now the BREAKPOINT_OFFSET is gone and it's implemented as you originally suggested. Implementation is much simpler now.
Smoketests work on Mac (x86) Linux x86 and Linux PPC64le. Unfortunately you'll have to take a look again.
Sorry for the mess,
Gunter

apangin:
Hi Gunter,
Thank you for updating the PR.
It looks like my recent commit with `BREAKPOINT_OFFSET` caused confusion, sorry about that. It was intentional change though. I realized that adding `offsetBPaddr` in `resolveSymbol` is not quite right either, because the breakpoint offset should not be added when the symbol offset is specified explicitly, e.g.

- `-e malloc` should resolve to `&malloc + 8`, while
- `-e malloc+16` should resolve to `&malloc + 16`, not 24.

So I came up with a solution which is somewhere between your first and the final implementation, and committed those changes in the shared code. I believe no other changes in the shared code are needed now, so I expect your PR will include
 - `arch.h`
 - `stackFrame_ppc64.cpp`
 - `Makefile`

Please correct me if I missed something.

apangin:
When trying to build on Ubuntu 21.04, I got the following error:
```
src/stackFrame_ppc64.cpp:26:51: error: invalid use of incomplete type ‘struct pt_regs’
   26 |     return (uintptr_t&)_ucontext->uc_mcontext.regs->nip;
```
Seems like we need to include `<signal.h>`.
Not sure why this is relevant only for ppc64.

ghaug:
Hi Andrei,
thanks for your patience and sorry again, that was some bad communication from my side :( And yes, you're correct only changes in these three files.
Regarding the inclusion of  <signal.h>: It's not required on SuSE/PPC64le. I don't have access to an Ubuntu machine. But anyway, including it won't hurt.

apangin:
Good. So would you please update the PR accordingly? Or better, create a new one with all changes squashed, as this PR seems already messed in some strange way.
I could do this myself, but it will be fair if your authorship remains in git history.
Thanks.

apangin:
BTW, I got access to Power virtual machines in Unicamp Minicloud, so I can occasionally verify that PPC port builds and runs.
This will not be "officially" supported platform though, but I added [Community supported builds](https://github.com/jvm-profiling-tools/async-profiler/#community-supported-builds) sections where you can put links to your builds, if you wish.

ghaug:
Yes, I'll create a new one, will be easier after this mess :)

apangin:
Obsoleted by #459

PR #413, opened 2021-04-05
Read log file via /proc/pid/root

Jongy:
This way, it works for containers as well.

Only case where it doesn't work is when `/tmp` is an absolute symlink (therefore `/proc/pid/root/tmp` will point to e.g `/tmp2` which is not available in the host). If `/tmp` is an absolute symlink and we're profiling a host process, it will, of course, work just fine (so we're not regressing it here).

Tested it manually by profiling a container with `-e cpu`. Without this change, `profiler.sh` exits with no output. With this change, I get the expected `[WARN] perf_event_open failed: Operation not permitted` lines.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -77,27 +77,33 @@ jattach() {
     if [ $RET -ne 0 ]; then
         if [ $RET -eq 255 ]; then
             echo "Failed to inject profiler into $PID"
-            if [ "$(uname -s)" = "Darwin" ]; then
+            if [ "$UNAME_S" = "Darwin" ]; then
                 otool -L "$PROFILER"
             else
                 LD_PRELOAD="$PROFILER" /bin/true
             fi
         fi
 
+        # try both LOG and LOG_HOST, in case we don't have permissions for LOG_HOST but LOG
+        # is in our namespace.
         if [ -f "$LOG" ]; then
             cat "$LOG" >&2
             rm "$LOG"
+        elif [ "$LOG" != "$LOG_HOST" ] && [ -f "$LOG_HOST" ]; then
+            cat "$LOG_HOST" >&2
+            rm "$LOG_HOST"
         fi
         exit $RET
     fi
 
-    rm -f "$LOG"
+    rm -f "$LOG_HOST" "$LOG"
     mirror_output
     set -e
 }
 
 OPTIND=1
 SCRIPT_DIR="$(cd "$(dirname "$0")" > /dev/null 2>&1; pwd -P)"
+UNAME_S=$(uname -s)
 JATTACH=$SCRIPT_DIR/build/jattach
 PROFILER=$SCRIPT_DIR/build/libasyncProfiler.so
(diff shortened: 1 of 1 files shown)


profiler.sh
```diff
@@ -259,6 +259,7 @@ else
     esac
 fi
 LOG=/tmp/async-profiler-log.$$.$PID
+LOG_HOST=/proc/$PID/root/$LOG
```
Jongy, line 262: Need to use it only if Linux (otherwise, just `LOG`)

Jongy: Fixed

apangin:
AFAIK, `/proc/pid/root` is not always accessible by an unprivileged user.

apangin:
Also, just for the information, I'm going to get rid of `profiler.sh` entirely in favor of a binary launcher (with the integrated jattach functionality). So, fixing the script right now will be likely in vain.

apangin:
> AFAIK, `/proc/pid/root` is not always accessible by an unprivileged user.

Indeed; see [proc(5)](https://man7.org/linux/man-pages/man5/proc.5.html)

>               In a multithreaded process, the contents of the
>               /proc/[pid]/root symbolic link are not available if the
>               main thread has already terminated (typically by calling
>               pthread_exit(3)).
> 
>               Permission to dereference or read (readlink(2)) this
>               symbolic link is governed by a ptrace access mode
>               PTRACE_MODE_READ_FSCREDS check; see ptrace(2).

Jongy:
> AFAIK, `/proc/pid/root` is not always accessible by an unprivileged user.

I had in mind that we're running either as root, or as the same user, so access should be okay.
But indeed, `PTRACE_MODE_READ_FSCREDS` access might be stricter comparing to the simple UID/GID check the JVM employs, and is affected by a whole lot of other settings that we can't assume about (`PR_SET_DUMPABLE` state, maybe `ptrace_scope`, user namespaces...).

This is only an added measure to support more cases, so it might as well be "try `/tmp/async-profiler-log` and if that file doesn't exist, try `/proc/pid/root/...`". That sounds simple enough & legit to me.

> Also, just for the information, I'm going to get rid of profiler.sh entirely in favor of a binary launcher (with the integrated jattach functionality). So, fixing the script right now will be likely in vain.

Yeah, I recall I've seen it in one of the issues. Still, this fix is so small & simple, yet very useful, so I thought it's worth the time.

Jongy:
> This is only an added measure to support more cases, so it might as well be "try `/tmp/async-profiler-log` and if that file doesn't exist, try `/proc/pid/root/...`". That sounds simple enough & legit to me.

Fixed it that way.

If you're planning to make the binary launcher change before the next async-profiler release, then it's indeed in vain... if not, it's very useful, the log-to-file feature is very nice and there's no reason not to have it for containers.

apangin:
Thank you for patch. I've slightly modified it to access not only log file, but also a temporary output file through /proc/pid/root, too.

PR #428, opened 2021-05-07
Use time types instead of long.

uri-canva:
This fixes the build on macOS:

Before:
```
src/itimer.cpp:57:34: error: non-constant-expression cannot be narrowed from type 'long' to '__darwin_suseconds_t' (aka 'int') in initializer list [-Wc++11-narrowing]
    struct itimerval tv = {{sec, usec}, {sec, usec}};
                                 ^~~~
src/itimer.cpp:57:34: note: insert an explicit cast to silence this issue
    struct itimerval tv = {{sec, usec}, {sec, usec}};
                                 ^~~~
                                 static_cast<__darwin_suseconds_t>( )
src/itimer.cpp:57:47: error: non-constant-expression cannot be narrowed from type 'long' to '__darwin_suseconds_t' (aka 'int') in initializer list [-Wc++11-narrowing]
    struct itimerval tv = {{sec, usec}, {sec, usec}};
                                              ^~~~
src/itimer.cpp:57:47: note: insert an explicit cast to silence this issue
    struct itimerval tv = {{sec, usec}, {sec, usec}};
                                              ^~~~
                                              static_cast<__darwin_suseconds_t>( )
2 errors generated.
```

This is because on macOS `suseconds_t` is `int`, not `long`.

### DIFF
--- a/src/itimer.cpp
+++ b/src/itimer.cpp
@@ -52,8 +52,8 @@ Error ITimer::start(Arguments& args) {
 
     OS::installSignalHandler(SIGPROF, signalHandler);
 
-    long sec = _interval / 1000000000;
-    long usec = (_interval % 1000000000) / 1000;
+    time_t sec = _interval / 1000000000;
+    suseconds_t usec = (_interval % 1000000000) / 1000;
     struct itimerval tv = {{sec, usec}, {sec, usec}};
     
     if (setitimer(ITIMER_PROF, &tv, NULL) != 0) {


apangin:
What compiler and OS version is this?
I haven't seen such an error with clang on macOS 10.12 to 11.

uri-canva:
This is built with nix on CI here: https://hydra.nixos.org/build/142022256. I can repro the failure when building with nix, but not when building with the regular apple toolchain on macOS 11.3.1.

You can see it defined here though (`__darwin_suseconds_t`): https://opensource.apple.com/source/xnu/xnu-2782.30.5/bsd/sys/_types.h.auto.html, and I've found some other projects hitting the same issue in their code, and the documentation on both linux and macOS refers to those types as `time_t` and `suseconds_t` (https://man7.org/linux/man-pages/man2/getitimer.2.html) so this seems correct, sort of like using `size_t` instead of `int` or `long`.

apangin:
OK, thanks for fixing that.

PR #434, opened 2021-06-01
Travis CI: use Ubuntu 18.04 (Bionic Beaver)

simon04:
Ref: https://docs.travis-ci.com/user/reference/bionic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: cpp
 
-dist: precise
+dist: bionic
 
 sudo: required
 


apangin:
Thanks.
I've merged the PR, as working CI is certainly better than broken. I have two concerns though:
- I'd prefer to build on as old distribution as possible. Binaries need to be linked with a rather old verion of glibc in order to work on a wide range of operating systems.
- travis-ci.org will be shut down in a few days. We'll need to migrate to travis-ci.com or another CI provider.

martin-g:
>  We'll need to migrate to travis-ci.com

the drawback of moving to `.com` is that it is kinda `trial` - you will have 10000 free credits for build jobs and then you will have to start paying. TravisCI said it will be free for OSS projects but unfortunately it is not :-/

> or another CI provider.

I saw that you started using GitHub Actions. I'd recommend CircleCI because of their support for Linux ARM64 build nodes but if this is not essential then GHA is good too!

PR #442, opened 2021-06-10
Add differential flame graph generator to converter.jar

h-cusworth:
Differential flame graphs can be used to quickly compare two sets of collapsed stacks which can be really use if you want to see the impact of a change or contrast different profiles using different configurations (see http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html for more info on the basic idea). I've added a differential flame graph generator to the converter.jar. It is based on FlameGraph.java with added logic to parse the two files, add the comparison data to each frame and colour each frame.

The size and shape of the differential flame graph will be the same size and shape as the normal flame graph from the first profile, but it will be coloured to show areas of difference with the second profile: gold stack traces do not appear in the second profile, red traces appear more and blue traces appear less. The intensity of the red/blue reflects the size of the difference. This 'diff' value is calculated by normalising both sample counts for each trace, e.g. for a frame x which appears in both graphs: diff value for x = (samples of x in data set 1 / data set 1 total samples) - (samples of x in data set 2 / data set 2 total samples). Note that the diff values are also then normalised to the colour space i.e. there will always be a frame that is the reddest red. This means you can’t make visual comparisons between differential graphs (but you can compare the raw diff values). 

Because some stack traces appear in one profile but not the other, its best to generate a pair of differential graphs - one based on each normal flamegraph (you do this by just switching the order you pass input1.collapsed and input2.collapsed).

### DIFF
--- a/src/converter/Main.java
+++ b/src/converter/Main.java
@@ -25,6 +25,7 @@ public class Main {
         System.out.println();
         System.out.println("Available converters:");
         System.out.println("  FlameGraph input.collapsed output.html");
+        System.out.println("  DiffFlameGraph input1.collapsed input2.collapsed output.html");
         System.out.println("  jfr2flame  input.jfr       output.html");
         System.out.println("  jfr2nflx   input.jfr       output.nflx");
     }
--- a/src/converter/DiffFlameGraph.java
+++ b/src/converter/DiffFlameGraph.java
@@ -1,0 +1,489 @@
+/*
+ * Copyright 2020 Andrei Pangin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


apangin:
Thank you for the PR.
The feature is indeed useful, but I'm afraid, the proposed implementation has too much duplicated code. In particular, anyone who wants to improve the Flame Graph UI, will need to make changes separately in two places (or even in 3 places, since there is also a C++ generator). I'd suggest to refactor the code in order to reduce maintenance overhead. I can try to do it myself, but not in the nearest future, as there are other priority tasks on my list.

h-cusworth:
Hi Andrei,
Thank you for the feedback, I've refactored so that the diff graphs can be generated by passing a --diff flag to FlameGraph. If there are any more implementation changes you'd prefer let me know and I'll take a look :)

src/converter/Main.java
```diff
@@ -25,6 +25,7 @@ public static void main(String[] args) {
         System.out.println();
         System.out.println("Available converters:");
         System.out.println("  FlameGraph input.collapsed output.html");
+        System.out.println("  Flamegraph --diff input1.collapsed input2.collapsed output.html");
```
mfussenegger, line 28: I hope you don't mind the drive-by review. I was just testing this functionality and encountered this.


```suggestion
        System.out.println("  FlameGraph --diff input1.collapsed input2.collapsed output.html");
```

Lowercased `G` would result in an error: `Error: Could not find or load main class Flamegraph`

h-cusworth: Not at all, thanks for catching it!

apangin:
Superseded by #1553

PR #443, opened 2021-06-14
Add feature checking for debug symbols

PaulBGD:
I'm not sure how helpful this is, but it's a necessary function that I needed for my application so I thought I'd PR it in case it's something others are interested in.

This PR adds a new `AsyncProfiler.check(String)boolean` method, which can support checking for different built-in features. The only feature check I added initially is checking for debug symbols, as it allows knowing ahead of time whether you'll be able to do allocation profiling.

My C++ is very poor, so I think it may be better to move the debug symbol lookups to their own utility instead of copying and slightly modifying the code from the allocation profiler. Please let me know if/how this should be changed.

### DIFF
--- a/src/api/one/profiler/AsyncProfiler.java
+++ b/src/api/one/profiler/AsyncProfiler.java
@@ -112,6 +112,12 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
         }
     }
 
+    /**
+     * Checks whether a feature is supported by the profiler
+     *
+     * @param feature Profiling feature, see {@link Features}
+     * @return true if the feature is supported
+     */
     @Override
     public boolean check(String feature) {
         return check0(feature);
--- a/src/api/one/profiler/AsyncProfiler.java
+++ b/src/api/one/profiler/AsyncProfiler.java
@@ -112,6 +112,11 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
         }
     }
 
+    @Override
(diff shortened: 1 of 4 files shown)


apangin:
Thank you for your improvement proposals.
I'd kindly suggest to open an issue for discussion before submitting feature PRs. This may often help in finding a good approach to implement this or another feature.

As to the proposed feature, in fact, it was already implemented a while ago, but didn't receive much attention, and thus got accidentally broken with the advent of multi-event profiling. I've just pushed a small change to fix it back.

I'm talking about `check` command, which is used to test if the specified profiling configuration is supported. It's similar to your proposal, but works in broader scope, e.g. it can test if perf events are available in the current environment.
Now it works for `alloc` and `lock` modes, too.
For example, to check if allocation profiling is available, run
```
try {
    profiler.execute("check,alloc");
    // OK
} catch (Exception e) {
    // Allocation profiling is unavailable
}
```

apangin:
Also note that new native methods in the Profiler API are strongly discouraged.
New functionality should be implemented with the commands available through `execute` method.

apangin:
Given that the requested feature is already available, I'm closing the PR without merge.
Thanks.

PaulBGD:
Appreciate the response, it does seem like that'll work for my use-case. I do still think an API like this would be cool for checking what features are supported, but I understand wanting to put everything inside execute.

PR #459, opened 2021-09-10
port async profiler 2.0 to Power Linux little endian (linuxppc64le), step 2

ghaug:
Hi Andrei,
as discussed, here is the new pull request. I include `<signal.h>` in stackFrame_ppc64.cpp now. On SuSE it works, of course, but I wasn't able to check this on ubuntu. 
Have a nice WE,
Gunter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ API_JAR=async-profiler.jar
 CONVERTER_JAR=converter.jar
 
 CFLAGS=-O3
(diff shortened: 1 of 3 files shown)


apangin:
Perfect! The updated PR looks good, and I've verified it builds and runs file.

apangin:
While testing, I found one more ppc64 specific difference. Returned address is saved 2 machine words after frame pointer, where on other architectures it is 1 word after.
So I fixed that by introducing another architecture-dependent constant, `FRAME_PC_SLOT`.

Issue #272, opened 2019-12-02
Clarify samples count between -e cpu and -e itimer


franz1981:
Hi,

thanks for this awesome profiler!
I cannot found any info on the doc re how to interpret samples count while using `-e cpu` and `-e itimer`. I wouldn't like it to became a blog about how sampling profilers (should) work, but I admit that writing 2 lines of doc there could help to understand the difference in samples count a user can expect depending on the context.
We can start a discussion here and I can send a PR to add it, if is welcome :+1: 

thanks,
Franz

apangin:
Hi Franz,

Thank you for your interest in async-profiler and its internals.

Both `cpu` and `itimer` mode measure the CPU time spent by the running threads. For example, if an application uses 2 cpu cores, each with 30% utilization, and the sampling interval is 10ms, then the profiler will collect about 2 * 0.3 * 100 = 60 samples per second.

In other words, 1 profiling sample means that one CPU was actively running for N nanoseconds, where N is the profiling interval.

 - `itimer` mode is based on [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) syscall, which [ideally] generates a signal every given interval of the CPU time consumed by the process.
 - `cpu` mode relies on PMU-based sampling capabilities (like Intel PEBS or AMD IBS). The idea is the same - to generate a signal every `N` nanoseconds of CPU time, which in this case is achieved by configuring PMU to generate an interrupt every `K` CPU cyles.

apangin:
So, in ideal world, both `itimer` and `cpu` should collect the same number of samples. Typically the profiles indeed look very similar.

[In some cases](https://github.com/golang/go/issues/14434) `cpu` profile appears a bit more accurate though, since the signal is delivered exactly to the thread that overflowed a hardware counter.

franz1981:
Thanks @apangin!

I've written a simple experiment to understand this:
```java
public class SamplingExperiment {

    public static void main(String[] args) {
        final long targetNanos = TimeUnit.MILLISECONDS.toNanos(10);
        while (true) {
            spinWork(targetNanos);
            //I can measure latency of spinWork and use to sleep too
            sleep(targetNanos);
        }
    }

    private static void sleep(long nanos) {
        long remaining = nanos;
        while (remaining > 0) {
            final long start = System.nanoTime();
            LockSupport.parkNanos(remaining);
            final long elapsed = System.nanoTime() - start;
[...]

Issue #184, opened 2018-12-22
cpu profiling delivers wrong result


oehme:
I just profiled some Gradle builds using async-profiler and I noticed that the `cpu` profile has some really odd hot spots that definitely don't match reality:

<img width="1200" alt="cpu-profile" src="https://user-images.githubusercontent.com/1191797/50374931-a5e6bf00-05f5-11e9-8330-1356179dd860.png">

You can see `open()` calls completely dominating the profile. I cross-checked this with the `wall` profiler (focussing only on the thread in question). The open calls are highlighted in purple. As you can see they really only make up ~1% of the runtime.

<img width="1198" alt="wall-profile" src="https://user-images.githubusercontent.com/1191797/50374936-c6af1480-05f5-11e9-802d-66fd3c2358bd.png">

This is a single-threaded, CPU-bound program, so there should be very little difference between the `cpu` and `wall` results. I cross-checked with JFR, JProfiler and YourKit and all of them agree with the `wall` profile. I also did an experiment and completely removed this code path from the code base. It disappeared from the `cpu` profile, making it look more like the `wall` profile, however runtime was completely unaffected as I expected. This shows that there is in fact no significant time spent in this method.

You can reproduce these measurements with the following commands:

- `git clone git@github.com:gradle/gradle.git`
- `./gradlew install -Pgradle_installPath=/usr/local/gradle-source-build largeJavaMultiProject` (this will take some time)
- `cd subprojects/performance/build/largeJavaMultiProject`
- repeatedly run `/usr/local/gradle-source-build/bin/gradle help` while async-profiler is attached to the Gradle daemon process, then dump the results to SVG

I'm on macOS High Sierra and tried both Oracle Java 8 and OpenJDK 11 with the same results. Let me know if there's anything else I can do to get to the bottom of this.

apangin:
Sounds strange.
Does the problem happen on Linux as well?
Is it possible to get some reduced test case?

oehme:
I'll try it on Linux. Constructing a reduced case might be tough, as I have no clue what combination of calls is causing this pattern. I'll try, but the steps above should be relatively easy to follow, they just take a bit of time to execute since Gradle is a big project.

oehme:
I get the correct results on Ubuntu 18.04, so this seems MacOS specific.

apangin:
Thank you for the input. That's what I suspected.
[...]

PR #460, opened 2021-09-13
Default to itimer profiler when perf profiler isn't available

luhenry:
Today, it defaults to wall time profiler since d78e0b3a517593ced9af0fbaa073105f379c228d. This commit doesn't make it clear why that change of behavior, especially on Linux.

It is more useful to switch to itimer as it's behavior is closer to perf sampling interval.

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -884,7 +884,7 @@ Engine* Profiler::selectEngine(const char* event_name) {
     if (event_name == NULL) {
         return &noop_engine;
     } else if (strcmp(event_name, EVENT_CPU) == 0) {
-        return PerfEvents::supported() ? (Engine*)&perf_events : (Engine*)&wall_clock;
+        return PerfEvents::supported() ? (Engine*)&perf_events : (Engine*)&itimer;
     } else if (strcmp(event_name, EVENT_WALL) == 0) {
         return &wall_clock;
     } else if (strcmp(event_name, EVENT_ITIMER) == 0) {


apangin:
Hi @luhenry,

What is the motivation to switch to `itimer`? In my observation, it's the other way round: `wall` engine gives more accurate profile (which is closer to `perf_events`) comparing to `itimer`. Distribution of `itimer` signals isn't always fair between threads, especially on macOS, where they seem highly biased towards system calls (#184). On Linux, `itimer` is not that bad, but still produces skewed results sometimes (#272).

Despite its name, `wall` engine is not only for wall-clock profiling. It can also act as a CPU profiler by choosing those samples that hit runnable threads only.

luhenry:
`perf` is overall a CPU profiler, aka it's sampling what happens when on CPU. `itimer` (as used in `async-profiler`) is similar since it's also using a CPU clock and will trigger a signal once X ms has been spent executing on CPU.

`wall` clock however is not a CPU profiler in the same way. It will only sample threads which are currently running on CPU, but it won't sample every X ms of _on-CPU execution time but every X ms of _wall time_.

That difference is what concerns me, and which could potentially introduce bias in what we end up measuring. For example, what happens if a method ends up executing every 10 ms, while the sampling interval is also every 10 ms. If this method executes for 1 ms, then `itimer` and `perf` will get it right and only sample every 10 executions, but what's the guarantee `wall` will not sample every invocation of the method or, even worst, never sample any invocation of the method because it's just out of sync. Is it something you've observed in practice?

apangin:
OK, I understood your concern. It makes sense, and I agree that the scenario you describe is theoretically possible. If `itimer` signals were fair, making this a default engine would be a good choice. The problem is, they are not fair at all (#184, #272). So changing the default engine to `itimer` does not necessary make results better. On macOS, `itimer` is almost always worse. On Linux, it depends. Furthermore, `ITIMER_PROF` is not always available, e.g. on WSL 1. Therefore I'm not convinced that we should prefer `itimer` to wall/cpu.

PR #467, opened 2021-09-27
Add memory leak profiler

luhenry:
The approach is fairly simple:
 - Store a weak reference to a sample of allocated objects
 - Dump the list of objects which are still live after a certain
interval (not completely implemented just yet)


luhenry:
Opened against the wrong repo, sorry for the noise.

PR #468, opened 2021-09-29
Introduced system property to workaround JDK-8173361

apangin:


src/profiler.cpp
```diff
@@ -868,7 +871,7 @@ Error Profiler::start(Arguments& args, bool reset) {
 
     updateSymbols(args._ring != RING_USER);
 
-    _safe_mode = args._safe_mode;
+    _safe_mode = VM::_safe_mode | args._safe_mode;
```
eyalkoren, line 874: Proposing a small enhancement: you can use this combined safe mode to decide per profiling session whether to enable or disable `JVMTI_EVENT_COMPILED_METHOD_LOAD` and `JVMTI_EVENT_COMPILED_METHOD_UNLOAD`, thus protecting existing users that already set `(_safe_mode & 14) == 14` from the related JVM bug without requiring to use the system property.

apangin: Per-session decision isn't enough, unfortunately.
async-profiler needs information about all compiled methods, not only those compiled during profiling session. So it enables `CompiledMethodLoad` events just once at the agent startup and never switches them off. Enable/Disable does not work, as the information will  be inconsistent.

In theory, it's possible to get the list of all compiled methods from scratch at the beginning of every profiling session. I don't like this solution though, as it will cause delays on each `start` command. I plan to re-implement this part in 2.x anyway.

eyalkoren: I see, thanks for explaining. So maybe you can only limit it for the ON --> OFF switch for sessions and not vice versa. This avoids the inconsistency problem and still protects existing `safemode` users.

You can safely ignore this proposal, though, the `AsyncProfiler.safemode` system property does the job perfectly 🙂 .

PR #477, opened 2021-10-14
Add 'Engine::onThreadStart' and 'Engine::onThreadEnd' methods

KirillTim:
It is useful to propagate `onThreadStart` and `onThreadEnd` events from profiler to engine. It allows a new engine to handle such events. It also simplifies current `Profiler::onThreadStart` and `Profiler::onThreadEnd` implementations.

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -32,6 +32,12 @@ Error Engine::start(Arguments& args) {
 void Engine::stop() {
 }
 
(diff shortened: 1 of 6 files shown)


apangin:
The code in question has been refactored as a result of OpenJ9 support.
There are no more per-engine `onThreadStart` routines.

PR #479, opened 2021-10-14
'Error::_message' is std::string now

KirillTim:
Automatically release string resource in destructor


toktarev:
From what I see Error accepts string constant in all places;
No need to release string constants in C/C++.

apangin:
This change doesn't seem to address any known issue - I'd better leave the existing code.

PR #485, opened 2021-10-30
Add Hot/Cold FlameGraph generator to converter.jar

tetsuzawa:
I've added a [Hot/Cold FlameGraph](https://www.brendangregg.com/FlameGraphs/hotcoldflamegraphs.html) generator to the converter.jar.

The result of wall clock profiling recorded in jfr format by async-profiler contains thread state information.
The Hot/Cold FrameGraph generated by async-profiler is very effective because it can visualize a stack trace with the same context on-CPU and off-CPU.

Example:
[image](https://tetsuzawa.github.io/async-profiler-hot-cold-flamegraph-example/demo.html)

The darker colors indicate on-CPU and lighter colors indicate off-CPU.
The color fill of the parent stack is determined by the ratio of the number of on-CPU samples in the child stacks.

I created this in https://github.com/tetsuzawa/async-profiler-hot-cold-flamegraph-example.


apangin:
While this is an interesting feature, I can't really accept it in its current form, as the implementation involves much copy-pasted code, thus doubling maintenance costs.
Thank you anyways.

apangin:
Please also note that async-profiler accepts only those contributions that transfer all IP rights to the project owner, i.e. all new files need to retain the following copyright header:
```
 * Copyright 2021 Andrei Pangin
```

tetsuzawa:
Hi @apangin 

Thank you for your comment, I have refactored FlameGraph to be able to generate a Hot/Cold FlameGraph by passing the --hotcold flag to FlameGraph.
The Hot/Cold FlameGraph does not support collapsed format, so it is currently only valid when running jfr2flame.

If there are any problems with this changes, please let me know.

PR #486, opened 2021-10-31
Added option to profile specific CPU

pveentjer:
By default async-profiler will profile all CPU's.

A new option has been added to profile only a specific CPU. This can be used like this:

```
--cpu <cpu>
```

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -503,7 +503,7 @@ long PerfEvents::_interval;
 Ring PerfEvents::_ring;
 CStack PerfEvents::_cstack;
 
-int PerfEvents::createForThread(int tid) {
+int PerfEvents::createForThread(int tid, int cpu) {
     if (tid >= _max_events) {
         Log::warn("tid[%d] > pid_max[%d]. Restart profiler after changing pid_max", tid, _max_events);
         return -1;
@@ -557,7 +557,7 @@ int PerfEvents::createForThread(int tid) {
     if (FdTransferClient::hasPeer()) {
         fd = FdTransferClient::requestPerfFd(&tid, &attr);
     } else {
-        fd = syscall(__NR_perf_event_open, &attr, tid, -1, -1, 0);
+        fd = syscall(__NR_perf_event_open, &attr, tid, cpu, -1, 0);
     }
 
     if (fd == -1) {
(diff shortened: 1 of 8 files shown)


src/perfEvents_linux.cpp
```diff
@@ -557,7 +557,7 @@ int PerfEvents::createForThread(int tid) {
     if (FdTransferClient::hasPeer()) {
         fd = FdTransferClient::requestPerfFd(&tid, &attr);
     } else {
-        fd = syscall(__NR_perf_event_open, &attr, tid, -1, -1, 0);
+        fd = syscall(__NR_perf_event_open, &attr, tid, cpu, -1, 0);
```
apangin, line 560: When setting `cpu` argument here, it also needs to be supported in FdTransfer protocol. Otherwise, there will be an inconsistent behavior between using and not using `--fdtransfer` option.

src/profiler.cpp
```diff
@@ -120,13 +120,13 @@ void Profiler::addRuntimeStub(const void* address, int length, const char* name)
     CodeHeap::updateBounds(address, (const char*)address + length);
 }
 
-void Profiler::onThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
+void Profiler::onThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread, int cpu) {
```
apangin, line 123: The signature of this method matches JVM TI [ThreadStart](https://docs.oracle.com/en/java/javase/11/docs/specs/jvmti.html#ThreadStart) event. Therefore, the value of `cpu` is undefined here, which probably results in `EINVAL` when coming to `perf_event_open` syscall.

apangin:
Oops.. It seems I started reviewing too early, as the code doesn't even compile.

pveentjer:
That is weird. I'll recheck.

apangin:
You may find the executed commands and the exact error in [CI logs](https://github.com/jvm-profiling-tools/async-profiler/runs/4060620820?check_suite_focus=true)

apangin:
I'll close this PR for now, since it cannot be merged in the current form, and has been inactive for a while.
Feel free to reopen it or submit a new PR, once the mentioned problems are fixed. Thanks.

PR #503, opened 2021-11-13
Fail arguments parsing upon unknown argument

Jongy:
When using `profiler.sh` this probably can't happen, but for others writing other frontends, this can catch sad errors like typos in passed arguments, which are currently ignored :( Also, it might catch errors in existing frontends, who knows.

It would be nicer if the returned error included the argument name, but I see that the `Error` class expects a long-living `char *` pointer, and does no memory management on its own, so I can't pass a stack buffer (which is deallocated by the point the string is used) or a heap buffer (which will be leaked). Additionally, the `Log` object is not yet initialized, so I can't use it either.
A static string is still better than nothing, hence I'm proposing this PR as is.

~~I can maybe replace `} else {` with `#define DEFAULT` so it looks nicer?~~ done

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -307,6 +307,9 @@ Error Arguments::parse(const char* args) {
 
             CASE("reverse")
                 _reverse = true;
+
+            } else {
+                msg = "Unknown argument";
         }
     }
 
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -46,6 +46,8 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 
 #define CASE(s)        } else if (arg_hash == HASH(s "            ")) {
 
+#define DEFAULT()      } else {
+
 
 // Parses agent arguments.
 // The format of the string is:
@@ -308,7 +310,7 @@ Error Arguments::parse(const char* args) {
             CASE("reverse")
                 _reverse = true;
 
-            } else {
+            DEFAULT()
                 msg = "Unknown argument";
         }
     }


apangin:
Even though the fix looks trivial, I wouldn't call it absolutely harmless.

Ignoring unsupported parameters solved certain compatibility issues, and I used to exploit this before. Recent example: our network runs hundreds Java applications, they update in different time and thus may have different versions of async-profiler agent. After introducing a new `sched` option, we added it to our profiling system, not caring if the target supports it or not. On newer applications we automatically started collecting profiles broken down by scheduling policy; older application profiles remained as they were.

I admit that my example is probably not very good; I just wanted to say that the proposed change is not really minor - may be, it's better to postpone it until the next major release.

A special problem is the handling of an empty option, which is treated as invalid now. Consider a script that builds a profiler command line like this:
```
PARAMS="$PARAMS,bar"
```
It may easily happen that a command line starts with `,` or contains two commas in a row: `foo,,bar`. Async-profiler parser should definitely tolerate this.

Jongy:
Hmm, I understand from your example on how it can be exploited easier configuration.

On the other hand - it sounds like the `sched` example is an opt-in feature that works in best-effort mode, so you are okay with it working where possible. What if you're adding a new argument that's not best-effort? For example, changes the profiling configuration in a non-compatible way with the older configuration; or enabling stability features like `--safemode`. For those, you want to be absolutely sure that you get what you request.

The empty options case - we can easily make the profiler ignore them, yeah.

All in all, I agree that it's not a minor change and it may break existing frontends which are not prepared. async-profiler is yours and it's up to you if you want it :) I think it's an important robustness improvement for the frontend side. We can definitely wait for the next major release.

apangin:
Despite (dis)advantages of both approaches, I agree that failing on unknown options is better overall.
Please make sure the parser tolerates extra commas, and I'll definitely accept the change for async-profiler 3.0.

apangin:
I start working on the next major release, so it's a good time to merge this change.

PR #505, opened 2021-11-16
Refactor jattach such that it can be structured as library

etep:
Refactor jattach such that it can be more easily used as a library in other open source projects. Essentially, we just pull main() into its own file.

Signed-off-by: Pete Stevenson <jps@pixielabs.ai>


apangin:
jattach sources were copied verbatim from another project of mine - [`jattach`](https://github.com/apangin/jattach/)
(I think about integrating jattach into async-profiler as a git submodule).

So if you are interested in jattach, please check that another project.

BTW, jattach already has the special make target `dll` to build it as a shared library.

PR #514, opened 2021-12-03
Remove double close() call

Jongy:
fclose() on an fdopen()ed file closes the underlying file descriptor.

Tested with this program:
```
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>
#include <fcntl.h>

int main(void) {
    int fd = open("/dev/null", O_RDONLY);
    FILE *fp = fdopen(fd, "r");

    fclose(fp);
    close(fd);
}
```

stracing it results in:
```
openat(AT_FDCWD, "/dev/null", O_RDONLY) = 3
fcntl(3, F_GETFL)                       = 0x8000 (flags O_RDONLY|O_LARGEFILE)
brk(NULL)                               = 0x562a0bb5d000
brk(0x562a0bb7e000)                     = 0x562a0bb7e000
close(3)                                = 0
close(3)                                = -1 EBADF (Bad file descriptor)
```

Also from `fdopen(3)`: `The file descriptor is not dup'ed, and will be closed when the stream created by fdopen() is closed.`

Bug is present since #411 (v2.5). Sorry for that :(

The race here is unlikely but present, for the first time AP starts and loads the kernel symbols, there is a slight chance that another thread receives the same fd number from some `open` `accept` etc, and then we close it for that thread.

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -378,7 +378,6 @@ void Symbols::parseKernelSymbols(NativeCodeCache* cc) {
     }
 
     fclose(f);
-    close(fd);
 }
 
 void Symbols::parseLibraries(NativeCodeCache** array, volatile int& count, int size, bool kernel_symbols) {


Jongy:
Hmm, this is more serious that I previously thought. Per my tests:
1. async-profiler tries to read kernel symbols also in `itimer` mode (as long as you don't pass `alluser`). Which is strange because to my understanding it won't receive any kernel symbols in this mode.
2. kernel symbols are re-read upon every `start` operation (I was thinking async-profiler reads them just once and caches them?)

apangin:
> async-profiler tries to read kernel symbols also in itimer mode

Yeah, async-profiler calls `updateSymbols` before engine selection. I'll check if I can move the call after and update kernel symbols only for perf_events engine.

> kernel symbols are re-read upon every start operation (I was thinking async-profiler reads them just once and caches them?

Yes, it caches symbols. However, if it wasn't able to read the kernel symbols once, it'll try again upon the next `start` command, in case a user discovers missing kernel symbols and updates the configuration to make them available. This is intentional.

apangin:
Thanks for the double close fix. Merged.

apangin:
Fixed profiler to read kernel symbols only for perf_events engine.

Jongy:
>  However, if it wasn't able to read the kernel symbols once, it'll try again upon the next start command

Hmm, I guess my observation was incomplete - I saw that there is an `open` call on `/proc/kallsyms` upon every `start`, but since the container I tested on is not privileged, all addresses were read as `0000....` so `_have_kernel_symbols` was never flagged on :facepalm: that explains.

PR #518, opened 2021-12-12
Incremental jfr

Artyomcool:
This PR introduces a file format that really close to the JFR but has less overhead when flushing chunks.
Main idea is to dump only new stacktaces/methods/constants/etc. File format is compatible with jfr2flame tool but incompatible with FR.
To try it just use -f somefilename.ijfr. Then jfr2flame converter must be used to get flame graph. It can be run without profiling stop.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -328,6 +328,10 @@ Error Arguments::parse(const char* args) {
         if (_output == OUTPUT_SVG) {
             return Error("SVG format is obsolete, use .html for FlameGraph");
         }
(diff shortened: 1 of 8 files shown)


jbachorik:
Just a bit of info here - there is no need to modify the JFR file format. 
It actually supports incrementally outputting the constant pools already. It is done via `jdk.Checkpoint` event (event id 1)  and there may be any number of such events in a chunk. Internally, JFR is using the concept of epochs to achieve double buffering (from a quick glance the approach in this PR is similar).

PR #526, opened 2021-12-27
Build flame graph from events in time range

pnf:
It's useful sometimes to examine the profile within a specific time range.  This requires knowing the start time of the recording before aggregating, and it seems that the easiest way to do this is to iterate over all the events to find the earliest.  I tried extracting the `ActiveRecording` metadata, but the values didn't make sense - strings were plausible, but durations were too long.  It also seems a little weird to have the time range extracted via the `FlameGraph` constructor, but, again, it was easiest.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -97,6 +97,9 @@ fdtransfer() {
 }
 
 jattach() {
+    if [ -n "$VERBOSE" ]; then
+      echo Executing: "$JATTACH" "$PID" load "$PROFILER" true "$1,log=$LOG"
+    fi
     set +e
     "$JATTACH" "$PID" load "$PROFILER" true "$1,log=$LOG" > /dev/null
     RET=$?
@@ -139,6 +142,7 @@ OUTPUT=""
 FORMAT=""
 PARAMS=""
 PID=""
+VERBOSE=""
 
 while [ $# -gt 0 ]; do
     case $1 in
@@ -260,6 +264,9 @@ while [ $# -gt 0 ]; do
         --safe-mode)
             PARAMS="$PARAMS,safemode=$2"
             shift
+            ;;
+        --verbose)
+            VERBOSE=1
             ;;
         [0-9]*)
             PID="$1"
--- a/src/api/one/profiler/AsyncProfiler.java
+++ b/src/api/one/profiler/AsyncProfiler.java
@@ -30,6 +30,11 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
     private AsyncProfiler() {
     }
 
+    // Assume it was loaded with -agentpath
+    public static AsyncProfiler getPreloadedInstance() {
+        instance = new AsyncProfiler();
+        return instance;
+    }
     public static AsyncProfiler getInstance() {
         return getInstance(null);
     }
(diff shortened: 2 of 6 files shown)


apangin:
Hi Peter,
This is a useful feature indeed. I already have a draft alternative implementation though, where the range is specified in the absolute (wall clock) timestamps. Will push it some time soon.

pnf:
Hi -
That wound great.  It would be slightly more great to be able to specify time either absolutely or relative to start... 
I've also refactored a bit locally to allow jfr2flame to emit collapsed stacks, suitable for loading into (for example) speedscope.  Is that in your pipeline too?
Somewhat related: I initially tried to get time-scoped flame graphs by loading async-profiler jfr into jmc, but results were inconsistent - for a random example, _widening_ the range to encompassing gc events (with shenandoah fwiw) resulted in empty graphs in jmc (but not of course in jfr2flame). That led me to wonder how well defined and stable the jfr format is, whether async-profiler's jfr output is slightly different from the jdk's, and whether hard-coding a sequence of `getVar...` calls is generally safe.

apangin:
> I've also refactored a bit locally to allow jfr2flame to emit collapsed stacks, suitable for loading into (for example) speedscope. Is that in your pipeline too?

Not really. In my opinion, converting jfr to collapsed format makes little sense: collapsed stacks format is much bigger in size, but at the same time completely misses timeline details. Instead, it would be better to add jfr support directly in speedscope.

Besides Mission Control, there is a nice JFR viewer in IntelliJ IDEA. Also, my colleagues are working on another cool visualization solution for JFR.

> widening the range to encompassing gc events (with shenandoah fwiw) resulted in empty graphs in jmc

I'm not sure I've understood the issue. So far I could successfully open with JMC all JFRs produced by async-profiler, except files larger than 2GB: JMC does not support chunks > 2GB. If possible, please send me a problematic JFR file - I'll check what's wrong with it (or with JMC).

krzysztofslusarski:
Collapsed stack has some benefits. It is a text format so yo can do any text transformation you want, before conversion to flame graph. 
@pnf - I wrote JFR -> Collapsed stack converter - https://github.com/krzysztofslusarski/collapse-jfr - you can take a look if it is suitable for you

pnf:
I had two reasons for generating collapsed stack format: text-modification, as you say, like domain-specific fqcn abbreviations; and loading into external viewers like speedscope.  
If you’re generating stacks anyway to feed into flamegraphs, it’s trivial to dump them out as text instead. 

Le Ven, 31 déc., 2021, à 03:11, Krzysztof Ślusarski a écrit:

pnf:
> I'm not sure I've understood the issue. So far I could successfully open with JMC all JFRs produced by async-profiler, except files larger than 2GB: JMC does not support chunks > 2GB. If possible, please send me a problematic JFR file - I'll check what's wrong with it (or with JMC).
@apangin - probably not worth reproducing.  I was using home-built loom from what was then `HEAD`, starting and dumping the profile via the API and some combination of GC parameters I can't recall.  If I see it again, I'll upload the jfr.

Speaking of the API, I thought you should be able to get an `AsyncProfiler` instance when running  as an agent, without specifying a load-path again.

apangin:
> Speaking of the API, I thought you should be able to get an AsyncProfiler instance when running as an agent, without specifying a load-path again.

I committed this fix, too.
Seems like all questions in this PR are now addressed, so I can close it.

PR #530, opened 2022-01-14
Support default pattern highlighting in flame graph

rmarsch:
- Add an argument to flame graph args to pass in a regex pattern that will generate the graph with all matches highlighting in a different color
- Primary use case would be to highlight user application stack lines from other java stack lines

It's often the case that seeing the difference between user code (often your application code is all under a shared package) and non-user code quickly in a flame graph has value. The search function does let users manually do this, however it's likely nice to be able to search for other things while still having this basic distinction.

Also in a larger automated operation of generating flame graphs as part of tooling, it can be nice to pre-search a relevant pattern for users, such as having application lines versus 3rd party java lines as mentioned above.

Here's what the color selection looks like. I tried to stay away from the existing 5 colors and the highlight color. It seemed like green, yellow, orange, red, and blue were covered. Highlighting presents as pink. I opted for a light purple: 
image


src/converter/FlameGraph.java
```diff
@@ -55,6 +57,8 @@ public FlameGraph(String... args) {
                 minwidth = Double.parseDouble(args[++i]);
             } else if (arg.equals("--skip")) {
                 skip = Integer.parseInt(args[++i]);
+            } else if (arg.equals("--highlightstacks")) {
+                highlightStacksPattern = Pattern.compile(".*" + args[++i] + ".*");
```
rmarsch, line 61: Was on the fence whether the `.*` setup to allow single string matches should be pushed down to the argument or not

apangin, line 61: Where is `highlightstacks` field initialized, if no argument specified on the command line? Seems like there will be a NullPointerException in such case.

src/converter/FlameGraph.java
```diff
@@ -173,7 +177,9 @@ private String stripSuffix(String title) {
     }
 
     private int frameType(String title) {
-        if (title.endsWith("_[j]")) {
+        if (highlightStacksPattern.matcher(title).matches()) {
```
apangin, line 180: I'm afraid, pattern matching on every single frame name may notably decrease conversion performance. At least, there should be a fast path to skip regex stuff when no `highlightStacksPattern` specified.

src/converter/FlameGraph.java
```diff
@@ -173,7 +177,9 @@ private String stripSuffix(String title) {
...
+            return 5;
```
apangin, line 181: The problem is, such highlighting discards the information about the original frame type.
I'd rather not add a new frame type here, but do highlighting on the client side. The easiest way will be to initialize `pattern` variable in the generated JavaScript code.

rmarsch: To the best of my reading, using `pattern` would prevent the ability of using the search box in the generated page. The frame type only seems to be used to colorize the stack, which is the intent of this use case. Let me know if I have misinterpreted how this might work in the client-side.

The goal was to provide a way for a user to define a custom frame type that matches their needs. For example, they may want a classification of frames for "application domain calls" which all match a specific package name. I know that matches the goal of my organization, to separate the frame type for the code our organization owns in the graph from those outside of our organization, while still being able to use the search field to search for other interesting elements in the graph.

Artyomcool: It could be better to put your pattern(s) with their colors to generated JS code:
```
    "\tfunction f(level, left, width, type, title) {\n"
    "\t\tlevels[level].push({left: left, width: width, color: getColor(palette[type]), title: title});\n"
    "\t}\n"
```
Here we can easily match any number of patterns against the title and choose appropriate colors without touching original type.

rmarsch: Okay, I've rewritten it to just change the palette color, feeding the pattern into the javascript function and applying the extra color style if anything matches.

apangin:
Added `--highlight` option to the FlameGraph converter to automatically search and highlight frames matching the given RegExp. This reuses the search mechanics and the corresponding purple color.

At this point, there a no plans to add different frame types / colors for highlighting. If this is to be implemented in future, there should be some general way to provide a custom palette.

PR #537, opened 2022-01-27
Look in plt sections for the global offset table too

ghaug:
Hi Andrei,
unfortunately the 'Patch Global Offset Table to hook libc functions' change broke the PPC64le port. The symbol is located in .plt on PPC64le. The fix is simple, and should not brake anything else. I did it together with my colleague Johannes (@parttimenerd) who will continue to help maintaining the port.

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -228,7 +228,8 @@ loaded:
 
         // Find the bounds of the Global Offset Table
         ElfSection* got = findSection(SHT_PROGBITS, ".got.plt");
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for pointing this out.
I'll verify the fix and integrate it shortly.

apangin:
I verified that `.plt` section has `NOBITS` attribute on ppc64le, but `PROGBITS` on other supported platforms.
So the added condition is true only on ppc64le, which is fine.
I added a few comments and merged the changes. Thanks.

PR #544, opened 2022-02-03
Improved arguments description during error's construction

toktarev:
"Unknown argument" is very non-informative description.

Tried to improve usability.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -118,7 +118,8 @@ Error Arguments::parse(const char* args) {
     }
     char* args_copy = strcpy(_buf + EXTRA_BUF_SIZE, args);
 
-    const char* msg = NULL;    
+    const char* msg = NULL;
+    const char* unknownArgument = NULL;
 
     for (char* arg = strtok(args_copy, ","); arg != NULL; arg = strtok(NULL, ",")) {
         char* value = strchr(arg, '=');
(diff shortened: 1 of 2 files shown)


toktarev:
PR was motivated by the current problem.

I see "Unknown Argument" but don't understand which argument is actually wrong.

toktarev:
Close for now, looks like thing are harder than it looks like at first time.

toktarev:
Checked, it works.

toktarev:
@apangin ,
I realize that you probably do not consider this PR as something useful.

Or my code breaks your thoughts about my programming skills and how the code should be written.

Anyway - let's just close PR if it is hopeless :)

apangin:
Unfortunately, the problem is not as trivial as it may seem.

- The implementation is not quite correct: it violates the "rule of three" and thus does not work with the assignment operator.
- `Error` was meant a simple immutable value type that can be compared, returned from a function, etc. Passing arbitrary dynamic messages requires bigger refactoring, which I didn't want to do at the moment. The proposed ad-hoc fix does not look best to me, as it covers just one specific error and requires duplicating `unknownArgument = NULL;` line for every other case.

toktarev:
---The implementation is not quite correct: it violates the "rule of three" and thus does not work with the assignment operator.

I agree. But current implementation always passes some char * constant into the Error and thus we don't care how to release it. Adding dynamic messages we should in some cases release this message, but in other cases don't.
I agree - looked hacky, but didn't find another way taking into account the current design.

---unknownArgument = NULL;
I agree - this is evil, but It looked like the smallest evil among other fast fixes.

---Passing arbitrary dynamic messages requires bigger refactoring, which I didn't want to do at the moment

Could you please example the actual reasons  - why you do not want to do this refactoring?

Jongy:
> The implementation is not quite correct: it violates the "rule of three" and thus does not work with the assignment operator.

Hmm, I have never seen the "rule of three" applied on programming this way... Out of curiosity, do you mean https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming) ?

apangin:
@Jongy Yes. Since the given PR added resource management in constructor/destructor, it also needs copy assignment operator, or the below code will crash:
```
{
    Error e = Error("abc", "def");
    Error e1 = e;
}
```
(there are actual usages of Error assignment throughout the async-profiler code)

toktarev:
I closed this PR because of "unknownArgument=NULL".

toktarev:
The idea of new refactoring:
We replace abstraction Error to abstraction Status.

We work with our standard messages using their codes (and corresponding codes) and as result we don't need to think about if we should release this char * or not.

Our constructor looks like Status(Code code, Slice message).

Where the Slice is the wrapper under "char * msg, size_t len"

PR #545, opened 2022-02-04
Distinguish inlined frames

parttimenerd:
Async profiler doesn't distinguish frame types (compiled/inlined/interpreted) for HotSpot, which is relevant information for developers. With this change the frame type is tracked at least for top frames.

The basic principle used to gather the frame type is simple: If the pc of topmost java frame is within the interpreter, the top frame is interpreted. If not, the methodID of the topmost java frame is looked up in the code cache model. Then, the vframe stack trace obtained from the VM is walked till the methodID is found. This is the topmost physical frame. The ones possibly found above this are virtual frames of inlined methods.

The Trie is also extended to take the type information on the account (and to count the number of occassions a node is considered as inlined/compiled/interpreted), this information is also presented in the flamegraph: A node is colored based on the type of the majority of occasions.

The JFR output furthermore supports the collected type information.
There are a few extensions to the simple output functions (summary, flat and traces) which are self-explanatory.

This change is compatible with the OpenJ9 support.

This PR is the work of @ghaug and me, the differences to the prior PR (#209) are:
- Less overall modifications
- Flamegraph generation is far simpler (no gradients) 
- JFR support

image
_Parf of a flamegraph created for a program running on OpenJDK 8, the turquiose colored frames are inlined_


apangin:
Interesting. I've been working on a similar feature at the moment.

parttimenerd:
What do you mean with similar? Maybe we could combine our efforts?

apangin:
I'm researching the possibility of stack walking without AsyncGetCallTrace, which will provide information about frame kinds for the entire stack. As the first step, I planned to obtain JIT information for the top frames, just like you suggest in this PR. I'll definitely check your achievements in this direction, thank you for your work.

parttimenerd:
Seems to be superseded by https://github.com/jvm-profiling-tools/async-profiler/commit/d7a2a4fc8ba5aecd9b6d4bb4b5f1609d37ab4a8c

PR #550, opened 2022-02-11
Only use pinned memory for Kernel stack traces

luhenry:
This change removes the requirement for pinned memory when only
gathering user-mode stacktraces.

Pinned memory is a precious resource on some systems (some Linux
distributions only allow for a few hundred kilobytes) and changing the
limite requires admin privileges. This is particularly problematic when
running in a container without CAP_SYS_ADMIN.

In these limited cases, it's then better not to rely on pinned memory
when not required, and instead end up using perf only as a CPU-time
based trigger signaling mechanism.

### DIFF
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -48,9 +48,9 @@ enum Counter {
 };
 
 enum Ring {
-    RING_ANY,
-    RING_KERNEL,
-    RING_USER
+    RING_KERNEL = 1 << 1,
+    RING_USER   = 1 << 2,
+    RING_ANY    = RING_KERNEL | RING_USER,
 };
 
 enum Style {
--- a/src/itimer.cpp
+++ b/src/itimer.cpp
@@ -36,10 +36,18 @@ void ITimer::signalHandler(int signo, siginfo_t* siginfo, void* ucontext) {
 void ITimer::signalHandlerJ9(int signo, siginfo_t* siginfo, void* ucontext) {
     if (!_enabled) return;
 
-    J9StackTraceNotification notif;
-    notif.num_frames = _cstack == CSTACK_NO ? 0 : _cstack == CSTACK_DWARF
-        ? StackWalker::walkDwarf(ucontext, notif.addr, MAX_J9_NATIVE_FRAMES)
-        : StackWalker::walkFP(ucontext, notif.addr, MAX_J9_NATIVE_FRAMES);
+    J9StackTraceNotification notif = { .num_frames = 0 };
+    if (_cstack != CSTACK_NO) {
+        switch (_cstack) {
+        case CSTACK_DWARF:
+            notif.num_frames += StackWalker::walkDwarf(ucontext, notif.addr + notif.num_frames, MAX_J9_NATIVE_FRAMES - notif.num_frames);
+            break;
+        case CSTACK_FP:
+        case CSTACK_DEFAULT:
+            notif.num_frames += StackWalker::walkFP(ucontext, notif.addr + notif.num_frames, MAX_J9_NATIVE_FRAMES - notif.num_frames);
+            break;
+        }
+    }
     J9StackTraces::checkpoint(_interval, &notif);
 }
 
--- a/src/perfEvents.h
+++ b/src/perfEvents.h
@@ -46,7 +46,7 @@ class PerfEvents : public Engine {
     const char* title();
     const char* units();
 
-    static int walk(int tid, const void** callchain, int max_depth);
+    static int walkKernel(int tid, const void** callchain, int max_depth);
     static void resetBuffer(int tid);
 
     static bool supported();
--- a/src/perfEvents.h
+++ b/src/perfEvents.h
@@ -46,7 +46,7 @@ class PerfEvents : public Engine {
     const char* title();
     const char* units();
 
-    static int walkKernel(int tid, const void** callchain, int max_depth);
+    static int walk(int tid, const void** callchain, int max_depth);
     static void resetBuffer(int tid);
 
     static bool supported();
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -449,7 +449,7 @@ struct PerfEventType {
     ((perf_hw_cache_id) | PERF_COUNT_HW_CACHE_OP_READ << 8 | PERF_COUNT_HW_CACHE_RESULT_MISS << 16)
 
 PerfEventType PerfEventType::AVAILABLE_EVENTS[] = {
-    {"cpu",          DEFAULT_INTERVAL, PERF_TYPE_SOFTWARE, PERF_COUNT_SW_CPU_CLOCK},
+    {"cpu",          DEFAULT_INTERVAL, PERF_TYPE_SOFTWARE, PERF_COUNT_SW_TASK_CLOCK},
     {"page-faults",                 1, PERF_TYPE_SOFTWARE, PERF_COUNT_SW_PAGE_FAULTS},
     {"context-switches",            1, PERF_TYPE_SOFTWARE, PERF_COUNT_SW_CONTEXT_SWITCHES},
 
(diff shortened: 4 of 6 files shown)


jbachorik (approved):

apangin:
Thank you for working on async-profiler. While I understand the motivation of this change, I can hardly accept the PR in its current form.

To begin with, it breaks `--cstack lbr` feature.

Besides that, the idea of unconditional stack walking in userspace is questionalble. perf_events built-in callchain mechanism is fast and safe, so it generally should be preferred to manual stack walking.

Nevertheless, I agree that `mmap` is not needed when using `--cstack dwarf` along with `--alluser`. This part of the change definitely makes sense. I think we can also skip `mmap` for `--cstack fp --alluser`, if we add a new `--cstack perf` option (enabled by default).

In general, in restricted environments, we don't need to depend on perf_events at all. There is currently `-e itimer` mode for that, and I'm thinking of adding a new engine based on `timer_create(CLOCK_THREAD_CPUTIME_ID)`, which should be a more accurate alternative.

aalexand:
> and I'm thinking of adding a new engine based on timer_create(CLOCK_THREAD_CPUTIME_ID), which should be a more accurate alternative

FWIW, Golang recently [made a switch to that approach](https://github.com/golang/go/issues/35057).

luhenry:
> To begin with, it breaks --cstack lbr feature.

I'm not sure I understand how it breaks `--cstack lbr`. Today, if you pass this argument, you set `attr.exclude_callchain_user = 1;`.  `--cstack lbr` then only impacts Kernel stack unwinding.

With this change, `PerfEvents` still does the Kernel unwinding, still uses LBR when captured, and still only focuses on Kernel stack unwinding.

> Besides that, the idea of unconditional stack walking in userspace is questionalble. perf_events built-in callchain mechanism is fast and safe, so it generally should be preferred to manual stack walking.

Perf will only do FP-based stack unwinding in the kernel. It won't do DWARF ([Linus's position on DWARF is well known](https://lkml.org/lkml/2012/2/10/356)) and so is unlikely to ever support it. It also does stack unwinding in exactly the same way as it's done in user-space. Finally, as a signal is triggered nonetheless, it only changes the order of doing the unwinding before or after triggering the signal (before if doing Kernel space unwinding, after if doing user space unwinding).

Have you observed crashes related to FP-based or DWARF-based stack unwinding in async-profiler so far? It seems to me `async-profiler` is doing everything right to avoid a crash related to such issues with `SafeAccess::load` and checking the stack bounds.

I don't want this PR to block on that, so happy to go with a `--cstack perf` if you feel strongly about it.

> In general, in restricted environments, we don't need to depend on perf_events at all. There is currently -e itimer mode for that, and I'm thinking of adding a new engine based on timer_create(CLOCK_THREAD_CPUTIME_ID), which should be a more accurate alternative.

We've [implemented](https://github.com/DataDog/async-profiler/pull/23) such `setitimer`+`timer_create` mode, but we still observe strong syscall-bias. This is due to the underlying Kernel's POSIX timers not being tightly integrated with the scheduler as `perf_event_open` is for accurate CPU time measurements. It also suffers from time between samples being aligned on jiffies (multiple of 4ms, so 8ms or 12ms when requesting 10ms intervals).

We are then not relying on `itimer` mode for environments which don't have `perf_event_open` and would rather stay with the existing fallback to `WallClock` as it is more similar to how it's done in JFR. We also haven't observed syscall-bias with this approach, which is much more problematic from our point of view.

apangin:
> --cstack lbr then only impacts Kernel stack unwinding

Not really. With `PERF_SAMPLE_BRANCH_STACK` option, the branch stack is dumped to the ring buffer *in addition* to the callchain. It is then used to build a userspace call trace.

luhenry:
> Not really. With PERF_SAMPLE_BRANCH_STACK option, the branch stack is dumped to the ring buffer in addition to the callchain. It is then used to build a userspace call trace.

So do I understand correctly that despite `attr.exclude_callchain_user = 1;` at https://github.com/jvm-profiling-tools/async-profiler/pull/550/files#diff-7913edbe4a15c71181ca1593910bac5bcbf6d5c81c12c2f394c33111653357f1L591 in async-profiler today, the LBR is going to gather user-space LBR records?

luhenry:
I've updated the PR to take into account the case of `cstack=lbr`. Now, the cases for various values of `cstack` are the following:
- `lbr`: it uses `PerfEvent::walk` like it would previously for both kernel and user space callchains
- `fp`: it uses `PerfEvent::walk` for kernel callchains and `StackWalker::walkFP` for user callchains
- `dwarf`: it uses `PerfEvent::walk` for kernel callchains and `StackWalker::walkDwarf` for user callchains
- `default`: same as `fp` for `event_type == 0` (aka CPU profiling)
- `no`: it doesn't capture any native frames.

luhenry:
@apangin how could I help verify that this works as expected?

apangin:
What makes me feel uncomfortable in this PR is that it changes the default mechanism for no apparent reason.
In the most common use case - with the default stack walking mode and when the kernel traces are available - the existing kernel stack walker works well, and I hardly want to change it on thousands servers in our production environment.

Kernel callchain is stored in a ring buffer anyway, so I don't see a reason why not let it get the user callchain as well. This is fast and reliable, and uses exactly the same amount of mmap'ed memory.

As far as I understand the motivation of this PR, the original goal was to avoid mmap'ing perf fds in --all-user mode. So why not change the stack walker for this mode only?

BTW, was there a particular reason for replacing `PERF_COUNT_SW_CPU_CLOCK` with `PERF_COUNT_SW_TASK_CLOCK`? (Just curious, as I thought they behave the same way).

luhenry:
The reason is to lower the barrier to enable `perf_event_open`-based cpu sampling. Because perf is such a higher quality sampler than `itimer` and `wallClock` (due to the integration in the scheduler), I would bias towards making it as easy to use as possible. The fact that we then need to do stack unwinding in user-space is an unfortunate (but necessary) result of that.

In our continuous profiler use-case, most of our customers don't need the Kernel stacktraces. It's a very nice to have for sure, but not a requirement. Blocking the most useful use case from using the highest-quality CPU sampler then doesn't seem like the best tradeoff.

> As far as I understand the motivation of this PR, the original goal was to avoid mmap'ing perf fds in --all-user mode. So why not change the stack walker for this mode only?

I'm not sure I understand. With this PR, with `--all-user`, perf is pretty much only used as a signal-triggering mechanism as the Kernel component doesn't capture any stacktrace and simply triggers the `SIGPROF` signal. The `StackWalker` is then in charge of walking the stack in user-space.

> BTW, was there a particular reason for replacing PERF_COUNT_SW_CPU_CLOCK with PERF_COUNT_SW_TASK_CLOCK? (Just curious, as I thought they behave the same way).

Per the Linux kernel sources and [various discussions](https://stackoverflow.com/questions/23965363/linux-perf-events-cpu-clock-and-task-clock-what-is-the-difference), `PERF_COUNT_SW_CPU_CLOCK` is more akin to wall time profiler while `PERF_COUNT_SW_TASK_CLOCK` is closer to cpu time profiler.

apangin:
Sorry if I wasn't clear enough. Let me just implement this myself - hopefully, the code will explain better.
I commited the fix that should address your issue, while keeping the behavior of the default most common use case unchanged.

Now, if you want async-profiler to walk the native stack on its own, explicitly specify `cstack=fp` or `cstack=dwarf`.
In addition, if `fp` or `dwarf` mode is accompanied with `alluser` option, no perf ring buffer will be created at all.

Meanwhile, when kernel symbols are available and `cstack` is not set, async-profiler will create a ring buffer for each perf fd and let kernel fill it with the entire native callchain.

> most of our customers don't need the Kernel stacktraces

Likely because they are not aware of them :) Before async-profiler, people [were happy](https://twitter.com/blueos/status/997546653900034048) with safepoint based profilers that show no native frames, no VM threads, have large overhead, etc.

PR #551, opened 2022-02-11
Add loglevel argument

luhenry:
It allows to set the minimal log level to output. This is particularely
helpful if you're using async-profiler in production and where you don't
want to clubber your output with not always relevant log information.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -254,6 +254,12 @@ Error Arguments::parse(const char* args) {
             CASE("log")
                 _log = value == NULL || value[0] == 0 ? NULL : value;
 
+            CASE("loglevel")
+                if (value == NULL || value[0] == 0) {
+                    msg = "loglevel must not be empty";
+                }
+                _loglevel = value;
+
             CASE("fdtransfer")
                 _fdtransfer = true;
                 _fdtransfer_path = value;
--- a/src/javaApi.cpp
+++ b/src/javaApi.cpp
@@ -67,6 +67,8 @@ Java_one_profiler_AsyncProfiler_execute0(JNIEnv* env, jobject unused, jstring co
         return NULL;
     }
 
+    Log::open(args._log, args._loglevel);
+
     if (!args.hasOutputFile()) {
         std::ostringstream out;
         error = Profiler::instance()->runInternal(args, out);
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -18,13 +18,20 @@
 #include "log.h"
 #include "profiler.h"
 
-
-const char* const Log::LEVEL_NAME[] = {NULL, "TRACE", "DEBUG", "INFO", "WARN", "ERROR"};
+const char* const Log::LEVEL_NAME[] = {
+    "TRACE",
+    "DEBUG",
+    "INFO",
+    "WARN",
+    "ERROR",
+    "none",
+    NULL
+};
 
 FILE* Log::_file = stdout;
+LogLevel Log::_level = LOG_TRACE;
 
-
-void Log::open(const char* file_name) {
+void Log::open(const char* file_name, const char* level) {
     if (_file != stdout && _file != stderr) {
         fclose(_file);
     }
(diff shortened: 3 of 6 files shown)


src/arguments.cpp
```diff
@@ -254,6 +254,12 @@ Error Arguments::parse(const char* args) {
             CASE("log")
                 _log = value == NULL || value[0] == 0 ? NULL : value;
 
+            CASE("loglevel")
```
toktarev, line 257: I predict that after a couple of years this method will contain more than 500 "CASE" blocks.
Someone from the contributors really needs to do refactoring here using normal patterns (factory and so on ...)

src/arch.h
```diff
@@ -23,19 +23,18 @@ typedef unsigned short u16;
...
-
-static inline u64 loadAcquire(u64& var) {
+template <typename T>
+static inline T loadAcquire(T& var) {
     return __atomic_load_n(&var, __ATOMIC_ACQUIRE);
 }
 
-static inline void storeRelease(u64& var, u64 value) {
+template <typename T>
+static inline void storeRelease(T& var, T value) {
```
jbachorik, line 37: These changes don't seem to be related to the log level handling.
Can they be left out?

apangin: I agree that this change looks unrelated to the logging stuff.

I don't mind having `storeRelease` template for arbitrary types, but adding this would make sense with replacing all other `__atomic_store_n` usages with this template (which, I think, is better to do in a separate PR).

src/log.cpp
```diff
@@ -18,16 +18,22 @@
...
 
-
-const char* const Log::LEVEL_NAME[] = {NULL, "TRACE", "DEBUG", "INFO", "WARN", "ERROR"};
+const char* const Log::LEVEL_NAME[] = {
+#define LOG_LEVEL_STR(level) #level,
+LOG_LEVELS(LOG_LEVEL_STR)
+#undef LOG_LEVEL_STR
+    "none",
+    NULL
+};
```
apangin, line 27: I'm not a fan of higher order macros (though they are often used, for example, in HotSpot code).
They definitely make the code harder to read and they are not well supported in IDEs.
I can jusify such macros when they significantly reduce the amount of boilerplate, but this is not the case here.

apangin (commented):
Thank you for the contribution. I like the idea of `loglevel` option.
Please find my comments inline.

src/log.cpp
luhenry: I've removed the higher order macro.

src/arch.h
luhenry: I've reverted the template on `storeRelease` and replaced the use with `__atomic_store_n`.

src/javaApi.cpp
```diff
@@ -37,6 +37,7 @@ Java_one_profiler_AsyncProfiler_start0(JNIEnv* env, jobject unused, jstring even
         args._event = event_str;
         args._interval = interval;
     }
+    Log::open(args._log, args._loglevel);
```
apangin, line 40: Was this added intentionally? `start0` does not accept arguments other than `event` and `interval`, so `_log` and `_loglevel` are always empty here.

src/log.cpp
```diff
@@ -62,21 +85,55 @@ void Log::log(LogLevel level, const char* msg, va_list args) {
...
+    }
+
     va_list args;
     va_start(args, msg);
     log(LOG_INFO, msg, args);
     va_end(args);
 }
 
 void Log::warn(const char* msg, ...) {
+    if (LOG_WARN < _level) {
```
apangin, line 122: `log` function does not only print to a file/console, but also issues a corresponding JFR event, which is useful regardless of `loglevel`.

src/javaApi.cpp
luhenry: I've removed this one.

src/log.cpp
luhenry: I've moved that check into `Log::log`

apangin:
Thank you for the fixes; I merged the PR.
The new argument should be also mentioned in [the list](https://github.com/jvm-profiling-tools/async-profiler/blob/v2.7/src/arguments.cpp#L52), but I'll just add it myself.

PR #568, opened 2022-03-25
Add '--safe-mode' to the usage description.

vyazelenko:


franz1981 (approved):
Recently used, good idea!

apangin:
I intentionally left `--safe-mode` undocumented. This is an internal option for debugging; it should never be used normally.
If something does not work well without `--safe-mode` option - this indicates a bug that needs to be reported and fixed.

Furthermore, the meaning of `--safe-mode` bits can be changed in any version, and in fact, it has already been changed in v2.8.

vyazelenko:
Understood.

PR #576, opened 2022-04-03
Context ID functionality

krzysztofslusarski:
Adding two operations to Java API:
- ```setContextId(long contextId)```
- ```clearContextId()```

Use case:
There are java applications that want to achieve better performance by distributing the single request work into multiple threads. In such cases it is hard to find in profiler results where the time is wasted, since you have no information which thread executed which request. To make it traceable I would like to use external correlation id, so it is generated by client before passing work to other threads, then the worker thread would do:
```java
asyncProfiler.setContextId(correlationId);
actualWork();
asyncProfiler.clearContextId();
```
The context id is passed to custom field on execution sample, so we can post-filer it.

Other use cases I see is reactive programming, and in the future, loom project, distributed systems...

This PR is not finished, I just want to know, what do you think @apangin? If you like that functionality I can add this field to other profiling event.


apangin:
There have been many discussions around the profiling context recently - this in indeed a demanded feature. The problem is quite broad though, it's not just a matter of a thread-local string.

I haven't yet come up with a proper approach - this will be the topic of investigation. However, in the first approximation, I think, the context may include multiple values with different tags. An API should provide a control on what tags to record in the profile.

Implementation-wise I'd try to avoid JNI calls on the fast path, as the context may change quite often, and the JNI overhead can be significant. Also, AFAIK C++ thread locals are not async signal safe.

krzysztofslusarski:
> However, in the first approximation, I think, the context may include multiple values with different tags. An API should provide a control on what tags to record in the profile.

Sure, I can extend the implementation so we have context map where key would be mapped to JFR attribute key.

We can also have two maps, one global, one thread local.

> I'd try to avoid JNI calls on the fast path

I tried, but we need to pass that value to signal handler somehow. Either it is stored on C++ level or java level. Is it possible to pass java String to C++ without JNI? Or is there a signal safe way to get String from Java heap (like static fields)?

> C++ thread locals are not async signal safe

A map like ```std::map<int, std::string> _thread_names``` with guarding mutex would be a better solution? Or we should have that context at Java level?

luhenry:
Hi, chiming in a bit as I've worked on trying to integrate similar feature in the OpenJDK directly. I've just filled up a JEP at https://bugs.openjdk.java.net/browse/JDK-8284453, and I would love your feedback on it (not necessarily on this PR). I then wonder how similar the approach for async-profiler should be.

apangin:
@luhenry Thank you for submitting and linking the JEP - it makes a lot of sense.
I support the idea of supplementing events with a single context ID instead of the full contents. Otherwise the size of the event record is out of control, and contexts cannot be reused.

I think we can benefit from non-string context values in some cases, e.g. a stack trace saved in a context can automatically solve the problem of asynchronous stack traces mentioned in #575. On the other hand, a stack trace ID can be as well encoded as a string.

krzysztofslusarski:
@apangin please let me know if you want me to continue my work on that PR. If yes, just please, provide me answers from my previous comment and keep in mind that I'm Java developer. this is my first C++ PR in my life ;)

vasi-stripe:
> C++ thread locals are not async signal safe.

Thread locals are indeed not async signal safe in the general case! But [`initial-exec` ones are](https://maskray.me/blog/2021-02-14-all-about-thread-local-storage#async-signal-safe-tls), so that's an option.

vasi-stripe:
The GCP Java agent (also Apache 2.0) seems to use initial-exec thread locals this way: https://github.com/GoogleCloudPlatform/cloud-profiler-java/blob/master/third_party/javaprofiler/accessors.h#L83-L103

krzysztofslusarski:
@vasi-stripe thanks for that tip.

apangin:
@krzysztofslusarski
> please let me know if you want me to continue my work on that PR

If you'd like to, you're welcome to propose a solution. But the entire "profiling context" feature (I mean, something similar to what Ludovic suggested) may appear a lot more complex than the functionality you originally needed.

>  Is it possible to pass java String to C++ without JNI? Or is there a signal safe way to get String from Java heap (like static fields)?

I thought of a direct ByteBuffer to share the context data (or at least, the context ID) between Java and native. But if it helps simplicity, we can stick to JNI for the first implementation.

> A map like std::map<int, std::string> _thread_names with guarding mutex would be a better solution?

Neither `std::map` nor a mutex are safe to use inside a signal handler. pthread_getspecific/setspecific can be used with some caveats. Initial-exec TLS is not supposed to work with dlopen'ed libraries at all, but glibc seems to have a small surplus in static TLS specifically for these cases.

> Or we should have that context at Java level?

Yes, I suppose contexts could be managed at Java level and later synchronized with the native at convenient safe points. But as I said, I haven't come up with a complete solution yet, so implementation details are discussable.

krzysztofslusarski:
@apangin Thank you for your comment, I will try, we will see if I'm able to do it properly :)

vasi-stripe:
FWIW, I would find it valuable to just have an int64, which I could then interpret as I wish. Obviously it would be even better to have a full tagging solution, but it may be worth starting with something simple first.

src/profiler.cpp
```diff
@@ -1075,6 +1088,30 @@ Error Profiler::start(Arguments& args, bool reset) {
...
+
+    int size = strlen(ecid);
+    char *duplicate = (char*) malloc( size+ 1 );
+    strcpy(duplicate, ecid);
+    int status = pthread_setspecific(local_ecid_key, duplicate);
+    return status == 0 ? Error::OK : Error("Cannot set pthread_setspecific");
+}
+
+const char* Profiler::getEcid() {
+    pthread_once(&local_ecid_key_once, createEcidKey);
```
vasi-stripe, line 1105: I'm not sure if this is async-signal-safe.

krzysztofslusarski:
@apangin I switched to ```pthread_getspecific/setspecific```. Can you please look at this implementation if it is safe? 

Should I add a switch to enable this functionality? Like ```--conntext```? If it's not enabled then the ```pthread_getspecific``` won't be invoked.

For a bigger context, I thought about simple struct:
- either with array of ```char*```, so we can store attributes by numbers, the api would be ```profiler.setContext(int, String)``
- either with array of tuples ```<char*, char*>``` with API ```profiler.setContext(String, String)``
The side question - should that array grow in runtime, or for the first implementation we use static size?

src/profiler.cpp
krzysztofslusarski: Do you know if ```pthread_key_create``` should be invoked in every thread or just once in some static block? 

I found both implementation in the internet, and I'm not sure if this line is really needed.

luhenry:
Instead of using a Thread-Local, you can use the same approach as https://github.com/jvm-profiling-tools/async-profiler/blob/743959cde965d592e5ccee599a1f1fce9cc25f1d/src/perfEvents_linux.cpp#L806-L811. That's guaranteed to be thread-safe.

src/profiler.cpp
```diff
@@ -1075,6 +1088,30 @@ Error Profiler::start(Arguments& args, bool reset) {
...
 
+Error Profiler::setEcid(const char* ecid) {
+    pthread_once(&local_ecid_key_once, createEcidKey);
+
+    void *value = pthread_getspecific(local_ecid_key);
+    destroyEcidKey(value);
+
+    int size = strlen(ecid);
+    char *duplicate = (char*) malloc( size+ 1 );
+    strcpy(duplicate, ecid);
```
AdrK, line 1099: You already know the size, so you can use `memcpy` to avoid overhead of calculating the string size again (which is done inside the `strcpy`).

src/profiler.cpp
```diff
@@ -1075,6 +1088,30 @@ Error Profiler::start(Arguments& args, bool reset) {
     return error;
 }
 
+Error Profiler::setEcid(const char* ecid) {
+    pthread_once(&local_ecid_key_once, createEcidKey);
+
+    void *value = pthread_getspecific(local_ecid_key);
+    destroyEcidKey(value);
```
AdrK, line 1095: How do you know if there is always something under this pointer that has to be freed? Even in the case when this function is being called for the first time?

krzysztofslusarski: From a specification of [pthread_key_create](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_key_create.html):

> Upon key creation, the value NULL shall be associated with the new key in all active threads. Upon thread creation, the value NULL shall be associated with all defined keys in the new thread.

Since I use ```pthread_key_create``` on both get/set I thought it is safe to assume that it is nulled. The destroy function:
```c
static void destroyEcidKey(void* data) {
    if (data) {
        free(data);
    }
}
```

So here is a check if the pointer points to something.

That's for a theory, I'm using that feature for a week, and sometimes I get some garbage as ECID, so something must not working as supposed. I put Java UUID as ECID and here is a sample of what I get:
image

krzysztofslusarski: First column is an ECID.

krzysztofslusarski: To overcome this I'm post filtering the results:
https://github.com/krzysztofslusarski/jvm-profiling-toolkit/blob/309f9b3779df7e0d4f82f9d87847627f24d99124/jfr-parser/src/main/java/pl/ks/jfr/parser/JfrParserImpl.java#L336

krzysztofslusarski:
@apangin can you please comment this https://github.com/jvm-profiling-tools/async-profiler/pull/576#issuecomment-1097021656 - i would like to continue work on that PR.

BTW: this functionality is already used by me on a daily basis. I've already found some issues that caused high latency from time to time in distributed system.

src/profiler.cpp
apangin: `pthread_key_create` needs to be called just once, and not inside a signal handler

src/profiler.h
```diff
@@ -191,6 +192,8 @@ class Profiler {
     void shutdown(Arguments& args);
     Error check(Arguments& args);
     Error start(Arguments& args, bool reset);
+    Error setEcid(const char* ecid);
+    const char* getEcid();
```
apangin, line 196: `ECID` does not seem to be a commonly used abbreviation.

src/profiler.cpp
apangin: `strdup`

apangin: or just [`strdup`](https://man7.org/linux/man-pages/man3/strdup.3.html)

src/flightRecorder.cpp
```diff
@@ -1074,6 +1074,7 @@ class Recording {
         buf->putVar32(tid);
         buf->putVar32(call_trace_id);
         buf->putVar32(event->_thread_state);
+        buf->putUtf8(event->_ecid);
```
apangin, line 1077: This will not work with long strings (>120 characters).
As mentioned in [the comment](https://github.com/jvm-profiling-tools/async-profiler/pull/576#issuecomment-1092314951), it would be better to have a numeric context ID instead.

krzysztofslusarski: I believe you are right, for now >50% of overhead of ```setEcid``` is done by ```jni_GetStringUTFChars```, so that overhead can be eliminated.

src/profiler.h
krzysztofslusarski: I believe it is used in many Oracle technologies, but I can switch the name to simply correlationId.

vasi-stripe: Something like "ContextID" might be clear?

krzysztofslusarski:
@apangin I've switched to ```jlong``` with context id. In a simple test that solution works. I will test it under heavy load and let you know about overhead. The biggest overhead measured with ```String``` was 2,5%, I believe now it will be much smaller, since ```malloc``` is executed only once and there is no ```jni_GetStringUTFChars```. If you have any other comment just let me know.

src/profiler.cpp
```diff
@@ -1075,6 +1094,32 @@ Error Profiler::start(Arguments& args, bool reset) {
     return error;
 }
 
+Error Profiler::setContextId(u64 contextId) {
+    void *value = pthread_getspecific(local_context_key);
+
+    int status = 0;
+
+    if (!value) {
+        Context *context = (Context*) malloc(sizeof(Context));
```
AdrK, line 1103: Why is there still a need to use malloc?

krzysztofslusarski: ```pthread_setspecific``` needs a pointer to a thread-local variable. Is it safe to add a pointer to a ```jlong``` that is passed from JNI call?

krzysztofslusarski:
First overhead results 0,06% overhead - looks very promising. This is an additional overhead generated by native calls with ​~12870 inv per second per thread for each of ```setContextId``` and ```clearContextId```.

src/profiler.cpp
krzysztofslusarski: Code is not longer in use, resolving conversation.

src/flightRecorder.cpp
krzysztofslusarski: done

src/profiler.h
krzysztofslusarski: done

src/profiler.cpp
krzysztofslusarski: Code is not longer in use, resolving conversation.

src/profiler.cpp
krzysztofslusarski: done

src/profiler.cpp
AdrK: Well, a pointer is just a number, so I think that you could simply cast the `contextId` to the pointer type and use it as such. In this case though I would suggest to switch to `u32`.

krzysztofslusarski:
@apangin can you look at the API and the new implementation? Do you wish me to switch from ```u64``` to ```u32``` and remove the ```malloc```, or stay with the struct so we could evolve that functionality to (for example) multiple id's in context?

src/profiler.cpp
```diff
@@ -63,6 +64,20 @@ static J9WallClock j9_wall_clock;
...
+        free(value);
+    }
+}
+
+static void* createContext() {
+    (void) pthread_key_create(&local_context_key, destroyContext);
+    return NULL;
+}
+
+static void* init_local_context = createContext();
```
apangin, line 80: `__attribute__((constructor))` will make the intention clearer and will also avoid the unused variable
(but that's just a question of personal preferences)

src/profiler.cpp
apangin: I absolutely agree there is no need for memory allocation here.

`contextId` could remain `u64`, but what to do with Java API on 32-bit systems then? Either throw an exception if contextId does not fit `uintptr_t`, or just make it `u32` to avoid confusion, as @AdrK suggests. I don't mind either solution.

apangin:
@krzysztofslusarski Thank you for addressing the review comments. I added some more inline.
I believe `contextId` should remain a single number in any case. However, in future this number may encode multiple values. Consider it as an index in some dictionary of contexts. At some point this dictionary could be dumped in the output .jfr

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -208,8 +208,18 @@ private void filterThread(Thread thread, boolean enable) {
         }
     }
 
+    public void setContextId(long contextId) {
+        setContextId0(contextId);
+    }
+
+    public void clearContextId() {
+        clearContextId0();
```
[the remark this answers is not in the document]
krzysztofslusarski: Done

src/event.h
```diff
@@ -30,6 +30,7 @@ class Event {
 
 class ExecutionEvent : public Event {
   public:
+    u64 _context_id;
```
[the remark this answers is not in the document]
krzysztofslusarski: I moved fetching context id to ```flightRecorder.cpp```

src/instrument.cpp
```diff
@@ -616,6 +616,7 @@ void JNICALL Instrument::recordSample(JNIEnv* jni, jobject unused) {
 
     if (_interval <= 1 || ((atomicInc(_calls) + 1) % _interval) == 0) {
         ExecutionEvent event;
+        event._context_id = Profiler::instance()->getContextId();
```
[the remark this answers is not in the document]
krzysztofslusarski: I moved fetching context id to ```flightRecorder.cpp```

src/javaApi.cpp
```diff
@@ -55,6 +55,25 @@ Java_one_profiler_AsyncProfiler_stop0(JNIEnv* env, jobject unused) {
     }
 }
 
+extern "C" DLLEXPORT void JNICALL
+Java_one_profiler_AsyncProfiler_setContextId0(JNIEnv* env, jobject unused, jlong contextId) {
+    Error error = Profiler::instance()->setContextId(contextId);
+
+    if (error) {
+        JavaAPI::throwNew(env, "java/lang/IllegalStateException", error.message());
+    }
```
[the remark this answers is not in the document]
krzysztofslusarski: Now there can be an error when you run it on 32bit JVM

src/profiler.cpp
```diff
@@ -63,6 +64,20 @@ static J9WallClock j9_wall_clock;
 static ITimer itimer;
 static Instrument instrument;
 
+static pthread_key_t local_context_key;
+
+static void destroyContext(void* value) {
+    if (value) {
+        free(value);
+    }
```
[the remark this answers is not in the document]
krzysztofslusarski: Done

krzysztofslusarski:
@apangin I've fixed your comments. I also moved invocation of ```getContextId()``` to ```flightRecorder.cpp```. If you prefer previous approach (passing through ```Event``` field) just let me know.

Also added it to ```alloc``` and ```lock```.

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -208,8 +208,17 @@ private void filterThread(Thread thread, boolean enable) {
         }
     }
 
+    public void setContextId(long contextId) {
```
jbachorik, line 211: Can you add javadoc here explaining that `0` value is reserved for 'no-context' and as such should not be used to refer to a valid context?

jbachorik (changes requested):

src/flightRecorder.cpp
```diff
@@ -1075,6 +1075,7 @@ class Recording {
         buf->putVar32(tid);
         buf->putVar32(call_trace_id);
         buf->putVar32(event->_thread_state);
+        buf->putVar64(Profiler::instance()->getContextId());
```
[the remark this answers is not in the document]
krzysztofslusarski: > I apologise for complicating the matters

Nothing to apologies for, every comment that makes opensource software better is a good one. 

The other option would be to enable whole ```contextId``` feature by a flag that is disabled by default. If someone use ```contextId``` then there is a high chance that he/she is going to use different JFR viewer since JMC doesn't support ```contextId```. 

@apangin any thoughts on that?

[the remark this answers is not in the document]
jbachorik: > If someone use contextId then there is a high chance that he/she is going to use different JFR viewer since JMC doesn't support contextId

`contextId` is written as a standard event attribute. Since JFR format is self-describing you will be able to see the `contextId` attribute value for events in JMC. 

I would really strongly suggest using custom event IDs (above 10000 eg. so we have a plenty of headroom for any builtin native events as well as any user defined Java events) - it would make it less of a hack. Let's face it - the events are going to have a different structure now so it is just fair that they are distinguishable from the builtin ones.

And, as I already mentioned - we still can keep all the metadata so eg. the execution samples are picked by JMC to present the profiling view etc.

jbachorik:
> Inventing a completely new set of types just to add some optional data to existing events in order to make one particular tool happy - that's what seems odd to me.

Actually, this will be only a problem when you combine the platform and async-profiler recordings. But, it will be a slight problem nevertheless. So, if you ever load only the async-profiler part of the data you are golden. When you mix those recordings then any parser will have to make sense of the different schemas describing the same type ids.

If I want to be blunt it would be like overriding something like `T_INT` with a a special 'complex' int. Yes, you can do it. Yes, parsers will most probably deal with it. But it really does not seem right.

My proposal is to be a good guy from beginning and use different type ids when we are trying to overload the platform defined events. The metadata still needs to be maintained in async profiler - the only extra work is to use different long ids.

apangin:
async-profiler does not reuse IDs of JDK events. IDs may even intersect, as long as they are in different chunks, but async-profiler events never meet with JFR events in the same chunk. What matters is the event name. So I don't really see how changing `101` to `10101` would help anything. On the other hand, keeping the event IDs low (below 128) allows to encode them in one byte, thus reducing the overall size of the recording. Considering the above, I suggest to leave IDs as they were.

krzysztofslusarski:
> On the other hand, keeping the event IDs low (below 128) allows to encode them in one byte

That's probably wy my change to > 10000 didn't work. I didn't change the ```buf->put8```.

krzysztofslusarski:
@apangin Thank you :)

franz1981:
Thanks @krzysztofslusarski for this one, I'm so curious to make this to work for quarkus (the reactive stack on Red Hat - the company I work for)
Let me know if you have any idea on how to use it and I will happily setup something on our CI

krzysztofslusarski:
@franz1981 Do you want to use it on JVM or native image?

franz1981:
@krzysztofslusarski sorry, noticed the comment just now :O
JVM!

krzysztofslusarski:
I don't really see why this state of PR may not work on Quarkus at JVM, but I'm not a user of that framework. AsyncProfiler gives you an API with static methods:

```java
    public static AsyncProfiler getInstance() {
    // ...
    }

    public static synchronized AsyncProfiler getInstance(String libPath) {
    // ...
    }
```

Those return an instance of a class that has new methods.

krzysztofslusarski:
@apangin What about that PR? Are you going to merge it?

apangin:
Hi @krzysztofslusarski,
Yes, my plan is to merge this for the next major version of the profiler.
However, before non-trivial code contributions can be accepted, you'll be asked to sign a Contributor License Agreement (CLA). This is a standard practice in many companies: e.g., to make the first contribution to OpenJDK, one needs to sign the [OCA](https://oca.opensource.oracle.com/). Async-profiler employs the same practice. I'm clarifying details about the process - will keep you informed. Thank you for patience.

krzysztofslusarski:
Sure. Thank you.

src/flightRecorder.cpp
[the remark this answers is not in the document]
ocadaruma: Looking forward to this feature will be released!

Regarding this discussion, I realized that converter's JfrReader would not work if context id is introduced, because though the JFR format itself is self-describing, JfrReader hard codes the parsing logic
https://github.com/async-profiler/async-profiler/blob/v2.9/src/converter/one/jfr/JfrReader.java#L155-L158

Should we fix JfrReader as well?

moming00:
hello @krzysztofslusarski, any thought on how to save the contextId into JFR when using AsyncProfiler in attach mode?

krzysztofslusarski:
To save context ID you need change the code of your application to use Async-profiler Java API. The first time you execute ```setContextId``` you need to have the native library loaded. If you want to have that functionality without loading the native library on startup your application's code needs to look like this:

```java
if (shouldSetContextId()) {
    asyncProfiler.setContextId(...);
}
```

and the method ```shouldSetContextId()``` should return ```true``` only if library is loaded (maybe also if profilier is started). I didn't tested it but I believe you can get list of loaded native libraries with some reflection on class loader class, or you can do it the way https://github.com/async-profiler/async-profiler/blob/master/src/api/one/profiler/AsyncProfiler.java#L49-L52. To check if profiler is running you probably can use ```asyncProfiler.execute()``` method and pass ```status``` command. 

To be honest in my opinion that doesn't make sense. If you are changing you application code to work with context ID Async-profiler you can load the library at startup as well.

moming00:
> To save context ID you need change the code of your application to use Async-profiler Java API. The first time you execute `setContextId` you need to have the native library loaded. If you want to have that functionality without loading the native library on startup your application's code needs to look like this:
> 
> ```java
> if (shouldSetContextId()) {
>     asyncProfiler.setContextId(...);
> }
> ```
> 
> and the method `shouldSetContextId()` should return `true` only if library is loaded (maybe also if profilier is started). I didn't tested it but I believe you can get list of loaded native libraries with some reflection on class loader class, or you can do it the way https://github.com/async-profiler/async-profiler/blob/master/src/api/one/profiler/AsyncProfiler.java#L49-L52. To check if profiler is running you probably can use `asyncProfiler.execute()` method and pass `status` command.
> 
> To be honest in my opinion that doesn't make sense. If you are changing you application code to work with context ID Async-profiler you can load the library at startup as well.

thanks for that, will give it a try.

moming00:
> To save context ID you need change the code of your application to use Async-profiler Java API. The first time you execute `setContextId` you need to have the native library loaded. If you want to have that functionality without loading the native library on startup your application's code needs to look like this:
> 
> ```java
> if (shouldSetContextId()) {
>     asyncProfiler.setContextId(...);
> }
> ```
> 
> and the method `shouldSetContextId()` should return `true` only if library is loaded (maybe also if profilier is started). I didn't tested it but I believe you can get list of loaded native libraries with some reflection on class loader class, or you can do it the way https://github.com/async-profiler/async-profiler/blob/master/src/api/one/profiler/AsyncProfiler.java#L49-L52. To check if profiler is running you probably can use `asyncProfiler.execute()` method and pass `status` command.
> 
> To be honest in my opinion that doesn't make sense. If you are changing you application code to work with context ID Async-profiler you can load the library at startup as well.

@krzysztofslusarski I've tried the approach as you suggested. But it seems the Java API is unable to recognize the attached native library, the AsyncProfiler.getInstance() continues to fail with UnsatisfiedLinkError  exception.
Also I tried the methods mentioned in below link to retrieve all the loaded libraries after attaching AsyncProfiler to my process but returned me nothing.
https://stackoverflow.com/questions/1007861/how-do-i-get-a-list-of-jni-libraries-which-are-loaded

Any thoughts on this symptom?

krzysztofslusarski:
@apangin can we use Java API with externally attach profiler?

apangin:
Java API works when the profiler library is loaded by any means, including `-agentpath` argument or dynamic attach. But if the library is not loaded, `AsyncProfiler.getInstance()` will attempt to load it on its own and will throw `UnsatisfiedLinkError` when it cannot.

This seems reasonable to me. Do you propose a different behavior?

krzysztofslusarski:
@apangin , no, that looks ok to me, thanks for the answer.

krzysztofslusarski:
@moming00 I don't know what your issue is, sorry, I tested the behavior with small program:

```java
    public static void main(String[] args) throws IOException {
        System.out.println("Waiting for attaching profiler");
        System.in.read();
        AsyncProfiler.getInstance();
    }
```

When I attach the profiler from outside, while waiting on ```System.in.read()'``, and then debug the ```AsyncProfiler.getInstance()``` it's works. Id doesn't try to load it again.

moming00:
> @moming00 I don't know what your issue is, sorry, I tested the behavior with small program:
> 
> ```java
>     public static void main(String[] args) throws IOException {
>         System.out.println("Waiting for attaching profiler");
>         System.in.read();
>         AsyncProfiler.getInstance();
>     }
> ```
> 
> When I attach the profiler from outside, while waiting on `` System.in.read()'``, and then debug the ``AsyncProfiler.getInstance()``` it's works. Id doesn't try to load it again.

thanks @krzysztofslusarski for your time, this information already helped a lot. I'll just try to investigate further and see why getInstance cannot detect the loaded lib in my case. =)

apangin:
@mrk-andreev There were many discussions around contextual profiles lately, and more ideas were brought up. We are looking forward to adding support for tagged profiling events in async-profiler, however, it will be likely implemented quite differently from what is proposed in this PR.

Ideally, instead of a single numeric ID, we would like to support multiple tags in the form of arbitrary strings.
Context information should be decoupled from other JFR events on one hand, but on the other hand, all events should benefit from additional context automatically without a requirement for having a dedicated field in each event.
Finally, implementation of contextual profiling should not have any visible performance impact - in particular, we should avoid calling a native method on each context switch.

Once new API is added, it will be hard to roll it back without affecting customers, so we need to be very careful with our design decisions. With that in mind, I don't think we can integrate this PR in its current form.

heoq:
Hi! We somehow manage to do some contextual profiling by changing the thread name. More precisely, we append the thread name for different contexts with different tags and use some scripts for spliting the flamegrah with the `--thread` flag. Is this way OK for the performance?

apangin:
@heoq Context ID feature is discussed in #1289. We have a prototype available in the [context-events](https://github.com/async-profiler/async-profiler/tree/context-events) branch. You are welcome to give it a try, we'll be happy to hear your feedback.

As for the approach to change thread names - this is totally fine if it works for you.

PR #586, opened 2022-05-10
Target JDK 11

NeQuissimus:
These changes target JDK11 now that Java 7 is a deprecated target

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /test/*.class
 .vscode
 *.iml
+target/*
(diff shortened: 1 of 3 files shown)


apangin:
async-profiler works with JDK 7+. This is intentional.

Starting from a new major release (3.0), the minimum requirement will be probably JDK 8. A number of OpenJDK vendors offer JDK 8 support until 2030. There are still many customers running JDK 8 in their production systems, and I have no plans to prevent them from using async-profiler.

NeQuissimus:
Makes sense, I will close this then :)

PR #587, opened 2022-05-10
Add pprof converter

NeQuissimus:
Relates to https://github.com/jvm-profiling-tools/async-profiler/issues/343 and probably https://github.com/jvm-profiling-tools/async-profiler/issues/134

Assists by @dalehamel @bmansoob

This add a pprof converter.
Only CPU stats are carried over onto a minimal set of fields from the pprof proto definition.

We noticed flame graphs converted directly and those created via this new pprof and the Golang pprof tool contain slightly different percentages, which may be due to rounding errors along the way.

### DIFF
--- a/src/converter/jfr2pprof.java
+++ b/src/converter/jfr2pprof.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2022 Andrei Pangin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import one.jfr.ClassRef;
+import one.jfr.Dictionary;
+import one.jfr.JfrReader;
+import one.jfr.MethodRef;
+import one.jfr.StackTrace;
+import one.jfr.event.ExecutionSample;
+
+import one.proto.Proto;
+
+/**
+ * Convert a JFR file to pprof
+ * <p>
+ * Protobuf definition: https://github.com/google/pprof/blob/44fc4e887b6b0cfb196973bcdb1fab95f0b3a75b/proto/profile.proto
+ */
+public class jfr2pprof {
+
+    public static class Method {
+        public final byte[] name;
+
+        public Method(final byte[] name) {
+            this.name = name;
+        }
+
+        @Override
+        public int hashCode() {
+            return Arrays.hashCode(name);
+        }
+
+        @Override
+        public boolean equals(final Object other) {
+            if (null == other) {
+                return false;
+            }
+
+            if (other instanceof Method) {
+                final Method that = (Method) other;
+                return Arrays.equals(this.name, that.name);
+            } else {
+                return false;
+            }
+        }
+    }
+
+    private final JfrReader reader;
+
+    public jfr2pprof(final JfrReader reader) {
+        this.reader = reader;
+    }
+
+    public static final byte[] METHOD_UNKNOWN = "[unknown]".getBytes();
+
+    // Profile IDs
+    public static final int PROFILE_SAMPLE_TYPE = 1;
+    public static final int PROFILE_SAMPLE = 2;
+    public static final int PROFILE_LOCATION = 4;
+    public static final int PROFILE_FUNCTION = 5;
+    public static final int PROFILE_STRING_TABLE = 6;
+    public static final int PROFILE_TIME_NANOS = 9;
+    public static final int PROFILE_DURATION_NANOS = 10;
+    public static final int PROFILE_COMMENT = 13;
+    public static final int PROFILE_DEFAULT_SAMPLE_TYPE = 14;
+
+    // ValueType IDs
+    public static final int VALUETYPE_TYPE = 1;
+    public static final int VALUETYPE_UNIT = 2;
+
+    // Sample IDs
+    public static final int SAMPLE_LOCATION_ID = 1;
+    public static final int SAMPLE_VALUE = 2;
+
+    // Location IDs
+    public static final int LOCATION_ID = 1;
+    public static final int LOCATION_LINE = 4;
+
+    // Line IDs
+    public static final int LINE_FUNCTION_ID = 1;
+    public static final int LINE_LINE = 2;
+
+    // Function IDs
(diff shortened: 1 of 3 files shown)


dalehamel:
To test this, we used test/Target.class, and profiled for 5 seconds

```
java -cp test/ Target
```

Then in another terminal

```
./profiler.sh -d 5 -o jfr -f profile.jfr 21549 # the pid from above
```

giving this jfr output [profile.jfr.gz](https://github.com/jvm-profiling-tools/async-profiler/files/8662978/profile.jfr.gz) 

Running that through the existing flamegraph converter, we get [out.html.gz](https://github.com/jvm-profiling-tools/async-profiler/files/8662977/out.html.gz)

Which looks like:
Screen Shot 2022-05-10 at 11 37 54 AM

Running through the new converter we get [output.pb.gz](https://github.com/jvm-profiling-tools/async-profiler/files/8662976/output.pb.gz) and visualizing this with `go tool pprof -http=:8082 ./output.pb` we get a similar flamegraph (note that in this visualization, stacks "grow" down):


Screen Shot 2022-05-10 at 11 39 28 AM


Note that the percentages between the built-in flamegraph output and the pprof one are slightly off. My theory (haven't verified) is that the built-in one does the percentage based on samples seen, and the pprof one uses the time values.

Note that I gzipped all files here so that github would accept them.

apangin:
Thank you for the proposed PR. There is indeed a demand for such converter (even though I personally believe JFR format is much better for Java profiles). Please find my review comments inline.

apangin:
> Note that the percentages between the built-in flamegraph output and the pprof one are slightly off. My theory (haven't verified) is that the built-in one does the percentage based on samples seen, and the pprof one uses the time values.

Interesting. I thought that timestamp information is irrelevant for an aggregate flame graph. It would be nice to find out the actual reason.

dalehamel:
> Thank you for the proposed PR. There is indeed a demand for such converter (even though I personally believe JFR format is much better for Java profiles).

Yes this is more a matter of compatibility with other tools than a question of the merit of the format itself, as a number of systems for visualizing profiling data tend to accept pprof.

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,243 @@
...
+    // Line IDs
+    public static final int LINE_FUNCTION_ID = 1;
+
+    // Function IDs
+    public static final int FUNCTION_ID = 1;
+    public static final int FUNCTION_NAME = 2;
+
+    // De-dupe string table
+    private long addToStringTable(final List<String> stringTable, final String value) {
+        final int idx = stringTable.indexOf(value);
```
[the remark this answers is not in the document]
NeQuissimus: I still need the elements in order, so looks like a `LinkedHashMap` will do the trick

NeQuissimus:
I addressed all the review comments. Thank you for the close look. @apangin

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,257 @@
...
+
+        // Used to de-dupe
+        final Map<Method, Long> functions = new HashMap<>();
+        final Map<Method, Long> locations = new HashMap<>();
+
+        final List<ExecutionSample> samples = reader.readAllEvents(ExecutionSample.class);
+        final Dictionary<StackTrace> stackTraces = reader.stackTraces;
+        final long startTicks = reader.startTicks;
+
+        final Map<byte[], Integer> stringTable = new LinkedHashMap<>(); // Will retain order of insertion
```
apangin, line 133: Strictly speaking, the order of insertion does not matter as long as we can later order keys by the corresponding values.
I'm not sure how many strings this map can save at all. If an experiment shows that savings are negligible, I would get rid of this map entirely.

apangin, line 133: Strictly speaking, the order of insertion does not matter as long as we can later order keys by the corresponding values.
I'm not sure how many strings this map can save at all. If an experiment shows that savings are negligible, I would get rid of this map entirely.

NeQuissimus: I ran a few experiments, looks like the map is not worth it.

NeQuissimus:
@apangin I removed the string map altogether now and confirmed the output is still valid.
All the string table work is now inline directly into the `Proto` instance, we no longer need to iterate anything :)

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,243 @@
...
+import one.proto.Proto;
+
+/**
+ * Convert a JFR file to pprof
+ * <p>
+ * Protobuf definition: https://github.com/google/pprof/blob/44fc4e887b6b0cfb196973bcdb1fab95f0b3a75b/proto/profile.proto
+ */
+public class jfr2pprof {
+
+    public class Method {
```
apangin, line 41: Can be `static`

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,243 @@
...
+            final int sampleThreadId = jfrSample.tid;
+            final int sampleStackTraceId = jfrSample.stackTraceId;
+            final StackTrace stackTrace = stackTraces.get(sampleStackTraceId);
+            final long[] methods = stackTrace.methods;
+            final byte[] types = stackTrace.types;
+
+            // Need to reverse the order of the elements; index 0 must be the leaf node
+            int current = 0;
+
+            final double nanosSinceLastSample = (jfrSample.time - previousTime) * 1_000_000_000 / reader.ticksPerSec;
```
apangin, line 158: The result of `long / long` is stored into `double`

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,243 @@
...
+            final Proto sample = new Proto(1_000).field(SAMPLE_VALUE, (long)nanosSinceLastSample);
+
+            while (current < methods.length) {
```
apangin, line 161: Isn't it a regular `for` loop?

src/converter/jfr2pprof.java
```diff
@@ -0,0 +1,243 @@
...
+
+    // Line IDs
+    public static final int LINE_FUNCTION_ID = 1;
+
+    // Function IDs
+    public static final int FUNCTION_ID = 1;
+    public static final int FUNCTION_NAME = 2;
+
+    // `Proto` instances are mutable, careful with reordering
+    public void dump(final FileOutputStream out) throws Exception {
```
apangin, line 107: I'd change the type of `out` to `OutputStream`, as this method is not specific to `FileOutputStream`.

apangin (commented):
Thanks! Here is another portion of minor comments, hopefully, the last.

src/converter/jfr2pprof.java
NeQuissimus: I guess we want to round it into a `long` anyways...

apangin:
Merged! Thanks for the contribution.

apangin:
> We noticed flame graphs converted directly and those created via this new pprof and the Golang pprof tool contain slightly different percentages, which may be due to rounding errors along the way.

> Note that the percentages between the built-in flamegraph output and the pprof one are slightly off.

While refactoring converters in #905, I found that the pprof converter was broken: instead of cpu time or counter, it recorded timestamp delta as a sample value, whereas pprof renderers accumulated that deltas, producing a kind of random total sum. In benchmarks, where cpu load is stable, timestamp deltas are also stable, and therefore result profile appears more or less plausible. However, in real applications percentages could be totally random.

I rewrote pprof converter from scratch: now it aggregates samples as designed by the protocol and records either counter or cpu time (depending on `--total` parameter). As a result, output files are now much smaller in size, and percentages match those in .html and .jfr format.

PR #606, opened 2022-06-04
add support for asychronous invocation stacks

pnf:
Extend async-profiler to capture stacks that are provided by the program being profiled. This is intended to be used for stacks of awaiters in asynchronous code.

The application or library will invoke

setAwaitStackId(long replacementId, long signal, long insertionId)

right before launching a new future, and the arguments will be stored in a thread-local AwaitData structure. The insertionId should be the jmethodID of the method invoking the future's body, for example the super.run method a subclassed Runnable, and when the profiler finds this method in the stack, it will:

    Change frame.method_id to match the replacementId argument.
    Change frame.bci to BCI_AWAIT_INSERTION
    Store the value of the signal argument in the thread-local structure for retrieval by the application.
    When the run returns, the app calls

long getAwaitSampledSignal();

and checks if it matches the value of the signal argument above, indicating that a sample occurred while this method was running. If so, it will then invoke

saveAwaitFrames(int frameType, long[] ids, int nids)

where the ids array holds insertionId, followed by the current awaiter stack - which of course the application/library is responsible for maintaining. This will then be saved in _call_trace_storage. (frameType indicates whether the elements of ids are (const char*) pointers or jmethodIDs).

When dumping output or storing jfr, we search for frame.bci==BCI_AWAIT_INSERTION that match the 0th element of a stored await trace, and we insert its frames.

The failure modes are intended to be graceful. For example, we might flush to jfr after setting the BCI_AWAIT_INSERTION frame, but before the full await stack has been set. In that case, we simply won't have the extra frames inserted.

See https://github.com/pnf/asyncstacks for a working example. There, the insertionId is indeed the jmethodID of a Runnable#run, the signal the value of an atomic counter, and replacementId is the running long hash of the await stack, resulting in a flamegraph:
image

THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE APACHE LICENSE V.2.0. YOU MAY OBTAIN A COPY OF THE LICENSE AT https://www.apache.org/licenses/LICENSE-2.0.
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.


pnf:
Here's an example using Loom structured concurrency from Java 19: https://github.com/pnf/loom-async-profile.

As with the previous scala async/await example, the cpu is all consumed in a contrived method designed to get some native action onto the stack:
image
```
    static double burnCpu(long ms, double x)  {
        var t = System.nanoTime() + ms * 1000 * 1000;
        while(System.nanoTime() < t) {  // repeat loop as long as necessary to use up requested time
            for(int i=0; i < 10000; i++) { x = x * 1.1; x = Math.sin(x);  i -=1;   x *= (1.0/1.1); }
        }
        return x;
    }
    double expensive1(); return burnCpu(1000, 1.1); }
``` 
These expensive methods are then invoked asynchronously via scoped futures, e.g.
```
    double foo() throws InterruptedException {
        var s = newScope();
        var f1 = s.fork(this::expensive1);
        var f2 = s.fork(this::expensive2);
        s.join();
        return f1.resultNow() + f2.resultNow();
    }
```
resulting in sampled async-profiler stacks that comprise native, java and asynchronous future invocations like:
image
analogous to the stacks that the intellij debugger can display
image

This works by [overriding](https://github.com/pnf/loom-async-profile/blob/main/src/main/java/loom/examples/TaggedTaskScope.java#L101)  `StructuredTaskScope#fork` to accumulate a stack of such forks and embed them in an extension of `Callable` whose [call method](https://github.com/pnf/loom-async-profile/blob/main/src/main/java/loom/examples/TaggedTaskScope.java#L69) will call `AsyncProfiler.saveAwaitFrames` if and only if a sample occurred during the call.  This toy implementation could of course be optimized, as suggested by comments in the source.

The full flame graph of this example looks like:

image

apangin:
This PR has been inactive for a long time, so I'm closing it.
The idea looks exciting, but the implementation does not fit existing async-profiler design, unfortunately. I'm happy to discuss more if there is still interest in this feature.

PR #619, opened 2022-07-14
Add basic CI for release 

cuichenli:
Used this github action for personal usage and feeling this project may need it. But feel free to reject it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,71 @@
+name: Release
+on:
+  push:
+    tags:
+    - "v*.*.*"
+
+jobs:
+  build-mac:
+    runs-on: macos-latest
+    steps:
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for the proposal. This is a definitely useful change, however, it does not work for my case right now.

In order to provide highly compatible builds that work on a wide range of platforms (including older CentOS 6 with kernel 2.6), I use the specific environment and the specific toolchain that is not available out of the box in GitHub Actions. Yes, some customers still use an aged environment, and I need to support them, too.

Furthermore, macOS builds need to be signed and notarized, that also complicates GA-based release process.

apangin:
While the proposal makes total sense, I cannot make use of it for the mentioned reasons. Therefore, I'm closing PR without merge.

Issue #634, opened 2022-08-11
fdtransfer with net=host setting


Jongy:
fdtransfer creates a UDS socket in the net NS of the target program. The UDS is per PID. This has the hidden assumption that same PIDs (i.e different processes in different PID namespaces that share the same NSPID) do not share network namespaces.

If 2 processes do share network NS (e.g 2 Java containers running with --net=host parameter) - fdtransfer will run in the shared netns but it will actually target only one of the processes (because it is pairted with the real PID, not NSPID).

I have this case on a real deployment so I'm thinking what can be done.

An idea I got now while writing this post (@apangin let me know what you think about it, I can open a PR) is to include the `/proc/pid/ns/pid` inode (the number you get when you `readlink` this file). It is guaranteed to be different between different PID namespaces. Unprivileged processes can readlink their `ns/pid` and get the number, and fdtransfer can do so as well - together they provide a unique identifier across the system. This way, no matter if the net namespace is shared or not, it will be unique.

apangin:
The proposed solution may work, but are you sure it's universally applicable to all supported Linux flavors?
What if `/proc/self/ns/pid` does not exist or is not accessible inside a container?

On the other hand, the socket name does not need to relate to `pid` at all. It can be any unique string, as long as we pass the very same string both to `fdtransfer` and to the target process (as an agent argument).

Jongy:
> The proposed solution may work, but are you sure it's universally applicable to all supported Linux flavors? What if `/proc/self/ns/pid` does not exist or is not accessible inside a container?

If `/proc/self/ns/pid` doesn't exist - I don't read the number at all (and use 0) which makes sense because if no PID namespaces exist, then we don't need this number of uniqueness. I implemented in https://github.com/Jongy/async-profiler/commit/12e079989e85080e1ec00781d5ba3dcef320cf51.
AFAIK this will always be present in containers (its' under standard procfs), and the process' PID NS is always accessible for itself to read.

But, in any case ...

> On the other hand, the socket name does not need to relate to `pid` at all. It can be any unique string, as long as we pass the very same string both to `fdtransfer` and to the target process (as an agent argument).

I realized that as well while I was writing the code. This requires more changes in the code but will keep the API with OS simpler and will also ensure fdtransfers of different `profiler.sh` invocations no longer need fight for the same socket path (so in case an fdtransfer program remains hanging, it doesn't affect subsequent runs of async-profiler).

That solution makes more sense, I'll implement that

PR #635, opened 2022-08-14
Generate randomized fdtransfer path per async-profiler invocation

Jongy:
Closes: #634

This solves the problem described in #634.

I sanity-tested it by running `for p in $(pgrep java); do sudo ./profiler.sh -d 10 -e cpu -t -f xxx-$p.html -a -o flamegraph $p -i 50us --fdtransfer & ; done` on a system with 2 containerized Java apps; both are the same PID in their PID namespace. On `master`, it fails with `FdTransfer listen(): Address already in use`. On this branch, it passes.

### DIFF
--- a/profiler.sh
+++ b/profiler.sh
@@ -93,7 +93,7 @@ check_if_terminated() {
 
 fdtransfer() {
     if [ "$USE_FDTRANSFER" = "true" ]; then
-        FDTRANSFER_PATH="@async-profiler-$PID-$(od -An -N3 -i /dev/random | xargs)"
+        FDTRANSFER_PATH="@async-profiler-$(od -An -N3 -i /dev/random | xargs)"
         PARAMS="$PARAMS,fdtransfer=$FDTRANSFER_PATH"
         "$FDTRANSFER" "$FDTRANSFER_PATH" "$PID"
     fi
--- a/profiler.sh
+++ b/profiler.sh
@@ -93,7 +93,9 @@ check_if_terminated() {
 
 fdtransfer() {
     if [ "$USE_FDTRANSFER" = "true" ]; then
-        "$FDTRANSFER" "$PID"
+        FDTRANSFER_PATH="@async-profiler-$PID-$(od -An -N3 -i /dev/random | xargs)"
+        PARAMS="$PARAMS,fdtransfer=$FDTRANSFER_PATH"
+        "$FDTRANSFER" "$FDTRANSFER_PATH" "$PID"
     fi
 }
 
@@ -131,6 +133,7 @@ SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_BIN")" > /dev/null 2>&1; pwd -P)"
 JATTACH=$SCRIPT_DIR/build/jattach
 FDTRANSFER=$SCRIPT_DIR/build/fdtransfer
 USE_FDTRANSFER="false"
+FDTRANSFER_PATH=""
 PROFILER=$SCRIPT_DIR/build/libasyncProfiler.so
 ACTION="collect"
 DURATION="60"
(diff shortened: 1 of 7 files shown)


profiler.sh
```diff
@@ -93,7 +93,9 @@ check_if_terminated() {
 
 fdtransfer() {
     if [ "$USE_FDTRANSFER" = "true" ]; then
-        "$FDTRANSFER" "$PID"
+        FDTRANSFER_PATH="@async-profiler-$(od -An -N3 -i /dev/random | xargs)"
```
Jongy, line 96: `od -An -N3 -i /dev/random | xargs` is the most generic way I found to get a random number, stripped of spaces. (`$RANDOM` works in bash but not sh)

src/fdtransfer/fdtransferServer.cpp
```diff
@@ -68,7 +68,13 @@ bool FdTransferServer::bindServer(struct sockaddr_un *sun, socklen_t addrlen, in
         }
     }
 
-    if (bind(_server, (const struct sockaddr*)sun, addrlen) < 0 || listen(_server, 1) < 0) {
+    if (bind(_server, (const struct sockaddr*)sun, addrlen) < 0) {
```
Jongy, line 71: I split it here because I kept getting `FdTransfer listen()` errors, and some error codes overlap between `listen` and `bind`, so best to keep them as separate messages.

src/fdtransfer/fdtransfer.h
```diff
@@ -56,28 +56,15 @@ struct perf_fd_response {
...
-
 static inline bool socketPath(const char *path, struct sockaddr_un *sun, socklen_t *addrlen) {
     const int path_len = strlen(path);
     if (path_len > sizeof(sun->sun_path)) {
         return false;
     }
     memcpy(sun->sun_path, path, path_len);
+    if (sun->sun_path[0] == '@') {
+        sun->sun_path[0] = '\0';
+    }
```
apangin, line 67: I'd do that exactly the same way :)
Nice to see it's now one `socketPath` function.

src/arguments.cpp
```diff
@@ -271,6 +271,9 @@ Error Arguments::parse(const char* args) {
 
             CASE("fdtransfer")
                 _fdtransfer = true;
+                if (value == NULL) {
+                    msg ="fdtransfer path must not be empty";
```
apangin, line 275: missing space after `=`

I'd also add `if (value == NULL || value[0] == 0)`

src/fdtransfer/fdtransferServer.cpp
```diff
@@ -267,7 +273,7 @@ bool FdTransferServer::sendFd(int fd, struct fd_response *resp, size_t resp_size
     return true;
 }
 
-static int single_pid_server(int pid) {
+static int single_pid_server(int pid, const char *path) {
```
apangin, line 276: `single_pid_server` previously switched to the target's `net` namespace.
Now it depends on the path type, whether to switch or not.

src/arguments.cpp
Jongy: Hmm I compared it to `safemode` which doesn't do the check. Why do some have `value[0] == 0` and some do not?

Anyway, changed

src/fdtransfer/fdtransferServer.cpp
Jongy: Correct. I'll add a check here

profiler.sh
Jongy: Since the script uses `/bin/sh` it cannot use `-o pipefail`, and thus if `od` or `xargs` are missing the script will continue instead of getting an error. I suppose this can be worked around by doing this without a subshell...

Jongy:
@apangin do you need any other changes on my part here?

apangin:
It's good enough - merged.

PR #640, opened 2022-08-29
Adjust 'j9thread_self' lookup code

jbachorik:
For OpenJ9 distributions and OpenJDK17+ the current integration is not working properly.

The reason is that the `j9thread_self` symbol was renamed to `j9thread_self@plt` in more recent versions JDK versions. This lead the detection mechanism to fail and the integration not work.

The proposed fix is very simple - just lookup the symbol as `j9thread_self*` to allow the suffix.

With the fix in place the integration works properly once again - tested for Semeru distributions of OpenJDK 17 (17.0.4.1) and OpenJDK 18 (18.0.2).

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -138,7 +138,7 @@ bool VM::init(JavaVM* vm, bool attach) {
     Profiler* profiler = Profiler::instance();
     profiler->updateSymbols(false);
 
-    _openj9 = !is_hotspot && J9Ext::initialize(_jvmti, profiler->resolveSymbol("j9thread_self"));
+    _openj9 = !is_hotspot && J9Ext::initialize(_jvmti, profiler->resolveSymbol("j9thread_self*"));
     _can_sample_objects = !is_hotspot || hotspot_version() >= 11;
 
     CodeCache* lib = isOpenJ9()


apangin:
This sounds strange.

`*@plt` are synthetic symbols added by async-profiler for PLT entries. There are no such symbols in the original .so files. Also, if there were no `j9thread_self`, `j9thread_self@plt` could not have appeared either. You may check
```
nm $JDK_DIR/lib/default/*.so | grep j9thread_self
```

I just verified that async-profiler works out of the box with IBM Semeru Runtime 17.0.4.1 and 18.0.2.1.

jbachorik:
Yes, you are right. This seems to be a problem in our fork - the J9 integration should be unmodified but without the modified lookup it will fail.
I am closing this PR as it is not relevant for the main line. 
Sorry for the inconvenience ...

jbachorik:
Actually, this seems to be an issue in 2.8.3 

```
bits@jaroslavbachorik-0:~/dd/async-profiler$ git branch
* (HEAD detached at v2.8.3)
  jb/j9_fix_openjdk17
  jb/tid_in_tls
  jb/wclock_experiments
  main
  master
bits@jaroslavbachorik-0:~/dd/async-profiler$ ./test/smoke-test.sh
+ '[' -z /home/bits/.sdkman/candidates/java/17.0.4.1-sem ']'
++ dirname ./test/smoke-test.sh
+ cd ./test
+ '[' Target.class -ot Target.java ']'
+ FILENAME=/tmp/java.trace
+ JAVAPID=1249298
+ sleep 1
+ /home/bits/.sdkman/candidates/java/17.0.4.1-sem/bin/java Target
Starting...
+ ../profiler.sh -f /tmp/java.trace -o collapsed -d 5 1249298
JVMJ9TI064E Agent initialization function Agent_OnAttach failed for library /home/bits/go/src/github.com/DataDog/async-profiler/build/libasyncProfiler.so, return code 200
[ERROR] Could not find VMThread bridge. Unsupported JVM?
```

The same thing works fine in master, though.

jbachorik:
Cherry-picking https://github.com/jvm-profiling-tools/async-profiler/commit/9cec0765cdccf4597b36d5c25eced650d2d150e0 for our fork did the trick.

Issue #423, opened 2021-04-29
Cannot build using jdk17


navyxliu:
I got error message as follows. 

```
../jdk/build/linux-x86_64-server-release/images/jdk/bin/javac -source 6 -target 6 -d build/api src/api/one/profiler/Events.java src/api/one/profiler/AsyncProfiler.java src/api/one/profiler/AsyncProfilerMXBean.java src/api/one/profiler/Counter.java
warning: [options] bootstrap class path not set in conjunction with -source 6
error: Source option 6 is no longer supported. Use 7 or later.
error: Target option 6 is no longer supported. Use 7 or later.
make: *** [build/async-profiler.jar] Error 2
```

I saw you guys made it configurable in #285 .  I change it from 6 to 7 and get passed. 
Why do we need to specify a source version here?

apangin:
async-profiler supports a wide range of JDK versions, starting from JDK 6, so we need to compile Java classes with the minimum supported bytecode level.

In order to build it with a recent JDK, run
```
make JAVAC_RELEASE_VERSION=7
```

apangin:
For now, I've made javac use Java 6 target only when building `release`.

navyxliu:
your patch LGTM. Thanks for solving it!

ijuma:
I think it would make sense to drop Java 6 support in the next release. All the major vendors dropped support for that a while back. Even Java 7 is only one year away from the end of extended support by Oracle (and Red Hat's support seems to end this month).

apangin:
@ijuma Already did this a few hours ago :)
https://github.com/jvm-profiling-tools/async-profiler/commit/ff49ccccb73c2a5e677653da67f269c4daa7057d#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R19

ijuma:
Awesome!

PR #655, opened 2022-10-07
Allow to set Bytecode version for Java compile in makefile

parttimenerd:
Reintroduces the `JAVAC_RELEASE_VERSION` to allow developers to specify the Java bytecode version. The rational is that the minimally configurable byte code version goes from 7 to 8 in the JDK 20. This change allows building async-profiler with these JDKs too by using `make JAVAC_RELEASE_VERSION=8`, this his helpful when developing profiling APIs in the JDK.

This is related to #423, but just setting the bytecode version to 8 would be hard to justify as Azul supports Java 7 till 2027.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,10 @@ MERGE=true
 
 JAVAC=$(JAVA_HOME)/bin/javac
 JAR=$(JAVA_HOME)/bin/jar
-JAVAC_OPTIONS=-source 7 -target 7 -Xlint:-options
+ifeq ($(JAVAC_RELEASE_VERSION),)
+  export JAVAC_RELEASE_VERSION=7
+endif
+JAVAC_OPTIONS=-source $(JAVAC_RELEASE_VERSION) -target $(JAVAC_RELEASE_VERSION) -Xlint:-options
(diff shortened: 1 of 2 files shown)


parttimenerd:
It's especially important for my jdk-profiling-tester which is used to find bugs in the ASGCT implementation.

Makefile
```diff
@@ -17,7 +17,10 @@ MERGE=true
 
 JAVAC=$(JAVA_HOME)/bin/javac
 JAR=$(JAVA_HOME)/bin/jar
-JAVAC_OPTIONS=-source 7 -target 7 -Xlint:-options
+ifeq ($(JAVAC_RELEASE_VERSION),)
+  export JAVAC_RELEASE_VERSION=7
+endif
```
apangin, line 22: `ifeq` is redundant: command line arguments will override the value anyway.
I'd also prefer a shorter name (easier to type), e.g., `JAVAC_TARGET`.

parttimenerd: You're right, fixed it.

apangin:
Thanks. Merged

PR #656, opened 2022-10-11
Update README.md

mab:
IntelliJ profiler documentation moved, updates URL.


apangin:
New URL leads to an IDEA promo page that has no reference to async-profiler.

I'd rather remove the paragraph entirely. After all, there are many products that use async-profiler under the hood (e.g., gProfiler, Pyroscope, Elastic APM and more), there's no need to highlight just one of them.

apangin:
Got the correct URL from IDEA profiler team.

mab:
Thanks. That was quick. Just talked to the IDEA guys.

PR #675, opened 2022-11-07
Increase frame buf size

guybortnikov:
I use using async-profiler, and I print the fully qualified name and the full method signature, on several use cases the method signature has been truncated (because it has many arguments will long fully qualified names)

This PR changes the truncation of the method signature to use the remaining capacity of the buffer instead of 255 const. Also, it increases the frame name buffer size from 800 bytes to 1kb.

### DIFF
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -188,7 +188,7 @@ char* FrameName::javaMethodName(jmethodID method) {
         result = javaClassName(class_name + 1, strlen(class_name) - 2, _style);
         strcat(result, ".");
         strcat(result, method_name);
-        if (_style & STYLE_SIGNATURES) strcat(result, truncate(method_sig, 255));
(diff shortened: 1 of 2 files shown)


richardstartin:
Hi @guybortnikov - I'm curious what the motivation for this change is - could you add it to the PR description please?

guybortnikov:
@richardstartin 
> Hi @guybortnikov - I'm curious what the motivation for this change is - could you add it to the PR description please?

Thanks for the feedback, added the motivation to the PR description.

apangin:
> It's not the first time I see this issue, however, up to the moment it wasn't a real problem. My point is, text-based output formats (including collapsed format) are designed mainly for visualization for end user (who doesn't parse 1KB strings by eye anyway). For computer processing, there is a jfr format.
On the other hand, the suggested PR does not fix the root cause, just makes the issue appear less often.

I committed the fix to allow unlimited length signatures in collapsed format.

PR #681, opened 2022-11-15
Reading the program headers of some dynamic libraries may cause a segment fault

yanglong1010:
The program headers of some dynamic libraries are not stored in the segment with execution permission, so it may not be possible to directly calculate the starting address of the program headers by adding the starting address of the current segment and the file offset of the program headers, which will cause a segment fault.
In this case, we can mmap the program headers in the file into memory to read the correct data.

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -145,6 +145,8 @@ class ElfParser {
     const char* _base;
     const char* _file_name;
     ElfHeader* _header;
+    const char* pheaders;
+    int mmap_length;
     const char* _sections;
     const char* _vaddr_diff;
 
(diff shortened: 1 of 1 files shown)


yanglong1010:
> Thank you for working on the issue, however, I cannot accept the suggested pull request.
> 
> First, the mentioned issue is already addressed in `dl-iterate` branch.
> 
> Second, the proposed change does not align well with the architecture of ElfParser. ELF parsing currently proceeds in two steps: one handles the loaded DSO in memory, another parses the corresponding shared library file. You added yet another file parsing in the first step which essentially targets the cases when the underlying file is not accessible (deleted).
> 
> Let me close the PR and let's continue discussion in the issue.

Thank you for your explanation, now I understand the first stage is to deal with the file not exist situation.

PR #685, opened 2022-11-26
Add accept timeout as 3rd parameter of fdtransfer

Jongy:
Attaching may take more than 10 seconds if the target process is handling previous attach requests / is under heavy load. fdtransfer is started before jattach, and will wait for at most 10 seconds before bailing. Afterwards, jattach eventually succeeds but AP fails with the error `FdTransferClient connect(): Connection refused` because fdtransfer already went down.

This fix allows drivers to control the timeout. I'm not sure how to apply it in `profiler.sh` - what we'd want is, after jattach returned, if no connection was made to fdtransfer, it should exit. So this is actually "inifinite" timeout because jattach has no timeout - and trapping upon jattach exit. Possibly another PR.

### DIFF
--- a/src/fdtransfer/fdtransferServer.cpp
+++ b/src/fdtransfer/fdtransferServer.cpp
@@ -273,7 +273,7 @@ bool FdTransferServer::sendFd(int fd, struct fd_response *resp, size_t resp_size
     return true;
 }
 
-static int single_pid_server(int pid, const char *path) {
+static int single_pid_server(int pid, const char *path, unsigned int timeout) {
     // get its nspid prior to moving to its PID namespace.
     int nspid;
     uid_t _target_uid;
@@ -301,7 +301,7 @@ static int single_pid_server(int pid, const char *path) {
         }
     }
 
-    if (!FdTransferServer::bindServer(&sun, addrlen, 10)) {
+    if (!FdTransferServer::bindServer(&sun, addrlen, timeout)) {
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


Jongy:
@apangin , can we merge it for the next release? or it will wait for the next major as it is an API change?

apangin:
The project does not accept external code contributions at the moment.
This is temporary. I'll let you know once the issue is resolved.

Jongy:
Gotcha, np

apangin:
Is the PR still relevant, considering the binary launcher changes? `fdtransfer` functionality is now embedded in `asprof`.
BTW, I increased the timeout to 30 seconds.

Jongy:
Yes, I think it is. Attaching currently has no timeout, and as in some cases it may take beyond 10 seconds (and even beyond 30 seconds), it's convenient to be able to increase fdtransfer's timeout (perhaps with a flag `--fdtransfer-timeout`). What do you think? I can update the PR to mainline.

apangin:
I wonder if timeout needs to be configured at all. Once forked jattach has returned, we know that either fdtransfer had already accepted connection or, if jattach failed, fdtransfer is no longer needed and can be killed.

Jongy:
Yes, but if jattach needs more than 10 or even 30 seconds, I agree that's a bit extreme... but in such cases the fdtransfer needs to have a timeout that is as long as the timeout jattach has; or maybe have no timeout at all, e.g run it as a thread which is stopped e.g by closing the listening socket after jattach returns.

apangin:
> or maybe have no timeout at all, e.g run it as a thread which is stopped by closing the listening socket after jattach returns

Yeah, that's what I meant. The control process knows exactly when jattach returns, so it also knows the deadline for accepting fdtransfer connection.

apangin:
This has been idle for a while, let me close the PR.
Feel free to update and reopen if it's still relevant.

PR #692, opened 2022-12-02
Improve tree view display

aleris:
- fix text wrapping for long traces when horizontal scroll is present
- add vertical line aligned on expland/collapse icon for clearer tree
view
- use ⊕ and ⊖ instead of + and - for better visibility
- use monospace font for better numerical alignment
- fix form search on enter key press
- header form small alignment improvements

**Wrapping**

Before:

image

After:

image

**Form & Alignment**

Before:

image

After:

image

### DIFF
--- a/src/res/tree.html
+++ b/src/res/tree.html
@@ -1,18 +1,25 @@
 <!DOCTYPE html>
 <html lang='en'>
 <head>
-<title>Tree view</title>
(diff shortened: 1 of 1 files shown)


apangin:
Tree view got a slightly refreshed design. I hope it looks better now.

Issue #702, opened 2022-12-27
Large flame graphs not user friendly


Cheejyg:
When generating large flame graphs with many stack traces, the UI elements (mainly the **search** and **status** with number of samples) are not user friendly because they are at the edge of the HTML. i.e. you would have to scroll all the way up to use the search function, and have to scroll all the way down to even see the status. Moreover, you can't see the status for elements that are high up in the flame graph because the status bar is always at the bottom of the page.

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/7922964/209644637-3203bbe5-f599-4c4f-9463-ec735f053b6d.png">
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/7922964/209645646-e4dedc8b-58b6-4098-85c6-c044b2d02a6b.png">

apangin:
The status bar and search results are now always visible at the bottom of the screen, no matter how large the flame graph is.

<img width="650" alt="visible" src="https://user-images.githubusercontent.com/1749416/224571814-787e6fa4-36fe-4c1e-81b0-46e00e4b9bba.png">

PR #704, opened 2022-12-27
fix position of search and status bar

Cheejyg:
Large flame graphs not user friendly #702


apangin:
I implemented this in 91691ce0396a79b923672bbfa1dc15c6e7e87368 in a slightly different but a similar way.

PR #712, opened 2023-02-08
Chore: document chunksize and chunktime

lujiajing1126:
Add [default values](https://github.com/jvm-profiling-tools/async-profiler/issues/360) in the `README.md`.

And according to the impl, the chunksize must be larger than 262,144 bytes. This also has to be mentioned in the document.


apangin:
Thank you for updating the documentation.

*"and must be larger than 262,144 bytes"* is not quite accurate: it is permitted to specify a lower value, even though the result chunk will be bigger. As you noticed, this is an implementation detail, which is subject to change and therefore need not be documented. The doc already mentions that the limits are approximate.

PR #713, opened 2023-02-08
jfr2pprof: Add allocations and lock data

NeQuissimus:
This is basically a follow-up to #587 

The previous PR was limited to converting CPU measurements, this one is adding allocations and contended locks.
Running the converter will create a file for each type of measurement, so that JFR files containing mixed data are untangled.

/cc @federico-quintas

### DIFF
--- a/src/converter/jfr2pprof.java
+++ b/src/converter/jfr2pprof.java
@@ -244,25 +244,49 @@ public class jfr2pprof {
     }
 
     public static void main(String[] args) throws Exception {
-        if (args.length < 3) {
-            System.out.println("Usage: java " + jfr2pprof.class.getName() + " [cpu|alloc|lock] input.jfr output.pprof");
+
+        final Arguments arguments = new Arguments(args);
+        if (arguments.input == null || arguments.output == null) {
+            System.out.println("Usage: java " + jfr2pprof.class.getName() + " [options] input.jfr output.pprof");
+            System.out.println();
+            System.out.println("options can be one or more of the following:");
+            System.out.println("  --cpu        CPU");
+            System.out.println("  --alloc      Allocation");
+            System.out.println("  --lock       Lock contention");
+            System.out.println();
+            System.out.println("If no options are given, --cpu is assumed");
             System.exit(1);
         }
 
-        final String type = args[0];
-        if (!TYPE_CPU.equals(type) && !TYPE_ALLOC.equals(type) && !TYPE_LOCK.equals(type)) {
-            System.out.println("Usage: java " + jfr2pprof.class.getName() + " [cpu|alloc|lock] input.jfr output.pprof");
-            System.exit(1);
+        final List<String> types = new ArrayList<>();
+
+        if (arguments.cpu) {
+            types.add(TYPE_CPU);
         }
 
-        File dst = new File(args[2]);
-        if (dst.isDirectory()) {
-            dst = new File(dst, new File(args[1]).getName().replace(".jfr", ".pprof"));
+        if (arguments.alloc) {
+            types.add(TYPE_ALLOC);
         }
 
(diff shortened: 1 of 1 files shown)


NeQuissimus:
@apangin I think this is good to go now

src/converter/jfr2pprof.java
```diff
@@ -56,8 +60,20 @@ public boolean equals(final Object other) {
...
+
     public static final byte[] METHOD_UNKNOWN = "[unknown]".getBytes();
 
+    public static final byte[] VALUE_TYPE_CPU = "cpu".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_TYPE_ALLOC = "allocations".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_TYPE_LOCK = "lock".getBytes(StandardCharsets.UTF_8);
+
+    public static final byte[] VALUE_UNIT_CPU = "nanoseconds".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_UNIT_ALLOC = "count".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_UNIT_LOCK = "nanoseconds".getBytes(StandardCharsets.UTF_8);
```
apangin, line 75: How come the value is "count" for allocations, but "nanoseconds" for locks? I see the converter code handles both events exactly the same way.
Also, does it make sense to include "bytes" in the allocation profile?

src/converter/jfr2pprof.java
```diff
@@ -56,8 +60,20 @@ public boolean equals(final Object other) {
...
 
+    public static final String TYPE_CPU = "cpu";
+    public static final String TYPE_ALLOC = "alloc";
+    public static final String TYPE_LOCK = "lock";
+
     public static final byte[] METHOD_UNKNOWN = "[unknown]".getBytes();
 
+    public static final byte[] VALUE_TYPE_CPU = "cpu".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_TYPE_ALLOC = "allocations".getBytes(StandardCharsets.UTF_8);
+    public static final byte[] VALUE_TYPE_LOCK = "lock".getBytes(StandardCharsets.UTF_8);
```
apangin, line 71: These fields (as well as the following ones) are redundant: they are never reused in the code.

src/converter/jfr2pprof.java
```diff
@@ -209,19 +244,49 @@ private byte[] getMethodName(final long methodId, final byte methodType) {
...
-             final FileOutputStream out = new FileOutputStream(dst)) {
-            new jfr2pprof(jfr).dump(out);
+        if (types.isEmpty()) {
+            types.add(TYPE_CPU);
+        }
+
+        for (final String type : types) {
+          File dst = new File(arguments.output + "." + type);
+
+          if (dst.isDirectory()) {
```
apangin, line 282: This code was intended for a case when a user specifies a directory as an output argument. However, here `dst` has already been altered with the type suffix.

src/converter/jfr2pprof.java
```diff
@@ -209,19 +244,49 @@ private byte[] getMethodName(final long methodId, final byte methodType) {
...
+              dst = new File(dst, new File(arguments.input).getName().replace(".jfr", ".pprof" + "." + type));
```
apangin, line 283: I'd put `type` ahead of the file extension and only for the case when multiple types are given.

apangin:
Hi, is this PR still relevant?

NeQuissimus:
It's very relevant as the functionality otherwise does not exist. I have not had time to update this but will try to get to it.

apangin:
OK, no worries. I will leave the PR open and review it when updated. Thank you.

apangin:
Obsoleted by #905

PR #718, opened 2023-02-23
Support read JFR file chunk by chunk

lujiajing1126:
If the JFR file is large, for example, `alloc` event is enabled, it may cost large heap space to process millions of events.

This PR intends to amortize memory consumption by allowing users to read a single chunk once.

API:

As is used by `readAllEvents` in the `JfrReader.java`,

```java
    public <E extends Event> List<E> readAllEvents(Class<E> cls) {
        Chunks<E> chunks = readChunks(cls);
        ArrayList<E> events = new ArrayList<>();
        for (final Chunk<E> chunk : chunks) {
            for (final E event : chunk) {
                events.add(event);
            }
        }
        Collections.sort(events);
        return events;
    }
```

Still questions: (Excuse for my poor understanding of the JFR spec)

In the current impl, I noticed only `types` and `typesByName` are cleared. However, according to [the file format](https://www.morling.dev/blog/jdk-flight-recorder-file-format/), `Chunk` should be self-contained. Does it mean that we can clear all intermediate states, e.g. classes, symbols, methods when we start to read a new Chunk?

### DIFF
--- a/src/converter/jfr2flame.java
+++ b/src/converter/jfr2flame.java
@@ -53,7 +53,7 @@ public class jfr2flame {
         long startTicks = args.from != 0 ? toTicks(args.from) : Long.MIN_VALUE;
         long endTicks = args.to != 0 ? toTicks(args.to) : Long.MAX_VALUE;
 
-        for (Event event; (event = jfr.readEvent(eventClass)) != null; ) {
+        for (Event event; (event = jfr.readEvent(eventClass).getEvent()) != null; ) {
             if (event.time >= startTicks && event.time <= endTicks) {
                 if (threadState < 0 || ((ExecutionSample) event).threadState == threadState) {
                     agg.collect(event);
--- a/src/converter/one/jfr/Chunk.java
+++ b/src/converter/one/jfr/Chunk.java
@@ -1,0 +1,53 @@
+package one.jfr;
+
+import one.jfr.event.Event;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+public class Chunk<E extends Event> implements Iterable<E> {
+    private final List<E> events;
+
+    final int nextChunkPos;
+
+    private Chunk(Builder<E> bld) {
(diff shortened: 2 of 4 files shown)


apangin:
The problem is known and I have already addressed it in my recent not-yet-published work.
I pushed the relevant change - it introduces the new flag in `JfrReader`: `stopAtNewChunk` that breaks reading between chunks. Without the flag set, everything works as before.

This change is tiny comparing to the grand refactoring in the proposed PR, and it does not introduce any performance overhead.

apangin:
I tried your implementation, but unfortunately it does not work with my JFR files (produces empty HTML). So, in the presense of the alternative, let me close this PR.

apangin:
> In the current impl, I noticed only types and typesByName are cleared. However, according to [the file format](https://www.morling.dev/blog/jdk-flight-recorder-file-format/), Chunk should be self-contained. Does it mean that we can clear all intermediate states, e.g. classes, symbols, methods when we start to read a new Chunk?

Yes, chunks are self-contained. However, if we create a new copy of all constant pool items with every new chunk, this will be a big waste of memory, especially for long recordings. `JfrReader` relies on the implementation details of the particular JFR writer. Not good for universal solution, but quite efficient for regular async-profiler and JFR recordings.

lujiajing1126:
> I tried your implementation, but unfortunately it does not work with my JFR files (produces empty HTML). So, in the presense of the alternative, let me close this PR.

I cherry-picked from an internal repo. Maybe some nits. I've tried your commit, it worked well!

> Yes, chunks are self-contained. However, if we create a new copy of all constant pool items with every new chunk, this will be a big waste of memory, especially for long recordings. `JfrReader` relies on the implementation details of the particular JFR writer. Not good for universal solution, but quite efficient for regular async-profiler and JFR recordings.

Thanks for your explanation. I understand JfrReader in this repo may only work for async-profiler, but it is fast and cost less memory!

PR #721, opened 2023-03-02
Compile with -std-c++2b without errors

alkis:
Make codebase compile with `-std-c++2b`.


apangin:
Thanks for the fix, it has been now merged.

Issue #729, opened 2023-03-21
Statically link libstdc++ at least for musl builds


rrva:
A lot of jdk docker images are
based on alpine without libstdc++ preinstalled. It is great to be able to use libasyncprofiler without having to modify the docker image that java is running on.

Please link libstdc++ statically but keep other dependencies like musl libc dynamically linked.

I can make a pull request if this is something that would be ok to have in all musl release builds, I am guessing something like this

`g++ -shared -o your_library.so your_object_files.o -static-libstdc++ -Wl,-Bdynamic -lc
`

Jongy:
We're building AP with `-static-libstdc++` on musl in our async-profiler fork used in gProfiler. For reference [this is the change](https://github.com/Granulate/async-profiler/commit/49080a221b46e312041947ebbf7845dc054b4438). Note that it greatly increases the size (300-400k -> 8mb).

As a generic solution we might want to have 2 separate AP builds, one static and one not static, because not all musl-based JVMs lack libstdc++, so we can use the appropriate one. (to reduce memory overhead & loaded code, etc).

apangin:
As @Jongy noticed, static linking may significantly impact the library size - this is not something that can be done unconditionally. At the same time, I don't want to multiply the number of builds.
Ideal solution would be to get rid of `libstdc++` dependency at all. This is doable but requires a fair amount of routine.

rrva:
Would it not be possible to add the libstdc++.so to the jar file which already contains libasyncprofiler.so ? I bet there is a way to set LD_LIBRARY_PATH so it would only load this copy if libstdc++ is non-existent on the system, and libstdc++ should be known at build time which version of it is the compatible one, so it could be included in the jar. This way, there is no increase in memory usage that can be seen with static linking, but asyncprofiler is usable on stock docker images where libstdc++ is not preinstalled (most notably a lot of alpine based openjdk distributions). This or the rewrite where libstdc++ is removed completely (but I can understand if that is a larger task)

apangin:
You can build `libasyncProfiler.so` statically linked with `libstdc++` with no modifications to async-profiler source code by passing custom `CXXFLAGS` value to `make` command.

We have no plans to provide multiple builds of `libasyncProfiler.so` that differ only by used compiler/linker flags. At the same time we welcome changes that will _decrease_ the number of produced binaries and thus improve user experience and reduce maintenance overhead. As mentioned above, one of the possible approaches to that is getting rid of `libstdc++` dependency altogether.

[...]

PR #735, opened 2023-03-26
Prune to most "important" stacks - FOR DISCUSSION

pnf:
Adds a new `--prune NUM` option to be used with `collapsed` format.  Leaves will be pruned from the stack graph in order of increasing `_total` until only `NUM` remain; when all of a node's children have been pruned, it is added back to the queue.  With `--prune `5`, the `/tmp/java.trace` from the smoke test looks like:
```
Target.main;Target.method1 42
Target.main;Target.method2 46
Target.main;Target.method3;java/io/File.list;java/io/File.normalizedList;java/io/UnixFileSystem.list;Java_java_io_UnixFileSystem_list;JNU_CopyObjectArray;jni_DeleteLocalRef 17
Target.main;Target.method3;java/io/File.list;java/io/File.normalizedList;java/io/UnixFileSystem.list;Java_java_io_UnixFileSystem_list;JNU_CopyObjectArray;jni_GetObjectArrayElement 26
Target.main;Target.method3;java/io/File.list;java/io/File.normalizedList;java/io/UnixFileSystem.list;Java_java_io_UnixFileSystem_list;newSizedString8859_1;jni_NewString;java_lang_String::create_oop_from_unicode(unsigned short*, int, Thread*);java_lang_String::create_from_unicode(unsigned short*, int, Thread*);java_lang_String::basic_create(int, bool, Thread*) 25
```
Note, for example, that `java_lang_String::basic_create` would typically not show up in a list of most-sampled stacks, because the samples are fragmented over a dozen contributing child stacks.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -163,6 +163,11 @@ Error Arguments::parse(const char* args) {
             CASE("collapsed")
                 _output = OUTPUT_COLLAPSED;
 
+            CASE("prune")
+                _output = OUTPUT_COLLAPSED;
+                if(value == NULL || (_pruned = atoi(value)) <= 0)
+                  msg = "Invalid pruning";
+
             CASE("flamegraph")
                 _output = OUTPUT_FLAMEGRAPH;
 
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -169,6 +169,7 @@ class Arguments {
     CStack _cstack;
     Clock _clock;
     Output _output;
+    int _pruned;
     long _chunk_size;
     long _chunk_time;
     const char* _jfr_sync;
@@ -216,6 +217,7 @@ class Arguments {
         _cstack(CSTACK_DEFAULT),
         _clock(CLK_DEFAULT),
         _output(OUTPUT_NONE),
+        _pruned(0),
         _chunk_size(100 * 1024 * 1024),
         _chunk_time(3600),
         _jfr_sync(NULL),
--- a/src/flameGraph.cpp
+++ b/src/flameGraph.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <algorithm>
+#include <queue>
 #include <vector>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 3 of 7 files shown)


apangin:
Hi Peter, I appreciate your interest in improving async-profiler.

If you would like to suggest a new feature, you are welcome to submit an issue or start a discussion first. PRs like this have little chance to get accepted, unfortunately. A quick look reveals multiple problems, starting from the compilation issues to the insufficient motivation for the change and the justification of the chosen algorithm.

It's worth mentioning that `collapsed` output format is left only for compatibility reasons, its usage for new features is deprecated. I'd also prefer not to bring more stuff from libstdc++, since this conflicts with another initiative to get rid of libstdc++ in async-profiler completely (#729). Thank you for your understanding.

pnf:
Hi -
I didn't really expect the PR to be merged as-is, but it's easier for me to express the idea if it has a concrete form. 
The motivation is that I often want a way to summarize long samples over thousands of processes. I imagined squinting at a series of mini-flamegraphs comprising, say, 30 second intervals, into which one could drill for further details. The squinted-at flamegraph is conceptually coarser, which basically means the flame tips have been lopped off such that the bases that remain represent more significant fractions of the total samples. There are various algorithms for doing this, but they all come down to pruning the flame graph tree. The algorithm in the PR is one I'm already using in Java to process text dumps of collapsed stacks obtained via the API, but the cost of this is hugely dominated by parsing, so it seemed simpler to do the whole thing in c++.
In any case, I find the summaries to be useful, and they also seem to help motivate others to examine the samples in more detail, so we'll either continue to do the processing in java or maintain our own fork, which in any case we have to do already for awaiter stacks.

apangin:
Thank you for the background, the problem you describe makes sense. Async-profiler already has `--minwidth` option for building coarse flame graphs, so I'd be interested in a comparsion of your proposed approach to the existing one.

In any case, this feature is more suitable for the out-of-process converter than for the core agent that works in the context of a Java application. I agree that `collapsed` format is too verbose and expensive for parsing. Did you consider dumping output in `jfr` format?

pnf:
You’re right that doing this out of process would seem to make more sense. In our present case, we’re capturing the thumbnails (as lists of strings) simultaneously with snapshots of a bunch of app-specific measurements that are not easily obtained externally.  


Le Mer, 29 mars, 2023, à 10:43, Andrei Pangin a écrit:

apangin:
Considering the above discussion, I'm closing the PR without plans to merge.

Issue #664, opened 2022-10-21
Consider making converters part of the public API


cristiangreco:
Hi, thanks for the great work here!

Would you consider making the various converter methods (jfr2pprof, jfr2nflx, etc) part of a public API? Right now they can only be used as CLI tools with `converter.jar`, but it'd be useful to have the various `dump()` methods exposed as part of then `one.profiler` package in `async-profiler.jar`. Alternatively, even moving the converter code to `one.jfr` package within  `converter.jar` itself would make them more easily accessible.

I'm happy to open a PR if you're ok with the change but have other priorities.

apangin:
Thank you for your interest in async-profiler.
It wasn't my plan to make the converter a part of public API: its implementation, internals and even CLI is subject to change in any minor update.
I may publish converter as a separate maven artifact, and it's OK to use it in your applications, but there is no intention to merge it with async-profiler API which is supposed to be stable.

cristiangreco:
Thanks for your quick reply @apangin! It makes sense to continue keeping the converter tools non-public to avoid breaking changes, especially not to make them part of the async-profiler API. I'd be happy with any solution that makes the `dump` methods more easily accessible (e.g. out of the `jfr2pprof` main class, maybe part of `one.jfr.format` or something).

apangin:
I'm not sure I understand the point - how moving to another package will make `dump` method "more easily accessible"?

cristiangreco:
Suppose I'm importing the jars from a local folder:

```gradle
dependencies {
    // ...
    implementation files('libs/async-profiler.jar')
    implementation files('libs/converter.jar')
}
```

I'd like to do something similar to what `src/converter/jfr2pprof.java` does, i.e.

```java
JfrReader jfr = new JfrReader(path);
FileOutputStream out = new FileOutputStream(dst)
new jfr2pprof(jfr).dump(out); // or similar
```

However can't do that as `jfr2pprof` class is not part of any package (ideally I'd just prefer to use `dump`, as `jfr2pprof` is implementation detail of the converter tools). This is why I was thinking that moving the dump logic into some API like e.g. `src/converter/one/jfr/format/pprof.java` would help.

Hope this clarifies a bit. Maybe there's a simpler solution I'm not seeing here!

cristiangreco:
Hi @apangin did you have any further thought about this use case perhaps? Thanks!

apangin:
[...]

PR #738, opened 2023-03-29
Add missing package declarations.

tjgq:
So that these files can be imported from other projects.


apangin:
Converters are not part of public API. They are deliberately put in the top level package. See #664 for discussion.

Apart from that, the proposed PR is incomplete, as it's not enough just to add "package" without actually moving the files and updating all relevant references in the code and documentation.

PR #739, opened 2023-03-31
Flush if needed before writing UTF-8 symbols and thread names

richardstartin:
Since the max UTF-8 string length is 8191 and flushes are only triggered when there is less than 4KiB left in each JFR buffer, an (albeit abnormally long) thread name or symbol longer than 4KiB can wrap around and corrupt the offset at the start of the next buffer, unless the buffer is the last buffer in which case it will just overflow.


apangin:
Thanks for pointing this out.
I pushed a slightly changed fix.

PR #743, opened 2023-04-19
Fix build when the main folder contains `flame.html` or `tree.html` files

parttimenerd:
Tiny fix for the case that the current async-profiler folder contains `flame.html` or `tree.html` or files. The build script currently takes these files instead of the files from `src/res`, causing an error during the flame graph generation. This is problematic because `flame.html` is not an uncommon name for created flame graph files.

### DIFF
--- a/src/flameGraph.cpp
+++ b/src/flameGraph.cpp
@@ -26,8 +26,8 @@
 // Browsers refuse to draw on canvas larger than 32767 px
 const int MAX_CANVAS_HEIGHT = 32767;
 
-INCBIN(FLAMEGRAPH_TEMPLATE, "flame.html")
-INCBIN(TREE_TEMPLATE, "tree.html")
+INCBIN(FLAMEGRAPH_TEMPLATE, "src/res/flame.html")
+INCBIN(TREE_TEMPLATE, "src/res/tree.html")
 
 
 class StringUtils {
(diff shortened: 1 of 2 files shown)


src/flameGraph.cpp
```diff
@@ -26,8 +26,8 @@
 // Browsers refuse to draw on canvas larger than 32767 px
 const int MAX_CANVAS_HEIGHT = 32767;
 
-INCBIN(FLAMEGRAPH_TEMPLATE, "flame.html")
-INCBIN(TREE_TEMPLATE, "tree.html")
+INCBIN(FLAMEGRAPH_TEMPLATE, "src/res/flame.html")
+INCBIN(TREE_TEMPLATE, "src/res/tree.html")
```
apangin, line 30: Did you test it with XCode on macOS?

Is `-I src/res` include path no longer needed?

parttimenerd: > Did you test it with XCode on macOS?

I tested it with XCode (creating a project that uses the external build script) and it works.

> Is -I src/res include path no longer needed?

You're right, it is no longer needed.

apangin:
Oops, you apparently committed extra files

parttimenerd:
Oh, thanks for spotting this, I updated the commit.

apangin:
Looks good now. Thanks for the fix.

PR #744, opened 2023-04-25
Fix error when compiling against C++20.

tjgq:
This code appears to be ill-formed in C++20 (clang interprets the `<` as the start of a template parameter list).


tjgq:
It appears that https://github.com/async-profiler/async-profiler/pull/721 would also fix this issue; feel free to go with that one instead.

Issue #724, opened 2023-03-12
Binary launcher


apangin:
### Background

async-profiler currently includes the following components:
 - `libasyncProfiler.so` - profiler engine that works in the context of Java process;
 - `jattach` - utility to inject the profiler engine into the target JVM;
 - `fdtransfer` - helper for profiling containerized applications;
 - `profiler.sh` - profiler command-line interface written in shell script.

The package has non-standard structure: with binaries in the `build` folder and CLI together with the documentation in the root folder.

### Proposal

1. Reshape the package to a more conventional structure:
   - `/bin` directory with the profiler launcher;
   - `/lib` with `libasyncProfiler.so`.
2. Make the launcher a binary executable `asprof` instead of `profiler.sh`.
3. Incorporate the functionality of `jattach `and `fdtransfer` in the launcher, so the package contains no other executables besides `asprof`.

### Motivation

1. Intuitive usage.
2. Simpler deployment / integration.
3. Less typing.
4. Preparation for building packages in commonly used formats: RPM, DEB, etc.
5. More flexibility, e.g., the binary launcher may in future switch between host and container namespaces to provide seamless user experience: automatically copy the profiler library into the container and get the output from the container.
6. The same statically linked executable may work both with glibc and musl.

### Non-goals

1. There is no intention to change existing command line syntax, except for minor nuances.

### Target release

async-profiler 3.x

apangin:
@v-thakkar The proposed change does not affect the API of the agent library. `libasyncProfiler.so` should remain intact.

`jattach` will not be bundled with async-profiler anymore, however, it will be still possible to load `libasyncProfiler.so` using a separate [`jattach`](https://github.com/jattach/jattach) binary.

apangin:
Binary launcher has been integrated; directory structure has been changed; `profiler.sh`, `jattach` and `fdtransfer` are no longer bundled with async-profiler.
Futher work (e.g. packaging and better container support) will be done in separate issues.

PR #752, opened 2023-05-29
#724: Fix readme following removal of profiler.sh

kabunchi:


apangin:
I've updated the documentation. This required a bit more changes, since README previously referred to `profiler.sh` script, while `asprof` is not a script.

PR #765, opened 2023-07-06
change default interval 100hz->~99hz to avoid lockstep sampling

tomershafir:
Common logic is more likely to execute periodically at 100hz rather than 99hz, thus lockstep sampling is less likely to occur.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,7 @@ The following is a complete list of the command-line options accepted by
 * `-i N` - sets the profiling interval in nanoseconds or in other units,
   if N is followed by `ms` (for milliseconds), `us` (for microseconds),
   or `s` (for seconds). Only CPU active time is counted. No samples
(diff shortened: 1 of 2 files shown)


apangin:
Async-profiler does not use fixed sampling *frequency*. Instead, it works with a fixed *interval* between samples. Furthermore, this interval is measured in CPU time (i.e. unstalled processor cycles), not in the absolute time.

Having a round value for the interval helps to evaluate results sanity for a short profiling intervals, e.g., a process consuming 100% of one CPU core should produce ~500 samples during 5 seconds profiling session. For longer profiling sessions, there is no "lockstep" issue for the reasons mentioned above. If you have a real world example where changing the interval from 10ms to 10.10101ms solves some issue, please provide more details. Otherwise, changing the defaults that have been used by many customers for a long time, seem to bring more harm than benefits.

PR #788, opened 2023-08-02
Feat/add native stack depth max opt

tomershafir:
Adding a config option to control native stack depth. This allows a user to increase resolution of native parts in the profile, or to reduce unwind time and space overhead where possible.

Following curr idioms, I added `--cstackdepth` launcher opt mapped to `cstackdepth=DEPTH` format, replacing `MAX_NATIVE_FRAMES` and the j9 equivalent usage with the injected config while retaining the 128 default, following the curr procedural and object design.

Testing vm:
```
$ uname -a
Linux tomersha 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:29:30 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

$ java --version
openjdk 17.0.7 2023-04-18
OpenJDK Runtime Environment (build 17.0.7+7-Ubuntu-0ubuntu122.04.2)
OpenJDK 64-Bit Server VM (build 17.0.7+7-Ubuntu-0ubuntu122.04.2, mixed mode, sharing)

JAVA_HOME=/usr/lib/jvm/java-17-openjdk-arm64

$ g++ --version
$ cc --version
(Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

$ make --version
GNU Make 4.3
Built for aarch64-unknown-linux-gnu
```
Make builds, smoke suite tests pass, and basic itimer and perf manual sanity tests pass.

### DIFF
--- a/src/j9StackTraces.h
+++ b/src/j9StackTraces.h
@@ -22,14 +22,12 @@
 #include "arguments.h"
 
 
-const int MAX_J9_NATIVE_FRAMES = 128;
-
 struct J9StackTraceNotification {
     void* env;
     u64 counter;
     int num_frames;
     int reserved;
-    const void* addr[MAX_J9_NATIVE_FRAMES];
+    const void **addr;
 
     size_t size() {
         return sizeof(*this) - sizeof(this->addr) + num_frames * sizeof(const void*);
@@ -41,6 +39,7 @@ class J9StackTraces {
   private:
     static pthread_t _thread;
     static int _max_stack_depth;
+    static int _max_native_stack_depth;
     static int _pipe[2];
 
     static void* threadEntry(void* unused) {
--- a/src/j9StackTraces.h
+++ b/src/j9StackTraces.h
@@ -28,10 +28,6 @@ struct J9StackTraceNotification {
     int num_frames;
     int reserved;
     const void **addr;
-
-    size_t size() {
-        return sizeof(*this) - sizeof(this->addr) + num_frames * sizeof(const void*);
-    }
 };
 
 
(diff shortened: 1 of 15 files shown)


apangin:
For the PR to be reviewed, please provide a meaningful description along with the problem statement and the outline of the proposed solution.

tomershafir:
@apangin done

src/j9StackTraces.h
```diff
@@ -22,14 +22,12 @@
...
 
-const int MAX_J9_NATIVE_FRAMES = 128;
-
 struct J9StackTraceNotification {
     void* env;
     u64 counter;
     int num_frames;
     int reserved;
-    const void* addr[MAX_J9_NATIVE_FRAMES];
+    const void **addr;
```
apangin, line 30: Has the PR been tested with J9? I doubt, since this change does not look correct: the code for exchanging stack trace notifications assumes the notification is one contiguous block of memory.

tomershafir: sorry, missed that. Replaced notification::size() with sizeof() to write/read the buff pointer to/from the pipe.

Tested itimer and cpu sanity with j9 runtime and it looks correct.

apangin: The fix is not that simple, unfortunately. Now, the code in the signal handler allocates a temporary on-stack buffer and passes a pointer to it to another thread through a pipe. By the time another thread reads the notification, the pointer may become dangled, resulting in garbage reads or even a crash.

tomershafir: indeed, sorry again. For the feature, in the curr design its not possible to use pipe interface with continuous data bytes r/w because the arg value is dynamic, and because malloc is unsafe in sig handler. The 2 I see options are (1) considering pipe alternative or (2) considering safe global buff manager. Seems like a big effort, given probable relatively small real world benefit, where native stack traces are not so big due to the procedural C/C++ nature, and the static 128 limit shouldnt incur high overhead in the other direction. wdyt?

Issue #783, opened 2023-07-27
accept SIGTERM signal


zdyj3170101136:
for now, async profiler only accept SIGINT signal.

but systemctl stop send SIGTERM signal in default, it would be better if async profiler accept SIGTERM signal

apangin:
The should be easy to fix. Feel free to submit a PR.

zdyj3170101136:
> This should be easy to fix. Feel free to submit a PR.

sorry, i am familary with golang language, not familar with c++.

apangin:
Copied from https://github.com/async-profiler/async-profiler/pull/794#issuecomment-1674809948

> After thinking a bit more on the issue, I found the purpose of this change arguable.
> What is the actual use case for handling SIGTERM this way?
> 
> SIGINT is obvious: it's to interrupt an interactive profiling session. But SIGTERM is meant for services. asprof itself is not a service - it's a CLI utility. If it is wrapped into some service - it's the job of that service to handle interrupt signals.
>

zdyj3170101136:
> Copied from [#794 (comment)](https://github.com/async-profiler/async-profiler/pull/794#issuecomment-1674809948)
> 
> > After thinking a bit more on the issue, I found the purpose of this change arguable.
> > What is the actual use case for handling SIGTERM this way?
> > SIGINT is obvious: it's to interrupt an interactive profiling session. But SIGTERM is meant for services. asprof itself is not a service - it's a CLI utility. If it is wrapped into some service - it's the job of that service to handle interrupt signals.

1，not anyone know that asprof only accept SIGINT signal.
2,  not every process could handler process interrupt correctly, there are always some bug, and it is not easy to test.
3, systemd.kill  have lots of config https://www.freedesktop.org/software/systemd/man/systemd.kill.html, it is not easy to write a right config.

apangin:
Added `SIGTERM` handling

PR #794, opened 2023-08-11
Handle `SIGTERM` in the launcher

TimPushkin:
Makes the launcher also gracefully handle `SIGTERM` addressing #783

### DIFF
--- a/src/launcher/main.cpp
+++ b/src/launcher/main.cpp
@@ -214,7 +214,7 @@ static int duration = 60;
 static int pid = 0;
 static volatile unsigned long long end_time;
 
-static void sigint_handler(int sig) {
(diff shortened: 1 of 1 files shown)


apangin:
After thinking a bit more on the issue, I found the purpose of this change arguable.
What is the actual use case for handling SIGTERM this way?

SIGINT is obvious: it's to interrupt an _interactive_ profiling session. But SIGTERM is meant for services. `asprof` itself is not a service - it's a CLI utility. If it is wrapped into some service - it's the job of that service to handle interrupt signals.

TimPushkin:
I don't have an actual use case in mind, just decided to help with the open issue. Maybe @zdyj3170101136 can better explain his use for it.

PR #812, opened 2023-09-08
Add Async-Profiler Testing Framework

Michael-Jingzhong-Fan:
Added async-profiler testing framework. Testing framework tested to pass for all combinations of: Corretto; 8, 11, 17, 21, (22 Nightly if available). OS; Alpine Linux, Amazon Linux, MacOS. CPU Architecture; Arm64, x64

Change aborted: Includes additions to Output object of methods for more convenient asserts and outputs. 

New change: OAssert static class for more convenient assert and output logging


test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,169 @@
...
+                continue;
+            }
+
+            log.log(Level.INFO, "Running " + m.getDeclaringClass().getName() + '.' + m.getName() + "...");
+            try (final TestProcess p = new TestProcess(test, libExt)){
+                Object holder = (m.getModifiers() & Modifier.STATIC) == 0 ? m.getDeclaringClass().newInstance() : null;
+                m.invoke(holder, p);
+                log.info("OK");
+            } catch (CommandFail e) {
+                log.log(Level.WARNING, "Command call failed. Stderr: " + e.getStderr().toString(), e.getCause());
```
apangin, line 142: Why `CommandFail` should be handled differently than other exceptions?

test/test/lock/LockTests.java
```diff
@@ -0,0 +1,17 @@
...
+import one.profiler.test.Output;
+import one.profiler.test.Test;
+import one.profiler.test.TestProcess;
+
+public class LockTests {
+
+    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true)
+    public void datagramSocketLock(TestProcess p) throws Exception {
+        Output out = p.profile("-e cpu -d 3 -o collapsed --cstack dwarf"); 
+        out.assertRatioGreater("(Unsafe..ark|Unsafe..npark|PlatformEvent::.ark|PlatformEvent::.npark)", 0.1);
```
apangin, line 12: `Unsafe..ark` matches both Java method `Unsafe.park` and a native function `Unsafe_Park`. The test intends to check the presence of the native one. I know it fail on Alpine now, but that's fine - it's a real issue that needs to be fixed.

test/test/recovery/RecoveryTests.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test(mainClass = Suppliers.class, debugNonSafepoints = true)
+    public void suppliers(TestProcess p) throws Exception {
+        Output out = p.profile("-d 3 -e cpu -o collapsed --safe-mode 31");
+        out.assertRatioGreater("unknown_Java", 0.2);
+
+        out = p.profile("-d 3 -e cpu -o collapsed");
+        out.assertRatioGreater("itable stub", 0.01);
+        out.assertRatioGreater("Suppliers.loop" ,0.6);
```
apangin, line 50: whitespace formatting

apangin (changes requested):

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,278 @@
...
+            System.setProperty("java.util.logging.SimpleFormatter.format", "%4$s: %5$s%6$s%n");
+        }
+
+        if (args.length == 0) {
+            System.out.println("Usage: java " + Runner.class.getName() + " TestName ...");
+            System.exit(1);
+        }
+
+        Logger rootLogger = Logger.getLogger("");
+        String logLevel =  args[0];
```
apangin, line 229: I don't want making log level a required argument. It can be a system property instead (set with `-D` command line option).
Probably, we don't need fine-grained configuration for the logging level at all: this can be either default (INFO) or verbose (ALL).

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,278 @@
...
+        }
+        Boolean retainLogs;
+        if (args[1].equals("true")){
+            retainLogs = true;
+        } else if (args[1].equals("false")) {
+            retainLogs = false;
+        } else {
+            log.log(Level.WARNING, "retainLogS option is malformed. Got: " + args[1]);
+            retainLogs = false;
+        }
```
apangin, line 264: Like with log levels, it does not need to be a required argument.

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,266 @@
...
+        p.destroy();
+        
+        try {
+            waitForExit(p, 20);
+        } catch (TimeoutException e) {
+            log.log(Level.WARNING, "Failed to terminate child process", e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } finally {
+            try {
```
apangin, line 167: Too long block - move to a separate function.

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,266 @@
...
+                    File sourceDirectory = (tmpFiles.get("%pout") != null) ? tmpFiles.get("%pout") : tmpFiles.get("%out");
+                    File destinationDirectory = new File(directory, "stdout");
+                    sourceDirectory.renameTo(destinationDirectory);   
+                    
+                    sourceDirectory = (tmpFiles.get("%perr") != null) ? tmpFiles.get("%perr") : tmpFiles.get("%err");
+                    destinationDirectory = new File(directory, "stderr");
+                    sourceDirectory.renameTo(destinationDirectory);
+                    
+                    File profile = tmpFiles.get("%f");
+                    if (!(profile == null)) {
```
apangin, line 181: `if (profile != null)`

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,266 @@
...
+        
+        try {
+            waitForExit(p, 20);
+        } catch (TimeoutException e) {
+            log.log(Level.WARNING, "Failed to terminate child process", e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } finally {
+            try {
+                if (retainLog == true) {
```
apangin, line 168: `if (retainLog)`

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,266 @@
...
+                    String workingDirectory = System.getProperty("user.dir");
+                    File directory = new File(workingDirectory + "/build/test/logs/" + testName);
+                    directory.mkdirs();
+                    File sourceDirectory = (tmpFiles.get("%pout") != null) ? tmpFiles.get("%pout") : tmpFiles.get("%out");
+                    File destinationDirectory = new File(directory, "stdout");
+                    sourceDirectory.renameTo(destinationDirectory);   
+                    
+                    sourceDirectory = (tmpFiles.get("%perr") != null) ? tmpFiles.get("%perr") : tmpFiles.get("%err");
+                    destinationDirectory = new File(directory, "stderr");
```
apangin, line 177: In general, it's discouraged to reuse variables for different things.
In this particular case, I wouldn't introduce intermediate variables at all.

apangin (changes requested):

test/one/profiler/test/Runner.java
Michael-Jingzhong-Fan: It is not required from user perspective, the default is info. Or do you mean that you don't want it in the method signature?

apangin: I mean, not the required argument of `Runner` class. E.g., in case I want to debug some test in IDE without executing `make`.

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,284 @@
...
+            if (retainProfilesProperty.equals("true")) {
+                return true;
+            } else if (retainProfilesProperty.equals("false")) {
+                return false;
+            } else {
+                return true;
+            }
+        } else {
+            return true;
+        }
```
apangin, line 261: Can be a one-liner:
```
retainProfiles = !"false".equals(System.getProperty("retainProfiles"));
```

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            System.out.println("Usage: java " + Runner.class.getName() + " TestName ...");
+            System.exit(1);
+        }
+
+        Boolean retainLogs = configureLogging(args);
```
apangin, line 270: The return value for `configureLogging` is not obvious.
It's better to set some field instead. You can probably get `retainLogs` value in `TestProcess` static initializer, since `TestProcess` is the only place where this variable actually matters.

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,265 @@
...
+            
+            File errDirectory = (tmpFiles.get("%perr") != null) ? tmpFiles.get("%perr") : tmpFiles.get("%err");
+            errDirectory.renameTo(new File(directory, "stderr"));
+            
+            File profileDirectory = tmpFiles.get("%f");
+            if (profileDirectory != null) {
+                profileDirectory.renameTo(new File(directory, "profile"));
+            }
+        } catch (Throwable e) {
+            log.log(Level.WARNING, "Failed to retain logs: " + e.getMessage(), e.getStackTrace());
```
apangin, line 171: The last argument in all `log` statements with an exception should be the exception itself (not the stack trace).

test/one/profiler/test/JfrOutput.java
```diff
@@ -0,0 +1,140 @@
...
+            case "ContendedLock":
+                return countMethod(contendedLockAgg, method);
+            default: //customEvent case
+                return countMethod(customAgg, method);
+        }
+    }
+
+    public int countSTraces() {
+        return countMethod("");
+    }
```
apangin, line 139: Duplicates `countMethod()`.
Also, don't shorten StackTraces

test/one/profiler/test/Output.java
```diff
@@ -0,0 +1,82 @@
...
+
+package one.profiler.test;
+
+import java.util.Arrays;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+public class Output {
+    private final String[] lines;
+    public final JfrOutput jfrOut;
```
apangin, line 25: All methods of `Output` deal with `lines`. Why placing `jfrOut` here at all?

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        if (arch.length > 0 && !Arrays.asList(arch).contains(currentArch)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public static Boolean run(Method m, Map skipMap, Boolean retainlogs) throws Exception{
```
apangin, line 162: Use primitive `boolean` instead of `Boolean` wrapper when possible

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,261 @@
...
+
+            try (final TestProcess p = new TestProcess(test, libExt, testName);) {
+                Object holder = (m.getModifiers() & Modifier.STATIC) == 0 ? m.getDeclaringClass().newInstance() : null;
+                m.invoke(holder, p);
+                log.info("OK");
+            } catch (Throwable e) {
+                log.log(Level.FINE, e.getMessage());
+                log.log(Level.WARNING, "Test failed " + e.getCause().getMessage());
+                e.printStackTrace();
+                e.getCause().printStackTrace();
```
apangin, line 188: Log facility should print exception stack traces. Do not call `printStackTrace` manually.

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,261 @@
...
+        return passedAll;
+    }
+
+    public static Boolean run(Class<?> cls, Map skipMap, Boolean retainlogs) throws Exception{
+        Boolean passedAll = true;
+        for (Method m : cls.getMethods()) {
+            Boolean success = run(m, skipMap, retainlogs);
+            if (!success) {
+                passedAll = false;
+            }
```
apangin, line 201: Can be simplified: `passedAll &= run(...);`
Also, Boolean -> boolean

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,261 @@
...
+            System.setProperty("java.util.logging.ConsoleHandler.formatter", "java.util.logging.SimpleFormatter");
+        }
+        if (System.getProperty("java.util.logging.SimpleFormatter.format") == null) {
+            System.setProperty("java.util.logging.SimpleFormatter.format", "%4$s: %5$s%6$s%n");
+        }
+
+        String logLevelProperty = System.getProperty("logLevel");
+        if (logLevelProperty != null){
+            Level logLevel = Level.parse(logLevelProperty);
+            System.out.println("setting logLevel to " + logLevel);
```
apangin, line 217: No debug printing please

test/one/profiler/test/Runner.java
```diff
@@ -0,0 +1,261 @@
...
+        if (args.length == 0) {
+            System.out.println("Usage: java " + Runner.class.getName() + " TestName ...");
+            System.exit(1);
+        }
+
+        configureLogging(); 
+        Boolean retainlogs = !"false".equals(System.getProperty("retainLogs"));
+
+        String[] skips = System.getProperty("skip") != null ? System.getProperty("skip").split(",") : new String[0];
+        Map<String, String> skipMap = new HashMap<>();
```
apangin, line 237: Can be `HashSet`

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,367 @@
...
+    private void clearTempFiles() {
+        for (File file : tmpFiles.values()) {
+            file.delete();
+        }
+    }
+
+    private void moveLog(File file, File targetDirectory, String name) {
+        String fileName = file.getName();
+        String fileExtension = (fileName.lastIndexOf('.') != -1) ? fileName.substring(fileName.lastIndexOf('.')) : "";
+        fileExtension = (fileExtension == ".tmp") ? "": fileExtension; // Shave off .tmp
```
apangin, line 190: `String.equals`

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,367 @@
...
+            throw new IOException("Profiling call failed " + readPOut().toString() + readPErr().toString());
+        }
+
+        return readPOut();
+    }
+
+    public Output readFile(String fileId) throws IOException {
+        File f = tmpFiles.get(fileId);
+        String fileExtension = f.getName().replaceAll(".*\\.(\\w+)$", "$1");
+        if (fileExtension.equals("jfr")) {
```
apangin, line 292: Can be simplified: `if (f.getName().endsWith(".jfr"))`

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,367 @@
...
+            return readJfr(f);
+        } else {
+            return readFile(f);
+        }
```
apangin, line 296: That's a questionable solution. A test usually knows whether it expects JFR or Text output. Why not calling `readFile` vs. `readJfrFile` explicitly? They can return different objects: `Output` vs. `JfrOutput`, thus resolving my previous concern about embedding `JfrOutput` in `Output`.

test/one/profiler/test/TestProcess.java
```diff
@@ -0,0 +1,367 @@
...
+                    customAgg.collect(event);
+                    break;
+            }
+
+        }
+
+        JfrOutput jfrOut = new JfrOutput(readerObj, executionSampleAgg, allocationSampleAgg, liveObjectAgg, contendedLockAgg, customAgg); 
+        if (jfrOut == null) {
+            throw new IOException();
+        }
```
apangin, line 344: `new` expression can never return null.

test/test/cpu/CpuTests.java
```diff
@@ -0,0 +1,20 @@
...
+import one.profiler.test.TestProcess;
+import one.profiler.test.OsType;
+import one.profiler.test.JvmType;
+
+public class CpuTests {
+
+    @Test(mainClass = RegularPeak.class, jvmArgs = "-XX:+UseG1GC -Xmx1g -Xms1g", jvm = {JvmType.HOTSPOT})
+    public void regularPeak(TestProcess p) throws Exception {
+        p.profile("-e cpu -d 3 -f %f.jfr");
+        System.out.println("it closed already");
```
apangin, line 15: ?

apangin:
With the changes made by @roy-soumadipta, the test framework has been finally integrated. Thank you all!

PR #827, opened 2023-10-19
Converter: add warning when running with -jar instead of -cp

selckin:
Prevents users like me wasting time

### DIFF
--- a/src/converter/Main.java
+++ b/src/converter/Main.java
@@ -14,13 +14,41 @@
  * limitations under the License.
  */
 
+import java.util.Arrays;
+
 /**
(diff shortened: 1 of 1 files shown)


apangin:
Accepting both `-cp` and `-jar` would result in even better user experience.
Instead of printing a suggestion, converter can do the trick automatically.

selckin:
> Accepting both `-cp` and `-jar` would result in even better user experience. Instead of printing a suggestion, converter can do the trick automatically.

Updated

apangin:
I've merged PR with minor changes. Thank you for the contribution.

PR #844, opened 2023-11-23
Support alloc event in pprof converter

heart4lor:
Currently, the pprof converter only supports CPU events. This PR adds support for converting alloc events to pprof format.

### DIFF
--- a/src/converter/jfr2pprof.java
+++ b/src/converter/jfr2pprof.java
@@ -19,6 +19,8 @@ import one.jfr.Dictionary;
 import one.jfr.JfrReader;
 import one.jfr.MethodRef;
 import one.jfr.StackTrace;
+import one.jfr.event.AllocationSample;
+import one.jfr.event.Event;
 import one.jfr.event.ExecutionSample;
 import one.proto.Proto;
 
@@ -95,8 +97,18 @@ public class jfr2pprof {
         this.reader = reader;
     }
 
-    // `Proto` instances are mutable, careful with reordering
+
+
     public void dump(final OutputStream out) throws Exception {
+        dump(out, null);
+    }
+
+    // `Proto` instances are mutable, careful with reordering
+    public void dump(final OutputStream out, Class<? extends Event> eventClass) throws Exception {
+        if ((eventClass != ExecutionSample.class && eventClass != AllocationSample.class)) {
+            // Fallback to execution sample event
+            eventClass = ExecutionSample.class;
+        }
         // Mutable IDs, need to start at 1
         int functionId = 1;
         int locationId = 1;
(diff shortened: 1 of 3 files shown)


src/converter/jfr2pprof.java
```diff
@@ -116,26 +128,38 @@ public void dump(final OutputStream out) throws Exception {
...
-        profile.field(PROFILE_STRING_TABLE, "cpu".getBytes(StandardCharsets.UTF_8));
-        sampleType.field(VALUETYPE_TYPE, stringId++);
-
-        profile.field(PROFILE_STRING_TABLE, "nanoseconds".getBytes(StandardCharsets.UTF_8));
-        sampleType.field(VALUETYPE_UNIT, stringId++);
+        if (eventClass == AllocationSample.class) {
+            profile.field(PROFILE_STRING_TABLE, "alloc_space".getBytes(StandardCharsets.UTF_8));
+            sampleType.field(VALUETYPE_TYPE, stringId++);
+            profile.field(PROFILE_STRING_TABLE, "bytes".getBytes(StandardCharsets.UTF_8));
+            sampleType.field(VALUETYPE_UNIT, stringId++);
```
apangin, line 135: Why doesn't allocation profile have timestamped events?

src/converter/jfr2pprof.java
```diff
@@ -95,8 +97,18 @@ public jfr2pprof(final JfrReader reader) {
         this.reader = reader;
     }
 
-    // `Proto` instances are mutable, careful with reordering
+
+
```
apangin, line 101: No redundant whitespace, please.

apangin:
Note there is already another PR #713 for the similar functionality.
If you wish, you may contribute to the existing PR instead. I will likely accept one of these, whichever is ready first. Thanks.

heart4lor:
@apangin Hi, thanks for advices. I've resolved comments above. For `alloc` event, I kept the `nanosecond` field, and put allocated bytes key-value into [sample labels](https://github.com/google/pprof/blob/main/proto/profile.proto#L130C6-L130C6). What do you think about it? Please review again, thanks!

src/converter/jfr2pprof.java
```diff
@@ -116,26 +133,46 @@ public void dump(final OutputStream out) throws Exception {
...
+            final Proto sample = new Proto(1_000);
             final long nanosSinceLastSample = (jfrSample.time - previousTime) * 1_000_000_000 / reader.ticksPerSec;
-            final Proto sample = new Proto(1_000).field(SAMPLE_VALUE, nanosSinceLastSample);
+            sample.field(SAMPLE_VALUE, nanosSinceLastSample);
+            if (eventClass == AllocationSample.class) {
+                final long allocationSize = jfrSample.value();
+                final Proto label = new Proto(16)
+                        .field(LABEL_KEY, allocationSizeKeyId)
+                        .field(LABEL_NUM, allocationSize)
+                        .field(LABEL_NUM_UNIT, allocationSizeUnitId);
```
apangin, line 173: Why units are recorded with every sample? Can it be anything other than bytes?

src/converter/jfr2pprof.java
```diff
@@ -208,20 +245,28 @@ private byte[] getMethodName(final long methodId, final byte methodType) {
...
+            System.out.println("options support one of the following events:");
+            System.out.println("  --cpu         CPU events (default)");
+            System.out.println("  --alloc       Allocation events");
             System.exit(1);
         }
 
-        File dst = new File(args[1]);
+        Class<? extends Event> eventClass = args.alloc ? AllocationSample.class : ExecutionSample.class;
+
+        String output = Optional.ofNullable(args.output).orElse(args.input).replace(".jfr", ".pprof");
```
apangin, line 261: If output name is given, this will replace ".jfr" in the output name too, right? IMO, this is unexpected behavior.

src/converter/jfr2pprof.java
```diff
@@ -84,6 +87,12 @@ public boolean equals(final Object other) {
     // Line IDs
     public static final int LINE_FUNCTION_ID = 1;
     public static final int LINE_LINE = 2;
+    public static final int SAMPLE_LABEL = 3;
+
+    // Label IDs
+    public static final int LABEL_KEY = 1;
+    public static final int LABEL_NUM = 3;
+    public static final int LABEL_NUM_UNIT = 4;
```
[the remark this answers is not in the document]
apangin: Ah, never mind. Found the answer in [profile.proto](https://github.com/google/pprof/blob/main/proto/profile.proto#L130).

heart4lor:
@apangin All resolved, please review again. Thanks!

src/converter/jfr2pprof.java
```diff
@@ -116,26 +133,45 @@ public void dump(final OutputStream out) throws Exception {
...
+            profile.field(PROFILE_STRING_TABLE, "nanoseconds".getBytes(StandardCharsets.UTF_8));
+            sampleType.field(VALUETYPE_UNIT, stringId++);
+        }
 
-        profile.field(PROFILE_STRING_TABLE, "nanoseconds".getBytes(StandardCharsets.UTF_8));
-        sampleType.field(VALUETYPE_UNIT, stringId++);
+        profile.field(PROFILE_STRING_TABLE, "allocation_size".getBytes(StandardCharsets.UTF_8));
+        final int allocationSizeKeyId = stringId++;
+        profile.field(PROFILE_STRING_TABLE, "bytes".getBytes(StandardCharsets.UTF_8));
+        final int allocationSizeUnitId = stringId++;
```
apangin, line 151: Are these strings only need for allocation profile? Does it make sense to put them under `if (eventClass == AllocationSample.class)`?

src/converter/jfr2pprof.java
```diff
@@ -116,26 +133,45 @@ public void dump(final OutputStream out) throws Exception {
...
             final StackTrace stackTrace = stackTraces.get(jfrSample.stackTraceId);
             final long[] methods = stackTrace.methods;
             final byte[] types = stackTrace.types;
 
+            final Proto sample = new Proto(1_000);
             final long nanosSinceLastSample = (jfrSample.time - previousTime) * 1_000_000_000 / reader.ticksPerSec;
-            final Proto sample = new Proto(1_000).field(SAMPLE_VALUE, nanosSinceLastSample);
+            sample.field(SAMPLE_VALUE, nanosSinceLastSample);
+            if (eventClass == AllocationSample.class) {
+                sample.field(SAMPLE_LABEL, allocationLabel.field(LABEL_NUM, jfrSample.value()));
```
apangin, line 173: `field` method mutates the object; it's not correct to reuse `allocationLabel` this way.

src/converter/jfr2pprof.java
```diff
@@ -208,20 +244,28 @@ private byte[] getMethodName(final long methodId, final byte methodType) {
...
+            System.out.println("options support one of the following events:");
+            System.out.println("  --cpu         CPU events (default)");
+            System.out.println("  --alloc       Allocation events");
             System.exit(1);
         }
 
-        File dst = new File(args[1]);
+        Class<? extends Event> eventClass = args.alloc ? AllocationSample.class : ExecutionSample.class;
+
+        String output = Optional.ofNullable(args.output).orElse(args.input.replace(".jfr", ".pprof"));
```
apangin, line 260: I'd rewrite this with a regular `if`: `Optional` makes this harder to follow. Also, `args.input.replace` will be called regardless of the value of `output`.

apangin:
Hi, is this PR still relevant?

apangin:
Obsoleted by #905

PR #859, opened 2023-12-04
Make converter Arguments class public

sureshg:
Right now now [Arguments](https://github.com/async-profiler/async-profiler/blob/master/src/converter/Arguments.java) class is `package-private`, so is not possible to use `jfr2flame(JfrReader jfr, Arguments args)` programatically to create flamrgrpah (unless i copy the same class locally for non-module apps).  This PR is to make this class public to make it more API friendly.


apangin:
Thanks, merged

Issue #873, opened 2024-01-03
Regarding Java 11 and above, the accuracy issue of ObjectSampler in --total mode


yanglong1010:
hi Andrei

In this code, the number of calls to `testSmallAlloc` and `testLargeAlloc` is different, but their total memory allocations in bytes are close, which can also be proven from the actual running logs.

```java
public class AllocTest {
    private static volatile long n1 = 0;
    private static volatile long n2 = 0;
    private static Thread t1;

    public static void main(String[] args) throws Exception {
        testAlloc();
        while (true) {
            Thread.sleep(1000);
            System.out.println("small total: " + n1);
            System.out.println("large total: " + n2);
            System.out.println();
        }
    }

    private static void testAlloc() {
        t1 = new Thread(() -> {
            while (true) {
                for (int i = 0; i < 256; i++) {
                    n1 += testSmallAlloc(1024);
                }
                n2 += testLargeAlloc(1024 * 256);
            }
        });
        t1.start();
    }

    private static int testSmallAlloc(int size) {
        return realAlloc(size);
    }

    private static int testLargeAlloc(int size) {
        return realAlloc(size);
    }

    private static int realAlloc(int size) {
        byte[] arr = new byte[size];
        return arr.length;
    }
}

```
```bash
java11/bin/java -Xint AllocTest

small total: 2059834368
large total: 2060451840

small total: 4829180928
large total: 4829216768

small total: 7630267392
large total: 7630225408

small total: 10429923328
large total: 10429923328
````

Using `--total` mode, it is expected that the proportions of `testSmallAlloc` and `testLargeAlloc` should be close in the flame graph, but in fact `testLargeAlloc` is 99.91% while `testSmallAlloc` is only 0.81%.

``` bash
/root/async-profiler-fork/build/bin/asprof --alloc 256k -f static/profile.html --total -d 10 jps
```
![image](https://github.com/async-profiler/async-profiler/assets/24515208/c40cb5f0-60c9-4872-9ae2-3fb4d9416989)

The reason is that `size` is used as `counter` in `ObjectSampler::recordAllocation`, which causes only the array size to be recorded during `testSmallAlloc`, which is actually the `interval` in `JVMTI_EVENT_SAMPLED_OBJECT_ALLOC` (here we ignore the randomness of `interval` in `JVMTI`).
https://github.com/async-profiler/async-profiler/blob/915b09067c14e3b66826a988f7cbb768c4ab1e18/src/objectSampler.cpp#L163

[...]

PR #876, opened 2024-01-05
#873: the accuracy issue of ObjectSampler in --total mode

yanglong1010:
Hi Andrei

As discussed in [#873](https://github.com/async-profiler/async-profiler/issues/873).

Because [JEP 331](https://openjdk.org/jeps/331) doesn't return the actual pseudo-random `interval` in `JVM`, the `event._total_size` is the best we can use now.

I would appreciate it if you can help review this.


apangin:

PR #879, opened 2024-01-16
fix(jfr2pprof): Fix multichunk jfr symbols and location's line numbers

korniltsev:
- Do not use one.jfr.JfrReader#readAllEvents(java.lang.Class<E>) as it may trigger reading new chunk in multichunk jfrs, overwriting cp, producing wrong data(I observerd wrong class names). Note: it is still used in netflix code I did not touch.

- Use Location(method+line) as dedup key to fix line numbers

I've used this JFR file for testing, captured by Intellij IDEA Ultimate on Linux
[FastSlow_2024_01_16_180855.jfr.zip](https://github.com/async-profiler/async-profiler/files/13949433/FastSlow_2024_01_16_180855.jfr.zip)

### DIFF
--- a/src/converter/jfr2pprof.java
+++ b/src/converter/jfr2pprof.java
@@ -19,6 +19,7 @@ import one.jfr.Dictionary;
 import one.jfr.JfrReader;
 import one.jfr.MethodRef;
 import one.jfr.StackTrace;
+import one.jfr.event.Event;
 import one.jfr.event.ExecutionSample;
 import one.proto.Proto;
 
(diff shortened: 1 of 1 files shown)


src/converter/jfr2pprof.java
```diff
@@ -56,6 +57,29 @@ public boolean equals(final Object other) {
...
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Location location = (Location) o;
+            return line == location.line && Objects.equals(method, location.method);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(method, line);
```
apangin, line 79: `Objects.hash` is inefficient because of varargs - please rewrite the method without `Objects.hash`. As a general rule, `hashCode` should not allocate.

apangin (commented):
Thank you for the Pull Request. The changes are good except for the minor performance note.

src/converter/jfr2pprof.java
korniltsev: Thanks. Rewritten.

apangin:

PR #889, opened 2024-02-07
Fix Proto encoding of negative integers

cklin:
Currently the `Proto` class determines whether to emit a continuing byte for the integer value `n` this way:

```java
while (n > 0x7f) {
```

The intention of the comparison is to check whether `n` has any bit set below the lowest 7 bits.  Unfortunately it does not work if `n` is negative, which causes the loop to be skipped entirely.  When that happens, the `jfr2pprof` produces a protobuf profile that is rejected by `pprof`.

This PR fixes the bug so that `Proto` encodes negative integer values correctly.

### DIFF
--- a/src/converter/one/proto/Proto.java
+++ b/src/converter/one/proto/Proto.java
@@ -71,7 +71,7 @@ public class Proto {
         int length = n == 0 ? 1 : (38 - Integer.numberOfLeadingZeros(n)) / 7;
         ensureCapacity(length);
 
-        while (n > 0x7f) {
+        while ((n >>> 7) != 0) {
             buf[pos++] = (byte) (0x80 | (n & 0x7f));
             n >>>= 7;
         }
@@ -82,7 +82,7 @@ public class Proto {
         int length = n == 0 ? 1 : (70 - Long.numberOfLeadingZeros(n)) / 7;
         ensureCapacity(length);
 
-        while (n > 0x7f) {
+        while ((n >>> 7) != 0) {
             buf[pos++] = (byte) (0x80 | (n & 0x7f));
             n >>>= 7;
         }


src/converter/one/proto/Proto.java
```diff
@@ -71,7 +71,7 @@ public void writeInt(int n) {
         int length = n == 0 ? 1 : (38 - Integer.numberOfLeadingZeros(n)) / 7;
         ensureCapacity(length);
 
-        while (n > 0x7f) {
+        while (n & 0x7f != n) {
```
apangin, line 74: Why can a number be negative? In what cases?

src/converter/one/proto/Proto.java
```diff
@@ -82,7 +82,7 @@ public void writeLong(long n) {
         int length = n == 0 ? 1 : (70 - Long.numberOfLeadingZeros(n)) / 7;
         ensureCapacity(length);
 
-        while (n > 0x7f) {
+        while (n & 0x7f != n) {
```
apangin, line 85: A canonical way to check if an integer has higher bits set is
```
(n & ~0x7f) != 0
```
or
```
(n >>> 7) != 0
```

src/converter/one/proto/Proto.java
cklin: The case I hit was `nanosSinceLastSample` in `jfr2pprof.java` being negative.

I don't know if that is supposed to happen; all I can tell you is that it happened, with async-profiler 3.0 and the default clock source.

src/converter/one/proto/Proto.java
cklin: I changed the check based on your suggestion.

apangin:
Thank you for the fix. Merged.

PR #906, opened 2024-03-20
FlameGraph: search with Command+F key (⌘F) on mac

qwwdfsad:
* ctrl+f is already overridden
* On OS X the default way to search on page is command+f, would be nice to have it in flamegraphs as well


apangin:

PR #939, opened 2024-05-28
Make parsing of features more strict and prohibit ill-formed values

qwwdfsad:
Prevent errors in the case when a completely invalid value was passed (e.g. 'vtabl' or 'comptaks').

Rationale: typos in command-line scripts when the value is ill-formed before this commit are silently ignored. Taking into account that `--features` (potentially slightly) change the profiling result, such an error might skew the expected results and cause some debugging time that could've been avoided.

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -76,7 +76,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     interval=N       - sampling interval in ns (default: 10'000'000, i.e. 10 ms)
 //     jstackdepth=N    - maximum Java stack depth (default: 2048)
 //     signal=N         - use alternative signal for cpu or wall clock profiling
-//     features=LIST    - advanced stack trace features (vtable, comptask)"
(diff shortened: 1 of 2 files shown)


qwwdfsad:
Note that it doesn't account for partial failures (e.g. `-F vtabl2e,comptask `). I can try and fix it if you want to -- I see that arguments are concatenated using `+`, and it should be quite straightforward; though it's backwards-incompatible for Java API consumers (i.e. `Java_one_profiler_AsyncProfiler_execute0` invokes `parse` immediately without replacing `,` with `+`). Up to you

apangin:
Failing with an error may break forward compatibility when new features are introduced or removed. Such incompatibilities caused issues in the past for some customers. After that, I changed an error for an unknown argument to a warning.
Similarly, an unknown feature should print a warning at most, but should not prevent profiler from starting.

apangin:
Also, `probesp` is intentionally left undocumented - it's not stable and not recommended for production use.

qwwdfsad:
Thanks! I guess the PR can be safely closed in that case

Issue #955, opened 2024-07-12
Add an asprof argument to specify the path of libasyncProfier.so in the container


yanglong1010:
### Describe the feature

Hi, in 2.9 and bellow, there is an argument `--lib path` in `profiler.sh` to specify the path of libasyncProfier.so in the container.
But 3.0 and above, this argument is missing. I think this argument is necessary.

### Use Case

```bash
asprof -e cpu --libpath /tmp/libasyncProfiler.so 1234
```

### Proposed Solution

Add the argument back.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
The plan is to make `asprof` automatically provide `libasyncProfier.so` to the process inside a container - in this case, the option will be redundant.
However, given that there is no timeline for this feature, and since adding `--libpath` is straightforward, I don't mind if you add this option.

PR #956, opened 2024-07-15
#955 Add an asprof option to specify the path of libasyncProfier.so in the container

yanglong1010:
Hi, could I have a review of this PR. I would appreciate it.

### Description
Add --libpath option to specify the path of libasyncProfier.so in the container

### Related issues
https://github.com/async-profiler/async-profiler/issues/955

### Motivation and context
In async-profiler 2.9, there is an option --lib path in profiler.sh to specify the path of libasyncProfier.so in the container.
But in 3.0 and above, this option is missing. After [discussion](https://github.com/async-profiler/async-profiler/issues/955), I know that there is a plan is to make asprof automatically provide libasyncProfier.so to the process inside a container, but no timeline for this feature, adding --libpath is straightforward and acceptable.

### How has this been tested?
The tests below ran in both container and non-container scenarios, all passed:
- If --libpath is not used, the libasyncProfiler.so relative to asprof is used.
- If --libpath is used and the file exists, the specified libasyncProfiler.so is used.
- If --libpath is used and the file does not exist, asprof reports: `Target JVM failed to load /tmp/libasyncProfiler.so.1`.
---

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -82,6 +82,7 @@ static const char USAGE_STRING[] =
     "  --jfrsync config  synchronize profiler with JFR recording\n"
     "  --fdtransfer      use fdtransfer to serve perf requests\n"
     "                    from the non-privileged target\n"
+    "  --libpath path    full path to libasyncProfiler.so in the container\n"
     "\n"
     "<pid> is a numeric process ID of the target JVM\n"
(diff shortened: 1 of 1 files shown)


apangin:
Merged. Thank you for the contribution

Issue #929, opened 2024-05-12
Support debuginfod


mohe2015:
Hi,

it seems like [debuginfod](https://sourceware.org/elfutils/Debuginfod.html) is becoming a popular way to distribute debug information and e.g. for Arch Linux it's the only easy way. It would be great if this would be supported. The easiest way would probably be to be able to read already cached debuginfod information or use `debuginfod-find`.

Kind regards
Moritz

apangin:
Thank you for the proposal, it sounds interesting.
Async-profiler will hardly download anything on its own, but making use of cached debuginfod files definitely makes sense.

I will look into this sometime later, otherwise feel free to contribute a PR, if you wish.

apangin:
Thanks to @krk, async-profiler can now load symbols form debuginfod cache.

Trass3r:
> The easiest way would probably be to be able to read already cached debuginfod information or use debuginfod-find.

Ok so just for the record this has been implemented.
It won't automatically pull new symbols. (?!)

apangin:
@Trass3r Correct. async-profiler now reads cached debuginfod symbols, but does not automatically pull new ones into the cache.

PR #962, opened 2024-07-28
#929 Load symbols from debuginfod cache

krk:
### Description
Parse symbols from debuginfod cache as a fallback. Cache can be found at `$DEBUGINFOD_CACHE_PATH`, `$XDG_CACHE_HOME/debuginfod_client/` or `$HOME/.cache/debuginfod_client/`, whichever folder is found first.

### Related issues
https://github.com/async-profiler/async-profiler/issues/929

### Motivation and context
To simplify use cases when debuginfod is easier to use.

### How has this been tested?
Tested with the Dockerfile and sample java app that loads libcrypt and does SHA-512 hashing to keep CPU busy in the native-land: https://gist.github.com/krk/6df8ccd561bfa76a1a7e236e42baad1c

Without symbols loaded:

```
          ns  percent  samples  top
  ----------  -------  -------  ---
  5270020407   94.58%      527  /usr/lib/libcrypt.so.2.0.0
    39999796    0.72%        4  /usr/lib/libc.so.6
```

with `debuginfod-find debuginfo /lib64/libcrypt.so`:

```
          ns  percent  samples  top
  ----------  -------  -------  ---
  4690012040   84.32%      469  SHA512_Transform
   449998244    8.09%       45  _crypt_SHA512_Final
   100000082    1.80%       10  _crypt_SHA512_Update
    70129705    1.26%        7  /usr/lib/libc.so.6
```

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -166,6 +166,7 @@ typedef Elf32_Dyn  ElfDyn;
 
 
 static bool musl = false;
+static char _debuginfod_cache_buf[PATH_MAX] = {0};
 
 class ElfParser {
   private:
@@ -227,6 +228,9 @@ class ElfParser {
     bool loadSymbolsUsingDebugLink();
     void loadSymbolTable(const char* symbols, size_t total_size, size_t ent_size, const char* strings);
     void addRelocationSymbols(ElfSection* reltab, const char* plt);
+    const char* getDebuginfodCache();
+    bool parseFileFromDebug(char* path, const char* build_id, const int build_id_len);
+    bool parseFileFromDebuginfodCache(char* path, const char* build_id, const int build_id_len);
 
   public:
     static void parseProgramHeaders(CodeCache* cc, const char* base, const char* end, bool relocate_dyn);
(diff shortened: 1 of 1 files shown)


src/symbols_linux.cpp
```diff
@@ -457,7 +461,56 @@ void ElfParser::loadSymbols(bool use_debug) {
...
+    if(_debuginfod_cache_buf[0]) {
+        return _debuginfod_cache_buf;
+    }
+
+    const char* env_vars[] = {"DEBUGINFOD_CACHE_PATH", "XDG_CACHE_HOME", "HOME"};
+    const char* suffixes[] = {"", "debuginfod_client/", ".cache/debuginfod_client/"};
+
+    for (int i = 0; i < sizeof(env_vars); ++i) {
+        const char* env_val = getenv(env_vars[i]);
+        if (!env_val || strlen(env_val) == 0) {
```
apangin, line 474: `env_val[0]` is a better way to check for non-empty string, you already do this in other places.

src/symbols_linux.cpp
```diff
@@ -457,7 +461,56 @@ void ElfParser::loadSymbols(bool use_debug) {
...
+            p += snprintf(p, 3, "%02hhx", build_id[i]);
+        }
+        strcpy(p, "/debuginfo");
+
+        return parseFile(_cc, _base, path, false);
+    }
+    return false;
+}
+
+// Load symbols from /usr/lib/debug/.build-id/ab/cdef1234.debug, where abcdef1234 is Build ID.
```
apangin, line 512: I think it makes sense to list all 4 locations in the search order.

apangin (commented):
Thank you for working on this feature. The proposed solution makes total sense. I added several comments inline.

apangin:
Thank you for the contribution. Merged.

PR #975, opened 2024-08-21
Add flamegraph to collapsed conversion in test framework

roy-soumadipta:
### Description
* Addition of flamegraph to collapsed conversion for test assertions.
* Uncommenting flamegraph test assertions.

### Related issues
N/A

### Motivation and context
Flamegraph generation has changed leading to flamegraph test assertion failures. The test framework required changes to accommodate the flamegraph changes. 

### How has this been tested?
Tested on AL2_x86_64, aarch64 and macOs

---

### DIFF
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -5,6 +5,9 @@
 
 package one.profiler.test;
 
+import one.convert.Arguments;
+import one.convert.FlameGraph;
+
 import java.io.*;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
@@ -251,8 +254,8 @@ public class TestProcess implements Closeable {
         log.log(Level.FINE, "Profiling " + cmd);
 
         Process p = new ProcessBuilder(cmd)
-                .redirectOutput(createTempFile("%pout"))
-                .redirectError(createTempFile("%perr"))
+                .redirectOutput(createTempFile(PROFOUT))
+                .redirectError(createTempFile(PROFERR))
                 .start();
 
         waitForExit(p, 10);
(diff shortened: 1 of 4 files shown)


test/one/profiler/test/TestProcess.java
```diff
@@ -261,19 +264,43 @@ public Output profile(String args, boolean sudo) throws IOException, TimeoutExce
             throw new IOException("Profiling call failed: " + readFile(PROFERR));
         }
 
-        return readFile(PROFOUT);
+        return readFile(PROFOUT, args.split("\\s+"));
```
apangin, line 267: This looks strange. `args` here is the command line for `asprof`, but it is used as a command line for the converter.

test/one/profiler/test/TestProcess.java
```diff
@@ -261,19 +264,43 @@ public Output profile(String args, boolean sudo) throws IOException, TimeoutExce
...
-    public Output readFile(String fileId) {
+    public Output readFile(String fileId, String... args) {
         File f = getFile(fileId);
-        try (Stream<String> stream = Files.lines(f.toPath())) {
-            return new Output(stream.toArray(String[]::new));
+        String extension = getExtFromFile(f);
+        try {
+            if (extension.equals(".html")) {
+                for (int i = 0; i < args.length; i++) {
+                    args[i] = args[i].equalsIgnoreCase("flamegraph") ? "collapsed" : args[i];
```
apangin, line 280: It's not a good idea to modify the array in place

test/one/profiler/test/TestProcess.java
```diff
@@ -261,19 +264,43 @@ public Output profile(String args, boolean sudo) throws IOException, TimeoutExce
...
-        return readFile(PROFOUT);
+        return readFile(PROFOUT, args.split("\\s+"));
     }
 
     public File getFile(String fileId) {
         return tmpFiles.get(fileId);
     }
 
-    public Output readFile(String fileId) {
+    public Output readFile(String fileId, String... args) {
```
apangin, line 274: `readFile` should return file contents as is. Implicit conversion makes semantics of the method obscure.
Also, `.html` file will not be always a flame graph. For example, one of recent PRs adds heatmaps that also produce output in `.html` format.

test/test/alloc/AllocTests.java
```diff
@@ -30,11 +30,9 @@ public void allocTotal(TestProcess p) throws Exception {
         Output out = p.profile("-e alloc -d 3 -o collapsed --total");
         assert out.samples("java.util.HashMap\\$Node\\[]") > 1_000_000;
 
-        /* Commented out as part of disabling asserting JFR output.
         out = p.profile("stop -o flamegraph --total");
-        assert out.contains("f\\(\\d+,\\d+,\\d+,\\d,'java.lang.Long'\\)");
-        assert out.contains("f\\(\\d+,\\d+,\\d+,\\d,'java.util.HashMap\\$Node\\[]'\\)");
-        */
+        assert out.contains("java\\.lang\\.Long");
```
apangin, line 34: Implicit conversion of a flamegraph to collapsed output not only complicates the code, but also harms testing capabilities. For example, if I remove `-o flamegraph` from the above command, the test will still pass.
Tests should be explicit about output they expect. I propose adding `convert` method in the `Output` class which a test will explicitly call before calling `contains` etc.

Issue #974, opened 2024-08-21
Lookup::fillJavaMethodInfo may cause crash on JDK 8/11/17


yanglong1010:
### Describe the bug

Hi, Andrei

We recently encountered a crash. After analysis, I think this problem may occur on OpenJDK 8, 11, and 17. It will not occur on OpenJDK 18 and above because [JDK-8268364](https://bugs.openjdk.org/browse/JDK-8268364) avoids this problem at the JDK level.

- JDK 8
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fa5e1ef9108, pid=12810, tid=0x00007fa5234f4700
#
# JRE version: OpenJDK Runtime Environment (8.0_412-b08) (build 1.8.0_412-b08)
# Java VM: OpenJDK 64-Bit Server VM (25.412-b08 mixed mode linux-amd64 compressed oops)
# Problematic frame:

V  [libjvm.so+0x766108]  oopDesc::is_a(Klass*) const+0x48
V  [libjvm.so+0x7635ec]  jvmti_GetClassSignature+0x15c
C  [libasyncProfiler.so+0x35389]  Recording::finishChunk()+0x20c9
C  [libasyncProfiler.so+0x3599c]  FlightRecorder::stop()+0x9c
C  [libasyncProfiler.so+0x39559]  Profiler::stop(bool)+0x159
C  [libasyncProfiler.so+0x3b4af]  Profiler::runInternal(Arguments&, Writer&) [clone .constprop.772]+0x32f
C  [libasyncProfiler.so+0x3b6d1]  Profiler::run(Arguments&)+0x181
C  [libasyncProfiler.so+0x3d019]  Agent_OnAttach+0x209
V  [libjvm.so+0x7c12d9]  JvmtiExport::load_agent_library(AttachOperation*, outputStream*)+0x329
V  [libjvm.so+0x3328df]  attach_listener_thread_entry(JavaThread*, Thread*)+0x15f
V  [libjvm.so+0xb11d32]  JavaThread::thread_main_inner()+0x212
V  [libjvm.so+0x942a62]  java_start(Thread*)+0xf2
C  [libpthread.so.0+0x7ea5]  start_thread+0xc5
```

- JDK 11
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f73c96aa658, pid=31632, tid=31752
#
[...]

PR #981, opened 2024-09-02
Avoid JVM crash by deleting JNI refs after calling GetMethodDeclaringClass

krk:
This is a partial fix for https://github.com/async-profiler/async-profiler/issues/974, complementing JDK-8268364.

### How has this been tested?
Backported JDK-8268364 into Java 17 and tested with reproducer at [gist](https://gist.github.com/krk/a0800bac5bc5a01709be85637285a965):

```
javac Main.java
gcc -shared -I"$JAVA_HOME/include" -I"$JAVA_HOME/include/linux" -fPIC repro.cpp -orepro.so

# Low Xmx to pressure GC into unloading classes sooner.
java -agentpath:"$(pwd)/repro.so" -Xmx100m Main
```

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -134,6 +134,8 @@ class Lookup {
     Dictionary _symbols;
 
   private:
+    JNIEnv* _jni;
+
     void fillNativeMethodInfo(MethodInfo* mi, const char* name, const char* lib_name) {
         if (lib_name == NULL) {
             mi->_class = _classes->lookup("");
@@ -198,6 +200,9 @@ class Lookup {
             mi->_sig = _symbols.lookup("()L;");
         }
 
+        if (method_class) {
+            _jni->DeleteLocalRef(method_class);
+        }
         jvmti->Deallocate((unsigned char*)method_sig);
         jvmti->Deallocate((unsigned char*)method_name);
         jvmti->Deallocate((unsigned char*)class_name);
@@ -227,7 +232,7 @@ class Lookup {
 
   public:
     Lookup(MethodMap* method_map, Dictionary* classes) :
-        _method_map(method_map), _classes(classes), _packages(), _symbols() {
+        _method_map(method_map), _classes(classes), _packages(), _symbols(), _jni(VM::jni()) {
     }
 
     MethodInfo* resolveMethod(ASGCT_CallFrame& frame) {
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -80,7 +80,8 @@ FrameName::FrameName(Arguments& args, int style, int epoch, Mutex& thread_names_
     _cache_epoch((unsigned char)epoch),
     _cache_max_age(args._mcache),
     _thread_names_lock(thread_names_lock),
-    _thread_names(thread_names)
+    _thread_names(thread_names),
+    _jni(VM::jni())
 {
     // Require printf to use standard C format regardless of system locale
     _saved_locale = uselocale(newlocale(LC_NUMERIC_MASK, "C", (locale_t)0));
@@ -189,6 +190,9 @@ void FrameName::javaMethodName(jmethodID method) {
         _str.assign(buf);
     }
 
+    if (method_class) {
+        _jni->DeleteLocalRef(method_class);
+    }
     jvmti->Deallocate((unsigned char*)class_name);
     jvmti->Deallocate((unsigned char*)method_sig);
     jvmti->Deallocate((unsigned char*)method_name);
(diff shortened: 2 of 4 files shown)


src/flightRecorder.cpp
```diff
@@ -180,6 +180,7 @@ class Lookup {
         }
 
         jvmtiEnv* jvmti = VM::jvmti();
+        JNIEnv* jni = VM::jni();
```
apangin, line 183: `VM::jni()` is not an expensive call, but it's not free either - not a trivial getter like `VM::jvmti()`.
So, I'd create `JNIEnv*` once in `Lookup` constructor and cache it as a field.

src/frameName.cpp
```diff
@@ -167,6 +167,7 @@ void FrameName::javaMethodName(jmethodID method) {
     char* method_sig = NULL;
 
     jvmtiEnv* jvmti = VM::jvmti();
+    JNIEnv* jni = VM::jni();
```
apangin, line 170: Similar to above: obtain `JNIEnv*` once in `FrameName` constructor.

src/flightRecorder.cpp
krk: Thanks, cached `JNIEnv`.

src/frameName.cpp
krk: Cached as above.

src/frameName.cpp
```diff
@@ -80,7 +80,8 @@ FrameName::FrameName(Arguments& args, int style, int epoch, Mutex& thread_names_
     _cache_epoch((unsigned char)epoch),
     _cache_max_age(args._mcache),
     _thread_names_lock(thread_names_lock),
-    _thread_names(thread_names)
+    _thread_names(thread_names),
+    _jni(VM::jni())
```
apangin, line 84: There will be the same problem with native apps as above.
I suggest simply returning `NULL` from `VM::jni()` if `_vm == NULL`.

src/flightRecorder.cpp
```diff
@@ -198,6 +200,9 @@ class Lookup {
             mi->_sig = _symbols.lookup("()L;");
         }
 
+        if (method_class) {
+            _jni->DeleteLocalRef(method_class);
```
D-D-H, line 204: Hi,

After investigating why the call to `DeleteLocalRef` is added here(cannot completely fix the problem), I think we can fix the problem in OpenJDK upstream.

Do you plan to fix it? If not, we will file an issue at OpenJDK.

yanglong1010:
I ran the reproducer. The stack trace in `hs_err log` produced by this reproducer is different from the stack trace I provided in #974. The reproducer crashes in GC threads, [D-D-H](https://github.com/D-D-H) thinks this a JDK bug, `DeleteLocalRef ` can alleviate this problem, but it should also be fixed on JDK (including JDK 18 and above). 

The crash in #974 on `JDK 8/11/17` is due to lack of [JDK-8268364](https://bugs.openjdk.org/browse/JDK-8268364).

These are two different issues and need to be treated differently.
- #981 Need to fix all JDKs . Use DeleteLocalRef in async-profiler to alleviate.
- #974 Need to backport [JDK-8268364](https://bugs.openjdk.org/browse/JDK-8268364). Add workaround in async-profiler for released JDKs.

Am i right ?

apangin:
Yes, there are two crashes, but they are both parts of the same issue.
JDK-8268364 only mitigates one possible race, but does not solve the problem completely.

src/flightRecorder.cpp
krk: I have created a backport into 17 for JDK-8268364 at https://github.com/openjdk/jdk17u-dev/pull/2839. Do you propose a different fix?

D-D-H: > Do you propose a different fix?

If I understand correctly, the call to 'DeleteLocalRef' here to reduce the possibility of a crash. That crash is actually a bug in JDK and not fixed by JDK-8268364. So we should file a new issue in OpenJDK to fix it.

apangin: @D-D-H You are right: `DeleteLocalRef` is a cheap way to significantly decrease possibility of a crash, but fixing the root cause in OpenJDK is still highly desirable. I think it makes sense to integrate this change anyway, before the proper fix is released.

D-D-H: My colleague Liang created an issue for this problem: https://bugs.openjdk.org/browse/JDK-8339725

apangin: Cool, thanks!

Issue #960, opened 2024-07-18
"Profiler started" doesn't honor loglevel


geertn:
### Describe the bug

In this PR https://github.com/async-profiler/async-profiler/pull/551 a loglevel argument was added with reasoning "This is particularly helpful if you're using async-profiler in production and where you don't want to clubber your output with not always relevant log information"

Using Grafana Profiles we are looking at enabling "Continuous profiling" in production. However when enabling the profiler using Grafana Alloy agent we get a large amount of "Profiling started" log messages in our applications.

Looking at the [source](https://github.com/async-profiler/async-profiler/blob/6207d5dc865d9acdd8bfd8944177775d021f992c/src/profiler.cpp#L1660C17-L1660C28) this and related messages seems to be written out to STDOUT without looking at loglevel.

We don't like to enable profiling in production if this means we can't control logging (and logging format) of our application. 

### Expected vs. actual behavior

"Profiling started" and related logs should honor the log level.

### Reproduction Steps

1. Start Java test program, eg `java -jar 'renaissance-gpl-0.15.0.jar' scrabble`
2. Attach asprof `./asprof -L error <pid>`
3. Observe application stdout:

```
====== scrabble (functional) [default], iteration 5 completed (573.007 ms) ======
====== scrabble (functional) [default], iteration 6 started ======
Profiling started
GC before operation: completed in 58.073 ms, heap usage 336.025 MB -> 78.046 MB.
====== scrabble (functional) [default], iteration 6 completed (448.292 ms) ======
====== scrabble (functional) [default], iteration 7 started ======
```

### Async-profiler version

3.0

apangin:
Thank you for the report.
This is not a log message, but actual profiler output. E.g., this string is returned as a result of `execute` method; users of the profiler Java API may rely on this message.
The fix may not be that trivial. I'll see how to deal with this.

krk:
Resolved in https://github.com/async-profiler/async-profiler/pull/986.

PR #986, opened 2024-09-04
Profiling output respects loglevel.

krk:
### Description

"Profiling started" and other messages are now INFO loglevel and will respect the current loglevel.

### Related issues
https://github.com/async-profiler/async-profiler/issues/960

### Motivation and context
All messages should respect loglevel, "Profiling started" etc. messages did not.

### How has this been tested?
With different `loglevel` parameters:

```
java ... -agentpath: ...,loglevel=info ...
Profiling started
```

```
java ... -agentpath: ...,loglevel=warn ...
```

---

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -62,6 +62,15 @@ void Log::close() {
     }
 }
 
+void Log::writeRaw(LogLevel level, const char* msg, size_t len) {
+    if (level < _level) {
+        return;
+    }
+
+    fwrite(msg, 1, len, _file);
+    fflush(_file);
+}
+
 void Log::log(LogLevel level, const char* msg, va_list args) {
     char buf[1024];
     size_t len = vsnprintf(buf, sizeof(buf), msg, args);
--- a/src/writer.h
+++ b/src/writer.h
@@ -7,6 +7,7 @@
 #define _WRITER_H
 
 #include "asprof.h"
+#include "log.h"
 
 
 class Writer {
@@ -46,6 +47,16 @@ class FileWriter : public Writer {
 
     bool is_open() const {
         return _fd >= 0;
+    }
+
+    virtual void write(const char* data, size_t len);
+};
+
+class LogWriter : public Writer {
+    LogLevel _logLevel;
+
+  public:
+    LogWriter(LogLevel logLevel = LOG_INFO) : _logLevel(logLevel) {
     }
 
     virtual void write(const char* data, size_t len);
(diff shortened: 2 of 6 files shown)


apangin:
I'd argue that Profiler output should *not* respect `loglevel`.
Output and logs are two different streams. The former is where profiling results go to; results must be dumped regardless of the log level.

I think a better approach to the problem is to make "Profiler started" a log message, which it essentially is, with an exception of Java API, where this message is a valid response returned from `execute` method.

apangin:
When testing, please verify all 3 ways to run profiler:
1. as an agent;
2. using `asprof`;
3. via Java API.

krk:
Simplified the approach, introduced a new `LogWriter` class to be used instead of `FileWriter out(STDOUT_FILENO)`.

Tested as an agent, asprof and Java API.

```
❯ asprof start jps
Profiling started

❯ asprof start -L info jps
Profiling started

❯ asprof start -L warn jps
```

src/log.cpp
```diff
@@ -20,6 +20,9 @@ const char* const Log::LEVEL_NAME[] = {
 FILE* Log::_file = stdout;
 LogLevel Log::_level = LOG_INFO;
 
+int Log::fd() {
+    return _file ? fileno(_file) : 0;
```
apangin, line 24: `_file` must not be NULL, I think.

src/writer.h
```diff
@@ -34,9 +35,13 @@ class FileWriter : public Writer {
     int _fd;
     char* _buf;
     size_t _size;
+    bool _keepOpen = false;
```
apangin, line 38: This will issue a compiler warning with older gcc.
We should migrate to a newer C++ standard first, but that's a separate story.

src/writer.h
```diff
@@ -51,6 +56,16 @@ class FileWriter : public Writer {
     virtual void write(const char* data, size_t len);
 };
 
+class LogWriter : public FileWriter {
+    LogLevel _logLevel;
+  public:
+    LogWriter(LogLevel logLevel = LOG_INFO) : FileWriter(Log::fd(), true) {
```
apangin, line 62: This creates an implicit coupling with `Log::_file`, which lives its own life.
Currently, it's not a problem, since `Log` is always closed after the output file, but this dependency is fragile.
Alternatively, `LogWriter` could call `Log::writeRaw` or something.

apangin:
After this change, "Profiling started" message goes to `asprof` output. This is intended, but does not look perfect, given that `asprof` prints its own start/stop messages:
```
Profiling started
Profiling for 5 seconds
Done
```
I'm more are more inclined to believe this message is redundant and can be removed entirely except for the Java API call.

PR #990, opened 2024-09-09
Change flame graph ordering of frames from by-name to by-total

exi:
### Description
Changes frame ordering from by-name to by-total in flame graphs.

### Motivation and context
Usually when someone is using flame graphs, they are looking for performance problems. If there is a large number of frames, it is near impossible to visually make out the most expensive one. This orders them by total time, which is much easier to read.

### How has this been tested?
I habe been using this for some time for profiling/debugging.

---

### DIFF
--- a/src/flameGraph.cpp
+++ b/src/flameGraph.cpp
@@ -204,7 +204,7 @@ void FlameGraph::printFrame(Writer& out, u32 key, const Trie& f, int level, u64 
     for (std::map<u32, Trie>::const_iterator it = f._children.begin(); it != f._children.end(); ++it) {
         children.push_back(Node(it->first, _name_order[f.nameIndex(it->first)], it->second));
     }
-    std::sort(children.begin(), children.end(), Node::orderByName);
+    std::sort(children.begin(), children.end(), Node::orderByTotal);
 
     x += f._self;
     for (size_t i = 0; i < children.size(); i++) {


aalexand:
One disadvantage of grouping by total is that it makes comparing two flame graphs side by side more difficult since minor differences in timing can result in big changes in the shape of the flame graphs.

exi:
@aalexand Ok, I can see your point, which tells me that it should probably be configurable. It's just changing one line based on config.

apangin:
@exi Thank you for your proposal.

Indeed, there are reasons for by-total ordering, but there are reasons for alphabetic ordering as well. Which one is better is subjective. I personally find alphabetic order more convenient: it is stable and easier to read; classes from the same package and related methods come next to each other; frames of the same type are grouped together, logically splitting graph into flames of the same color, whereas a graph may look motley with by-total ordering. When searching for a pattern, alphabetic ordering is also preferable, since highlighted frames are often grouped close to each other rather than being scattered.

The mentioned problems are not relevant for the tree view. Note that the tree visualization already has child frames sorted by the number of samples.

From the technical perspective, the PR is incomplete, as it touches only built-in flame graph generator, but not the flame graph converter.

If you still wish to propose the feature (which I admit is useful in some cases), I suggest adding an option both in the flame graph generator and in the converter. Even better alternative would be to add a flame graph button that switches sort order of the already rendered frames. In this case, users can generate one flame graph and then switch frame order back and forth right in their browser.

exi:
@apangin i like the idea of the in-graph sorting! I'm not sure I'll have time for it in the coming days though since I already fixed the reason why i needed the profiler in the first place.

Would it be ok to report back in a couple of days to see if I can give it a shot?

apangin:
Yes, absolutely!

Artyomcool:
Ping me pls to port it into #944 ) sounds really great)

apangin:
This has been idle for a while, let me close the PR.
Feel free to update and reopen if it's still relevant.

PR #992, opened 2024-09-12
Enable jfr tests

roy-soumadipta:
### Description
* Add JFR to Collapsed conversion implementation.
* Enable disabled JFR tests.

### Related issues
N/A

### Motivation and context
Enable previously disabled JFR test with necessary corrections

### How has this been tested?
Tested on AL2_x86_64, AL2_aarch64 and M1 MacOs Sonoma

---

### DIFF
--- a/test/one/profiler/test/Output.java
+++ b/test/one/profiler/test/Output.java
@@ -71,10 +71,12 @@ public class Output {
         }
     }
 
-    public Output convertJfrToCollapsed(String input, Arguments args) throws IOException {
+    public Output convertJfrToCollapsed(String input, String... args) throws IOException {
         JfrToFlame converter;
         try (JfrReader jfr = new JfrReader(input)) {
-            converter = new JfrToFlame(jfr, args);
+            Arguments arguments = new Arguments(args);
+            arguments.output = "collapsed";
+            converter = new JfrToFlame(jfr, arguments);
             converter.convert();
         }
 
--- a/test/one/profiler/test/Output.java
+++ b/test/one/profiler/test/Output.java
@@ -71,7 +71,7 @@ public class Output {
         }
     }
 
-    public Output convertJfrToCollapsed(String input, String... args) throws IOException {
+    public static Output convertJfrToCollapsed(String input, String... args) throws IOException {
         JfrToFlame converter;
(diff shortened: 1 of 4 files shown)


test/test/cpu/CpuTests.java
```diff
@@ -5,17 +5,22 @@
...
 
 public class CpuTests {
 
-    @Test(mainClass = RegularPeak.class, enabled = false, jvmArgs = "-XX:+UseG1GC -Xmx1g -Xms1g", jvm = Jvm.HOTSPOT)
-    public void regularPeak(TestProcess p) throws Exception {
-        p.profile("-e cpu -d 3 -f %f.jfr");
-        Output out = p.readFile("%f");
-        assert out.contains("java/util/stream/SpinedBuffer\\.accept");
+    @Test(mainClass = RegularPeak.class)
+    public void regularPeakLinux(TestProcess p) throws Exception {
```
apangin, line 16: Redundant `Linux` in the name

test/test/cpu/CpuTests.java
```diff
@@ -5,17 +5,22 @@
...
+        Output out = p.profile("-e cpu -d 6 -f %f.jfr");
+        String jfrOutPath = p.getFile("%f").getAbsolutePath();
+        out = out.convertJfrToCollapsed(jfrOutPath, new Arguments("-o", "collapsed", "--to", "2500"));
```
apangin, line 19: Why do we need to specify `-o collapsed` among arguments, if the method name already suggests we are converting to collapsed? I suggest `convertJfrToCollapsed` creates `Arguments` from the passed array of strings and sets `args.output` to `collapsed` explicitly.
As a bonus, test writers won't need to call `new Arguments()` every time - they'll just pass String varargs:
```
convertJfrToCollapsed(jfrOutPath, "--to", "2500");
```

test/one/profiler/test/Output.java
```diff
@@ -69,6 +71,19 @@ public Output convertFlameToCollapsed() throws IOException {
         }
     }
 
+    public Output convertJfrToCollapsed(String input, Arguments args) throws IOException {
```
apangin, line 74: Can be `static`, since it does not rely on any instance members.

PR #996, opened 2024-09-13
Support tests on JDK > 17

parttimenerd:
### Description
Change the hard-coded use of `-source 7 -target 7` in the Makefile to using the Java version passed via `TARGET_RELEASE` (added for this very purpose in #655).

### Motivation and context
The compilation of some Java files still uses the hard-coded source and target version 7 in the Makefile, leading to issues with Java compilers from JDK 21 and newer, that don't support version 7 any more.

### How has this been tested?

By calling `make test TARGET_RELEASE=21` on JDK 21, it suceeds with this change without Java warnings.

---


Makefile
```diff
@@ -164,7 +164,7 @@ build/$(CONVERTER_JAR): $(CONVERTER_SOURCES) $(RESOURCES)
 	$(RM) -r build/converter
 
 %.class: %.java
-	$(JAVAC) -source 7 -target 7 -Xlint:-options -g:none $^
+	$(JAVAC) -source ${TARGET_RELEASE} -target ${TARGET_RELEASE} -Xlint:-options -g:none $^
```
krk, line 167: Is `TARGET_RELEASE` different than `JAVA_TARGET`?

test/one/profiler/test/TestProcess.java
```diff
@@ -85,6 +85,13 @@ public String[] inputs() {
         return this.inputs;
     }
 
+    private static boolean needsDynamicAgentLoadingFlag() {
+        String version = System.getProperty("java.version");
+        String[] versionParts = version.split("\\.");
+        int majorVersion = Integer.parseInt(versionParts[0]);
+        return majorVersion >= 21;
+    }
```
apangin, line 93: Target JDK is not necessary the same as the host JDK.

test/one/profiler/test/TestProcess.java
```diff
@@ -96,7 +103,10 @@ private List<String> buildCommandLine(Test test, Os currentOs) {
             cmd.add(substituteFiles(String.join(";", sh)));
         } else {
             cmd.add(System.getProperty("java.home") + "/bin/java");
-            cmd.add("-cp");
+            if (needsDynamicAgentLoadingFlag()) {
+                cmd.add("-XX:+EnableDynamicAgentLoading");
+            }
```
apangin, line 108: Why is this option needed? `EnableDynamicAgentLoading` is true even in JDK 24.
For the cases when dynamic agent loading is disabled, I'll implement a different solution in jattach.

src/helper/one/profiler/Server.class
apangin, line 1: What has changed in binary files in this PR?

Makefile
apangin, line 167: I don't see where TARGET_RELEASE is defined.

test/one/profiler/test/TestProcess.java
parttimenerd: Ok, I'll remove it

src/helper/one/profiler/Server.class
parttimenerd: I remove the change

Makefile
parttimenerd: good catch

test/one/profiler/test/TestProcess.java
parttimenerd: oh, you're right

Makefile
parttimenerd: I replaced it `JAVA_TARGET`

apangin:
I'm not sure what problem you are solving, to begin with.
I just cloned a fresh repo and successfully built it with JDK 23.
Helper classes are only recompiled if you change their sources, but if you do so, .class files (being part of the repo) should still be targeted the minimum supported version, which is currently JDK 7. I will probably change it to JDK 8 for the next major release.

parttimenerd:
The problem is running the tests with JDK 21+.

parttimenerd:
You're right, I must have made a different mistake.

PR #998, opened 2024-09-16
Implement builds on each commit using Github Actions

openorclose:
### Description

This action will build, test, and publish binaries for macos, linux x64, linux arm for every commit to the master branch. The binaries will be published to the workflow run.

See example: https://github.com/openorclose/async-profiler/actions/runs/11237144809

### Motivation and context

Allows consuming latest async-profiler binaries more easily

### How has this been tested?
Example workflow run: https://github.com/openorclose/async-profiler/actions/runs/11016088117
---


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/checkout@v4
+      - name: Build and Test
+        run: |
+          make release
+          sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute
+          make test
+      - name: Publish Artifacts
+        run: |
+          # publish mac binaries
+          curl -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" -H "Content-Type: application/zip" --data-binary "@async-profiler-3.0-macos-arm64.zip" "https://uploads.github.com/repos/${REPOSITORY}/releases/${RELEASE_ID}/assets?name=async-profiler-3.0-macos-arm64-${GITHUB_SHA}.zip"
```
apangin, line 40: Can we avoid hardcoding the file name?
Passing `COMMIT_TAG=true` to `make` will add the file suffix automatically.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,79 @@
...
+      # https://github.com/actions/checkout/issues/1590
+      # https://github.com/actions/checkout/issues/1590#issuecomment-2293211398
+
+      # - name: Checkout
+      #  uses: actions/checkout@v3
+
+      # Instead, we have to clone using git manually
+      - name: Build and Test
+        run: |
+          apt-get update && apt-get install -y git curl jq # curl for publishing artifacts, git for checkout out source, jq to parse github api responses
```
apangin, line 58: Since it needs to be executed in a container every time, maybe it's better to include these tools in the container itself?

Makefile
```diff
@@ -120,7 +120,9 @@ $(PACKAGE_NAME).tar.gz: $(PACKAGE_DIR)
 
 $(PACKAGE_NAME).zip: $(PACKAGE_DIR)
 	truncate -cs -`stat -f "%z" build/$(CONVERTER_JAR)` $(PACKAGE_DIR)/$(JFRCONV)
-	codesign -s "Developer ID" -o runtime --timestamp -v $(PACKAGE_DIR)/$(ASPROF) $(PACKAGE_DIR)/$(JFRCONV) $(PACKAGE_DIR)/$(LIB_PROFILER)
+ifneq ($(GITHUB_ACTIONS), true)
+    codesign -s "Developer ID" -o runtime --timestamp -v $(PACKAGE_DIR)/$(ASPROF) $(PACKAGE_DIR)/$(JFRCONV) $(PACKAGE_DIR)/$(LIB_PROFILER)
```
apangin, line 124: Indentation has changed - I guess this does not work if GITHUB_ACTIONS != true.

test/one/profiler/test/TestProcess.java
```diff
@@ -261,7 +261,7 @@ public Output profile(String args, boolean sudo) throws IOException, TimeoutExce
                 .redirectError(createTempFile(PROFERR))
                 .start();
 
-        waitForExit(p, 10);
+        waitForExit(p, 20);
```
apangin, line 264: Why did the waiting time increase?

test/test/smoke/SmokeTests.java
```diff
@@ -35,7 +35,7 @@ public void threads(TestProcess p) throws Exception {
 
     @Test(mainClass = LoadLibrary.class)
     public void loadLibrary(TestProcess p) throws Exception {
-        p.profile("-f %f -o collapsed -d 4 -i 1ms");
+        p.profile("-f %f -o collapsed -d 9 -i 1ms");
```
apangin, line 38: This will extend running time by 5 seconds, which is a lot. I'd prefer not to do that.
What is the problem with the test?

Makefile
openorclose: will change this back to a tab, thanks

test/test/smoke/SmokeTests.java
openorclose: https://github.com/openorclose/async-profiler/actions/runs/10851672132/job/30115952174#step:2:162

```
INFO: Running SmokeTests.loadLibrary...
WARNING: A JVM TI agent has been loaded dynamically (/Users/runner/work/async-profiler/async-profiler/async-profiler/build/bin/../lib/libasyncProfiler.dylib)
WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
WARNING: Dynamic loading of agents will be disallowed by default in a future release
Profiling started
test/smoke/LoadLibrary.main;java/lang/Thread.sleep 1
test/smoke/LoadLibrary.main;java/lang/Thread.sleep 2
thread_start;_pthread_start;thread_native_entry;Thread::call_run;WatcherThread::run;WatcherThread::sleep;Monitor::wait_without_safepoint_check;PlatformMonitor::wait;__psynch_cvwait 3
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/lang/Thread.sleep0;JVM_Sleep;JavaThread::sleep_nanos;PlatformEvent::park_nanos;__gettimeofday 1
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/lang/Thread.sleep0;JVM_Sleep;JavaThread::sleep_nanos;PlatformEvent::park_nanos;__psynch_cvwait 7
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/lang/Thread.sleep0;JavaThread::is_interrupted 1
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/lang/Thread.sleep0;JVM_Sleep;JavaThread::sleep_nanos;PlatformEvent::park_nanos;_pthread_cond_wait;_pthread_mutex_droplock 1
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/util/concurrent/TimeUnit.toNanos 1
test/smoke/LoadLibrary.main;java/lang/Thread.sleep;java/lang/Thread.sleep0;JVM_Sleep;JavaThread::sleep_nanos;pthread_mutex_unlock 1
thread_start;_pthread_start;thread_native_entry;Thread::call_run;ConcurrentGCThread::run;G1ServiceThread::run_service;G1ServiceThread::run_task 1
WARNING: SmokeTests.loadLibrary failed
java.lang.AssertionError
	at test.smoke.SmokeTests.loadLibrary(SmokeTests.java:41)
	at one.profiler.test.Runner.run(Runner.java:152)
	at one.profiler.test.Runner.run(Runner.java:[168](https://github.com/openorclose/async-profiler/actions/runs/10851672132/job/30115952174#step:2:169))
	at one.profiler.test.Runner.main(Runner.java:217)
```

`assert out.contains("Java_sun_management");` This assert fails

test/one/profiler/test/TestProcess.java
openorclose: for the change below to help the test to pass on Mac

.github/workflows/nightly.yml
openorclose: Makes sense

.github/workflows/nightly.yml
openorclose: I'll do the refactor

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  push:
+    branches:
+      - master
+env:
+  REPOSITORY: async-profiler/async-profiler
+  RELEASE_ID: 175290112
+jobs:
+  delete-all-but-last-five-commits:
```
krk, line 11: Why do we need to delete these?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,80 @@
...
+      # Instead, we have to clone using git manually
+      - name: Build and Test
+        run: |
+          apt-get update && apt-get install -y git curl jq # curl for publishing artifacts, git for checkout out source, jq to parse github api responses
+          git clone https://github.com/${REPOSITORY}.git && cd async-profiler && git checkout $GITHUB_SHA # checkout sources
+          make COMMIT_TAG=true CC=/usr/local/musl/x86_64/bin/musl-gcc -j release # build x86 binaries
+          sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute
+          make test
+          make clean
+          make COMMIT_TAG=true ARCH_TAG=arm64 CROSS_COMPILE=aarch64-linux-gnu- CC=/usr/local/musl/aarch64/bin/musl-gcc -j release # build arm64 binaries
```
krk, line 65: Can we use an arm64 instance while building for arm64, instead of cross-compilation? How do we test the arm64 artifact?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,80 @@
...
+      # https://github.com/actions/checkout/issues/1590#issuecomment-2293211398
+
+      # - name: Publish Artifacts
+      #   uses: softprops/action-gh-release@v2
+
+      # Instead, we have to manually call the REST endpoints at https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28
+      - name: Publish Artifacts
+        run: |
+          cd async-profiler
+          for name in $(find . \( -name "async-profiler*.tar.gz" -o -name "async-profiler*.zip" \) -exec basename {} \;) ; do
```
krk, line 77: To make possible errors visible, we need to fail the action if this for loop has 0 iterations, in other words, if the "Publish Artifacts" step publishes nothing.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,80 @@
...
+            # publish binary
+            curl -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" -H "Content-Type: application/zip" --data-binary "@${name}" "https://uploads.github.com/repos/${REPOSITORY}/releases/${RELEASE_ID}/assets?name=${name}"
+          done
+  build-and-publish-linux:
+    runs-on: ubuntu-latest
+    container:
+      image: docker://public.ecr.aws/async-profiler/asprof-builder:latest
+      options: --privileged # needed in order to use sysctl
+    steps:
+      # Unfortunately we cannot use this action to clone the repository due to the actions requiring Node 20 which this container doesn't support
```
krk, line 49: We should build Node 20 in the image, it will then use the libc available and work without requiring a recent libc.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,80 @@
...
+      #  uses: actions/checkout@v3
+
+      # Instead, we have to clone using git manually
+      - name: Build and Test
+        run: |
+          apt-get update && apt-get install -y git curl jq # curl for publishing artifacts, git for checkout out source, jq to parse github api responses
+          git clone https://github.com/${REPOSITORY}.git && cd async-profiler && git checkout $GITHUB_SHA # checkout sources
+          make COMMIT_TAG=true CC=/usr/local/musl/x86_64/bin/musl-gcc -j release # build x86 binaries
+          sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute
+          make test
```
krk, line 63: We can introduce a new Makefile target, `publish` and call it here instead, as the incantation is repeated for mac/x86_64/arm64.

krk (changes requested):

.github/workflows/nightly.yml
openorclose: Not yet, linux arm support is not yet out https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/nightly.yml
openorclose: Andrei wants to keep only the latest 5 commits

.github/workflows/nightly.yml
openorclose: Doesn't look like 18.04 supports Node 20, workarounds like building our own glibc are needed to support it https://github.com/nodesource/distributions/issues/1392

There were other things about trying to symlink node16 to node20, but they all seem like hacks

krk: We can directly build from https://github.com/nodejs/node/tree/v20.17.0, no need to build separate glibc.

.github/workflows/nightly.yml
krk: I can find linux arm64 here: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#specifications-for-general-larger-runners

test/test/smoke/SmokeTests.java
apangin: Thanks. Probably, the following loop runs much longer than 2 seconds on macOS GitHub runner:
```
for (int i = 0; i < 200; i++) {
    Thread.sleep(10);
}
```
If this is the case, the loop needs to be adjusted (e.g., sleeping longer, but less frequently).
Increasing test duration is not the best solution.

.github/workflows/nightly.yml
apangin: Yeah, we don't need to store all artifacts indefinitely. At the same time, it should be clear for a user which artifact is the latest. Currently, it's not obvious at all: there is a bunch of similar files which differ only by a commit hash.
Suggestions are welcome. One idea is to update release notes with the links to the most recent artifacts.

.github/workflows/nightly.yml
apangin: I don't have access to arm64 runners. Let me see if I can request them.

.github/workflows/nightly.yml
apangin: What do you mean? `sysctl kernel.perf_event_paranoid` is not needed on macOS. `musl-gcc` is not relevant either.

.github/workflows/nightly.yml
apangin: Would be nice if that works. Do you think it's enough to preinstall nodejs 20 in a container for the standard actions to work?

.github/workflows/nightly.yml
krk: We could add the day in `yyyy-MM-dd` format to the archive filename, latest should have the word latest in it:

* async-profiler-3.0-2024-10-15-xxxxxxxxx-linux-x64.tar.gz
* async-profiler-3.0-2024-10-16-xxxxxxxxx-linux-x64.tar.gz
* async-profiler-3.0-2024-10-17-xxxxxxxxx-linux-x64.tar.gz
* async-profiler-3.0-2024-10-18-xxxxxxxxx-linux-x64.tar.gz
* async-profiler-3.0-2024-10-19-xxxxxxxxx-linux-x64.tar.gz
* async-profiler-3.0-latest-linux-x64.tar.gz - same as async-profiler-3.0-2024-10-19-xxxxxxxxx-linux-x64.tar.gz

This way, the link to latest will not change by the hash or a date, and it will be clear which day commits belong to, without checking from git history.

.github/workflows/nightly.yml
krk: `sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute` is included in the macos build above in the `build-and-publish-mac` task.

Either way, we could encapsulate the build differences in the Makefile and keep the github actions simple, by calling `make publish` only.

.github/workflows/nightly.yml
openorclose: currently it runs on every commit, so day is not specific enough. But duplicating the latest one and naming it latest makes sense, thanks!

.github/workflows/nightly.yml
openorclose: I'm not sure if this can be configured actually, because the github action fails with

```
/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
```

so it seems like it is using the one provided by Github directly and can't be changed.
But will have to confirm once we actually get node20 running

.github/workflows/nightly.yml
openorclose: I think I can refactor this into the make script

openorclose:
Updated PR, adding support for building and testing arm and publishing to another release containing only 1 binary

Example run: https://github.com/openorclose/async-profiler/actions/runs/10996639420

test/test/smoke/LoadLibrary.java
```diff
@@ -11,8 +11,8 @@
 public class LoadLibrary {
 
     public static void main(String[] args) throws Exception {
-        for (int i = 0; i < 200; i++) {
-            Thread.sleep(10);
+        for (int i = 0; i < 20; i++) {
+            Thread.sleep(100);
```
apangin, line 15: Can you move test changes to a separate PR please?
as they aren't actually related to nightly builds.

test/test/lock/LockTests.java
```diff
@@ -22,15 +23,15 @@ private static void contendedLocks(TestProcess p, int interval, double minRatio)
...
     // 0 is equivalent to disabling sampling of locks, so all profiles are included.
-    @Test(mainClass = LockProfiling.class, inputs = { "0", "70" })
-    @Test(mainClass = LockProfiling.class, inputs = { "10000", "70" })
+    @Test(mainClass = LockProfiling.class, inputs = { "0", "70" }, os = Os.LINUX)
+    @Test(mainClass = LockProfiling.class, inputs = { "10000", "70" }, os = Os.LINUX)
 
     // Large (for the specific paylod) interval value skews the sampled lock
     // contention distribution.
-    @Test(mainClass = LockProfiling.class, inputs = { "1000000", "90" })
+    @Test(mainClass = LockProfiling.class, inputs = { "1000000", "90" }, os = Os.LINUX)
```
apangin, line 31: Lock tests sometimes fail on Linux, too.
Let's disable them for now and ask @krk to amend.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  build-test-publish:
+    strategy:
+      matrix:
+        include:
+          - runson: macos-latest # mac
+
+          - runson: ubuntu-latest # linux x64
+            image: docker://public.ecr.aws/async-profiler/asprof-builder-x86:latest
```
apangin, line 16: Can we handle both x64 and arm64 builds uniformly? I mean, this can also be moved to CodeBuild.

Makefile
```diff
@@ -187,3 +189,7 @@ native:
 
 clean:
 	$(RM) -r build
+
+publish-github-artifact: release
```
apangin, line 193: The logic for accessing GitHub API is now spread across Actions and Makefile, which is not good. This will not work in a standalone mode anyway, unless all the required variables are passed from GHA.
What was the reason for moving it here?

dockerfiles/github-action-nightly-builder-arm/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+ARG musl_src=musl-1.2.5
+ARG musl_sha256=a9a118bbe84d8764da0ea0d28b3ab3fae8477fc7e4085d90102b8596fc7c75e4
+
+ADD https://musl.libc.org/releases/${musl_src}.tar.gz /
+RUN echo ${musl_sha256} ${musl_src}.tar.gz | sha256sum -c
+
+RUN ["/bin/bash", "-c", "\
+    tar xfz ${musl_src}.tar.gz && \
+    cd /${musl_src} && \
```
apangin, line 15: Just wondering, why arm and x64 Dockerfiles are slightly different (e.g., in slashes in the path).

dockerfiles/github-action-nightly-builder-arm/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Image for building async-profiler release packages for arm64 on github actions
+
+FROM public.ecr.aws/codebuild/amazonlinux2-aarch64-standard:3.0
+
+RUN yum update -y && yum install -y libicu tar sudo patchelf make gcc-c++ libstdc++-static java-11-amazon-corretto
```
apangin, line 5: What is `libicu` for?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          # get the last commits_to_keep commit hashes in the format 
+          #    contains($HASH1) or contains($HASH2) or
+          # to insert into jq
+          last_n_hashes=$(git log  --format='contains("%h") or' --abbrev=8 -n${{ matrix.commits_to_keep }} | tr '\n' ' ')
+          echo $last_n_hashes
+          # get the ids of all published artifacts in the nightly release that are not for the latest commits_to_keep hashes
+          ids=$(curl -L  -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -H "X-GitHub-Api-Version: 2022-11-28"  "https://api.github.com/repos/${REPOSITORY}/releases/${{ matrix.release_id }}/assets" | jq ".[]|select(.name | ($last_n_hashes false) | not).id")
+          # delete all assets in the nightly release
+          for id in $ids; do curl -L -X DELETE -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${REPOSITORY}/releases/assets/${id}"; done
```
apangin, line 67: All these delete stuff gets a bit complicated. Would be simpler if we were to delete artifacts basing on their age rather than count? wdyt?

dockerfiles/github-action-nightly-builder-arm/Dockerfile
openorclose: libicu is needed for the github actions setup script needed by codebuild, will add a comment

.github/workflows/nightly.yml
openorclose: Possible, but codebuild does take a while longer to execute

test/test/lock/LockTests.java
openorclose: Will disable the tests in another PR

Makefile
openorclose: I moved it here to make the GHA script simpler, but now that the publishing happens only in one place I should be able to move it back to the script

dockerfiles/github-action-nightly-builder-arm/Dockerfile
openorclose: the x64 one didn't work initially with the commands from the original Dockerfile, I'm not sure why

.github/workflows/nightly.yml
openorclose: https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy

Maybe instead of an archive release we can publish the artifacts to the workflow directly instead? And we can set the retention date to something like 30 days. The benefit is that we don't have to worry about managing them since they will be automatically deleted

apangin: Yeah, I like the idea. This should makes things a lot cleaner.

openorclose: https://github.com/openorclose/async-profiler/actions/runs/11016317799

Implemented here, now people can access the latest binaries by going to the commit and clicking on the workflow run

.github/workflows/nightly.yml
apangin: https://github.com/openorclose/async-profiler/actions/runs/11016650956 suggests the opposite, but maybe it's because of a cached image.

It's worth noting that the original asbrof-builder image based on ubuntu:18 is only 292 MB, while your asprof-builder-x86 is 4.25 GB. Why is it so large and is it possible to shrink it?

dockerfiles/github-action-nightly-builder-arm/Dockerfile
apangin: OK. It would be good to bring both dockerfiles in sync if possible.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,45 @@
...
+  build-test-publish:
+    strategy:
+      matrix:
+        include:
+          - runson: macos-latest # mac
+            archive_postfix: macos-arm64
+
+          - runson: ubuntu-latest # linux x64
+            image: docker://public.ecr.aws/async-profiler/asprof-builder-x86:latest
+            arch: x86_64
```
apangin, line 18: `matrix.arch` is only used in musl-gcc path. If the path is the same for both architectures, the variable is not needed.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Build and Test
+        run: |
+          if [[ "${{ matrix.runson }}" != "macos-latest" ]]; then # configuration needed only for linux
+            sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute
+            CC=/usr/local/musl/${{ matrix.arch }}/bin/musl-gcc
+          fi
+          git config --global --add safe.directory $PWD # needed to avoid `fatal: detected dubious ownership in repository at '/__w/async-profiler/async-profiler'` error
+          make test # build and test
+          make clean # remove test files
+          make # build again without test files
```
apangin, line 39: No need to recompile twice

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,45 @@
...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: async-profiler-nightly-${{ github.sha }}-${{ matrix.archive_postfix }}
+          # Uploading to the workflow automatically zips up the files, so we should upload the build directory instead of an already zipped release.
+          # https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives
+          path: build/
```
apangin, line 45: I don't see where `release` target is called.
Contents of `/build` is not the same as the contents of the release archive.

README.md
```diff
@@ -31,6 +31,10 @@ Current release (3.0):
 async-profiler also comes bundled with IntelliJ IDEA Ultimate 2018.3 and later.  
 For more information refer to [IntelliJ IDEA documentation](https://www.jetbrains.com/help/idea/cpu-and-allocation-profiling-basic-concepts.html).
 
+Nightly releases (built on each commit to master, kept for 90 days):
+- Go to the **Publish Nightly Builds** workflow of the commit by clicking on its green tick and selecting one of the **Publish Nightly Builds / build-test-publish** workflows.
+  - In the Summary tab there will be an Artifacts section at the bottom containing binaries if tests pass.
```
apangin, line 36: As previously discussed, it's a good idea to store build artifacts at the default GHA location. But it still makes sense to publish the latest build artifacts in the `nightly` release. This is what users will consume. An archive with previous versions is mostly for developers.

.github/workflows/nightly.yml
openorclose: have reduced the docker image size to <400MB, and made it as similar as possible to the original one

.github/workflows/nightly.yml
openorclose: fixed this

.github/workflows/nightly.yml
openorclose: am now using `/usr/local/musl` as the path

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,61 @@
...
+      matrix:
+        include:
+          - runson: macos-latest # mac
+            archive: async-profiler-3.0-macos-arm64.zip
+
+          - runson: ubuntu-latest # linux x64
+            image: docker://public.ecr.aws/async-profiler/asprof-builder-x86:latest
+            archive: async-profiler-3.0-linux-x64.tar.gz
+
+          - runson: codebuild-test-${{ github.run_id }}-${{ github.run_attempt }} # linux arm
```
apangin, line 20: Back to CodeBuild vs. GitHub runners: is there a problem with using CodeBuild both for x86 and ARM? If a build takes longer, I'm curious what exactly takes time.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,61 @@
...
+      options: --privileged # needed in order to use sysctl
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build and Test
+        run: |
+          if [[ "${{ matrix.runson }}" != "macos-latest" ]]; then # configuration needed only for linux
+            sudo sysctl kernel.perf_event_paranoid=1 && sudo sysctl kernel.kptr_restrict=0 # allow tests to execute
+            make CC=/usr/local/musl/bin/musl-gcc release test -j
+          else
+            make release test || true # temporarily allow test errors on mac
```
apangin, line 34: If a test fails on mac, please tix the test instead (or disable it temporary and create an issue for the fix).
BTW, what is current behavior if a test fails on one platform but not on another? I'd prefer not to publish new release at all if any of tests fails on any platform.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,61 @@
...
+          fi
+        shell: bash
```
apangin, line 36: What is bash-specific here?

dockerfile-github-actions/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Image for building async-profiler release packages on github actions
+
+# Stage 0: download musl sources and build toolchains
+FROM public.ecr.aws/lts/ubuntu:24.04
```
apangin, line 4: I've chosen 18.04, since there were backward compatibility issues when building on newer distros.
Can you verify that:
1. Result binaries are compatible with GLIBC 2.14 on x86 and with GLIBC 2.17 on ARM64? I.e. there should be no symbols requiring newer GLIBC when running `nm build/lib/libasyncProfiler.so | grep GLIBC_2.[0-9][0-9]`.
2. Result binaries work on Alpine Linux.

.github/workflows/nightly.yml
apangin, line 34: Make sure mac binaries are built with `FAT_BINARY=true` and work both on x86 and ARM.

dockerfile-github-actions/Dockerfile
openorclose: will check, if not will revert back to 18.04 and try building node20 from source

.github/workflows/nightly.yml
openorclose: There's no problem, the extra time is from waiting for codebuild to actually pick up and set up the job, which takes about 2 minutes more vs github runners which pick up the job immediately

.github/workflows/nightly.yml
openorclose: `[[` is bash specific, but I can change that

dockerfile-github-actions/Dockerfile
openorclose: I see some that need newer GLIBC `U pthread_mutexattr_init@GLIBC_2.34`

openorclose:
I had to force push here since I already rebased it previously. Will merge from master in future changes.

Have reverted to 18.04 to use older glibc to compile async-profiler. Because of this, node20 doesn't work (it needs glibc 2.28) but is needed for github actions like checkout and upload. At first I installed a local version of node20 on the docker image, but this doesn't work as github actions can only use the mounted version of node20.

To workaround this, we can use patchelf on a codebuild instance to force node20 to use a separate self-compiled glibc2.28. This has to be done on a codebuild instance, since if we use a github runner the volume node20 is provided in is read-only.

apangin:
As I found yesterday, Debian 10 image has glibc 2.28 which should be sufficient to run Node 20, and it has gcc 8.3 which produces binaries compatible with older glibc, so it might be a solution.

openorclose:
I tried it with the debion 10 image, it works with node20  here is the result of `nm libasyncProfiler.so | grep GLIBC`:

```
nm libasyncProfiler.so | grep GLIBC
                 U __cxa_atexit@@GLIBC_2.17
                 w __cxa_finalize@@GLIBC_2.17
                 U __errno_location@@GLIBC_2.17
                 U __getdelim@@GLIBC_2.17
                 U __libc_current_sigrtmin@@GLIBC_2.17
                 w __pthread_key_create@@GLIBC_2.17
                 U __xstat@@GLIBC_2.17
                 U _setjmp@@GLIBC_2.17
                 U abort@@GLIBC_2.17
                 U calloc@@GLIBC_2.17
                 U clock_gettime@@GLIBC_2.17
                 U close@@GLIBC_2.17
                 U closedir@@GLIBC_2.17
                 U confstr@@GLIBC_2.17
                 U connect@@GLIBC_2.17
                 U dl_iterate_phdr@@GLIBC_2.17
                 U dladdr@@GLIBC_2.17
                 U dlclose@@GLIBC_2.17
                 U dlerror@@GLIBC_2.17
                 U dlsym@@GLIBC_2.17
                 U fclose@@GLIBC_2.17
                 U fcntl@@GLIBC_2.17
                 U fdopen@@GLIBC_2.17
                 U fflush@@GLIBC_2.17
                 U fgets@@GLIBC_2.17
                 U fopen@@GLIBC_2.17
                 U fprintf@@GLIBC_2.17
                 U fputc@@GLIBC_2.17
                 U fputs@@GLIBC_2.17
                 U free@@GLIBC_2.17
                 U freelocale@@GLIBC_2.17
                 U ftruncate@@GLIBC_2.17
                 U fwrite@@GLIBC_2.17
                 U getenv@@GLIBC_2.17
                 U getpid@@GLIBC_2.17
                 U getrlimit@@GLIBC_2.17
                 U gettext@@GLIBC_2.17
                 U gettimeofday@@GLIBC_2.17
                 U ioctl@@GLIBC_2.17
                 U localtime_r@@GLIBC_2.17
                 U longjmp@@GLIBC_2.17
                 U lseek@@GLIBC_2.17
                 U malloc@@GLIBC_2.17
                 U memchr@@GLIBC_2.17
                 U memcmp@@GLIBC_2.17
                 U memcpy@@GLIBC_2.17
                 U memmove@@GLIBC_2.17
                 U memset@@GLIBC_2.17
                 U mktime@@GLIBC_2.17
                 U mmap@@GLIBC_2.17
                 U mprotect@@GLIBC_2.17
                 U munmap@@GLIBC_2.17
                 U nanosleep@@GLIBC_2.17
                 U newlocale@@GLIBC_2.17
                 U open@@GLIBC_2.17
                 U opendir@@GLIBC_2.17
                 U pipe@@GLIBC_2.17
                 U posix_fadvise@@GLIBC_2.17
                 U pthread_cond_init@@GLIBC_2.17
                 U pthread_cond_signal@@GLIBC_2.17
                 U pthread_cond_timedwait@@GLIBC_2.17
                 U pthread_detach@@GLIBC_2.17
                 U pthread_getspecific@@GLIBC_2.17
                 U pthread_join@@GLIBC_2.17
                 U pthread_key_create@@GLIBC_2.17
                 U pthread_kill@@GLIBC_2.17
                 U pthread_mutex_init@@GLIBC_2.17
                 U pthread_mutex_lock@@GLIBC_2.17
                 U pthread_mutex_unlock@@GLIBC_2.17
                 U pthread_mutexattr_init@@GLIBC_2.17
                 U pthread_mutexattr_settype@@GLIBC_2.17
                 w pthread_once@@GLIBC_2.17
                 U pthread_self@@GLIBC_2.17
                 U pthread_setspecific@@GLIBC_2.17
                 U pthread_sigmask@@GLIBC_2.17
                 U pwrite@@GLIBC_2.17
                 U qsort@@GLIBC_2.17
                 U read@@GLIBC_2.17
                 U readdir@@GLIBC_2.17
                 U realloc@@GLIBC_2.17
                 U recvmsg@@GLIBC_2.17
                 U rewinddir@@GLIBC_2.17
                 U sched_getscheduler@@GLIBC_2.17
                 U send@@GLIBC_2.17
                 U sendfile@@GLIBC_2.17
                 U setitimer@@GLIBC_2.17
                 U setrlimit@@GLIBC_2.17
                 U setsockopt@@GLIBC_2.17
                 U sigaction@@GLIBC_2.17
                 U sigaddset@@GLIBC_2.17
                 U sigemptyset@@GLIBC_2.17
                 U snprintf@@GLIBC_2.17
                 U socket@@GLIBC_2.17
                 U sprintf@@GLIBC_2.17
                 U sscanf@@GLIBC_2.17
                 U stderr@@GLIBC_2.17
                 U stdout@@GLIBC_2.17
                 U strcasecmp@@GLIBC_2.17
                 U strchr@@GLIBC_2.17
                 U strcmp@@GLIBC_2.17
                 U strcpy@@GLIBC_2.17
                 U strdup@@GLIBC_2.17
                 U strerror@@GLIBC_2.17
                 U strlen@@GLIBC_2.17
                 U strncmp@@GLIBC_2.17
                 U strncpy@@GLIBC_2.17
                 U strndup@@GLIBC_2.17
                 U strrchr@@GLIBC_2.17
                 U strstr@@GLIBC_2.17
                 U strtod@@GLIBC_2.17
                 U strtok@@GLIBC_2.17
                 U strtol@@GLIBC_2.17
                 U strtoll@@GLIBC_2.17
                 U strtoul@@GLIBC_2.17
                 U strtoull@@GLIBC_2.17
                 U syscall@@GLIBC_2.17
                 U sysconf@@GLIBC_2.17
                 U time@@GLIBC_2.17
                 U timer_create@@GLIBC_2.17
                 U timer_delete@@GLIBC_2.17
                 U times@@GLIBC_2.17
                 U uname@@GLIBC_2.17
                 U unlink@@GLIBC_2.17
                 U uselocale@@GLIBC_2.17
                 U vsnprintf@@GLIBC_2.17
                 U write@@GLIBC_2.17
```

apangin:
@openorclose
> here is the result of nm libasyncProfiler.so | grep GLIBC

That's on ARM, right? 2.17 on ARM is fine. Should me max 2.14 on x86.

openorclose:
I updated the PR as per @apangin's comments to use debian 10.

Reverted back to using github runners for linux x64, as there sometimes is a bit of a delay when waiting for codebuild to pick up the action. See https://github.com/openorclose/async-profiler/actions/runs/11293506198/job/31411783750 and "show timestamps" to compare the start times for the checkout action. Github runners seem to be more consistent at picking up jobs immediately.


Example run: https://github.com/openorclose/async-profiler/actions/runs/11293596691/attempts/1 (macos tests are flakly)

openorclose:
> @openorclose
> 
> > here is the result of nm libasyncProfiler.so | grep GLIBC
> 
> That's on ARM, right? 2.17 on ARM is fine. Should be max 2.14 on x86.

Yeah, this is the max I saw on x86
```U memcpy@@GLIBC_2.14```

File: .github/workflows/nightly.yml [hunk 26]
```diff
@@ -0,0 +1,71 @@
...
+  build-test-publish:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runson: macos-13 # mac x64
+            archive: async-profiler-3.0-macos-x64.zip
+
+          - runson: macos-14 # mac arm
+            archive: async-profiler-3.0-macos-arm64.zip
```
apangin, line 18: Why have you changed it to build two separate binaries?
Previous async-profiler releases included just one muti-arch async-profiler package for macOS - I think this is convenient from UX perspective.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,71 @@
...
+            ;;
+            *) make CC=/usr/local/musl/bin/musl-gcc release test
+            ;;
+          esac
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.archive }}-${{ github.sha }}
+          path: ${{ matrix.archive }}
+      - uses: actions/upload-artifact@v4
+        if: always()
```
apangin, line 45: What's the purpose of `if: always()`?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,71 @@
...
+              throw new Error("Unable to get list of assets from release!");
+            }
+            await Promise.all(assets.data
+              .filter(({name}) => name === "${{ matrix.archive }}")
+              .map(({id}) => github.rest.repos.deleteReleaseAsset({...commonOptions, asset_id: id})));
+            github.rest.repos.uploadReleaseAsset({
+              ...commonOptions,
+              name: "${{ matrix.archive }}",
+              data: await fs.readFile("${{ matrix.archive }}"),
+            });
```
apangin, line 71: Missing new line character at EOF

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          case "${{ matrix.runson }}" in
+            macos*) make release test
+            ;;
+            *) make CC=/usr/local/musl/bin/musl-gcc release test
+            ;;
+          esac
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.archive }}-${{ github.sha }}
```
apangin, line 42: Can we shorten archive suffix? Full-length SHA is too long for a filename.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,71 @@
...
+          distribution: 'corretto'
+          java-version: '11'
+      - uses: actions/checkout@v4
+      - name: Build and Test
+        run: |
+          case "${{ matrix.runson }}" in
+            macos*) make release test
+            ;;
+            *) make CC=/usr/local/musl/bin/musl-gcc release test
+            ;;
```
apangin, line 38: Maybe use `-j` argument to parallelize build?

.github/workflows/nightly.yml
openorclose: it makes the step always run, so if the test fails the logs will still be uploaded

.github/workflows/nightly.yml
openorclose: I can change it back, should I only upload the build artifacts from the arm runner and only run the tests in x86?

openorclose:
example workflow run https://github.com/openorclose/async-profiler/actions/runs/11328480730

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,85 @@
...
+
+          - runson: codebuild-test-x86-${{ github.run_id }}-${{ github.run_attempt }}
+            archive: async-profiler-3.0-linux-x64.tar.gz
+
+          - runson: codebuild-test-${{ github.run_id }}-${{ github.run_attempt }}
+            archive: async-profiler-3.0-linux-arm64.tar.gz
+    runs-on: ${{ matrix.runson }}
+    needs: [build-test] # needed to get the short sha
+    steps:
+      - uses: actions/setup-java@v4
```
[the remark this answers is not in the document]
openorclose: They do, but the current CI uses 11 so I wanted to maintain consistency

openorclose:
Updated to test the macos binaries on both architectures on macos-14

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,104 @@
...
+          echo "SHORT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+  build-test-publish:
+    needs: [short-hash-getter] # needed to get the short hash
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runson: macos-14 # mac arm
+            extraarch: "{for_setup_java: 'x64', for_env: 'X64'}" # we test x64 also using rosetta
+            archive: async-profiler-3.0-macos.zip
```
apangin, line 29: Can we avoid hard-coding version number? It may change frequently in the future.

apangin: On the contrary, parametrizing `extraarch` is not necessary: it will be only x64 and only on macOS.

openorclose:
Have updated the PR, latest workflow run: https://github.com/openorclose/async-profiler/actions/runs/11363662940

PR #1004, opened 2024-09-20
Make profiler version and arch tag configurable

shipilev:
### Description

I am trying to cross-compile async-profiler binaries in my CI. I am doing this on x86_64 host. Unfortunately, `ARCH_TAG` is getting rewritten by the unconditional `uname -a`. Additionally, current build produces the binaries that claim they are `3.0`, but that is misleading.

### Related issues

No related issues?

### Motivation and context

Enough configuration to make cross-compiled builds work.

### How has this been tested?

The default invocation of `make clean release` seems to produce the same result. For cross-compilation testing, I added this patch into my build pipeline, and then set `PROFILER_VERSION=nightly ARCH_TAG=... CROSS_COMPILE=...`. It produces the following binaries: https://builds.shipilev.net/async-profiler/

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
-PROFILER_VERSION=3.0
+ifeq ($(PROFILER_VERSION),)
+  PROFILER_VERSION=3.0
+endif
 
 ifeq ($(COMMIT_TAG),true)
   PROFILER_VERSION := $(PROFILER_VERSION)-$(shell git rev-parse --short=8 HEAD)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,4 +1,6 @@
-PROFILER_VERSION=3.0
+ifeq ($(PROFILER_VERSION),)
```
krk, line 1: We can use `PROFILER_VERSION ?= 3.0`, which will set it to 3.0 only when the variable is unset.

shipilev: Right, we can! Done so.

PR #1008, opened 2024-09-25
#1007: Optimize wall clock profiling

### LINKED ISSUE #1007: Optimize wall clock profiling
### Describe the feature

Wall clock profiling mode (`-e wall`) is currently implemented by a dedicated thread periodically sampling *all* application threads, regardless of their status. This means, if an application has 1000 threads and wall clock profiler runs with 100ms interval, profiler will record 10K samples per second, sending 10K signals and writing about 150kB data every second. This may cause notable performance overhead, both in CPU usage (signals are expensive) and in JFR file size.

In wall clock mode, most of the sampled threads are sleeping, e.g., idle thread pools waiting for incoming tasks. Signals will awake them, causing even more overhead by letting kernel manage threads that would be otherwise intact.

We can significantly reduce impact of wall clock profiling if we skip sampling threads that are known to be idle.

### Use Case

- Lower wall clock profiling overhead for applications with thousands of threads.
- Make smaller wall clock intervals feasible, thus improving profile accuracy.
- Reduce size of JFR recordings when wall clock mode is enabled.

### Proposed Solution

After sampling a thread in IDLE state, record its total CPU usage.
Next time, if CPU usage of this thread has not changed, we will not sample it again, considering the thread stays IDLE at the very same location we've seen it before. We will start sampling this thread again if its CPU time changes or after 1000 skipped iterations, whatever occurs first.

Introduce new JFR event `profiler.WallClockSample` having all the fields of `jdk.ExecutionSample` event plus a new `int samples` field containing a counter for skipped samples. Instead of recording 500 similar `ExecutionSample` events for a sleeping thread, we will issue only one `ExecutionSample` event and one `WallClockSample` event with samples=499.
Amend `JfrReader` to seamlessly handle `WallClockSample` events as if there was a single event per sample as before. Nothing should change from user's perspective; all implementation details should be encapsulated.

If for some reason previous behavior is desired, wall clock optimization can be disabled by adding `nobatch` profiler argument.

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

apangin:
### Description

Optimize wall clock profiling by skipping redundant samples when we know thread CPU time has not changed since the last sample.

### Related issues

Resolves #1007.

### Motivation and context

Make wall clock profiling safe for production use, even for applications with thousands of threads.

### How has this been tested?

Run benchmarks with new wall clock mechanism and legacy wall clock (`nobatch`) and compare flame graphs.
Verify the number of `ExecutionSample` and `WallClockSample` events manually.

---

### DIFF
--- a/src/arch.h
+++ b/src/arch.h
@@ -16,6 +16,10 @@ static inline u64 atomicInc(volatile u64& var, u64 increment = 1) {
     return __sync_fetch_and_add(&var, increment);
 }
 
+static inline int atomicInc(volatile u32& var, int increment = 1) {
+    return __sync_fetch_and_add(&var, increment);
+}
+
 static inline int atomicInc(volatile int& var, int increment = 1) {
     return __sync_fetch_and_add(&var, increment);
 }
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -59,6 +59,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     live             - build allocation profile from live objects only
 //     lock[=DURATION]  - profile contended locks overflowing the DURATION ns bucket (default: 10us)
 //     wall[=NS]        - run wall clock profiling together with CPU profiling
+//     nobatch          - legacy wall clock sampling without batch events
 //     collapsed        - dump collapsed stacks (the format used by FlameGraph script)
 //     flamegraph       - produce Flame Graph in HTML format
 //     tree             - produce call tree in HTML format
@@ -334,6 +335,9 @@ Error Arguments::parse(const char* args) {
 
             CASE("live")
                 _live = true;
+
+            CASE("nobatch")
+                _nobatch = true;
 
             CASE("allkernel")
                 _ring = RING_KERNEL;
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -182,6 +182,7 @@ class Arguments {
     bool _threads;
     bool _sched;
     bool _live;
+    bool _nobatch;
     bool _fdtransfer;
     const char* _fdtransfer_path;
     int _style;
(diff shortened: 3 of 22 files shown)


petrbouda:
Does it make sense to record other events than `STATE_DEFAULT` into jdk.ExecutionSample ?
Screenshot 2024-10-29 at 15 52 19

apangin:
@petrbouda The meaning of `jdk.ExecutionSample` remains unchanged: it may represent both CPU and Wall clock events depending on `state` field as before. `profiler.WallClockSample` is an _additional_ way to represent multiple wall clock samples as one event.

[`JfrReader`](https://github.com/async-profiler/async-profiler/blob/116504c9f75721911b2f561e29eda065c224caf6/src/converter/one/jfr/JfrReader.java) transparently converts both event types to [`ExecutionSample`](https://github.com/async-profiler/async-profiler/blob/116504c9f75721911b2f561e29eda065c224caf6/src/converter/one/jfr/event/ExecutionSample.java).

I agree it would be clearer to leave `jdk.ExecutionSample` only for CPU events and `profiler.WallClockSample` for Wall clock events. I will probably change this in the future.

petrbouda:
Hi Andrei, thank you for your answer. I've already looked into JfrReader to understand your parsing logic. 
Preparing support in [Jeffrey](https://github.com/petrbouda/jeffrey) :) ... this is a cool feature, thanks a lot!

apangin:
It was actually trivial: I modified profiler to always emit different events for CPU vs. Wall clock samples.
Now wall clock engine writes `profiler.WallClockSample` only, and CPU engine writes `jdk.ExecutionSample` with `STATE_DEFAULT`.

Note: if `nobatch` option is specified, both CPU and Wall clock profiler emit `jdk.ExecutionSample` as before.

petrbouda:
Cool! It makes recordings much smaller 🙂

Issue #928, opened 2024-05-09
Why Cannot StackTrace Be Found Using stackTraceId After LiveObject Is Enabled?


243300852:

243300852:
After liveobject is enabled，Occasional occurrence stackTrace can't be found in dictionary

apangin:
Thank you for your interest in async-profiler.
To report a bug, please include the following information:
- OS version, JDK version and async-profiler version;
- command line to start the profiler;
- expected behavior versus actual behavior;
- exact error message if any;
- sequence of actions that leads to the problem;
- an example program or any other information that helps to reproduce the issue.

243300852:
Thanks for the reply，The questions are as follows:
    In the case of disable liveobject ,Line 196 of the code in the figure， modes.LIVEOBJECT = false ，The memory collection report is correct.
    But in the case of enable liveobject  , Line 196 of the code in the figure， modes.LIVEOBJECT = true,  Occasional Error Reported During Memory Collection. When the backend receives the JFR file for parsing, stackTrace cannot be found in the dictionary using stackTraceId. As a result, the parsing fails and the Java backend reports a null pointer error, as shown in the following figure.

243300852:
Thanks for the reply，The questions are as follows:
In the case of disable liveobject ， modes.LIVEOBJECT = false ，The specific command line code is as follows，The memory collection report is correct.
But in the case of enable liveobject ,  modes.LIVEOBJECT = true, Occasional Error Reported During Memory Collection. When the backend receives the JFR file for parsing, stackTrace cannot be found in the dictionary using stackTraceId. As a result, the parsing fails and the Java backend reports a null pointer error, as shown in the following figure.
I tried different JDK versions, such as JDK11 JDK17.0.1 and so on have this problem.
The command line code is as follows:
std::stringstream command;
    command << "start,jfr=7"
            << ",jstackdepth=" << CNP_Constants::MAX_FRAME_COUNT()
            << ",threads=true";
    if (modes.CPU) {
        command << ",event=cpu,interval="
                << (timeIntvl.empty() ? std::to_string(CNP_Constants::DEFAULT_TIME_INTERVAL()) : timeIntvl) << "ms";
    }
    if (modes.MEMORY) {
        command << ",alloc="
                << (dataIntvl.empty() ? std::to_string(CNP_Constants::DEFAULT_DATA_INTERVAL()) : dataIntvl) << "k";
    }
    if (modes.LATENCY) {
[...]

PR #1009, opened 2024-09-25
Reset CallTraceStorage counters before reporting live objects.

krk:
Allows collecting allocation and live object traces at the same time. 
Also fixes "sometimes missing" stack traces in allocations.

### Related issues
https://github.com/async-profiler/async-profiler/issues/928

### Motivation and context
Enabling live object tracing missed stack traces "sometimes".

### How has this been tested?
Unit tests asserting on live object's size with varying GC probability. 

For the stack traces, JFR recording is created and allocation samples are verified to have stack traces in the JFR.

Sample test assertion *without* the patch, that is observed "sometimes":

```
INFO: Running AllocTests.livenessJfr...
WARNING: AllocTests.livenessJfr failed
java.lang.AssertionError: Stack trace missing for id 4939
        at test.alloc.AllocTests.livenessJfr(AllocTests.java:139)
        at one.profiler.test.Runner.run(Runner.java:152)
        at one.profiler.test.Runner.run(Runner.java:168)
        at one.profiler.test.Runner.main(Runner.java:217)
```

---


src/callTraceStorage.cpp
```diff
@@ -289,3 +289,18 @@ void CallTraceStorage::add(u32 call_trace_id, u64 counter) {
...
+void CallTraceStorage::resetCounters() {
+     for (LongHashTable* table = _current_table; table != NULL; table = table->prev()) {
+        u64* keys = table->keys();
+        CallTraceSample* values = table->values();
+        u32 capacity = table->capacity();
+
+        for (u32 slot = 0; slot < capacity; slot++) {
+            if (keys[slot] != 0) {
+                CallTraceSample& s = values[slot];
+                storeRelease(s.counter, 0);
```
apangin, line 302: Don't we need to reset `s.samples` too?
Check that flame graph is generated correctly both with and without `--total` option.

test/test/alloc/AllocTests.java
```diff
@@ -55,4 +61,69 @@ public void humongous(TestProcess p) throws Exception {
...
+    @Test(mainClass = RandomBlockRetainer.class, args = "0.9", inputs = { "true" })
+    @Test(mainClass = RandomBlockRetainer.class, args = "1.0", inputs = { "true" })
+    public void liveness(TestProcess p) throws Exception {
+        final long TOTAL_BYTES = 50000000;
+        final double tolerance = 0.10;
+
+        final boolean live = Boolean.parseBoolean(p.inputs()[0]);
+        final double keepChance = live ? Double.parseDouble(p.test().args()) : 1.0;
+
+        Output out = p.profile("--alloc 1k --total -o collapsed" + (live ? " --live" : ""));
```
apangin, line 87: `--live` is supported on JDK 11+ only. The does should not run or should fail on JDK 8.

src/objectSampler.cpp
```diff
@@ -100,6 +100,10 @@ class LiveRefs {
         jvmtiEnv* jvmti = VM::jvmti();
         Profiler* profiler = Profiler::instance();
 
+        // Set CallTraceStorage counters to zero, so only live objects
+        // will be reported to collapsed and flamegraph outputs.
+        profiler->resetCounters();
```
apangin, line 105: The idea is to reset counters only for non-JFR recording.
In JFR recording, each sample is recorded individually, so accumulated counters are not actually used.

test/one/profiler/test/Assert.java
```diff
@@ -12,6 +12,9 @@ public class Assert {
     private static final Logger log = Logger.getLogger(Assert.class.getName());
 
     public static void isGreater(double value, double threshold) {
+        if (value <= threshold) {
+            throw new AssertionError("Expected " + value + " > " + threshold);
+        }
```
apangin, line 17: Redundant lines after merge?

krk: Good catch, yes!

test/one/profiler/test/Assert.java
```diff
@@ -25,6 +25,12 @@ public static void isGreater(double value, double threshold, String message) {
         }
     }
 
+    public static void isGreaterOrEqual(double value, double threshold) {
+        if (value < threshold) {
+            throw new AssertionError("Expected " + value + " >= " + threshold);
+        }
+    }
```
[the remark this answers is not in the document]
krk: These are used in the current commit, happy to replace them with #1027 when it is merged.

PR #1011, opened 2024-09-26
Obtain can_generate_sampled_object_alloc_events JVMTI capability only when needed

openorclose:
### Description

Only one JVM TI agent can hold can_generate_sampled_object_alloc_events capability at a time. To avoid multiple profiler instances from conflicting with each other, profiler should obtain the capability only when needed and release it as soon as a profiling session finishes.

### Motivation and context

This allows multiple different async profiler instances to use the object sampler on the same java process one after the other. 

### How has this been tested?
Verifying that the test fails without implementation changes: https://github.com/openorclose/async-profiler/actions/runs/11050192717/job/30697437468#step:6:51

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1001,7 +1001,7 @@ Engine* Profiler::selectEngine(const char* event_name) {
 }
 
 Engine* Profiler::selectAllocEngine(long alloc_interval, bool live) {
-    if (VM::addCanSampleObjectsCapability()) {
+    if (VM::addSampleObjectsCapability()) {
         return &object_sampler;
     } else if (VM::isOpenJ9()) {
         return &j9_object_sampler;
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1001,7 +1001,7 @@ Engine* Profiler::selectEngine(const char* event_name) {
 }
 
 Engine* Profiler::selectAllocEngine(long alloc_interval, bool live) {
-    if (VM::addCanSampleObjectsCapability() == 0) {
+    if (VM::addCanSampleObjectsCapability()) {
         return &object_sampler;
     } else if (VM::isOpenJ9()) {
         return &j9_object_sampler;
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1001,7 +1001,7 @@ Engine* Profiler::selectEngine(const char* event_name) {
 }
 
 Engine* Profiler::selectAllocEngine(long alloc_interval, bool live) {
-    if (VM::canSampleObjects()) {
+    if (VM::addCanSampleObjectsCapability() == 0) {
         return &object_sampler;
     } else if (VM::isOpenJ9()) {
         return &j9_object_sampler;
--- a/src/vmEntry.h
+++ b/src/vmEntry.h
@@ -93,8 +93,6 @@ class VM {
     static bool _openj9;
     static bool _zing;
 
-    static const jvmtiCapabilities canSampleObjectsCapability;
-
     static jvmtiError (JNICALL *_orig_RedefineClasses)(jvmtiEnv*, jint, const jvmtiClassDefinition*);
     static jvmtiError (JNICALL *_orig_RetransformClasses)(jvmtiEnv*, jint, const jclass* classes);
 
@@ -145,12 +143,16 @@ class VM {
         return _zing;
     }
 
-    static int addCanSampleObjectsCapability() {
-        return _jvmti->AddCapabilities(&canSampleObjectsCapability);
+    static bool addCanSampleObjectsCapability() {
+        jvmtiCapabilities capabilities = {0};
+        capabilities.can_generate_sampled_object_alloc_events = 1;
+        return _jvmti->AddCapabilities(&capabilities) == 0;
     }
 
-    static int relinquishCanSampleObjectsCapability() {
-        return _jvmti->RelinquishCapabilities(&canSampleObjectsCapability);
+    static void relinquishCanSampleObjectsCapability() {
+        jvmtiCapabilities capabilities = {0};
+        capabilities.can_generate_sampled_object_alloc_events = 1;
+        _jvmti->RelinquishCapabilities(&capabilities);
     }
 
     static void JNICALL VMInit(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread);
--- a/src/vmEntry.h
+++ b/src/vmEntry.h
@@ -143,13 +143,13 @@ class VM {
         return _zing;
     }
 
-    static bool addCanSampleObjectsCapability() {
+    static bool addSampleObjectsCapability() {
         jvmtiCapabilities capabilities = {0};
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -173,6 +173,9 @@ build-test: all build/$(TEST_JAR)
 
 test: all build/$(TEST_JAR)
 	echo "Running tests against $(LIB_PROFILER)"
+	cp -r build build_copy # make a copy to allow us to test how different copies of async profiler interact with one another
+	rm -rf build/build_copy
+	mv build_copy build
```
apangin, line 178: This juggling is needed for one specific test. Makefile should not really care - it's a job of a test.

src/profiler.cpp
```diff
@@ -995,8 +995,11 @@ Engine* Profiler::selectEngine(const char* event_name) {
 
 Engine* Profiler::selectAllocEngine(long alloc_interval, bool live) {
     if (VM::canSampleObjects()) {
-        return &object_sampler;
-    } else if (VM::isOpenJ9()) {
+        if (VM::addCanSampleObjectsCapability() == 0) {
```
apangin, line 998: Does it really need to be under `canSampleObjects`? Can we just try acquire capability and if it returns an error, fall back to a different engine.

test/one/profiler/test/TestProcess.java
```diff
@@ -247,11 +247,24 @@ public Output profile(String args) throws IOException, TimeoutException, Interru
...
     public Output profile(String args, boolean sudo) throws IOException, TimeoutException, InterruptedException {
+        return profile(args, sudo, "build/bin/asprof");
+    }
+
+    public Output profileWithCopy(String args) throws IOException, TimeoutException, InterruptedException {
+        return profileWithCopy(args, false);
+    }
+
+    // this profiles with a different copy of the profiler located in build_copy
+    public Output profileWithCopy(String args, boolean sudo) throws IOException, TimeoutException, InterruptedException {
```
apangin, line 258: It's not necessary to use different `asprof`.
There is `--libpath` option where you can specify an alternate path to `libasyncProfiler.so`.

src/profiler.cpp
openorclose: will remove canSampleObjects

test/test/alloc/AllocTests.java
```diff
@@ -55,4 +59,16 @@ public void humongous(TestProcess p) throws Exception {
...
     }
+
+    @Test(mainClass = MapReaderOpt.class, jvmVer = {11, Integer.MAX_VALUE})
+    public void objectSamplerWtihDifferentAsprofs(TestProcess p) throws Exception {
+        Thread.sleep(1000);
+        Output out = p.profile("-e alloc -d 3 -o collapsed");
+        // _[k] suffix in collapsed output corresponds to jdk.ObjectAllocationOutsideTLAB, which means alloc tracer is being used
+        assert !out.contains("_\\[k\\]"); // we are using alloc tracer instead of object sampler, should definitely not happen on first profiling call
+        Path asprofCopy = Path.of("/tmp/libasyncProfiler." + p.currentOs().getLibExt());
+        Files.copy(Path.of("build/lib/libasyncProfiler." + p.currentOs().getLibExt()), asprofCopy, StandardCopyOption.REPLACE_EXISTING);
```
apangin, line 70: It's better to encapsulate the logic for getting library path in `TestProcess`, especially since TestProcess already has it.

test/test/alloc/AllocTests.java
```diff
@@ -55,4 +59,16 @@ public void humongous(TestProcess p) throws Exception {
...
         assert out.contains("G1CollectedHeap::humongous_obj_allocate");
     }
+
+    @Test(mainClass = MapReaderOpt.class, jvmVer = {11, Integer.MAX_VALUE})
+    public void objectSamplerWtihDifferentAsprofs(TestProcess p) throws Exception {
+        Thread.sleep(1000);
+        Output out = p.profile("-e alloc -d 3 -o collapsed");
+        // _[k] suffix in collapsed output corresponds to jdk.ObjectAllocationOutsideTLAB, which means alloc tracer is being used
+        assert !out.contains("_\\[k\\]"); // we are using alloc tracer instead of object sampler, should definitely not happen on first profiling call
+        Path asprofCopy = Path.of("/tmp/libasyncProfiler." + p.currentOs().getLibExt());
```
apangin, line 69: Temporary file should have a unique name (not to clash with an existing library and to be able to run multiple test processes simultaneously). The file should be deleted on exit.

test/test/alloc/AllocTests.java
```diff
@@ -55,4 +59,16 @@ public void humongous(TestProcess p) throws Exception {
         assert out.contains("java/io/ByteArrayOutputStream.toByteArray;");
         assert out.contains("G1CollectedHeap::humongous_obj_allocate");
     }
+
+    @Test(mainClass = MapReaderOpt.class, jvmVer = {11, Integer.MAX_VALUE})
+    public void objectSamplerWtihDifferentAsprofs(TestProcess p) throws Exception {
+        Thread.sleep(1000);
```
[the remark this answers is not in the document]
openorclose: I've removed the sleep

test/test/alloc/AllocTests.java
```diff
@@ -55,4 +60,16 @@ public void humongous(TestProcess p) throws Exception {
         assert out.contains("java/io/ByteArrayOutputStream.toByteArray;");
         assert out.contains("G1CollectedHeap::humongous_obj_allocate");
     }
+
+    @Test(mainClass = MapReaderOpt.class, jvmVer = {11, Integer.MAX_VALUE})
+    public void objectSamplerWtihDifferentAsprofs(TestProcess p) throws Exception {
+        Output out = p.profile("-e alloc -d 3 -o collapsed");
+        // _[k] suffix in collapsed output corresponds to jdk.ObjectAllocationOutsideTLAB, which means alloc tracer is being used
+        assert !out.contains("_\\[k\\]"); // we are using alloc tracer instead of object sampler, should definitely not happen on first profiling call
+        File asprofCopy = File.createTempFile("libAsyncProfiler", p.currentOs().getLibExt());
```
apangin, line 69: The name is lib**a**syncProfiler.so (small `a`)
Or you can just extract a file name from `p.profilerLibPath()`

src/vmEntry.h
```diff
@@ -144,8 +143,16 @@ class VM {
...
 
-    static bool canSampleObjects() {
-        return _can_sample_objects;
+    static bool addCanSampleObjectsCapability() {
+        jvmtiCapabilities capabilities = {0};
+        capabilities.can_generate_sampled_object_alloc_events = 1;
+        return _jvmti->AddCapabilities(&capabilities) == 0;
+    }
+
+    static void relinquishCanSampleObjectsCapability() {
```
apangin, line 152: A bit too complicated name. I know it matches JVMTI function, but I think `releaseSampleObjectsCapability` would be better. The other function will be correspondigly `addSampleObjectsCapability`.

PR #1015, opened 2024-10-07
Add Copyright header validator during build

roy-soumadipta:
### Description
The commit includes the following changes
* Add a linter logic to check for copyright header in relevant files.
* Add a target to run as the initial target during any build.

### Related issues
N/A

### Motivation and context
We have had instances of files getting merged without copyright headers. This change is to ensure build fails for a developer if a copyright header is missed for any relevant source file.

### How has this been tested?
Tested by adding a file with no copyright header, incorrect copyright header and misplaced copyright header.

---


checkstyle/validator/CopyrightValidator.java
```diff
@@ -0,0 +1,95 @@
...
+                                " * Copyright The async-profiler authors\n" +
+                                " * SPDX-License-Identifier: Apache-2.0\n" +
+                                " */\n";
+
+    private static final String expectedJattachCopyrightHeader = "/*\n" +
+                                " * Copyright The jattach authors\n" +
+                                " * SPDX-License-Identifier: Apache-2.0\n" +
+                                " */\n";
+
+    private static final Set<String> fileExtensionsToCheck = new HashSet<>(Arrays.asList("c", "h", "cpp", "java"));
```
krk, line 35: It might be a good idea to move "which files to validate" concern to the `Makefile` and pass a list of files as args to `CopyrightValidator`. 

This way `CopyrightValidator` will only be responsible for validating the passed file(s), not navigating directories or filtering by extensions. On the other hand, we could include/exclude files to validate from the `Makefile` if need be.

PR #1016, opened 2024-10-08
Add GHA to validate required license headers.

roy-soumadipta:
### Description
* Add a license validation config.
* Add a github workflow file to run copyright header
### Related issues
N/A

### Motivation and context
Add a linter checks during pushes and PR creation

### How has this been tested?
Tested with a file having no header and incorrect header

---


.github/workflows/linters.yml
```diff
@@ -0,0 +1,30 @@
+# Copyright The async-profiler authors.
+# SPDX-License-Identifier: Apache-2.0
```
apangin, line 2: I don't think header is necessary for configs like this.

.licenserc.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - '.gitignore'
+      - '.github/'
+      - 'demo/'
+      - 'target/'
+      - 'src/res/'
+      - '**/*.md'
+      - 'LICENSE'
+      - 'Dockerfile'
+      - 'JavaHome.class'
+      - '.licenserc.yaml'
```
apangin, line 35: Too many things to ignore. The list will definitely change often. Can we use explicit include list instead?

roy-soumadipta: Addressed and tested in dev box. Will reflect in next commit.

.github/workflows/linters.yml
```diff
@@ -0,0 +1,13 @@
...
+  - push
+  - pull_request
+
+jobs:
+  license-headers-linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check license Headers
+        uses: apache/skywalking-eyes/header@v0.6.0
```
openorclose, line 13: Should we pin the commit of this 3rd party action? Looks like it's part of the apache foundation so it may be trustworthy enough.

roy-soumadipta: Yes seems trustworthy and widely used. Will leave it to Andrei's discretion about pinning the commit or not

PR #1021, opened 2024-10-11
Log test args and inputs.

krk:
### Description
While running `make test`, print test args and inputs from annotations if present.

### Motivation and context
Test logs for each test should be unique:

```
INFO: Running XTests.abc
INFO: Running XTests.abc
INFO: Running XTests.abc
```

vs.

```
INFO: Running XTests.abc inputs: [0]...
INFO: Running XTests.abc inputs: [10000]...
INFO: Running XTests.abc inputs: [1000000]...
```

### How has this been tested?
`make test`

---

### DIFF
--- a/test/one/profiler/test/Runner.java
+++ b/test/one/profiler/test/Runner.java
@@ -139,7 +139,10 @@ public class Runner {
                 continue;
             }
 
-            log.log(Level.INFO, "Running " + testName + "...");
+            log.log(Level.INFO, "Running " + testName +
+                (test.args() != null && test.args().length() > 0 ? " args: " + test.args() : "" ) +
+                (test.inputs() != null && test.inputs().length > 0 ? " inputs: [" + String.join(" ", test.inputs()) + "]": "" ) +
+                 "...");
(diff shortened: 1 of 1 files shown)


test/one/profiler/test/Runner.java
```diff
@@ -139,7 +139,10 @@ private static boolean run(Method m) {
                 continue;
             }
 
-            log.log(Level.INFO, "Running " + testName + "...");
+            log.log(Level.INFO, "Running " + testName +
+                (test.args() != null && test.args().length() > 0 ? " args: " + test.args() : "" ) +
+                (test.inputs() != null && test.inputs().length > 0 ? " inputs: [" + String.join(" ", test.inputs()) + "]": "" ) +
```
roy-soumadipta, line 144: Nit: Maybe "with inputs" instead of just "inputs"?

krk: Then we may need `with args` too. Given that this is printed only during running tests, I would keep them brief.

PR #1023, opened 2024-10-11
Next generation test runner

krk:
1. Runner prints PASS/SKIP/FAIL, regardless of loglevel. Test status are output of the test runner program, used `println` instead of `log` in `Runner.java`.
2. Print test duration for each test.
3. Print test number / total number of tests.
4. Names of failed tests and summary at the end.

### Motivation and context
Improve visibility and structure in the test runner output.

### How has this been tested?
`make test`

Before:

```
INFO: Running RecoveryTests.stringBuilder...
INFO: OK
INFO: Running RecoveryTests.numbers...
INFO: OK
INFO: Running RecoveryTests.suppliers...
INFO: OK
```

After:

```
SKIP [27/30] PmuTests.pmuIncompatible took 0.0 ms
INFO: Running JfrTests.regularPeak...
PASS [28/30] JfrTests.regularPeak took 7087.2 ms
INFO: Running JfrTests.parseRecording...
PASS [29/30] JfrTests.parseRecording took 4087.1 ms
INFO: Running JfrTests.parseMultiModeRecording...
Profiling started
PASS [30/30] JfrTests.parseMultiModeRecording took 1586.1 ms

Total test duration: 107708.8 ms
Results Summary:
PASS: 26
SKIP: 4
FAIL: 0
TOTAL: 30
INFO: Test output and profiles are available in build/test/logs directory
```

with failures:

```
INFO: Running LockTests.datagramSocketLock...
FAIL [1/4] LockTests.datagramSocketLock took 4166.7 ms: java.io.IOException: Profiling call failed: [ERROR] Unsupported event type
INFO: Running LockTests.raceToLocks inputs: [0]...
PASS [2/4] LockTests.raceToLocks inputs: [0] took 2756.3 ms
INFO: Running LockTests.raceToLocks inputs: [10000]...
PASS [3/4] LockTests.raceToLocks inputs: [10000] took 2754.6 ms
INFO: Running LockTests.raceToLocks inputs: [1000000]...
PASS [4/4] LockTests.raceToLocks inputs: [1000000] took 2753.9 ms

Failed tests:
LockTests.datagramSocketLock

Total test duration: 12431.4 ms
Results Summary:
PASS: 3
SKIP: 0
FAIL: 1
TOTAL: 4
INFO: Test output and profiles are available in build/test/logs directory
Exception in thread "main" java.lang.RuntimeException: One or more tests failed
at one.profiler.test.Runner.main(Runner.java:261)
```

---


test/one/profiler/test/Runner.java
```diff
@@ -206,21 +233,35 @@ public static void main(String[] args) throws Exception {
...
-            String testName = arg;
-            if (testName.indexOf('.') < 0 && Character.isLowerCase(testName.charAt(0))) {
-                // Convert package name to class name
-                testName = "test." + testName + "." + Character.toUpperCase(testName.charAt(0)) + testName.substring(1) + "Tests";
+        List<RunnableTest> allTests = getRunnableTests(args);
+        final int testCount = allTests.size();
+
+        int i = 1;
+        long totalTestDuration = 0;
+        Map<TestStatus, Integer> statusCounts = new HashMap<>();
```
apangin, line 241: EnumMap

test/one/profiler/test/RunnableTest.java
```diff
@@ -0,0 +1,35 @@
+package one.profiler.test;
```
apangin, line 1: File headers everywhere.

test/one/profiler/test/Runner.java
```diff
@@ -197,6 +199,31 @@ private static void configureSkipTests() {
         }
     }
 
+    private static String formatTestMessage(RunnableTest rt, TestResult result, int i, int testCount, long durationNs) {
+        String duration = String.format("%.1f ms", durationNs / 1e6);
+        String message = result.status().toString() + " [" + i + "/" + testCount + "] " + rt.testInfo() + " took " + duration;
```
apangin, line 204: Can be just a single `String.format` for better readability. Also, calling `toString()` is redundant.

test/one/profiler/test/Runner.java
```diff
@@ -128,45 +127,48 @@ public static boolean enabled(Test test) {
...
         }
-        return passedAll;
+        return rts;
     }
 
-    private static boolean run(Class<?> cls) {
-        boolean passedAll = true;
-        for (Method m : cls.getMethods()) {
-            passedAll &= run(m);
+    private static List<RunnableTest> getRunnableTests(String[] args) throws ClassNotFoundException {
```
apangin, line 161: Tests that do not match current OS or current architecture are also considered runnable - is that intended?

test/one/profiler/test/Runner.java
```diff
@@ -206,21 +233,35 @@ public static void main(String[] args) throws Exception {
...
+            long durationNs = System.nanoTime() - start;
+
+            statusCounts.put(result.status(), statusCounts.getOrDefault(result.status(), 0) + 1);
+            totalTestDuration += durationNs;
+            String message = formatTestMessage(rt, result, i, testCount, durationNs);
+            if (result.status() == TestStatus.FAIL) {
+                failedTests.add(rt.testInfo());
             }
-            passedAll &= run(Class.forName(testName));
+            System.out.println(message);
```
apangin, line 254: Runner only used logging before your change.
I don't mind println, but let's then revise existing messages: which should remain optional logs and which should be always printed.

apangin:
That's a good proposal, btw 👍

test/one/profiler/test/Runner.java
krk: Yes, we collect all tests regardless of filtering here to count the total and report non-matching tests as skipped.

test/one/profiler/test/Runner.java
krk: I’ve kept only `"Running " + rt.testInfo() + "..."` as an info log, while all other outputs are directed to stdout via println.

Are there any specific println statements you think should be made optional?

To further reduce the output, we could revisit this in a separate PR. A possible enhancement would be to introduce a verbosity flag. Runner would captures all stdout/stderr output from tests themselves and print them only when the flag is enabled. By default, the runner would print only "Running...", the test result (PASS/FAIL/SKIP), and a summary at the end of all tests.

test/one/profiler/test/Runner.java
apangin: This means I'll never see 100% tests passing... okay

test/one/profiler/test/Runner.java
apangin: OK, agreed. Let's leave it as is.

test/one/profiler/test/Runner.java
```diff
@@ -197,6 +199,31 @@ private static void configureSkipTests() {
         }
     }
 
+    private static String formatTestMessage(RunnableTest rt, TestResult result, int i, int testCount, long durationNs) {
+        String duration = String.format("%.1f ms", durationNs / 1e6);
+        String message = String.format("%s [%d/%d] %s took %s", result.status(), i, testCount, rt.testInfo(), duration);
```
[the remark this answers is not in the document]
krk: `duration` was used in two places at some point, not anymore. thanks.

test/one/profiler/test/Runner.java
krk: I have added percentage and skip reasons to the summary:

```
Total test duration: 24.4 ms
Results Summary:
PASS: 35
SKIP: 3
        ArchMismatch: 1
        OsMismatch: 2
FAIL: 0
TOTAL: 38 (100.00% PASS)
```

test/one/profiler/test/SkipReason.java
```diff
@@ -0,0 +1,16 @@
...
+
+public enum SkipReason {
+    Disabled,
+    ArchMismatch,
+    OsMismatch,
+    JvmMismatch,
+    JvmVersionMismatch,
+    SkipByClassName,
+    SkipByMethodName,
+}
```
apangin, line 16: Two-level test status is bit of overengineering, IMO. Much code for a small (although useful) feature.
I don't think there is much value in differentiating tests skipped by a class name or a method name (especially since we'll likely unify this with a regex or a glob pattern). Difference between OS/Arch and Jvm/JvmVersion is also subtle. The only difference that probably makes sense is between not applicable tests vs. manually disabled tests. We can have two TestStatuses for them, or we can simply exclude not applicable tests from the total count during the first pass.
Actually, the previous version was already good. I'd just address the problem of 100% pass rate with one of the approaches mentioned above. Wdyt?

test/one/profiler/test/TestResult.java
```diff
@@ -0,0 +1,60 @@
...
+public class TestResult {
+    private final TestStatus status;
+    private final Throwable throwable;
+    private final SkipReason skipReason;
+
+    public TestResult(TestStatus status) {
+        this(status, (Throwable) null);
+    }
+
+    public TestResult(TestStatus status, Throwable throwable) {
```
apangin, line 17: What about leaving just one private constructor? Outside this class, new instances should be created only via static methods: pass, fail, etc.

test/one/profiler/test/SkipReason.java
krk: I made the skip details visible, because in every configuration there would be some skipped tests. While in the latest code this doesn't affect the 100% pass rate calculation, it may raise suspicion to someone reading the run results.

I like the idea of having two types of skip, manual and config mismatch.

krk: Summary would look like:

```
Results Summary:
PASS: 35
SKIP (disabled): 0
SKIP (config mismatch): 3
FAIL: 0
TOTAL: 38 (100.00% PASS)
```

apangin: Yes! Just a minor reordering, maybe?
```
Results Summary:
PASS: 35 (100%)
FAIL: 0
SKIP (disabled): 0
SKIP (config mismatch): 3
TOTAL: 38
```

test/one/profiler/test/Runner.java
```diff
@@ -100,73 +99,87 @@ private static int detectJvmVersion() {
         return Integer.parseInt(prop);
     }
 
-    public static boolean enabled(Test test) {
+    public static TestResult getSkipResult(RunnableTest rt) {
```
apangin, line 102: Will it be better to split the method into two: `boolean enabled()` + `boolean applicable()` or `configMatches()`?

test/one/profiler/test/TestResult.java
```diff
@@ -0,0 +1,44 @@
...
+    private final TestStatus status;
+    private final Throwable throwable;
+
+    private TestResult(TestStatus status, Throwable throwable) {
+        if ((status == TestStatus.FAIL) ^ (throwable != null)) {
+            throw new IllegalArgumentException();
+        }
+
+        this.status = status;
+        this.throwable = null;
```
apangin, line 18: throwable

test/one/profiler/test/Runner.java
```diff
@@ -197,6 +191,25 @@ private static void configureSkipTests() {
...
     }
 
+    private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List<String> failedTests, long totalTestDuration, int testCount) {
+        int fail = statusCounts.getOrDefault(TestStatus.FAIL, 0);
+        if (fail > 0) {
+            System.out.println("\nFailed tests:");
+            failedTests.forEach(System.out::println);
+        }
+
+        Integer pass = statusCounts.getOrDefault(TestStatus.PASS, 0);
```
apangin, line 201: int

test/one/profiler/test/RunnableTest.java
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.profiler.test;
+
+import java.lang.reflect.Method;
+
+public class RunnableTest {
+    private Method m;
+    private Test test;
```
apangin, line 12: final

test/one/profiler/test/TestResult.java
krk: which?

PR #1032, opened 2024-10-18
Refactor nightly builds to delete previous release and create a new release

openorclose:
### Description

Refactor nightly builds to delete previous release and create a new release with new binaries

### Motivation and context

The current implementation always uploads nightly builds to the same nightly release. This nightly release is tagged to a particular commit and remains unchanged, causing confusion to users. Furthermore, the source code files can't be modified for a release, so only the binaries get updated, causing a mismatch between source code and binaries

### How has this been tested?

Example workflow run: 

https://github.com/openorclose/async-profiler/actions/runs/11404731913

---

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,8 +3,7 @@ name: Publish Nightly Builds
 on:
   push:
     branches:
-      # - master
-      - recreate_release
+      - master
 
 env:
   JAVA_DISTRIBUTION: 'corretto'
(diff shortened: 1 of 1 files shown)


apangin:
@openorclose Thank you for updating the release action - it's a big deal!

Several follow-up comments:
 1. Do we still need to compute short `hash` at `build-test` step? We update suffix at `release` step anyway.
 2. The name format should be `async-profiler-3.0-hash-arch.tar.gz` (`hash` should go before `arch`, that is, be a part of the full version number).
 3. Will you also upload .jar artifacts from `build/jar`?

Issue #984, opened 2024-09-03
Add cpp level test


Jianqqiu-TCS:
### Describe the issue

As we can see, async-profiler only provide java test, we cannot see the percentage of test coverage code，is it necessary to add some other test like google test

apangin:
Thank you for your suggestion.
Pull requests are welcome.

Jianqqiu-TCS:
> Thank you for your suggestion. Pull requests are welcome.

Do we have some other method to see the percentage of test code coverage by our java tests, which may be also a good idea

apangin:
Async-profiler project currently does not measure code coverage, but we are open to contributions that will add support for this.

apangin:
A framework for C++ unit tests has been implemented in #1036.
Initial set of unit tests has been added, see `test/native/*`.

PR #1036, opened 2024-10-22
Unit test fw from scratch and PerfEventType::forName tests

krk:
### Description
Implement a C++ unit test runner from scratch and some tests.

### Related issues
https://github.com/async-profiler/async-profiler/issues/984

### Motivation and context
Be able to write unit tests in C++, in addition to the current tests.

### How has this been tested?
```
make test # or make test-cpp to only run cpp tests.
```

Sample test runner output:

```
Running ForName_uretprobe @ test/native/perfEvents_linux_test.cpp:219
PASS [37/38] ForName_uretprobe took 0.0 ms
Running ForName_uretprobe_offset @ test/native/perfEvents_linux_test.cpp:225
PASS [38/38] ForName_uretprobe_offset took 0.0 ms
Total C++ test duration: 0.1 ms
Total successful assertions: 193
PASS: 36 (94.7%)
FAIL: 0
SKIP: 2
TOTAL: 38
```

assertion:

```
Running ForName_raw_pmu_033 @ test/native/perfEvents_linux_test.cpp:243
Assertion failed: (event_type->config == 0x33),
        actual values: event_type->config = 13091 (0x3323), 0x33 = 51 (0x33)
        at test/native/perfEvents_linux_test.cpp:246
FAIL [31/38] ForName_raw_pmu_033 took 0.0 ms
```

Sample coverage report with `make coverage` using `gcov` and `gcovr`, found in `build/test/coverage`:

image

---


apangin:
Thanks for looking into this.

While I see benefits from C++ unit tests in general, I'm not convinced that an example provided is a good justification for the proposed solution. I'd suggest against unit tests like these, as they provide little to none value while doubling the required development efforts.

Arguments parsing is a good candidate for testing with the existing Java test framework. For example, it can include several tests that validate arguments when passed to asprof, agent, or through the Profiler API. Effective values of the argument can be obtained from the JFR [ActiveSetting event](https://github.com/async-profiler/async-profiler/blob/08f29f82ffa2a9a67700fce54df76b9e9bef37bb/src/flightRecorder.cpp#L805) or by verifying the corresponding functionality. This approach will not only cover the same C++ code, but will also test the entire end-to-end workflow, focusing on features, realistic use cases and publicly facing controls.

As to the library itself, I'm a bit concerned about adding 1MB of sources (more than the entire async-profiler codebase) and tens of thousands lines of code for trivial things like an assert-style macro.

Also, some customers that embed async-profiler in their products may not be happy about an extra license. There were such concerns before, so I even had to [rewrite](https://github.com/async-profiler/async-profiler/commit/f7ef0e97b2f2425b1295b75ed0e07448cc6b86ba) a significant chunk of code from scratch to get rid of an extra 3rd party dependency.

krk:
Thank you for the detailed feedback.

To clarify, the test case involving `Arguments` wasn't intended as a justification for the proposed solution - it was meant to demonstrate a simple use case of the proposed framework.

I understand the concerns around adding third-party dependencies, and I’m currently working on a simpler, custom implementation from scratch.

I believe we can separate the discussion around **what to test in C++** from the concept of **enabling unit tests in C++**. 

There are many areas in the codebase where I see value in unit testing, particularly in logic that crosses a certain complexity threshold. For example, classes such as `CallTraceStorage`, `CodeCache`, `Demangle`, `Dictionary`, `Dwarf`, and others could benefit from C++ unit tests without the need to start the JVM or trigger integration tests with asprof or the agent. C++ unit tests can provide value by catching issues earlier and simplifying the debugging process.

That said, the primary goal of this PR is not to add all possible unit tests but to **enable the addition of unit tests in future PRs**. The simple test involving argument parsing is just a demonstration of how the framework can be used with assertion macros. I’m happy to remove the `arguments_test.cpp` file from this PR or at any time.

apangin:
I'm fully with you on the concept. I just suggest not to add dummy tests; let's find something useful to test first, then choose or create a framework that satisfies requirements.

src/callTraceStorage.h
```diff
@@ -40,7 +40,7 @@ struct CallTraceSample {
         return *this;
     }
 
-    bool operator<(const CallTraceSample& other) const {
+    bool operator>(const CallTraceSample& other) const {
         return counter > other.counter;
     }
```
krk, line 45: I noticed this in the new unit test:

```
    left.counter = 11;
    right.counter = 100;
    ASSERT(left < right); // was failing.
```

A smaller counter value doesn't make the `CallTraceSample` smaller. This operator is used only from `profiler.cpp` while dumping traces to the text output format. I flipped the sort order there instead and kept the operator implementation unsurprising.

This was further tested by dumping traces and comparing them before/after the change.

test/native/test_runner_main.cpp
```diff
@@ -0,0 +1,7 @@
+#include "test_runner.hpp"
+
+int main()
+{
+    TestRunner::instance()->runAllTests();
```
krk, line 5: Sample output:

```
Running CallTraceSampleOps
PASS [1/1] CallTraceSampleOps

Total C++ test duration: 0.0 ms
Total successful assertions: 5
PASS: 1 (100.0%)
FAIL: 0
SKIP: 0
TOTAL: 1
```

with an assertion:

```
Running CallTraceSampleOps
Assertion failed: left.samples == 9991
        at test/native/callTraceSample_test.cpp:26
FAIL [1/1] CallTraceSampleOps

Total C++ test duration: 0.0 ms
Total successful assertions: 1
PASS: 0 (0.0%)
FAIL: 1
SKIP: 0
TOTAL: 1
```

krk:
Sample output after the latest changes:

```
Running ForName_uretprobe @ test/native/perfEvents_linux_test.cpp:219
PASS [37/38] ForName_uretprobe took 0.0 ms
Running ForName_uretprobe_offset @ test/native/perfEvents_linux_test.cpp:225
PASS [38/38] ForName_uretprobe_offset took 0.0 ms
Total C++ test duration: 0.1 ms
Total successful assertions: 193
PASS: 36 (94.7%)
FAIL: 0
SKIP: 2
TOTAL: 38
```

assertion:

```
Running ForName_raw_pmu_033 @ test/native/perfEvents_linux_test.cpp:243
Assertion failed: (event_type->config == 0x33),
        actual values: event_type->config = 13091 (0x3323), 0x33 = 51 (0x33)
        at test/native/perfEvents_linux_test.cpp:246
FAIL [31/38] ForName_raw_pmu_033 took 0.0 ms
```

Sample coverage report with `make coverage` using `gcov` and `gcovr`, found in `build/test/coverage`:

image

krk:
Sample test coverage artifacts: https://github.com/krk/async-profiler/actions/runs/11633671239/artifacts/2134141824

test/native/test_runner.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    for (auto& pair : testCases())
+    {
+        auto& testCase = pair.second;
+
+        printf("Running %s @ %s:%d\n", testCase.name.c_str(), testCase.filename.c_str(), testCase.lineno);
+
+        bool forceSkip = hasOnly && !testCase.only;
+        auto start = std::chrono::high_resolution_clock::now();
```
apangin, line 71: Try not to overuse `auto` - types are *very* important for readability.
It's OK to use `auto` for lengthy types such as iterators in a narrow scope.

test/native/test_runner_main.cpp
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "test_runner.hpp"
+
+int main()
```
apangin, line 8: Redundant class, imo.
TestRunner is essentially an entry point, you may just add `main` there.

apangin:
Please update the coding style in alignment with the rest of the project.
It's mostly based on [hotspot-style.md](https://github.com/openjdk/jdk/blob/master/doc/hotspot-style.md).
In particuar:
 - brace style;
 - fileNames;
 - local_variables and _class_fields.

apangin:
Please also update PR description - it still refers to Catch2.

Makefile
```diff
@@ -14,17 +14,20 @@ API_JAR=jar/async-profiler.jar
 CONVERTER_JAR=jar/jfr-converter.jar
 TEST_JAR=test.jar
 
-CC=$(CROSS_COMPILE)gcc
-CXX=$(CROSS_COMPILE)g++
+CC ?= $(CROSS_COMPILE)gcc
+CXX ?= $(CROSS_COMPILE)g++
```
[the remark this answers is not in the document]
krk: This is to also accept environment variables in addition to makefile variables. Use cases is when `CC` and `CXX` is set as env vars, e.g. to a more recent or non-default compiler, they are used without changing the command line.

Makefile
```diff
@@ -14,17 +14,20 @@ API_JAR=jar/async-profiler.jar
...
 STRIP=$(CROSS_COMPILE)strip
 
-CFLAGS=-O3 -fno-exceptions
-CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden
+CFLAGS_EXTRA ?=
+CXXFLAGS_EXTRA ?=
+CFLAGS=-O3 -fno-exceptions -Wformat=2 $(CFLAGS_EXTRA)
```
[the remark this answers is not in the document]
krk: As without `Wformat` locally, all tests pass but CI failed. With it, I could see the same messages locally. We can try setting `Wall` in a new PR.

src/perfEvents_linux.cpp
```diff
@@ -191,7 +191,7 @@ struct PerfEventType {
         return 0;
     }
 
-    // Breakpoint format: func[+offset][/len][:rwx][{arg}]
+    // Breakpoint format: func[+offset][/len][:r|w|x][{arg}]
```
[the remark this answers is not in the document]
krk: That was my expectation as well, I matched the docs to the code. Currently, `rx` and `wx` is parsed as `HW_BREAKPOINT_RW`. If this is a bug, I can revert this comment change and propose to handle the fix in a separate PR.

test/native/perfEvents_linux_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+TEST_CASE(ForName_raw_pmu_u64)
+{
+    PerfEventType* event_type = PerfEventType::forName("rF00FFFFF0123999");
+    ASSERT_RAW(event_type, "rNNN", 0xF00FFFFF0123999);
+}
+
+TEST_CASE(ForName_symbol)
+{
+    PerfEventType* event_type = PerfEventType::forName("strcmp");
+    const __u64 strcmp_addr = (__u64)(uintptr_t)dlsym(RTLD_DEFAULT, "strcmp");
```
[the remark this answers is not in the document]
krk: If we add `-rdynamic` linker option, we can. What would be the intention of this test? Please note that we are not loading libasyncProfiler shared object in the tests.

test/native/test_runner.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <unistd.h>
+#include <fcntl.h>
+
+TestRunner* TestRunner::_instance = NULL;
+
+TestRunner* TestRunner::instance()
+{
+    if (!_instance)
+    {
+        _instance = new TestRunner();
```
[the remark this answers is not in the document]
krk: No specific reason, converting to static.

test/native/perfEventsLinuxTest.cpp
```diff
@@ -0,0 +1,269 @@
...
+  ASSERT_BP(event_type, HW_BREAKPOINT_RW, 22291, 1, 0);
+}
+
+TEST_CASE(ForName_Breakpoint_Addr_HexOffset) {
+  PerfEventType* event_type = PerfEventType::forName("mem:0x123+0x22000");
+  ASSERT_BP(event_type, HW_BREAKPOINT_RW, 0x22123, 1, 0);
+}
+
+TEST_CASE(ForName_Breakpoint_Addr_HexOffset_DecLen) {
+  PerfEventType* event_type = PerfEventType::forName("mem:0x123+0x22000/16");
```
apangin, line 105: To be pedantic, correct breakpoint length can be only 1, 2, 4 or 8.
It's OK for the parser to accept other values though - they will be rejected later by perf_event_open API. A test should verify either 100% correct value or 100% incorrect value; for other values, behavior is implementation defined and therefore should not occur in assertions.

test/native/perfEventsLinuxTest.cpp
```diff
@@ -0,0 +1,269 @@
...
+  const __u64 mmap_addr = (__u64)(uintptr_t)dlsym(RTLD_DEFAULT, "mmap");
+  PerfEventType* event_type = PerfEventType::forName("mem:mmap+0x22000/0x16:x");
+
+  ASSERT_NE(mmap_addr, 0);
+  ASSERT_BP(event_type, HW_BREAKPOINT_X, mmap_addr + 0x22000, 0x16, 2);
+}
+
+TEST_CASE(ForName_Breakpoint_Symbol) {
+  PerfEventType* event_type = PerfEventType::forName("mem:strcmp");
+  const __u64 strcmp_addr = (__u64)(uintptr_t)dlsym(RTLD_DEFAULT, "strcmp");
```
apangin, line 164: Just curious: is it intentional that the order of forName/dlsym differs in the above tests?

test/native/perfEventsLinuxTest.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+#ifdef __linux__
+
+#include "callTraceStorage.h"
+#include "perfEvents.h"
+#include "testRunner.hpp"
+#include <linux/perf_event.h>
+#include <stdio.h>
+
+#define ASSERT_EVENT_TYPE(event_type, name_, type_, default_interval_, config_, config1_, config2_, counter_arg_) \
```
apangin, line 14: I see you've changed indentation to 2 spaces. Async-profiler uses 4 spaces - this is one of the differences from HotSpot coding style; sorry I haven't mentioned that in advance.

.github/workflows/nightly.yml
```diff
@@ -32,15 +32,23 @@ jobs:
           distribution: ${{ env.JAVA_DISTRIBUTION }}
           java-version: ${{ env.JAVA_VERSION }}
           # architecture: not specifying this defaults to architecture of the runner
+      - name: Install gcovr
```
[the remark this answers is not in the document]
krk: Where do we build these images?

test/native/perfEventsLinuxTest.cpp
krk: It is not, changed to be consistent.

test/native/perfEvents_linux_test.cpp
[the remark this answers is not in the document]
apangin: No-no, the point is to test breakpoints on private symbols, i.e. ensure that profiler resolves symbols that are _not_ in the dynamic symbol table.
I think you can just take an address of some function (`&func`) in your own code.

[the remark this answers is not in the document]
krk: As the profiler is not started in these tests, symbols are not parsed via `dl_iterate_phdr` and `Profiler::_native_libs` is empty, causing early exit in `Profiler::resolveSymbol`.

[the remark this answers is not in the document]
apangin: Call `Profiler::updateSymbols()`

.github/workflows/nightly.yml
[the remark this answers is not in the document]
krk: This is resolved, updated images for linux and do a brew install for macos.

src/perfEvents_linux.h
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _PERFEVENTS_LINUX_HPP
```
apangin, line 6: I assume this was moved to a separate file to make it possible to call PerfEventType functions from a test, right?
I'm afraid this approach does not scale well. Ideally, unit tests should be completely non-intrusive, i.e., do not require refactoring. Is it possible to avoid exposing implementation details through interfaces solely for the purpose of testing? Perhaps, by merging sources.

Dockerfile
```diff
@@ -13,20 +33,37 @@ ARG musl_sha256=a9a118bbe84d8764da0ea0d28b3ab3fae8477fc7e4085d90102b8596fc7c75e4
...
     cd /${musl_src} && \
     ./configure --disable-shared --prefix=/usr/local/musl && \
     make -j`nproc` && make install && make clean && \
     ln -s /usr/include/$(arch)-linux-gnu/asm /usr/include/{asm-generic,linux} /usr/local/musl/include/"]
 
-# Stage 1: install build tools + copy musl toolchain from the previous step
+# Build a recent python for gcovr.
+ENV PATH="$PATH:/opt/python3/bin"
+WORKDIR /tmp
+RUN curl -fsSL https://www.python.org/ftp/python/3.13.0/Python-3.13.0.tgz | tar xzf - && cd Python-3.13.0 && \
```
apangin, line 47: Why building python from scratch? Cannot we just apt install gcovr?

Dockerfile
```diff
@@ -13,20 +33,37 @@ ARG musl_sha256=a9a118bbe84d8764da0ea0d28b3ab3fae8477fc7e4085d90102b8596fc7c75e4
...
+# Stage 1: install build tools + copy musl toolchain and python3 from the previous step
 FROM public.ecr.aws/debian/debian:10-slim
 
-# The following command should be exactly the same as at stage 0 to benefit from caching.
-# libicu-dev is needed for the github actions runner
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    sudo libicu-dev patchelf curl make g++ openjdk-11-jdk-headless && \
+        curl \
+        g++ \
+        g++-aarch64-linux-gnu \
```
apangin, line 58: Why is cross-toolchain needed?

Dockerfile
krk: Recent versions of gcovr requires a recent python. apt install gets gcovr from 2013, there is was a release last month.

Dockerfile
krk: Good catch, this was from a merge conflict, removing.

src/perfEvents_linux.h
apangin: Don't build fat binary for unit tests then? Alternatively, merge sources to a temporary file rather than pipe or use some other build-time trick.

Extracting headers is undesirable, because it's more code, additional manual work, and exhibiting of implementation details that could otherwise stay local to the place of use.

Makefile
```diff
@@ -172,12 +178,39 @@ build/$(CONVERTER_JAR): $(CONVERTER_SOURCES) $(RESOURCES)
...
+test-cpp: build-test-cpp
+	echo "Running cpp tests..."
+	build/test/cpptests
+
+test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
 	$(JAVA) $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(TESTS)
 
+coverage: override FAT_BINARY=false
+coverage: clean-coverage
```
apangin, line 206: Add all non-file targets to `.PHONY`

test/native/testRunner.cpp
```diff
@@ -0,0 +1,93 @@
...
+int TestRunner::runAllTests() {
+    int passed = 0;
+    int failed = 0;
+    int skipped = 0;
+    int total_assertions = 0;
+    int i = 1;
+    double total_duration = 0;
+    const int total_tests = testCases().size();
+
+    const bool redirected = !isatty(fileno(stdout));
```
apangin, line 42: `STDOUT_FILENO`

test/native/testRunner.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+        printf("Running %s @ %s:%d\n", test_case.name.c_str(), test_case.filename.c_str(), test_case.line_no);
+
+        std::chrono::high_resolution_clock::time_point start = std::chrono::high_resolution_clock::now();
+        if (!force_skip) {
+            test_case.test_function();
+        } else {
+            test_case.skipped = true;
+        }
+        std::chrono::duration<double, std::milli> duration = std::chrono::high_resolution_clock::now() - start;
```
apangin, line 68: IIUC, `std::chrono::high_resolution_clock` is not required to be monotonic and therefore is not suitable for measuring operation duration.
`OS::nanotime()` available in the profiler itself would be a simpler and better alternative.

PR #1039, opened 2024-10-23
Initialize rem_* in DwarfParser.parseInstructions.

krk:
### Description
`rem_*` variables *may* not be initialized in all code paths.

### How has this been tested?
```
make test
```

---

### DIFF
--- a/src/dwarf.cpp
+++ b/src/dwarf.cpp
@@ -138,10 +138,10 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
     int fp_off = DW_SAME_FP;
     int pc_off = -EMPTY_FRAME_SIZE;
 
-    u32 rem_cfa_reg;
-    int rem_cfa_off;
-    int rem_fp_off;
-    int rem_pc_off;
+    u32 rem_cfa_reg = 0;
+    int rem_cfa_off = 0;
+    int rem_fp_off = 0;
+    int rem_pc_off = 0;
 
     while (_ptr < end) {
         u8 op = get8();
--- a/src/dwarf.cpp
+++ b/src/dwarf.cpp
@@ -138,10 +138,10 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
     int fp_off = DW_SAME_FP;
     int pc_off = -EMPTY_FRAME_SIZE;
 
-    u32 rem_cfa_reg = 0;
-    int rem_cfa_off = 0;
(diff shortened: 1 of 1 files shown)


src/dwarf.cpp
```diff
@@ -138,10 +138,10 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
...
     int pc_off = -EMPTY_FRAME_SIZE;
 
-    u32 rem_cfa_reg;
-    int rem_cfa_off;
-    int rem_fp_off;
-    int rem_pc_off;
+    u32 rem_cfa_reg = 0;
+    int rem_cfa_off = 0;
+    int rem_fp_off = 0;
+    int rem_pc_off = 0;
```
apangin, line 144: What problem does this fix?
`rem` values are valid only after they have been assigned to by `DW_CFA_remember_state` instruction. Otherwise, `0` is just as much garbage as any other random value.

krk: This is from a static analyzer finding. Hypothetically, there may be a corrupt dwarf file, causing uninitialized values to be used, which would add an invalid frame description.

apangin: I understand. My point is it's not a fix anyway. I.e., when using 0 instead of uninitialized values, it will be just a similar invalid frame descriptor.

apangin: Initializing with an [empty frame descriptor](https://github.com/async-profiler/async-profiler/blob/08f29f82ffa2a9a67700fce54df76b9e9bef37bb/src/dwarf.cpp#L136-L139) would be _slightly_ better, but does not change the situation fundamentally. I mean, bogus dwarf will result in an invalid or truncated stack trace anyway.

krk: I was able to make DwarfParser use uninitialized variables with this `eh_frame_hdr`, which could lead to undefined behavior:

```
00000000: 0103 0333 6100 0000 0100 0000 0100 0000  ...3a...........
00000010: 3000 0000 0000 0000 0000 0000 0000 0000  0...............
00000020: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000030: 1400 0000 0400 0000 0900 0000 0400 0000  ................
00000040: 000b 0000 00                             .....
```

```
DwarfParser p("name", 0, eh_frame_hdr);
```

I haven't investigated if invalid or malicious values could lead to a crash.

Updated the code to use the empty frame descriptor.

Issue #833, opened 2023-11-02
Add time-to-safepoint JFR event


apangin:
### Background

One of async-profiler features is time-to-safepoint profiling (`--ttsp` option) which enables collection of CPU or wall-clock events only in the time window between a safepoint has been requested and all Java threads have been stopped. This helps to diagnose the cause of long time-to-safepoint pauses.

`--ttsp` option is the most useful with JFR output format, since it preserves all samples with timestamps in a way that they can be matched with the corresponding pauses. However, to trace these pauses, user needs to enable safepoint logging and then manually match JFR samples with the VM logs.

It would be helpful to emit an additional JFR event for each time-to-safepoint pause to make it possible to analyze pauses using only the profiler recording independently of the VM logs.

### Proposal

Introduce new event in async-profiler recordings for time-to-safepoint pauses. In fact, this event can be generalized for any combination of `--begin/--end` triggers, not just for safepoints.

### Implementation

Add `profiler.Window` JFR event with the following fields:
 - startTime
 - duration
 - eventThread

It will be emitted for every time window between `--begin` and `--end` traps. In particular, for `--ttsp` feature, each `profiler.Window` event will correspond to the time interval between the execution of `SafepointSynchronize::begin` and `RuntimeService::record_safepoint_synchronized` functions - in other words, each event will denote a corresponding time-to-safepoint pause.

apangin:
In this example, we see that a `Profiling Window` event with the duration of 32 ms is followed by 4 `Method Profiling Sample` events all with the same stack trace ending with `MonitorList::unlink_deflated`.

<img width="512" alt="jfr" src="https://github.com/async-profiler/async-profiler/assets/1749416/3fb5756e-44dd-4aca-b9a8-cb66e55da57e">

The above information suggests that the long time-to-safepoint pause (32 ms) was caused by the VM code doing monitor deflation. This already helped us to diagnose [JDK-8319048](https://bugs.openjdk.org/browse/JDK-8319048).

richardstartin:
[...]

PR #1046, opened 2024-11-05
Add --nostop to always start profiler recording even when --ttsp or --start/--end are used

openorclose:
### Description
- Added `--nostop` to always start profiler recording even when `--ttsp` or `--start`/`--end` are used

### Related issues

https://github.com/async-profiler/async-profiler/issues/833

### Motivation and context

In particular, this allows us to continue recording events in --ttsp mode

### How has this been tested?

Added a test for the original ttsp mode and also the newly added flag.

---


test/test/jfr/JfrTests.java
```diff
@@ -74,4 +75,80 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
...
+     */
+    @Test(mainClass = Ttsp.class)
+    public void timebetween(TestProcess p) throws Exception {
+        p.profile("-d 5 --timebetween SafepointSynchronize::begin RuntimeService::record_safepoint_synchronized -f %f.jfr");
+        if (!containsSamplesOutsideWindow(p)) {
+            throw new RuntimeException("Expected to find samples outside of ttsp window, but did not find any");
+        }
+    }
+
+    private boolean containsSamplesOutsideWindow(TestProcess p) throws Exception {
```
apangin, line 121: Can be shortened and simplified a bit.
1. `events` list is redundant; we may collect profiler windows and execution samples in the first pass.
2. Profiler windows can be stored in a `TreeMap` as `startTime`->`endTime`. This will provide sorting and lookup for the closest window out of the box.
3. Search loop can be simplified with `stream().anyMatch()`.

test/test/jfr/Ttsp.java
```diff
@@ -0,0 +1,44 @@
...
+package test.jfr;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.*;
+import java.io.BufferedOutputStream;
+import java.io.FileOutputStream;
+
+class Ttsp {
```
apangin, line 15: This example needs some comments. How does it expose long TTSP issue? Why getAllStackTraces() is called in a loop? Writing 1GB files in a loop may not be a good thing to do in a test - IO is not what we are testing here. Furthermore, not all systems may even have a spare gigabyte in /tmp volume.

apangin:
Thank you for the PR.
The feature is indeed useful, but I think it can be implemented in a simpler way - added some comments inline.

test/test/jfr/Ttsp.java
openorclose: Thank you, I adapted this example from an online article regarding debugging ttsp issues, will rewrite with explanation

test/test/jfr/JfrTests.java
openorclose: thanks!

src/main/main.cpp
```diff
@@ -77,7 +77,9 @@ static const char USAGE_STRING[] =
     "  --clock source    clock source for JFR timestamps: tsc|monotonic\n"
     "  --begin function  begin profiling when function is executed\n"
     "  --end function    end profiling when function is executed\n"
-    "  --ttsp            time-to-safepoint profiling\n"
+    "  --timebetween a b record when a is executed, and then when b is executed, and duration between\n"
+    "  --onlyttsp        only time-to-safepoint profiling \n"
+    "  --includettsp     profile as per normal, but include time-to-safepoint information\n"
```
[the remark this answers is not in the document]
openorclose: Will use nostop for now, anyone else feel free to suggest others

test/test/jfr/Ttsp.java
```diff
@@ -0,0 +1,41 @@
...
+package test.jfr;
+
+import java.util.Random;
+
+class Ttsp {
+
+    static private byte loop() {
+        byte[] byteArray = new byte[1024 * 1024 * 1024];
+        for (int i = 0; i < 10000; i++) {
+            new Random().nextBytes(byteArray);
```
apangin, line 15: `Random.nextBytes` contains a loop with a safepoint check - it does not cause long time-to-safepoint pauses.

test/test/jfr/Ttsp.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+        return byteArray[0];
+    }
+
+    public static void main(String[] args) throws Exception {
+        new Thread(() -> {
+            while (true) {
+                System.gc();
```
apangin, line 23: `ThreadMXBean.dumpAllThreads` is a much cheaper way to force global safepoint.

test/test/jfr/Ttsp.java
```diff
@@ -0,0 +1,41 @@
...
+                try {
+                    Thread.sleep(20);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }).start();
+
+        Thread.sleep(1000);
```
apangin, line 32: Relying on `Thread.sleep` in tests in error-prone.

test/test/jfr/Ttsp.java
```diff
@@ -0,0 +1,41 @@
...
+
+        new Thread(() -> {
+            while (true) {
+                loop();
+            }
```
apangin, line 37: A loop in this example is not very useful. The only long ttsp pause happens the very first time when allocating `new byte[1024 * 1024 * 1024]` array; there are no ttsp pauses afterwards.

apangin:
Changes look good, except for the test, which does not really demonstrate time-to-safepoint pauses.
I'll merge this PR and push a revised test afterwards.

PR #1049, opened 2024-11-08
Increase setup_lib_path buf size to PATH_MAX

krk:
### Description
Buffer smaller than `PATH_MAX` would cause silent truncation when path length is larger than the buffer size.

### Motivation and context
To eliminate possible silent filename truncation, which may cause an error or surprising behavior.

### How has this been tested?
```
make test
asprof start and check libpath value.
```

---

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -6,6 +6,7 @@
 #include <alloca.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <signal.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


src/main/main.cpp
```diff
@@ -16,8 +16,13 @@
 #include <unistd.h>
 #include "fdtransferServer.h"
 
+#ifdef __linux__
+#include <linux/limits.h>
```
apangin, line 20: There is a platform independent version: `<limits.h>`

PR #1051, opened 2024-11-08
Initialize CodeCache to NULL.

krk:
### Motivation and context
Prevents an unlikely uninitialized access.

### How has this been tested?
```
make test
```

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -686,7 +686,7 @@ static int parseLibrariesCallback(struct dl_phdr_info* info, size_t size, void* 
     }
 
     CodeCacheArray* array = (CodeCacheArray*)data;
-    CodeCache* cc;
+    CodeCache* cc = NULL;
     const char* image_base = NULL;
     u64 last_inode = 0;
     u64 cc_inode = 0;


apangin:
> Prevents an unlikely uninitialized access

There is no uninitialized access in the given code. Even if there was, initialization to NULL would not be the right fix: at best, it would make the crash more deterministic, but it would still be a crash.

If this is an attempt to suppress a false positive finding of a static analyzer - this is fine: I don't mind trivial changes to satify static analyzer, if they have no impact on readability, performance or anything else, but let's be explicit in this case.

krk:
This initialization addresses the static analyzer finding: "variable 'cc' may be uninitialized when used here." While there’s no actual uninitialized access in the current code, setting `cc` to `NULL` serves as a defensive measure. This ensures a deterministic crash if `cc` were ever accessed before initialization, preventing silent corruption of `CodeCache` that could otherwise lead to incorrect results or crashes elsewhere or here.

Although a crash from uninitialized access isn’t currently possible here, initializing to `NULL` not only satisfies the static analyzer but also protects against potential future code changes that might unintentionally introduce uninitialized access to `cc`.

PR #1064, opened 2024-11-26
nativemem - Native memory profiler

krk:
Native memory allocations profiling and memory leak reporting, via GOT patching.

There is a new argument in the profiler `nativemem`, and `jfrconv` gains native memory tracking and memory leak reporting.

## Usage

New argument `--nativemem [N]`/`nativemem[=N]` is used to enable the `MallocTracer` engine.
It can profile `malloc`, `calloc`, `realloc` and `free` invocations that are dynamically linked to the target.
When `N` is specified, only one allocation every `N` bytes is sampled.

### Profiling with `nativemem`

```
java -agentpath:"$ASPROF_LIB/libasyncProfiler.so=start,nativemem,file=/dev/stderr" <app>
```

A flame graph can be generated directly as the profiler output option, which includes all sampled malloc calls. Alternatively, use `jfrconv` to process a jfr file and chart only the memory leaks:

### Memory leak analysis via `jfrconv`

```
java -agentpath:"$ASPROF_LIB/libasyncProfiler.so=start,event=cpu,nativemem,file=malloc.jfr" <app>

jfrconv --total --nativemem --leak malloc.jfr malloc-leak.html

# without leak analysis, includes all collected samples:
jfrconv --total --nativemem malloc.jfr malloc.html
```

Generated flame graphs unit is in bytes if `total` is specified, otherwise in number of invocations.

### asprof

If the profiler is attached via `asprof`, `nativemem` profiling can be started and stopped multiple times for the same process as necessary.

```
asprof start -e nativemem --total <app>
asprof stop <app>
...
```

## How has this been tested?

* New tests added that call `malloc` functions with and without multi-threading.

### All allocations

From a test application.

image

### Memory leaks only

Profile of the same application, processed via `jfrconv`.

image

---


test/one/profiler/test/Runner.java
```diff
@@ -221,14 +222,21 @@ public static void main(String[] args) throws Exception {
 
         List<RunnableTest> allTests = getRunnableTests(args);
         final int testCount = allTests.size();
+        boolean hasOnly = false;
```
apangin, line 225: It's not very convenient to use and not very nice design-wise. I initially thought that in order to run selected tests, one would specify an argument with test names. This should already work: `make test TESTS=nativemem`. This mechanism can be further extended to allow execution of individual tests, not just test sets.

test/one/profiler/test/Test.java
```diff
@@ -26,6 +26,8 @@
 
     String jvmArgs() default "";
 
+    String env() default "";
```
apangin, line 29: Will `String[]` be better? (it's always hard to recall which separator to use)

test/test/nativemem/CallsRealloc.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static native long nativeMalloc(int size);
+
+    public static native long nativeRealloc(long addr, int size);
+
+    public static void main(String[] args) throws InterruptedException {
+        final boolean once = args.length > 0 && args[0].equals("once");
+
+        final int malloc_size = 1999993; // unlikely, prime size.
+        final int realloc_size = 30000170;// unlikely, prime size.
```
apangin, line 21: In Java it's `mallocSize` :)

test/test/nativemem/doesmalloc.c
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <jni.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+JNIEXPORT jlong JNICALL Java_test_nativemem_CallsMallocCalloc_nativeMalloc(JNIEnv* env, jclass clazz, jlong size) {
```
apangin, line 10: `malloc`, `free` and `realloc` are exposed via `sun.misc.Unsafe`.
If not `calloc`, the test could be implemented fully in Java.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+typedef void* (*realloc_t)(void*, size_t);
+static realloc_t _orig_realloc = NULL;
+
+typedef void (*free_t)(void*);
+static free_t _orig_free = NULL;
+
+static void* malloc_hook(size_t size) {
+    void* ret = _orig_calloc(1, size);
+    if (likely(ret && size)) {
+        MallocTracer::recordMalloc(ret, size, OS::nanotime());
```
apangin, line 35: Why do you pass `OS::nanotime()` as argument every time?
`recordMalloc` can take a timestamp. A delay should be negligible.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+    return ret;
+}
+
+#ifdef __linux__
```
apangin, line 40: Will macOS build break if you don't surround it with ifdefs?
Ultimately we want to support `nativemem` on macOS too.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+extern "C" WEAK DLLEXPORT void* malloc(size_t size) {
+    if (unlikely(!_orig_malloc)) {
+        return NULL;
```
apangin, line 43: When can this happen?

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+        MallocTracer::recordMalloc(ret, num * size, OS::nanotime());
+    }
+    return ret;
+}
+
+#ifdef __linux__
+extern "C" WEAK DLLEXPORT void* calloc(size_t num, size_t size) {
+    if (unlikely(!_orig_calloc)) {
+        // In some libc versions, dlsym may call calloc on error case in the ADDRESS_OF macro.
+        return NULL;
```
apangin, line 65: Does this really happen? Even if it does, I guess returning `NULL` will break something anyway.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+    if (likely(MallocTracer::initialized())) {
+        return realloc_hook(addr, size);
+    }
+    return _orig_realloc(addr, size);
+}
+#endif // __linux__
+
+static void free_hook(void* addr) {
+    _orig_free(addr);
+    MallocTracer::recordFree(addr, OS::nanotime());
```
apangin, line 102: `free` can be legitimately called for NULL. No need to record an event for this, I think.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+    _orig_malloc = ADDRESS_OF(malloc);
+    _orig_realloc = ADDRESS_OF(realloc);
+}
+
+bool MallocTracer::initialize() {
+    if (!__sync_bool_compare_and_swap(&_initialized, false, true)) {
+        return false;
+    }
+
+    CodeCache* lib = Profiler::instance()->findLibraryByName("libasyncProfiler");
```
apangin, line 138: It's not necessarily called like this. A library can have an arbitrary name, especially when extracted as a temporary file. Use `findLibraryByAddress` instead.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+    assert(lib);
+
+    lib->mark(
+        [](const char* s) -> bool {
+            return strncmp(s, "_ZL11malloc_hook", 16) == 0 || strncmp(s, "_ZL11calloc_hook", 16) == 0 || strncmp(s, "_ZL12realloc_hook", 16) == 0 || strncmp(s, " _ZL9free_hook", 13) == 0;
```
apangin, line 143: Break expression into several lines for better readability.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    return installHooks();
+}
+
+bool MallocTracer::patchLibs(bool install) {
+    if (!initialized()) {
+        return false;
+    }
+
+    MutexLocker ml(_patch_lock);
```
apangin, line 155: In theory, this may conflict with `Hooks::patchLibraries()` which has its own lock.
It makes sense to unify these two functions (also, to reduce copy-paste).
Can be done in a follow-up PR.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+void MallocTracer::recordFree(void* address, u64 time) {
+    MallocEvent event;
+    event._start_time = time;
+    event._address = (uintptr_t)address;
+    event._size = 0;
+
+    Profiler::instance()->recordSample(NULL, 0, MALLOC_SAMPLE, &event);
```
apangin, line 201: `free` calls are not sampled according to `_interval`.
It may be too expensive to record stack traces for each `free`. I therefore propose to record `free` events without a stack trace.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+Error MallocTracer::check(Arguments& args) {
+#ifndef __linux__
```
apangin, line 205: Try to avoid `ifdef`s in shared code when possible.
`OS::isLinux()` can help here (should be optimized by the compiler).

apangin (commented):
Great work!

src/converter/one/convert/JfrConverter.java
```diff
@@ -61,10 +61,15 @@ protected EventAggregator collectEvents() throws IOException {
 
         long startTicks = args.from != 0 ? toTicks(args.from) : Long.MIN_VALUE;
         long endTicks = args.to != 0 ? toTicks(args.to) : Long.MAX_VALUE;
+        boolean allowZeroes = !args.nativemem;
+        IEventReader reader = eventFilter != null ? eventFilter : jfr;
```
apangin, line 65: Need to revise this approach - it will not work for multi-chunk JFRs.
An address map should persist between chunks.

test/test/nativemem/doesmalloc.c
krk: Thanks, then I will keep JNI usage for all of them for consistence.

src/mallocTracer.cpp
krk: Delay wasn't negligible, `~50 ns` in my tests. This tiny delay caused a corresponding `free` to be recorded before `malloc` for some instances, breking the causality when events are sorted by time. I originally thought this was the cost of taking a stack trace.

src/mallocTracer.cpp
krk: When malloc tracer is not initialized. The `.ctors` usage may have made this unreachable, I will test it.

src/mallocTracer.cpp
krk: This could happen in the first `dlsym` from `.ctors`, which can call `calloc` in the tested libc implementation on an error case.

Returning NULL does not break anything as libc has a safe fallback. Libc will ignore failing `calloc` in this specific case and use another libc-internal variable instead.

src/mallocTracer.cpp
apangin: I mean, a timestamp can be taken in `MallocTracer::recordMalloc` before getting stack traces etc. There is nothing else in between - this should not take more than a couple of nanoseconds. In any case, I don't see how this can break the order, since `malloc` does not return before an event is emitted, and `free` is not called before `malloc` returns.

src/mallocTracer.cpp
apangin: With that many linux-specific ifdefs, it's better to reorganize the code and put it in a platform-specific file.
But I would prefer just removing ifdefs if this does not do any harm on macOS.

src/mallocTracer.cpp
apangin: I thought you moved initialization of `_orig` pointer to a constructor exactly to solve the mentioned problem. If not, will a lazy resolution work the same way?

README.md
```diff
@@ -27,8 +28,8 @@ Current release (3.0):
 - Linux x64: [async-profiler-3.0-linux-x64.tar.gz](https://github.com/async-profiler/async-profiler/releases/download/v3.0/async-profiler-3.0-linux-x64.tar.gz)
 - Linux arm64: [async-profiler-3.0-linux-arm64.tar.gz](https://github.com/async-profiler/async-profiler/releases/download/v3.0/async-profiler-3.0-linux-arm64.tar.gz)
 - macOS x64/arm64: [async-profiler-3.0-macos.zip](https://github.com/async-profiler/async-profiler/releases/download/v3.0/async-profiler-3.0-macos.zip)
-- Converters between profile formats: [converter.jar](https://github.com/async-profiler/async-profiler/releases/download/v3.0/converter.jar)  
-  (JFR to Flame Graph, JFR to pprof, collapsed stacks to Flame Graph)
+- Converters between profile formats: [converter.jar](https://github.com/async-profiler/async-profiler/releases/download/v3.0/converter.jar)
```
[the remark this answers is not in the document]
krk: Changed to a list.

docs/ConverterUsage.md
```diff
@@ -1,58 +1,62 @@
...
-       
-       
+
+
 JFR options:
     --cpu              Generate only CPU profile during conversion
     --wall             Generate only Wall clock profile during conversion
     --alloc            Generate only Allocation profile during conversion
+    --nativemem        Generate native memory allocation profile
+    --leak             Only include memory leaks in nativemem
```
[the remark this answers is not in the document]
krk: We could have `alloc=native`, `alloc=jvm`, `alloc=both` options perhaps?

src/mallocTracer.cpp
krk: free was not called before malloc, yet its timestamp was ~50ns lower than malloc's. Will take another look.

src/mallocTracer.cpp
krk: There is a cycle we need to break: First `dlsym` call from `_orig_malloc = ADDRESS_OF(malloc)` can invoke calloc once. When `NULL` is returned, it can recover: https://github.com/bminor/glibc/blob/5daf0c98152bdc54719a3ef62302204705e16fa6/dlfcn/dlerror.c#L141

I fixed this in the new version by using `OS::safeAlloc` only while in `getOrigAddresses` method. Keeping returning `NULL` as it would be semantically correct and better than crashing when any `_orig_*` is NULL.

test/one/profiler/test/Runner.java
krk: Removing `only`.

src/mallocTracer.cpp
krk: Removing ifdefs segfault on macos, adding back for now: https://github.com/async-profiler/async-profiler/actions/runs/12073081069/job/33668452329

There are 4 ifdefs in this file, only on the exported weak functions. I doubt the MacOS problem is in their implementation. We can keep them in the same file, ifdefs to be removed with mac support.

src/mallocTracer.cpp
krk: malloc is not reached, see calloc comments below.

src/converter/one/convert/JfrConverter.java
krk: `IEventReader` instance persists between `collectEvents` calls, as it is created in the `JfrConverter` ctor.

krk: Went with a different design around the `EventAggregator` class.

test/test/nativemem/doesmalloc.c
```diff
@@ -0,0 +1,27 @@
...
+    return (jlong)(intptr_t)ptr;
+}
+
+JNIEXPORT jlong JNICALL Java_test_nativemem_Native_realloc(JNIEnv* env, jclass clazz, jlong addr, jlong size) {
+    void* ptr = realloc((void*)addr, (size_t)size);
+    return (jlong)(intptr_t)ptr;
+}
+
+JNIEXPORT void JNICALL Java_test_nativemem_Native_free(JNIEnv* env, jclass clazz, jlong addr) {
+    free((void*)addr);
```
apangin, line 26: Does this compile on a 32-bit system? I guess an intermediate conversion to `intptr_t` is needed.

.assets/images/nativemem_flamegraph.png
apangin, line 1: Can we have a more realistic flamegraph?
I used to show native memory leaks on [this example](https://github.com/apangin/java-profiling-presentation/blob/master/src/demo10/GetImage.java#L15-L16) (strictly speaking, it's not fully a leak, since finalizers are supposed to free native resources, but in practice, finalizers can be very unpredictable).

docs/ProfilingModes.md
```diff
@@ -83,6 +83,36 @@ $ gdb $JAVA_HOME/lib/server/libjvm.so -ex 'info address UseG1GC'
 This command's output will either contain `Symbol "UseG1GC" is at 0xxxxx`
 or `No symbol "UseG1GC" in current context`.
 
+## Native memory leaks
+
+It is possible to profile `malloc` and `mmap` calls in Java context before, not always helpful. A large amount of allocations does not yet mean a leak, in case all the allocated memory is released on time.
```
apangin, line 88: Something wrong with the first sentence, can you rephrase it?

docs/ProfilingModes.md
```diff
@@ -83,6 +83,36 @@ $ gdb $JAVA_HOME/lib/server/libjvm.so -ex 'info address UseG1GC'
...
+
+The profiling mode `nativemem` records `malloc`, `realloc`, `calloc` and `free` calls with the addresses, so that allocations can be matched with frees. This helps to focus the profile report only on unfreed allocations, which are the likely to be a source of a memory leak.
+
+Example:
+
+```
+asprof start --nativemem[=N] <YourApp>
+
+asprof stop -f <app.jfr> <YourApp>
```
apangin, line 97: With JFR output, filename must be given in `start` command.

test/test/nativemem/Native.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.nativemem;
+
+public class Native {
+    static {
+        System.loadLibrary("doesmalloc");
```
apangin, line 10: Unobvious name. Maybe `jnimalloc`?

src/mallocTracer.cpp
```diff
@@ -0,0 +1,248 @@
...
+static void* malloc_hook(size_t size) {
+    void* ret = _orig_calloc(1, size);
+    if (likely(ret && size)) {
+        MallocTracer::recordMalloc(ret, size);
+    }
+    return ret;
+}
+
+#ifdef __linux__
+extern "C" WEAK DLLEXPORT void* malloc(size_t size) {
```
apangin, line 41: Does nativemem profiler work with non-Java apps if async-profiler is LD_PRELOAD'ed?

Makefile
```diff
@@ -188,19 +191,24 @@ else
...
-build-test-java: all build/$(TEST_JAR)
+build-test-java: all build/$(TEST_JAR) build-test-java-libs
 
 build-test-cpp: build/test/cpptests
 
 build-test: build-test-cpp build-test-java
 
+build-test-java-libs:
+	@mkdir -p $(TEST_LIB_DIR)
+	$(CC) -shared -fPIC $(INCLUDES) -I$(realpath src) -o $(TEST_LIB_DIR)/libdoesmalloc.$(SOEXT) test/test/nativemem/doesmalloc.c
```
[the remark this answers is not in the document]
krk: Was useful in the multi-makefile version, removing.

.assets/images/nativemem_flamegraph.png
krk: Changed with a "ByteBuffer leak while reading a stream" scenario.

src/mallocTracer.cpp
krk: Yes, tested manually.

src/codeCache.h
```diff
@@ -24,13 +22,18 @@ enum ImportId {
     im_pthread_exit,
     im_pthread_setspecific,
     im_poll,
+    im_malloc,
```
r1viollet, line 25: In case it is helpful, here are some of the functions we had to hook: https://github.com/DataDog/ddprof/blob/bccb05d89ec9e8473cd7f4f26d4e4948a70839c8/src/lib/symbol_overrides.cc#L852-L899
Though I realize you might want to have a full coverage for now.

krk: Thanks for the link. We could implement these in another iteration or prioritize based on customer requests.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,248 @@
...
+        MallocEvent event;
+        event._start_time = OS::nanotime();
+        event._address = (uintptr_t)address;
+        event._size = size;
+
+        Profiler::instance()->recordSample(NULL, size, MALLOC_SAMPLE, &event);
+    }
+}
+
+void MallocTracer::recordFree(void* address) {
```
r1viollet, line 212: Am I correct in saying we do not have a sampling decision here (as we do not want to miss matching frees) ?
[Here](https://github.com/DataDog/ddprof/blob/bccb05d89ec9e8473cd7f4f26d4e4948a70839c8/src/lib/allocation_tracker.cc#L275) we proposed to record only free events for tracked addresses. However that meant having a lock free data structure containing tracked addresses.
The ideal solution might be to have more control over the allocator being used, to embed such a logic within the allocated object.

src/profiler.cpp
```diff
@@ -784,6 +793,7 @@ void* Profiler::dlopen_hook(const char* filename, int flags) {
     void* result = dlopen(filename, flags);
     if (result != NULL) {
         instance()->updateSymbols(false);
+        MallocTracer::installHooks();
```
r1viollet, line 796: While relying on a similar strategy, I recall we were not able to catch all loaded libraries. We eventually decided to put a timer that would trigger and recheck loaded libraries every X seconds (in addition to the dlopen hook). If this is of interest, I can find more details.

r1viollet:
Exciting to see memory profiling as part of async profiler! :clap:  I mentioned a few of the gotchas we encountered while implementing something similar in case this can be of interest.

src/mallocTracer.cpp
krk: Yes, we do not have a sampling decision for `free` and wanted to avoid having to keep a list of allocation addresses.

src/profiler.cpp
krk: It would be interesting to understand when this can happen.

r1viollet: I found the commit thanks to @nsavoire. A private symbol can be used to dlopen `__libc_dlopen_mode`.
https://github.com/DataDog/ddprof/commit/d0bb6c663d98a0be891d8e5be8e7aea8067daa9a

src/converter/one/convert/JfrConverter.java
```diff
@@ -21,31 +21,52 @@
 public abstract class JfrConverter extends Classifier {
     protected final JfrReader jfr;
     protected final Arguments args;
+    private final IEventAggregator leakAggregator;
```
apangin, line 24: It feels like a leaky abstraction (pun intended): we pretend it to be some abstract aggregator, although we know it can be only MallocLeakAggregator.

src/converter/one/convert/JfrConverter.java
```diff
@@ -21,31 +21,52 @@
...
+        }
+
+        if (leakAggregator != null) {
+            leakAggregator.finish();
+
+            if (args.grain > 0) {
+                leakAggregator.coarsen(args.grain);
+            }
+
+            convertChunk(leakAggregator);
```
apangin, line 51: An illustration to the above comment: we have a special case for leakAggregator, but trying to deal with it as with some generic aggregator.

Can we design it in a way that we either have a common sequence for any aggregator or don't pretend that leak processor is an aggregator at all and just make leak processor explicitly a different thing?

For example, to consolidate EventAggregator (currently created for each new chunk) and LeakAggegator (a single instance for the whole process) we may probably replace creation of a new EventAggregator with clearing/resetting the same instance of EventAggregator. (Just as an idea; I'm open for other options too).

src/converter/one/jfr/event/EventAggregator.java
```diff
@@ -38,7 +38,7 @@ public void collect(Event e) {
         collect(e, e.samples(), e.value());
     }
 
-    public void collect(Event e, long samples, long value) {
+    protected void collect(Event e, long samples, long value) {
```
apangin, line 41: This is a public API that is currently used in other projects.

src/converter/one/jfr/event/SortedEventAggregator.java
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.jfr.event;
+
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Collections;
+
+public abstract class SortedEventAggregator extends EventAggregator {
```
apangin, line 12: Redundant abstractions, IMO.
There are currently only two aggregators. If we go this way, we can just make MallocLeakAggregator a subclass of EventAggregator.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,248 @@
...
+    void* ret = OS::safeAlloc(num * size);
+    if (likely(ret)) {
+        memset(ret, 0, num * size);
+    }
+    return ret;
+}
+
+__attribute__((constructor)) static void getOrigAddresses() {
+    // Set malloc and calloc which may be called from libc during getOrigAddresses.
+    _orig_malloc = OS::safeAlloc;
```
apangin, line 137: So is it really necessary? I remember you were saying allocation in dlsym only happens on error path, and returning NULL does not break anything.

test/native/testRunner.hpp
```diff
@@ -68,6 +68,8 @@ struct TestCase {
 
 #define __ASSERT_OR_CHECK_OP(isAssert, val1, op, val2)                                                               \
     {                                                                                                                \
+        _Pragma("GCC diagnostic push");                                                                              \
+        _Pragma("GCC diagnostic ignored \"-Waddress\"");                                                             \
```
apangin, line 72: This is getting more and more intricate.
Would not it be better to have a special version of macros for strings?

src/mallocTracer.cpp
krk: For a specific libc version yes. Removing to see test results.

src/converter/one/convert/JfrConverter.java
krk: Removing the leak and special-cased invocations.

test/native/testRunner.hpp
krk: We could split it at some point.

src/converter/one/jfr/event/IEventAcceptor.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.jfr.event;
+
+public interface IEventAcceptor {
```
apangin, line 8: I understand why you added this, but imo this is unnecessary level of abstraction.
Fortunately, as per my other comment, the only use of `IEventAcceptor` (and therefore `IEventAcceptor` itself) can be safely removed.

src/mallocTracer.cpp
```diff
@@ -0,0 +1,235 @@
...
+    void* ret = _orig_malloc(size);
+    if (likely(ret && size)) {
+        MallocTracer::recordMalloc(ret, size);
+    }
+    return ret;
+}
+
+#ifdef __linux__
+extern "C" WEAK DLLEXPORT void* malloc(size_t size) {
+    if (unlikely(!_orig_malloc)) {
```
apangin, line 42: Should we check `MallocTracer::initialized()` first? (to make less comparisons on the fast path)

src/profiler.h
```diff
@@ -26,6 +26,14 @@
 #include "writer.h"
 
 
+# ifndef likely
+#  define likely(x)     (__builtin_expect(!!(x), 1))
+# endif
+
+# ifndef unlikely
+#  define unlikely(x)   (__builtin_expect(!!(x), 0))
+# endif
```
apangin, line 35: `profiler.h` is probably not the right place for these macros, they are only used in `mallocTracer.cpp`, you can put them there. Or, to make them available for a wider use, `arch.h` will fit better.

src/converter/one/jfr/event/MallocLeakAggregator.java
```diff
@@ -0,0 +1,78 @@
...
+
+            MallocEvent e = (MallocEvent) event;
+            if (e.size > 0) {
+                addresses.put(e.address, e);
+            } else {
+                addresses.remove(e.address);
+            }
+        }
+
+        return new ArrayList<>(addresses.values());
```
apangin, line 40: The list of "filtered" events is not actually used, is it? Therefore, creating `ArrayList` seems redundant.

src/converter/one/convert/JfrConverter.java
```diff
@@ -62,19 +70,20 @@ protected EventAggregator collectEvents() throws IOException {
...
+                    eventAggregator.collect(event);
                 }
             }
         }
+        eventAggregator.finishChunk();
 
         if (args.grain > 0) {
-            agg.coarsen(args.grain);
+            eventAggregator.coarsen(args.grain);
         }
```
apangin, line 84: `finishChunk` and `coarsen` can be moved out of `collectEvents`.

src/converter/one/convert/JfrConverter.java
```diff
@@ -21,31 +21,39 @@
...
-            convertChunk();
+            convertChunk(collectEvents());
+            eventAggregator.resetChunk();
         }
-    }
 
-    protected abstract void convertChunk() throws IOException;
+        eventAggregator.finish();
+        convertChunk(eventAggregator);
+        eventAggregator.resetChunk();
```
apangin, line 48: From the design perspective, it looks strange that `convertChunk/resetChunk` are called again after `finish`.
I understand they are needed only for `MallocLeakAggregator` to process the final filtered list, but I think this should be somehow hidden in `finish`.

src/converter/one/jfr/event/MallocLeakAggregator.java
krk: It is used in multi-chunk jfrs, filtered events will be sorted together with new events, before filtering again.

src/converter/one/convert/JfrConverter.java
krk: Moving them to their own method, to make the intent clearer.

apangin:
@linking12 Yes, please try a [nightly build](https://github.com/async-profiler/async-profiler/releases/tag/nightly).

PR #1070, opened 2024-12-01
add demangling of V0 Rust symbols

arielb1:
### Description

Uses the code from https://github.com/rust-lang/rustc-demangle/pull/75/files to allow demangling V0 ("new-style") Rust symbols

### Motivation and context

Currently, async-profiler is unable to demangle "new-style" Rust symbols. This PR adds a way of demangling them.

### How has this been tested?

Currently, the demangler has been fuzzed against the official Rust demangler and found to be byte-to-byte correct. Still waiting to test this PR in the real world (cc @Mark-Simulacrum).

---


src/demangle.cpp
```diff
@@ -117,14 +60,11 @@ void Demangle::cutArguments(char* s) {
...
-        const char* h = e - 16;
-        while ((*h >= '0' && *h <= '9') || (*h >= 'a' && *h <= 'f')) h++;
-        if (h == e) {
-            return demangleRust(s + 3, e - 19);
-        }
+    // try to demangle as Rust
+    struct demangle demangle;
+    rust_demangle_demangle(s, &demangle);
+    if (rust_demangle_is_known(&demangle)) {
+        return demangleRust(s, &demangle);
```
apangin, line 67: Why is demangling performed in two steps? Can be the logic encapsulated in one call?

src/rust-demangle.cpp
```diff
@@ -0,0 +1,2033 @@
...
+
+    const char *inner;
+    size_t inner_len;
+    if (s_len >= 2 && !strncmp(s, "_R", strlen("_R"))) {
+        inner = s+2;
+        inner_len = s_len - 2;
+    } else if (s_len >= 1 && !strncmp(s, "R", strlen("R"))) {
+        inner = s+1;
+        inner_len = s_len - 1;
+    } else if (s_len >= 3 && !strncmp(s, "__R", strlen("__R"))) {
```
apangin, line 148: Why are 3 options? Don't V0 symbols always start with `_R`?

src/demangle.cpp
```diff
@@ -25,80 +25,23 @@ char* Demangle::demangleCpp(const char* s) {
...
+        if (rust_demangle_display_demangle(demangle, result, demangled_size, true /* alternate */) == OverflowOk) {
+            return result;
         }
     }
-
-    *r = 0;
-    return result;
+    free(result);
+    // demangling Rust failed, just return the original string
+    return strdup(s);
```
apangin, line 44: The function is supposed to return `NULL` if it's not a valid Rust symbol. Then the caller attempts to demangle it as a C++ symbol, and if it also fails, returns the original string.

src/demangle.cpp
```diff
@@ -117,14 +60,11 @@ void Demangle::cutArguments(char* s) {
 }
 
 char* Demangle::demangle(const char* s, bool full_signature) {
```
apangin, line 62: Note that `Demangle::demangle` is called only for symbols starting with `_Z`.
IIUC, it's not the case for V0 format.

src/demangle.cpp
arielb1: because that's how it's done in the Rust code and the C code is a copy of the Rust code

src/rust-demangle.cpp
arielb1: weird MSVC/Mac behavior

src/demangle.cpp
arielb1: will change

src/demangle.cpp
arielb1: the new revision changes that

krk:
Great work! We could add some tests, using the C++ test runner as in: https://github.com/async-profiler/async-profiler/blob/master/test/native/perfEventsLinuxTest.cpp

src/demangle.cpp
apangin: Do we want to keep this similarity? In any case, I'd expect to have this logic encapsulated in `demangleRust` rather than in a more generic `Demangle::demangle`.

arielb1:
> Great work! We could add some tests, using the C++ test runner as in: https://github.com/async-profiler/async-profiler/blob/master/test/native/perfEventsLinuxTest.cpp

Any pointers on this? The original code does not have unit/integration tests.

src/demangle.cpp
arielb1: > Do we want to keep this similarity? In any case, I'd expect to have this logic encapsulated in demangleRust rather than in a more generic Demangle::demangle.

The code being this way means we don't try to demangle invalid Rust symbols as C++. I think that is correct.

arielb1:
added tests

arielb1:
review ping

krk:
For the testing part of the PR:

Thank you for adding tests! While we don't currently have formal unit testing guidelines, I appreciate your contribution toward improving test coverage.

If needed, you can find coverage reports in the GHA checks as a link, such as in `Test and Publish Nightly Builds / test (linux-x64,, corretto 21) (pull_request)` under the `Upload coverage report` step.

We don’t have a specific coverage percentage target, but it would be useful to add tests exercising some of the functions with zero coverage, especially in the following areas:

* Punycode-related code: `punycode_decode`, `code_to_utf8`, `utf8_next_char`.
* Constants-related: `printer_print_const`, `printer_print_const_str_literal`.

arielb1:
> We don’t have a specific coverage percentage target, but it would be useful to add tests exercising some of the functions with zero coverage, especially in the following areas:

There are good tests of that code in https://github.com/rust-lang/rustc-demangle

arielb1:
Copied some of these tests over to increase the coverage

src/demangle.cpp
```diff
@@ -25,80 +25,23 @@ char* Demangle::demangleCpp(const char* s) {
     return result;
 }
 
-char* Demangle::demangleRust(const char* s, const char* e) {
+char* Demangle::demangleRust(const char *s, struct demangle const *demangle, bool full_signature) {
```
apangin, line 28: Seems like `s` argument is redundant here: it is only used for initial size estimation. `struct demangle` already has everything needed to estimate result size even better.

src/demangle.cpp
```diff
@@ -25,80 +25,23 @@ char* Demangle::demangleCpp(const char* s) {
...
-
-        s = tmp;
-        if (s[0] == '_' && s[1] == '$') s++;
-
-        if (r > result) {
-            *r++ = ':';
-            *r++ = ':';
+    char* result = NULL, *new_result;
+    for (size_t demangled_size = strlen(s) * 3 / 2 + 1; demangled_size < 1000000; demangled_size *= 2) {
+        new_result = (char *)realloc(result, demangled_size);
```
apangin, line 32: `realloc` is not used for its intended purpose (no need to copy previous data).
The code will be even simpler if we use `malloc` here and `free` after `rust_demangle_display_demangle`.

src/demangle.cpp
```diff
@@ -117,14 +60,19 @@ void Demangle::cutArguments(char* s) {
...
-    const char* e = strrchr(s, 'E');
-    if (e != NULL && e - s > 22 && e[-19] == '1' && e[-18] == '7' && e[-17] == 'h') {
-        const char* h = e - 16;
-        while ((*h >= '0' && *h <= '9') || (*h >= 'a' && *h <= 'f')) h++;
-        if (h == e) {
-            return demangleRust(s + 3, e - 19);
-        }
+    // try to demangle as Rust
+    struct demangle demangle;
+    rust_demangle_demangle(s, &demangle);
```
apangin, line 65: This runs some non-trivial pre-parsing. I wonder if this may have performance impact, given that the code is also executed for every C++ symbol.

src/demangle.h
```diff
@@ -6,15 +6,19 @@
 #ifndef _DEMANGLE_H
 #define _DEMANGLE_H
 
+#include "rust-demangle.h"
```
apangin, line 9: How exactly Rust symbols are demangled is an implementation detail; I don't think Rust-specific headers should appear in the declaration part of Demangle class. In fact, nothing is really needed from `rust-demangle.h` here - you can just put a forward declaration of `struct demangle`.

src/demangle.h
```diff
@@ -6,15 +6,19 @@
...
 
 class Demangle {
   private:
     static char* demangleCpp(const char* s);
-    static char* demangleRust(const char* s, const char* e);
+    static char* demangleRust(const char *s, struct demangle const *demangle, bool full_signature);
```
apangin, line 14: I will not ask you to restyle `rust-demangle*` sources (maybe, except for filenames), since they came from a different project, but please don't change coding style of the existing sources.

src/rust-demangle.cpp
```diff
@@ -0,0 +1,2039 @@
...
+        }
+
+        // Advance by the length
+        for (size_t i = 0; i < len; i++) {
+            if (chars_len == 0) {
+                return DemangleInvalid;
+            }
+            c = *chars++;
+            chars_len--;
+        }
```
apangin, line 1748: Why doing this char-by-char instead of just incrementing by `len` at once?

test/native/demangleTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // C++ symbol
+    CHECK_EQ(Demangle::needsDemangling("_ZNKSbIwSt11char_traitsIwESaIwEE4_Rep12_M_is_sharedEv"), true);
+    // C symbols
+    CHECK_EQ(Demangle::needsDemangling("malloc"), false);
+    CHECK_EQ(Demangle::needsDemangling("_malloc"), false);
+}
+
+// demangle calls _cxa_demangle, which should support demangling Itanium-mangled symbols across all Linux systems,
+// but might have a different mangling format on some other platforms.
+#ifdef __linux__
```
apangin, line 25: What is Linux-specific here? Why can't it run on macOS?

test/native/demangleTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+TEST_CASE(Demangle_test_demangle_cpp) {
+    const char *s = Demangle::demangle("_ZNSt15basic_streambufIwSt11char_traitsIwEE9pbackfailEj", false);
+    CHECK_EQ (strcmp(s, "std::basic_streambuf<wchar_t, std::char_traits<wchar_t> >::pbackfail"), 0);
```
apangin, line 28: `CHECK_EQ` supports comparing strings - no need for `strcmp`.
Also, redundant space before `(`.

test/native/demangleTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+    free((void*)s);
```
apangin, line 29: Why not just `free(s)`?

test/native/demangleTest.cpp
arielb1: It didn't work for me for some reason and I tried it a few different ways. Do you have an example?

test/native/demangleTest.cpp
arielb1: doesn't compile

test/native/demangleTest.cpp
arielb1: I'm not sure if the demangling scheme for MacOS is the same as the one for Linux.

src/demangle.cpp
arielb1: it should be fairly fast. most C++ symbols don't start with _ZN and a digit, and even if they do it just follows the list of lengths.

src/demangle.cpp
arielb1: I think that is smarter

src/rust-demangle.cpp
arielb1: The code is slightly weird due to being copied from Rust code. changed it to be more idiomatic C

test/native/demangleTest.cpp
arielb1: found how to fix it

src/rust-demangle.cpp
arielb1: changed it to be more Rust

test/native/demangleTest.cpp
krk: An example is https://github.com/async-profiler/async-profiler/blob/f8e887eb9dcb998ca08c4757a2dbcae9395d9733/test/native/perfEventsLinuxTest.cpp#L16

arielb1: @krk 

I get errors like

```
        actual values: s = 94904844431968 (0x5650C18A3260), "std::basic_streambuf<wchar_t, std::char_traits<wchar_t> >::pbackfail" = 94904675958744 (0x5650B77F7FD8)
```

arielb1: It does work if you do it like this:
```
    char *s = Demangle::demangle("_ZNSt15basic_streambufIwSt11char_traitsIwEE9pbackfailEj", false);
    const char *e = "std::basic_streambuf<wchar_t, std::char_traits<wchar_t> >::pbackfail";
    CHECK_EQ(s, e);
```

krk: Trying to repro, I get this:

```
    const char *a = "asd";
    const char *b = "asdf";
    CHECK_EQ(a, b);

// Assertion failed: (a == b),
//         actual values: a = "asd", b = "asdf"
```

Could you share the code for the useless errors you got?

arielb1:
Just inline one of the string parameters in the tests.On 5 Dec 2024, at 15:40, Kerem Kat ***@***.***> wrote:﻿
@krk commented on this pull request.



In test/native/demangleTest.cpp:
> +    CHECK_EQ(Demangle::needsDemangling("_ZN12panic_unwind3imp5panic17exception_cleanup17he4cf772173d90f46E"), true);
+    // Rust v0-mangled symbol
+    CHECK_EQ(Demangle::needsDemangling("_RNvCs6KtT2fMGqXk_8infiloop4main"), true);
+    // C++ symbol
+    CHECK_EQ(Demangle::needsDemangling("_ZNKSbIwSt11char_traitsIwESaIwEE4_Rep12_M_is_sharedEv"), true);
+    // C symbols
+    CHECK_EQ(Demangle::needsDemangling("malloc"), false);
+    CHECK_EQ(Demangle::needsDemangling("_malloc"), false);
+}
+
+// demangle calls _cxa_demangle, which should support demangling Itanium-mangled symbols across all Linux systems,
+// but might have a different mangling format on some other platforms.
+#ifdef __linux__
+TEST_CASE(Demangle_test_demangle_cpp) {
+    const char *s = Demangle::demangle("_ZNSt15basic_streambufIwSt11char_traitsIwEE9pbackfailEj", false);
+    CHECK_EQ (strcmp(s, "std::basic_streambuf<wchar_t, std::char_traits<wchar_t> >::pbackfail"), 0);

Trying to repro, I get this:
    const char *a = "asd";
    const char *b = "asdf";
    CHECK_EQ(a, b);

// Assertion failed: (a == b),
//         actual values: a = "asd", b = "asdf"

Could you share the code for the useless errors you got?

krk:
Thanks, fixed in https://github.com/async-profiler/async-profiler/pull/1072

arielb1:
Updated PR to include a version of the test with the fix

test/native/demangleTest.cpp
apangin: It certainly works on macOS, async-profiler uses it.

src/demangle.cpp
apangin: > most C++ symbols don't start with _ZN and a digit

The majority of JVM symbols start with _ZN and a digit:
```
nm /usr/java/jdk-23.0.1/lib/server/libjvm.so | grep " _ZN[1-9]" | wc -l
34181
```

apangin: Some of these symbols are demangled as Rust producing a different result, e.g.:
```
_ZN12_GLOBAL__N_113single_threadE
Rust: _GLOBAL__N_1::single_thread
CPP: (anonymous namespace)::single_thread
```

arielb1:
So should I keep the old hash heuristic since it seems to work well enough?On 6 Dec 2024, at 19:12, Andrei Pangin ***@***.***> wrote:﻿
@apangin commented on this pull request.



In src/demangle.cpp:
> @@ -117,14 +60,19 @@ void Demangle::cutArguments(char* s) {
 }
 
 char* Demangle::demangle(const char* s, bool full_signature) {
-    // Check if the mangled symbol ends with a Rust hash "17h<hex>E"
-    const char* e = strrchr(s, 'E');
-    if (e != NULL && e - s > 22 && e[-19] == '1' && e[-18] == '7' && e[-17] == 'h') {
-        const char* h = e - 16;
-        while ((*h >= '0' && *h <= '9') || (*h >= 'a' && *h <= 'f')) h++;
-        if (h == e) {
-            return demangleRust(s + 3, e - 19);
-        }
+    // try to demangle as Rust
+    struct demangle demangle;
+    rust_demangle_demangle(s, &demangle);

Some of these symbols are demangled as Rust producing a different result, e.g.:
_ZN12_GLOBAL__N_113single_threadE
Rust: _GLOBAL__N_1::single_thread
CPP: (anonymous namespace)::single_thread

arielb1:
So if a symbol starts with _R or if it ends with a h17…E use the Rust demangling, otherwise use cxa_demangle.This will use cxa_demangle for legacy-mangled Rust symbols that have an LLVM termination, but I think it’s not too bad.On 6 Dec 2024, at 19:16, ***@***.*** wrote:﻿So should I keep the old hash heuristic since it seems to work well enough?On 6 Dec 2024, at 19:12, Andrei Pangin ***@***.***> wrote:﻿
@apangin commented on this pull request.



In src/demangle.cpp:
> @@ -117,14 +60,19 @@ void Demangle::cutArguments(char* s) {
 }
 
 char* Demangle::demangle(const char* s, bool full_signature) {
-    // Check if the mangled symbol ends with a Rust hash "17h<hex>E"
-    const char* e = strrchr(s, 'E');
-    if (e != NULL && e - s > 22 && e[-19] == '1' && e[-18] == '7' && e[-17] == 'h') {
-        const char* h = e - 16;
-        while ((*h >= '0' && *h <= '9') || (*h >= 'a' && *h <= 'f')) h++;
-        if (h == e) {
-            return demangleRust(s + 3, e - 19);
-        }
+    // try to demangle as Rust
+    struct demangle demangle;
+    rust_demangle_demangle(s, &demangle);

Some of these symbols are demangled as Rust producing a different result, e.g.:
_ZN12_GLOBAL__N_113single_threadE
Rust: _GLOBAL__N_1::single_thread
CPP: (anonymous namespace)::single_thread

apangin:
> if a symbol starts with _R or if it ends with a h17…E use the Rust demangling, otherwise use cxa_demangle

Sounds good to me

arielb1:
Anything else?

apangin:
Just a minor thing: perhaps, rename `rust-demangle.*` to `rustDemangle.*` or something (to match file naming convention of the project), and add a link to https://github.com/rust-lang/rustc-demangle in the source code to make it clear where the code originates from?

arielb1:
The C code now has a link - https://github.com/async-profiler/async-profiler/pull/1070/commits/9ac516a560397c2668820409d7d7bd85ad3cfecb#diff-6e3cd5c659e8a31192524378ab2bf3b2a71ffd6693fa346418d045fb29cb3188R9

Will rename the file.

arielb1:
updated to use the old (same as before this PR) heuristic

PR #1065, opened 2024-11-27
Fix #1060

swivelmargarita:
This commit attempts to fix #1060

Issue #1060, opened 2024-11-25
 Profiler fails to start if both `rela.plt` and `rela.dyn` have needed imports


swivelmargarita:
### Describe the bug

If the needed [imports](https://github.com/async-profiler/async-profiler/blob/master/src/codeCache.cpp#L185) happens to be residing in `rela.dyn` and `rela.plt` both, not only in one of them, `async-profiler` fails to start with the message `[ERROR] Could not set pthread hook`.

This is because[`ElfParser::parseDynamicSection()`]( https://github.com/async-profiler/async-profiler/blob/master/src/symbols_linux.cpp#L397) does not parse `rela.dyn` if it sees that `rela.plt` is present in the shared library.

Please see the Additional Information/Context.

#### Expected Behavior
The `async-profiler` to be run normally.

#### Actual Behavior
Error stating `async-profiler` couldn't set pthread hook
```
$ docker run async-profiler-liberica:latest
[ERROR] Could not set pthread hook


              |\      _,,,--,,_
             /,`.-'`'   ._  \-;;,_
  _______ __|,4-  ) )_   .;.(__`'-'__     ___ __    _ ___ _______
 |       | '---''(_/._)-'(_\_)   |   |   |   |  |  | |   |       |
 |    _  |    ___|_     _|       |   |   |   |   |_| |   |       | __ _ _
 |   |_| |   |___  |   | |       |   |   |   |       |   |       | \ \ \ \
 |    ___|    ___| |   | |      _|   |___|   |  _    |   |      _|  \ \ \ \
 |   |   |   |___  |   | |     |_|       |   | | |   |   |     |_    ) ) ) )
 |___|   |_______| |___| |_______|_______|___|_|  |__|___|_______|  / / / /
 ==================================================================/_/_/_/

:: Built with Spring Boot :: 3.3.5


2024-11-25T11:23:56.744Z  INFO 1 --- [           main] o.s.s.petclinic.PetClinicApplication     : Starting PetClinicApplication v3.3.0 using Java 21.0.5 with PID 1 (/app/petclinic.jar started by root in /app)
2024-11-25T11:23:56.747Z  INFO 1 --- [           main] o.s.s.petclinic.PetClinicApplication     : No active profile set, falling back to 1 default profile: "default"
2024-11-25T11:23:57.162Z  INFO 1 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Bootstrapping Spring Data JPA repositories in DEFAULT mode.
[...]

PR #1074, opened 2024-12-06
Parse both .rela.dyn and .rela.plt of libraries.

krk:
It is possible to have symbols in both `.rela.dyn` and `.rela.plt` sections, depending how the symbols are used. Furthermore, `.rela.dyn` can have `R_X86_64_64` relocation type among others.

async-profiler did not parse `.rela.dyn` section if `.rela.plt` was present.

Support parsing `.rela.dyn` `R_X86_64_64` symbols and both `.rela.dyn` and `.rela.plt` in the same ELF binary, to enable internal usage (hooking) of the imported functions.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1060

### Motivation and context
While investigating https://github.com/async-profiler/async-profiler/issues/1060 and the possible fix in https://github.com/async-profiler/async-profiler/pull/1065, noticed that `R_X86_64_64` relocation type is not supported. 

### How has this been tested?
New test added in `symbolsLinuxTest.cpp`.

---


src/symbols_linux.cpp
```diff
@@ -145,22 +145,29 @@ typedef Elf32_Dyn  ElfDyn;
...
+#  define R_DIRECT R_X86_64_64
 #elif defined(__i386__)
 #  define R_GLOB_DAT R_386_GLOB_DAT
+#  define R_DIRECT -1
 #elif defined(__arm__) || defined(__thumb__)
 #  define R_GLOB_DAT R_ARM_GLOB_DAT
+#  define R_DIRECT -1
 #elif defined(__aarch64__)
 #  define R_GLOB_DAT R_AARCH64_GLOB_DAT
+#  define R_DIRECT -1
```
apangin, line 157: Isn't there a AArch64 equivalent?
I see it generates `R_AARCH64_ABS64`

src/symbols_linux.cpp
```diff
@@ -394,17 +401,19 @@ void ElfParser::parseDynamicSection() {
                     _cc->addImport((void**)(base + r->r_offset), strtab + sym->st_name);
                 }
             }
-        } else if (rel != NULL && relsz != 0) {
+        }
+
+        if (rel != NULL && relsz != 0) {
```
apangin, line 406: Can the same function occur in multiple relocations? I believe it can. We don't necessarily need to handle such cases right now, but if this happens, `.rela.plt` should take precedence.

apangin (commented):
Great job on creating a test case that produces different relocation types in one binary.

src/symbols_linux.cpp
krk: I originally added something for all archs, but given I cannot test them, I focused on x86-64. I can add it for arm64.

src/symbols_linux.cpp
```diff
@@ -394,17 +401,19 @@ void ElfParser::parseDynamicSection() {
...
             // Shared library was built without PLT (-fno-plt)
             // Relocation entries have been moved from .rela.plt to .rela.dyn
             for (size_t offs = relcount * relent; offs < relsz; offs += relent) {
                 ElfRelocation* r = (ElfRelocation*)(rel + offs);
-                if (ELF_R_TYPE(r->r_info) == R_GLOB_DAT) {
+                if (ELF_R_TYPE(r->r_info) == R_GLOB_DAT || ELF_R_TYPE(r->r_info) == R_DIRECT) {
                     ElfSymbol* sym = (ElfSymbol*)(symtab + ELF_R_SYM(r->r_info) * syment);
                     if (sym->st_name != 0) {
-                        _cc->addImport((void**)(base + r->r_offset), strtab + sym->st_name);
+                        _cc->tryAddImport((void**)(base + r->r_offset), strtab + sym->st_name);
```
apangin, line 414: Maybe just switch the order of these two loops? This will ensure proper precedence automatically without override logic in CodeCache?

PR #1082, opened 2024-12-16
Regression: Cross-compilation does not work after CC/CXX changes

shipilev:
### Description

I noticed that my own cross-compiled binaries from https://builds.shipilev.net/async-profiler/ do not build cleanly anymore. In fact, there is a bug in `Makefile`, introduced by this hunk:
https://github.com/async-profiler/async-profiler/commit/3cf733d58970f8c130b659821e244150eef40ee8#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R19

The problem is that `CC` and `CXX` are pre-defined in GNU Make: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html. So, `?=` does not work well: it _never_ overwrites to `$(CROSS_COMPILE)...`, which results in binaries being compiled by host compiler, and only later `strip` discovers the binary format is wrong:

```
$ ARCH_TAG=aarch64 CROSS_COMPILE=/home/buildbot/deps/x-tools/aarch64-linux-gnu/bin/aarch64-linux-gnu- make clean release
rm -f -r build
mkdir -p build/bin
mkdir -p build/lib
for f in src/*.cpp; do echo '#include "'$f'"'; done |\
g++  -O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11  -U_FORTIFY_SOURCE -Wl,-z,defs -Wl,--exclude-libs,ALL -static-libstdc++ -static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections -fwhole-program -DPROFILER_VERSION=\"3.0\" -I/usr/lib/jvm/java-11-openjdk-amd64/include -Isrc/helper -I/usr/lib/jvm/java-11-openjdk-amd64/include/linux -fPIC -shared -o build/lib/libasyncProfiler.so -xc++ - -ldl -lpthread -lrt
cc  -O3 -fno-exceptions  -DPROFILER_VERSION=\"3.0\" -o build/bin/asprof src/main/*.cpp src/jattach/*.c
/home/buildbot/deps/x-tools/aarch64-linux-gnu/bin/aarch64-linux-gnu-strip build/bin/asprof
/home/buildbot/deps/x-tools/aarch64-linux-gnu/bin/aarch64-linux-gnu-strip: Unable to recognise the format of the input file `build/bin/asprof'
make: *** [Makefile:152: build/bin/asprof] Error 1
```

This PR reverts `?=` back to original form.

### Related issues

https://github.com/async-profiler/async-profiler/pull/1036 introduced the bug.

### How has this been tested?

Tested to build fine on `builds.shipilev.net` workers.

I am not actually sure it does not break some assumptions in https://github.com/async-profiler/async-profiler/pull/1036, and GHA workflows need maintainer approval to run :)

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,8 @@ API_JAR=jar/async-profiler.jar
 CONVERTER_JAR=jar/jfr-converter.jar
 TEST_JAR=test.jar
 
-CC ?= $(CROSS_COMPILE)gcc
-CXX ?= $(CROSS_COMPILE)g++
+CC=$(CROSS_COMPILE)gcc
+CXX=$(CROSS_COMPILE)g++
 STRIP=$(CROSS_COMPILE)strip
 
 CFLAGS_EXTRA ?=


roy-soumadipta (approved):

shipilev:
Actually, I might need to adjust the `CC=` in places like this:
https://github.com/async-profiler/async-profiler/blob/78f78cf6817a07f7c9a2353c42d67ab1a6c63921/.github/workflows/test-and-publish-nightly.yml#L79

Maybe a cleaner way out would be to drop `CROSS_COMPILE` completely, and just rely on `CC`/`CXX`/`STRIP` overrides.

apangin:
All good. Thanks for pointing out a bug and a quick fix.

shipilev:
Thanks! Looks like my binaries are built well now: https://builds.shipilev.net/async-profiler/

But I still have a question about this block:
https://github.com/async-profiler/async-profiler/blob/78f78cf6817a07f7c9a2353c42d67ab1a6c63921/.github/workflows/test-and-publish-nightly.yml#L79

Now that we override CC/CXX unconditionally to CROSS_COMPILE-d path, have we lost compilation with musl-gcc in the line above?

shipilev:
> Now that we override CC/CXX unconditionally to CROSS_COMPILE-d path, have we lost compilation with musl-gcc in the line above?

Ah no, I confused myself. In current master, if we override `CC` from a command line, make would ignore our definition in `Makefile`. Which is the same effect I think we originally wanted with `?=`.

Anyway, do you think `CROSS_COMPILE` still holds water? I think we can just drop all these three lines and ask users to configure their CC/CXX/STRIP instead:

```
CC=$(CROSS_COMPILE)gcc
CXX=$(CROSS_COMPILE)g++
STRIP=$(CROSS_COMPILE)strip
```

apangin:
@shipilev Right, it's possible to override a local Makefile assignment on the command line.
`?=` would also allow overriding it with an environment variable, but that's not something we absolutely need.

Of course, a user can override all of CC, CXX and STRIP. `CROSS_COMPILE` is just a shortcut for doing that in a common use case when all have the same prefix. I think we can leave it for convenience.

PR #1104, opened 2025-01-13
Mitigate cross site authentication attacks

openorclose:
Mitigate cross site authentication attacks

---

### DIFF
--- a/src/res/heatmap.html
+++ b/src/res/heatmap.html
@@ -242,7 +242,7 @@
 	<pre id="methods">/*methods:*/</pre>
 </div>
 <span class="status" style="border-right: none; border-top-left-radius: 4px; right:0;">
-	<span id="produced">Produced by <a href='https://github.com/async-profiler/async-profiler'>async-profiler</a>&nbsp;</span>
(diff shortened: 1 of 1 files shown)


src/res/heatmap.html
```diff
@@ -242,7 +242,7 @@
 	<pre id="methods">/*methods:*/</pre>
 </div>
 <span class="status" style="border-right: none; border-top-left-radius: 4px; right:0;">
-	<span id="produced">Produced by <a href='https://github.com/async-profiler/async-profiler'>async-profiler</a>&nbsp;</span>
+	<span id="produced">Produced by <a href='https://github.com/async-profiler/async-profiler' target='_blank' rel='noopener noreferrer'>async-profiler</a>&nbsp;</span>
```
apangin, line 245: I understand what "noopener noreferrer" mean, but could you please elaborate on why you have changed target to `_blank` and how authentication is compromised without that?

openorclose: Without the blank target, the authentication popup for the site might popup over the current page, which could result in user confusion about whether the authentication is for the current page or for the new page

apangin:
As discussed, this was a false positive finding from an automated scanner.

PR #1120, opened 2025-01-26
Demo for special pid by java api

kings1990:
### Description

Add the demo for profile special pid by java api

Hope this example can help other developers


apangin:
Can you please describe your use case in detail?
Why would a user want to use this approach instead of one of the standard ways to run async-profiler, such as `asprof`, `-agentpath` and Java API for embedded use?

apangin:
This PR has been inactive for a while, I'm closing it.

Issue #1106, opened 2025-01-14
Support TSC with no JVM


arielb1:
### Describe the feature

Currently TSC only seems to be supported in JVM mode. Support it in no-JVM mode as well.

### Use Case

Supporting TSC in non-JVM mode will allow linking tsc-based timestamps between profiling events and logs from the system being profiled in a consistent way.

### Proposed Solution

Should be a simple matter of programming. We just need some code that gets the tick frequency from CPUID

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
Resolved by #1123

PR #1123, opened 2025-02-03
support clock=tsc without a JVM

arielb1:
### Description

Supports clock=tsc without a JVM. See #1106.

The clock calibration logic has been copied from https://github.com/metrics-rs/quanta

### How has this been tested?

unit tests + tested locally

---

### DIFF
--- a/src/lockTracer.cpp
+++ b/src/lockTracer.cpp
@@ -36,16 +36,7 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
 
 
 Error LockTracer::start(Arguments& args) {
-    if (!VM::loaded()) {
-        return Error("lock tracer only works with the JVM loaded");
-    }
-
-    if (!TSC::frequencyAvailable()) {
-        // shouldn't really happen - frequency should always be available with the JVM loaded.
-        // check anyway.
-        return Error("lock tracer only works with the JVM's TSC frequency calibration");
-    }
-
+    // There is a JVM here, so TSC::frequency is calibrated from it
     _ticks_to_nanos = 1e9 / TSC::frequency();
     _interval = (u64)(args._lock * (TSC::frequency() / 1e9));
     _total_duration = 0;
--- a/src/lockTracer.cpp
+++ b/src/lockTracer.cpp
@@ -36,6 +36,16 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
 
 
 Error LockTracer::start(Arguments& args) {
+    if (!VM::loaded()) {
+        return Error("lock tracer only works with the JVM loaded");
+    }
+
+    if (!TSC::frequencyAvailable()) {
+        // shouldn't really happen - frequency should always be available with the JVM loaded.
+        // check anyway.
+        return Error("lock tracer only works with the JVM's TSC frequency calibration");
+    }
+
     _ticks_to_nanos = 1e9 / TSC::frequency();
     _interval = (u64)(args._lock * (TSC::frequency() / 1e9));
     _total_duration = 0;
--- a/src/tsc.cpp
+++ b/src/tsc.cpp
@@ -7,6 +7,7 @@
 #include "tsc.h"
 #include "vmEntry.h"
 
+
 bool TSC::_initialized = false;
 bool TSC::_available = false;
 bool TSC::_enabled = false;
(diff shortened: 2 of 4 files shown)


apangin:
Thank you for your PR.
Could you please provide a bit more context on what calibration logic does and why it is needed?
How do you check that TSC can be used for timestamps? Although TSC is reliable on modern x86 CPUs, this is not always the case for older processors.
Can you also clarify the ownership of the code? If the algorithm is copied from another project with a different license, I'm afraid I can't accept the PR without consent of the original author.

arielb1:
On AMD and ARM processors, there is no API to get the TSC frequency, so you need a calibration loop to measure it.The code currently does not check that the tsc is invariant (i.e. ticks at a constant real-world frequency). If you run it on an Intel processor with a non-invariant tsc, the ticks-per-second will be incorrect.On the other hand, this code only runs if you explicitly ask for clock source=tsc. I am not sure that silently reverting to wall clock time in that case is actually better. I don’t care either way.On 9 Feb 2025, at 2:37, Andrei Pangin ***@***.***> wrote:﻿
Thank you for your PR.
Could you please provide a bit more context on what calibration logic does and why it is needed?
How do you check that TSC can be used for timestamps? Although TSC is reliable on modern x86 CPUs, this is not always the case for older processors.
Can you also clarify the ownership of the code? If the algorithm is copied from another project with a different license, I'm afraid I can't accept the PR without consent of the original author.

arielb1:
It’s easy enough to add a check for non invariant tsc. Should I add one?If the calibration loop is only needed when Java lock profiling is used, and Java lock profiling is only used when there is a JVM, can I just remove the calibration loop code and instead put an assertion to ensure the tsc frequency is only used in JVM mode?On 10 Feb 2025, at 2:47, Andrei Pangin ***@***.***> wrote:﻿

you need a calibration loop to measure it

Not really. async-profiler automatically computes frequency at the end of each JFR chunk. The only case when the frequency is needed in advance is Java lock profiling, which requires translating ticks to nanoseconds on the go.

If you run it on an Intel processor with a non-invariant tsc, the ticks-per-second will be incorrect.

That's my concern exactly.

this code only runs if you explicitly ask for clock source=tsc

But this is not how it currently works. With your change, async-profiler will always attempt to use TSC, unless clock=monotonic is set explicitly.

apangin:
Yes, I believe calibration loop is unnecessary for non-Java processes, at least for now, while there are no events with thresholds measured in nanoseconds.

The simplest way to go forward will be to choose TSC for non-Java processes only when `clock=tsc` is set explicitly: `clock == CLK_TSC`. Alternatively, if a check for reliablie tsc is implemented, tsc can be selected by default.

arielb1:
Pushed a branch with no calibration code. Seems to work just fine.

arielb1:
added invariant tsc detection

src/lockTracer.cpp
```diff
@@ -36,6 +36,16 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
 
 
 Error LockTracer::start(Arguments& args) {
+    if (!VM::loaded()) {
+        return Error("lock tracer only works with the JVM loaded");
+    }
```
apangin, line 41: Not needed. This is already checked at a higher level:
https://github.com/async-profiler/async-profiler/blob/75aefa36c9a9c1568c94f3d7e48017b3b3bece1b/src/profiler.cpp#L1095-L1096

src/lockTracer.cpp
```diff
@@ -36,6 +36,16 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
...
+
+    if (!TSC::frequencyAvailable()) {
+        // shouldn't really happen - frequency should always be available with the JVM loaded.
```
apangin, line 44: This shouldn't happen indeed, and thus the code is redundant.
It may be replaced with `assert`. However, given that `frequencyAvailable()` is not used anywhere else, I'd get rid of this function entirely.

src/tsc.cpp
```diff
@@ -7,43 +7,47 @@
...
+            }
 
-            u64 frequency = env->CallLongMethod(env->GetStaticObjectField(cls, jvm), getTicksFrequency);
-            if (frequency > NANOTIME_FREQ) {
-                // Default 1GHz frequency might mean that rdtsc is not available
-                u64 jvm_ticks = env->CallStaticLongMethod(cls, counterTime);
-                _offset = rdtsc() - jvm_ticks;
-                _frequency = frequency;
+            env->ExceptionClear();
+            _initialized = true;
```
apangin, line 44: Why is it set to `true` only in the VM branch? The initialization code is supposed to run once in non-Java case too, isn't it?

arielb1: I didn't understand the code logic. fixed that.

arielb1: ed: when I wrote the code I tried to follow the pre-existing logic, which didn't set `_initialized` if the JVM was not loaded. I think that it is better to do it this way.

Issue #1121, opened 2025-01-29
nativemem doesn't hook all allocations


olivergillespie:
### Describe the bug

In my rust application, it seems that not all allocations are hooked. If I compare the results between jemalloc/jeprof and async-profiler's nativemem, some stacks are missing from async-profiler's results. I'd rather use async-profiler, if possible!

It seems to be that for some reason my code is not using the public malloc entry. I used `-e cpu` to monitor the same application, and here are two of the stacks that are missing from `nativemem`. They appear to not use the public malloc method:

```
__GI___libc_malloc
alloc::raw_vec::RawVecInner<A>::try_allocate_in
alloc::raw_vec::RawVecInner<A>::with_capacity_in
<T as alloc::slic::hack::ConvertVec>::to_vec
<alloc::borrow::Cow<str> as alloc::string::ToString>::to_string
...
my_proprietary_code
...
```

```
_int_malloc
_int_memalign
_mid_memalign
__posix_memalign
__rdl_alloc
thread_local::ThreadLocal<T>::insert
...
my_proprietary_code
...
```

One of these was the biggest allocator in my application.

I haven't managed to create a simple reproducer, I'm not sure if I have some link-time optimization settings interfering with things, or something else environmental. I appreciate this may be a problem with my application, not async-profiler, but somehow jemalloc does manage to hook everything (using glibc's dedicated [hooks](https://www.gnu.org/software/libc/manual/html_node/Hooks-for-Malloc.html)?).

### Async-profiler version

Building directly from master

olivergillespie:
Thanks! I'll try to test it this week.

apangin:
The fix has been confirmed.

PR #1128, opened 2025-02-06
Patch both rela.plt and rela.dyn for hooks

krk:
If a symbol has both plt and dyn relocations, both are patched now, e.g. for `malloc`. Otherwise, dyn invocations would not be hooked as plt has precedence in symbols_linux.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1121

### Motivation and context
Only hooking plt would be incomplete, when there are .rela.dyn usage as well, for the native memory tracer.

### How has this been tested?
New test `malloc_plt_dyn` added with a `malloc` invocation via both `plt` and `dyn` entries.

---


src/codeCache.h
```diff
@@ -103,7 +109,7 @@ class CodeCache {
     unsigned int _plt_offset;
     unsigned int _plt_size;
 
-    void** _imports[NUM_IMPORTS];
+    void** _imports[NUM_IMPORT_TYPES * NUM_IMPORTS];
```
apangin, line 112: Maybe `_imports[NUM_IMPORT_TYPES][NUM_IMPORTS]`? for better readability and to avoid manual index calculation.

src/codeCache.h
```diff
@@ -116,6 +122,15 @@ class CodeCache {
 
     void expand();
     void makeImportsPatchable();
+    void saveImport(ImportId id, void** entry);
+
+    inline void** getImport(ImportId id, ImportType type) {
```
apangin, line 127: I'm not a fan of adding `inline` everywhere, it's just noise.
But with two-dimensional `_imports`, this function is likely not needed at all.

Makefile
```diff
@@ -205,6 +206,10 @@ build-test-libs:
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libreladyn.$(SOEXT) test/native/libs/reladyn.c
 	$(CC) -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjnimalloc.$(SOEXT) test/native/libs/jnimalloc.c
 
+build-test-bins:
+	@mkdir -p $(TEST_BIN_DIR)
+	$(CXX) -o $(TEST_BIN_DIR)/malloc_plt_dyn test/native/bins/malloc_plt_dyn.c
```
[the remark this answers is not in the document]
krk: will use `gcc` directly here.

src/codeCache.cpp
```diff
@@ -168,42 +168,52 @@ const void* CodeCache::findSymbolByPrefix(const char* prefix, int prefix_len) {
     return NULL;
 }
 
+void CodeCache::saveImport(ImportId id, void** entry) {
+    for (int ty = 0; ty < NUM_IMPORT_TYPES; ty++) {
+        ImportType type = static_cast<ImportType>(ty);
```
apangin, line 173: Isn't this cast redundant?

Makefile
```diff
@@ -205,6 +206,11 @@ build-test-libs:
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libreladyn.$(SOEXT) test/native/libs/reladyn.c
 	$(CC) -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjnimalloc.$(SOEXT) test/native/libs/jnimalloc.c
 
+build-test-bins:
+	@mkdir -p $(TEST_BIN_DIR)
+	gcc -o $(TEST_BIN_DIR)/malloc_plt_dyn test/native/bins/malloc_plt_dyn.c
+	gcc -o $(TEST_BIN_DIR)/nativeApi -Isrc -ldl test/test/nativeapi/nativeApi.c
```
apangin, line 212: underscore_convention vs. camelCase - can we make naming consistent?

src/codeCache.cpp
```diff
@@ -214,22 +224,34 @@ void** CodeCache::findImport(ImportId id) {
...
-    void** entry = findImport(id);
-    if (entry != NULL) {
-        *entry = hook_func;
+    if (!_imports_patchable) {
+        makeImportsPatchable();
+        _imports_patchable = true;
+    }
+
+    for (int ty = 0; ty < NUM_IMPORT_TYPES; ty++) {
+        ImportType type = static_cast<ImportType>(ty);
```
apangin, line 237: same as above

src/codeCache.h
```diff
@@ -29,6 +29,12 @@ enum ImportId {
     NUM_IMPORTS
 };
 
+enum ImportType {
+    PRIMARY,
+    SECONDARY,
+    NUM_IMPORT_TYPES,
```
apangin, line 35: Please remove `,` after `NUM_IMPORT_TYPES` as it will always be the last constant

test/test/nativeapi/NativeapiTests.java
```diff
@@ -3,17 +3,17 @@
...
-package test.c;
+package test.nativeapi;
 
 import one.profiler.test.Os;
 import one.profiler.test.Output;
 import one.profiler.test.Test;
 import one.profiler.test.TestProcess;
 
-public class CTests {
+public class NativeapiTests {
```
apangin, line 13: Renaming will break our internal pipelines that have CTests in exclude list.

src/codeCache.cpp
krk: it is now, good catch.

Makefile
```diff
@@ -205,6 +206,11 @@ build-test-libs:
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libreladyn.$(SOEXT) test/native/libs/reladyn.c
 	$(CC) -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjnimalloc.$(SOEXT) test/native/libs/jnimalloc.c
 
+build-test-bins:
+	@mkdir -p $(TEST_BIN_DIR)
+	gcc -o $(TEST_BIN_DIR)/malloc_plt_dyn test/native/bins/malloc_plt_dyn.c
+	gcc -o $(TEST_BIN_DIR)/native_api -Isrc -ldl test/test/c/native_api.c
```
apangin, line 212: I thought you would rename `malloc_plt_dyn` instead :)

Anyhow, let's agree on where to store test-specific C sources. I'd prefer to keep them in a directory of a particular test (like `native_api.c` in `test/c` directory). A common directory may also do, but this should be consistent across tests.

krk: :) I moved `malloc_plt_dyn` to `nativemem` folder.

PR #1129, opened 2025-02-07
Add test name filtering to the java test runner

krk:
### Description
Test filter format is controlled by makefile variables `TESTS` and `SKIP`. Both are case-sensitive and accept the format and rules as:

### TESTS

1. If the argument exactly specifies an existing directory under `test/test`, include all tests in the specified directory.
    1. `make test-java TESTS=alloc`
1. If the argument is not an existing directory name, it is a `filter`. Supported filter types are:
    1. Exact match: `CstackTests.vmStructs`
    1. Starts with: `Cstac*`
    1. Ends with: `*Structs`
    1. Contains: `*vm*`
1. Filters are matched on `<ClassName>.<methodName>`.
1. Multiple arguments can be specified with space or comma.
    1. `make test-java TESTS='*alloc* CstackTests.vmStructs'`
    1. `make test-java TESTS='*alloc*,CstackTests.vmStructs'`
1. If a directory and a filter is specifies at the same time, tests in directories are included regardless of the filter. Filter applies to tests in other directories.
    1. `make test-java TESTS=alloc,CstackTests.vmStructs` matches all tests in `alloc` folder and any other test matching exactly `CstackTests.vmStructs`.
1. If only filter is specified and not directory, it applies to all tests.
    1. `make test-java TESTS='*Calloc*'` matches all tests with `Calloc` in their `<ClassName>.<methodName>`.

### SKIP

Skip does not accept directory names. Exclusion is via glob matching the `<ClassName>.<methodName>` as in `TESTS`:

1. `make test-java TESTS='*Calloc*' SKIP="*Tra*"`:
    1. Skips `NativememTests.canAsprofTraceMallocCalloc`
    1. Executes `NativememTests.canAgentFilterMallocCalloc`
1. `make test-java TESTS='*Calloc*' SKIP="*Tra* nativemem"`:
    1. `nativemem` is considered an exact match and not a directory name.
    1. Same set of tests run as above.
1. `make test-java SKIP='*lloc*'`
    1. Skip all tests containing `lloc` in their `<ClassName>.<methodName>`.

### How has this been tested?
```
make test-java TESTS=calloc
make test-java TESTS=nativemem
make test-java TESTS=CstackTests.vmStructs
make test-java TESTS='malloc CstackTests.vmStructs'
make test-java TESTS='*alloc* CstackTests.vmStructs'
make test-java TESTS=JfrTests
make test-java TESTS=jfrNoLeaks

# Run all nativemem and jfr tests that ends with the string 'Calloc' in their name:
make test-java TESTS='nativemem jfr *Calloc'

make test-java SKIP='*lloc*'
```

Use `LOG_LEVEL=FINE` to log parsed test declaration and filters:

```
make test-java TESTS='alloc *Malloc CstackTests.vmStructs' SKIP='*vm*' LOG_LEVEL=FINE
...
FINE: Selected directories: [alloc]
FINE: Test Filters: [*Malloc, CstackTests.vmStructs]
FINE: Skip Filters: [*vm*]
...
```

---


apangin:
There are already multiple ways to enable/disable tests: `TESTS` and `SKIP` variables and `enabled` annotation parameter. Introducing one more would only increase confusion. It's better to have a uniform way to enable/disable tests basing on a test name (which includes directory, class name and method name).

Here is my proposal, but I'm open to other suggestions:
1. `TESTS` and `SKIP` should follow the same syntax. Runner will find all tests that match `TESTS` and do not match `SKIP` selectors.
2. Either varible allows to specify multiple test selectors separated by a comma or whitespace.
3. Selector may consist of a directory name (`cpu`, `nativemem`), simple class name (`CpuTests`, `NativememTests`), method name (`jfrNoFree`) or a combination of simple class name and method (`NativememTests.jfrNoFree`).

This will be intuitive, simple to use, and backward compatible.

krk:
If only a single test is required to run during development, `enabled` will not cut it, as we may have to add `enabled = false` to all "other" test cases.

We could reintroduce the `only = true` from a previous cancelled PR. When `only = true` is present, only runs the tests with that attribute.

apangin:
Right, `enabled` is for different purpose, let's leave it aside.
I think the best way to run a single test would be `make test TESTS=SomeClass.singleMethod`

Makefile
```diff
@@ -211,7 +213,7 @@ test-cpp: build-test-cpp
 
 test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
-	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(TESTS)
+	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(TEST_DIRS) $(TEST_FILTER)
```
apangin, line 216: There are still two different variables. My idea was to have one, so a user can specify `TESTS=cpu` or `TESTS=ClassName` or `TESTS=ClassName.methodName`. No filters - just a list of tests to run. If it helps, make test specifier case-sensitive, so it's easy to disambiguate between directory names and class names.

test/one/profiler/test/Runner.java
```diff
@@ -210,6 +215,55 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
         System.out.println("TOTAL: " + testCount);
     }
 
+    private static RunnerDeclaration parseRunnerDeclaration(String[] args) {
+        List<String> testNames = new ArrayList<>();
+        List<String> testDirectories = new ArrayList<>();
+        List<String> testFilters = new ArrayList<>();
```
apangin, line 221: I think you overcomplicated it a bit. There is no division into directories, names and filters. Furthermore, there are no filters. Everything passed as an argument is just a test name. The rules to interpret a test name can be the following (you may suggest your version, but as usual - the simpler the better).
1. If it only consists of lowercase letters - it's a directory name (alternatively, you may check for directory existence, but this is not necessary). A directory name is unambiguously translated to the class name: `alloc` -> `test.alloc.AllocTests`.
2. Otherwise, if it ends with `Tests`, it's a class name. It may be a fully qualified class name or short class name. In the latter case, the package is prepended using the same basic rules: `AllocTests` -> `test.alloc.AllocTests`.
3. Otherwise, if it contains a dot, it's a class name with a method name.

test/one/profiler/test/Runner.java
```diff
@@ -210,6 +215,55 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+
+        // If 'arg' contains a period and not starts with a lowercase character, treat it as a testName.
+        // Otherwise, if it is an existing directory, consider it a directory.
+        // Else, treat it as a filter.
+        for (int i = 0; i < args.length; i++) {
+            String arg = args[i];
+
+            if(arg.indexOf('.') >= 0 && !Character.isLowerCase(arg.charAt(0))) {
```
apangin, line 229: Please, format the code. There is always a space between a keyword and `(`.

test/one/profiler/test/RunnerDeclaration.java
```diff
@@ -0,0 +1,38 @@
...
+        return directories;
+    }
+
+    public List<String> filters() {
+        return filters;
+    }
+
+    public boolean isFilterMatch(Method m) {
+        String fullNameLower = (m.getDeclaringClass().getName() + '.' + m.getName()).toLowerCase();
+        return filters().isEmpty() || filters().stream().anyMatch(f -> fullNameLower.contains(f.toLowerCase()));
```
apangin, line 36: To avoid ambiguity, I would not convert names to lowercase and would check for exact match rather than `contains`. Otherwise, `AllocTests.alloc` will match both `alloc` and `allocTotal`; `liveness` will also select `livenessJfrHasStacks` and so on. This is probably not what a user expects.

test/one/profiler/test/Runner.java
krk: That directory-name division was already in the code, I made it explicit by splitting `args` into `testNames` and `testDirectories`. `testFilters` is a new addition.

I am hoping to implement a `contains` matching. Current PR does not implement exact matching.

krk:
updated PR description with the latest changes.

test/one/profiler/test/Runner.java
```diff
@@ -210,6 +215,55 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+
+            File f = new File("test/test/" + args[i]);
+            if (!f.exists() || !f.isDirectory()) {
+                testFilters.add(args[i]);
+            } else {
+                testDirectories.add(args[i]);
+            }
+        }
+
+        if (testDirectories.isEmpty()) {
```
apangin, line 248: This conditional branch makes it even more confusing.
Maybe we can always get a list of all available tests and then let specifiers (or filters, whatever you call it) select matched ones?

apangin:
One more proposal: convert `,` to ` ` when passing `TESTS` list to the command line.
It is easier to specify `make test TESTS=cpu,alloc` instead of `make test TESTS='cpu alloc'`

test/one/profiler/test/Runner.java
```diff
@@ -210,6 +215,55 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+        List<String> testDirectories = new ArrayList<>();
+        List<String> testFilters = new ArrayList<>();
+
+        // If 'arg' contains a period and not starts with a lowercase character, treat it as a testName.
+        // Otherwise, if it is an existing directory, consider it a directory.
+        // Else, treat it as a filter.
+        for (int i = 0; i < args.length; i++) {
+            String arg = args[i];
+
+            if (arg.indexOf('.') >= 0 && !Character.isLowerCase(arg.charAt(0))) {
```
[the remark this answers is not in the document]
krk: This condition is met e.g. for `CstackTests.vmStructs`.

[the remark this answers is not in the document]
krk: Thanks, simplified and refactored.

krk:
Simplified and refactored. Added `SKIP` support with the glob format, updated PR desc.

Makefile
```diff
@@ -227,7 +228,7 @@ test-cpp: build-test-cpp
 
 test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
-	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(TESTS)
+	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(subst $(COMMA), ,$(TESTS))
```
apangin, line 231: Since `Runner` already scans test directories, `Makefile` no longer needs to do so. Perhaps `TESTS` can be initialized with `*`

test/one/profiler/test/Filter.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.profiler.test;
+
+import java.util.regex.Pattern;
+
+public abstract class Filter {
```
apangin, line 10: It's now a single-method utility class; I'd just move the method to RunnerDeclaration.

test/one/profiler/test/Runner.java
```diff
@@ -210,24 +204,59 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+        // If directories are specified, they are considered an exact match.
+        List<String> includeDirs = new ArrayList<>();
+
+        // Glob filters matching "ClassName.methodName".
+        List<String> testFilters = new ArrayList<>();
+
+        for (int i = 0; i < args.length; i++) {
+            String arg = args[i];
+            File f = new File("test/test/" + arg);
+            if (!f.exists() || !f.isDirectory()) {
```
apangin, line 220: RunnerDeclaration already gets a list of all directories elsewhere. Let's use list (or better, set) here.

test/one/profiler/test/RunnerDeclaration.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private final List<Pattern> skipFilters;
+    private final Set<String> skipLooksLikeClassName;
+    private final Set<String> skipLooksLikeMethodName;
+
+    public RunnerDeclaration(List<String> includeDirs, List<String> globs, List<String> skipGlobs) {
+        this.includeDirs = new HashSet<>(includeDirs);
+        this.filters = globs.stream().map(Filter::from).collect(Collectors.toList());
+        this.looksLikeClassName = globs.stream().filter(f -> !f.contains(".") && !f.contains("*") && Character.isUpperCase(f.charAt(0))).collect(Collectors.toSet());
+        this.looksLikeMethodName = globs.stream().filter(f -> !f.contains(".") && !f.contains("*") && Character.isLowerCase(f.charAt(0))).collect(Collectors.toSet());
```
apangin, line 33: I think all these 4 structures can be replaced with a single list of patterns as follows:
1. glob is a test dir => `GlobTests.*`
2. glob looks like a class name => `Glob.*`
3. glob looks like a method name => `*.glob`
4. otherwise => `glob`

test/one/profiler/test/Runner.java
krk: We need to check directory existence to distinguish between `testFilters` and `includeDirs`.

Makefile
krk: It is initialized as empty now.

test/one/profiler/test/Runner.java
apangin: You get complete list of test directories when Runner starts, no need to do filesystem checks again. `includeDirs` will go away as per my other comment.
So, the algorithm I imagine works as follows:
1. Obtain a set of test directories by listing `test/test/`.
2. Parse test specifiers passed in command line arguments, converting each one to `Pattern`.
3. Similarly, parse `skip` definition to get a list of Patterns to be skipped.
4. Use reflection to find all runnable tests from directories obtained at (1).
5. Go through a sorted list of runnable tests and execute them if the name matches any pattern from (2) and does not match any pattern from (3).

Does this make sense?

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,148 @@
...
+        String className = "test." + dir + "." + Character.toUpperCase(dir.charAt(0)) + dir.substring(1) + "Tests";
+        try {
+            return getRunnableTests(Class.forName(className));
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private List<RunnableTest> getRunnableTests(List<String> globs) {
+        return globs.stream().filter(g -> g.equals(g.toLowerCase())).flatMap(g -> getRunnableTests(g).stream()).sorted(Comparator.comparing(RunnableTest::testName)).collect(Collectors.toList());
```
apangin, line 87: Not readable. Streams are usually formatted as follows:
```
return globs.stream()
        .filter(g -> g.equals(g.toLowerCase()))
        .flatMap(g -> getRunnableTests(g).stream())
        .sorted(Comparator.comparing(RunnableTest::testName))
        .collect(Collectors.toList());
```

Given that `List<String> globs` is always a list of directories here, `filter` can be removed.

Also, `getRunnableTests` method has too many overloads, I'd inline this one in `getRunnableTests()` below.

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,148 @@
...
+        return Pattern.compile("^" + Pattern.quote(s) + "$");
+    }
+
+    private static List<Pattern> filterFromGlobs(List<String> globs) {
+        Set<String> result = new HashSet<>();
+        for (String g : globs) {
+            if (!g.contains(".") && !g.contains("*")) {
+                // all lowercase, folder name.
+                if (g.equals(g.toLowerCase())) {
+                    result.add(g.substring(0, 1).toUpperCase() + g.substring(1).toLowerCase() + "Tests.*");
```
apangin, line 128: `continue;` ?

But I'd rewrite all this loop without `continue` using a plain one-level if-else chain - this will be easier to follow, IMO:
```
if (g.contains(".") || g.contains("*")) {
   ...
} else if (...) {
   ...
} else if ...
```

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,138 @@
...
+    private static Stream<RunnableTest> getRunnableTests(String dir) {
+        String className = "test." + dir + "." + Character.toUpperCase(dir.charAt(0)) + dir.substring(1) + "Tests";
+        try {
+            List<RunnableTest> rts = new ArrayList<>();
+            for (Method m : Class.forName(className).getMethods()) {
+                for (Test t : m.getAnnotationsByType(Test.class)) {
+                    rts.add(new RunnableTest(m, t));
+                }
+            }
+            return rts.stream();
```
apangin, line 76: nit: returning `List` here and calling `stream()` in the context of the stream below will look a bit nicer.

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
```
apangin, line 20: Unused import

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,137 @@
...
+        this.skipGlobs = filterFromGlobs(skipGlobs);
+
+        log.log(Level.FINE, "Test Directories: " + allDirs);
+        log.log(Level.FINE, "Test Filters: " + globs);
+        log.log(Level.FINE, "Skip Filters: " + skipGlobs);
+    }
+
+    public static TestDeclaration parse(String[] args) {
+        // All available directories.
+        List<String> testDirectories = new ArrayList<>();
```
apangin, line 41: Dead code

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,137 @@
...
+            .flatMap(g -> getRunnableTests(g).stream())
+            .sorted(Comparator.comparing(RunnableTest::testName))
+            .collect(Collectors.toList());
+    }
+
+    public boolean matches(Method m) {
+        String name = m.getDeclaringClass().getSimpleName() + '.' + m.getName();
+
+        if (includeGlobs.isEmpty() || includeGlobs.stream().anyMatch(f -> f.matcher(name).matches())) {
+            return skipGlobs.isEmpty() || !skipGlobs.stream().anyMatch(f -> f.matcher(name).matches());
```
apangin, line 93: !anyMatch -> noneMatch

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,137 @@
...
+        Set<String> result = new HashSet<>();
+        for (String g : globs) {
+            if (allDirs.contains(g)) {
+                // all lowercase, folder name.
+                result.add(g.substring(0, 1).toUpperCase() + g.substring(1).toLowerCase() + "Tests.*");
+            } else if (g.contains(".") || g.contains("*")) {
+                result.add(g);
+            } else if (Character.isUpperCase(g.charAt(0))) {
+                // Looks like class name.
+                result.add(g.substring(0, 1).toUpperCase() + g.substring(1).toLowerCase() + ".*");
```
apangin, line 127: What's the purpose of these case manipulations?

test/one/profiler/test/TestDeclaration.java
```diff
@@ -0,0 +1,137 @@
...
+            } else if (Character.isLowerCase(g.charAt(0))) {
+                // Looks like method name.
+                result.add("*." + g.toLowerCase());
```
apangin, line 130: Why do we convert it to lowercase? AFAICS, we use case sensitive patterns.

test/one/profiler/test/TestDeclaration.java
krk: It was a misunderstanding, correcting to `result.add(g + ".*");`.

test/one/profiler/test/TestDeclaration.java
krk: It was a misunderstanding, correcting to `result.add("*." + g);`.

krk:
Unrelated failure:

> INFO: Running WallTests.cpuWall...
FAIL [71/71] WallTests.cpuWall took 8.953 s
INFO: Test output and profiles are available in build/test/logs directory
java.lang.AssertionError
Error: Exception in thread "main" java.lang.RuntimeException: One or more tests failed
	at test.wall.WallTests.cpuWall(WallTests.java:27)
	at one.profiler.test.Runner.main(Runner.java:211)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
make: *** [test-java] Error 1
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at one.profiler.test.Runner.run(Runner.java:128)

PR #1130, opened 2025-02-07
Support posix_memalign and aligned_alloc in the native memory tracker

krk:
In addition to `malloc`, `calloc`, `realloc` and `free`; now `posix_memalign` and `aligned_alloc` are supported for tracking native memory allocations.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1121

### How has this been tested?
New nativemem test added with `posix_memalign` and `aligned_alloc` usage.

---


src/mallocTracer.cpp
```diff
@@ -73,6 +81,22 @@ extern "C" void free_hook(void* addr) {
...
+extern "C" int posix_memalign_hook(void** memptr, size_t alignment, size_t size) {
+    int ret = _orig_posix_memalign(memptr, alignment, size);
+    if (MallocTracer::running() && ret == 0 && memptr && *memptr && size) {
+        MallocTracer::recordMalloc(*memptr, size);
+    }
+    return ret;
+}
+
+extern "C" void* aligned_alloc_hook(size_t alignment, size_t size) {
+    void* ret = _orig_aligned_alloc(alignment, size);
```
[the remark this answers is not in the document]
krk: 1. `posix_memalign` triggers `posix_memalign_hook` as well as `aligned_alloc_hook`.
2. `aligned_alloc` only triggers `aligned_alloc_hook`.

Tested on Alpine 3.21:

```
profiler.Malloc {
  startTime = 17:22:36.503 (2025-02-07)
  address = 0x7FECBB4654D0
  size = 28.6 MB
  eventThread = "main" (osThreadId = 190)
  stackTrace = [
    libasyncProfiler.so.Profiler::getNativeTrace() line: 0
    libasyncProfiler.so.Profiler::recordSample() line: 0
    libasyncProfiler.so.MallocTracer::recordMalloc() line: 0
    libasyncProfiler.so.aligned_alloc_hook() line: 0
    ld-musl-x86_64.so.1.posix_memalign() line: 0
    ...
  ]
}

profiler.Malloc {
  startTime = 17:22:36.503 (2025-02-07)
  address = 0x7FECBB4654D0
  size = 28.6 MB
  eventThread = "main" (osThreadId = 190)
  stackTrace = [
    libasyncProfiler.so.Profiler::getNativeTrace() line: 0
    libasyncProfiler.so.Profiler::recordSample() line: 0
    libasyncProfiler.so.MallocTracer::recordMalloc() line: 0
    libasyncProfiler.so.posix_memalign_hook() line: 0
    main.main() line: 0
    ...
  ]
}
```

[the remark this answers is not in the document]
apangin: OK, thanks. This needs to be fixed then?

src/mallocTracer.cpp
```diff
@@ -109,6 +133,8 @@ void MallocTracer::patchLibraries() {
         cc->patchImport(im_calloc, (void*)calloc_hook);
         cc->patchImport(im_realloc, (void*)realloc_hook);
         cc->patchImport(im_free, (void*)free_hook);
+        cc->patchImport(im_posix_memalign, (void*)posix_memalign_hook);
+        cc->patchImport(im_aligned_alloc, (void*)aligned_alloc_hook);
```
apangin, line 137: Don't forget to add these to `MallocTracer::initialize()`

src/mallocTracer.cpp
```diff
@@ -29,17 +29,33 @@ static realloc_t _orig_realloc = NULL;
...
+typedef void* (*aligned_alloc_t)(size_t, size_t);
+static aligned_alloc_t _orig_aligned_alloc = NULL;
+
+// In musl, posix_memalign calls aligned_alloc and it in turns calls malloc.
+// If in_musl_posix_memalign, do not recordMalloc in aligned_alloc_hook.
+// If in_musl_aligned_alloc, do not recordMalloc in malloc_hook.
+// Do not access thread_local variables in non-musl.
+static bool is_musl = false;
+static thread_local bool in_musl_posix_memalign = false;
+static thread_local bool in_musl_aligned_alloc = false;
```
apangin, line 44: `thread_local` may allocate (using malloc!), so it's problematic to use `thread_local` in `malloc` hooks.
Why not just skip these hooks entirely on musl? Will `malloc_hook` catch these functions?

src/mallocTracer.cpp
```diff
@@ -91,9 +137,14 @@ void MallocTracer::initialize() {
...
                 || strcmp(s, "realloc_hook") == 0
-                || strcmp(s, "free_hook") == 0;
+                || strcmp(s, "free_hook") == 0
+                || strcmp(s, "posix_memalign_hook") == 0
+                || strcmp(s, "aligned_alloc_hook") == 0;
         },
         MARK_ASYNC_PROFILER);
+
+    // _CS_GNU_LIBC_VERSION is not defined on musl
+    is_musl = confstr(_CS_GNU_LIBC_VERSION, NULL, 0) == 0 && errno != 0;
```
apangin, line 147: To avoid code duplication, can we move this logic to os_linux.cpp and expose `OS::isMusl()`?

src/mallocTracer.cpp
krk: In musl, the next-gen [aligned_alloc implementation](https://git.musl-libc.org/cgit/musl/tree/src/malloc/mallocng/aligned_alloc.c?h=v1.2.2) no longer relies on `malloc`, unlike the current-gen version, which uses `malloc` [in certain cases](https://git.musl-libc.org/cgit/musl/tree/src/malloc/oldmalloc/aligned_alloc.c#n22).

[As of 2012](https://git.musl-libc.org/cgit/musl/commit/src/malloc/posix_memalign.c?id=9bff7c133e73ecfb200614d7a7d386a164a1a61f), `posix_memalign` has consistently called `aligned_alloc` instead of using `malloc`, so hooking `posix_memalign` in musl is unnecessary.

However, since `aligned_alloc` no longer necessarily calls `malloc`, it still requires hooking to ensure proper behavior.

apangin: Wait, new `aligned_alloc` [does call](https://git.musl-libc.org/cgit/musl/tree/src/malloc/mallocng/aligned_alloc.c#n24) `malloc`, am I wrong?

krk: Sorry, yes. I haven't seen this in jfr profiles of `aligned_alloc` calls. Need to check this again.

src/mallocTracer.cpp
```diff
@@ -29,17 +29,33 @@ static realloc_t _orig_realloc = NULL;
...
+
+typedef void* (*aligned_alloc_t)(size_t, size_t);
+static aligned_alloc_t _orig_aligned_alloc = NULL;
+
+// In musl, posix_memalign calls aligned_alloc and it in turns calls malloc.
+// If in_musl_posix_memalign, do not recordMalloc in aligned_alloc_hook.
+// If in_musl_aligned_alloc, do not recordMalloc in malloc_hook.
+// Do not access thread_local variables in non-musl.
+static bool is_musl = false;
+static thread_local bool in_musl_posix_memalign = false;
```
r1viollet, line 43: We hit several issues with deadlocks and thread locals.
Our solution was to control the [creation and deletion of the thread local](https://github.com/DataDog/ddprof/blob/cf9ec330672eb0cf597a89559ed5948f940ae416/src/lib/allocation_tracker.cc#L41) variables with get/set_specific.
That implies hooking thread creation to ensure you can create these states. Though I think you already have such hooks.
We also added guards to avoid re-entrance (this avoids double counting mentioned bellow).
I'm not sure if this is preferable to your approach, though feel free to compare.

krk: Thanks, we are looking into a simpler implementation that does *not* avoid double-accounting, yet still has correct leak detection.

PR #1135, opened 2025-02-14
Fix wrong addresses of original functions with LD_PRELOAD'ed allocator

apangin:
### Description

Fix nativemem crashes when LD_PRELOAD'ing libjemalloc.so and libasyncProfiler.so together.

### Motivation and context

dlsym(RTLD_NEXT) may return wrong address depending on the order of preloaded libraries.
However, since async-profiler library no longer exposes malloc functions itself, the correct option will be RTLD_DEFAULT.

### How has this been tested?

```
LD_PRELOAD="libjemalloc.so libasyncProfiler.so" ASPROF_COMMAND=start,nativemem java -version
LD_PRELOAD="libasyncProfiler.so libjemalloc.so" ASPROF_COMMAND=start,nativemem java -version
LD_PRELOAD="libasyncProfiler.so" ASPROF_COMMAND=start,nativemem java -version

+ all regular tests
```

---

### DIFF
--- a/src/hooks.cpp
+++ b/src/hooks.cpp
@@ -187,3 +187,5 @@ void Hooks::patchLibraries() {
         cc->patchImport(im_pthread_exit, (void*)pthread_exit_hook);
     }
 }
+
+#undef ADDRESS_OF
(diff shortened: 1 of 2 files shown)


apangin:
Unfortunately, `RTLD_DEFAULT` does not always work: it may return an address of PLT stub instead of a real function, and that will cause infinite recursion in the hook.

Issue #379, opened 2020-12-14
Generate color key for flame graph


alblue:
When a flame graph is generated with the -out flamegraph.html option, display a legend at the bottom that explains the colours used.

It might also be an idea to say whether or not the graph was generated from wall clock time, cpu time, allocation mode, or whether the display is for the total amount of time or number of samples.

That way, when the file is shared to another person, they immediately know what type of graph it is and what the colours mean.

apangin:
Flame Graph title is generated now depending on the `-e` and `--total` options.

E.g.

- when profiling `-e wall` => the title is "Wall clock profile"
- when profiling `-e alloc --total` => the title is "Allocation profile (bytes)"

Sarkie:
@apangin 
What about the colour legend? 
Green is Java, yellow is C++, orange is kernel, and red is the remainder (native user-level, or kernel modules).

https://brendangregg.com/FlameGraphs/cpuflamegraphs.html

apangin:
![colors](https://github.com/async-profiler/async-profiler/assets/1749416/d9f0ce1a-ae3d-4f99-91c6-c3a7293ec8be)

I'll add the legend in HTML later.

wash-amzn:
> Green is Java, yellow is C++

There are actually two greens (one for JIT'ed, one for interpreted) and two yellows (one for C1 compiled and one for C++). There's also blue for inlined.

The two greens and two yellows are pretty close to eachother, plus there some randomness added to each frame's color. As a result, it isn't that easy to tell the difference between each pair of cases.

apangin:
Added color legend in #1140.

shakuzen:
Thanks for adding the legend. That definitely helps. I still think the feedback given by @wash-amzn above is valid:

> The two greens and two yellows are pretty close to eachother, plus there some randomness added to each frame's color. As a result, it isn't that easy to tell the difference between each pair of cases.

Should a separate issue be opened to consider solutions to that?

apangin:
@shakuzen Colors for the different types of Java frames are intentionally picked from the same palette, otherwise flamegraphs would look too variegated and difficult to read. One idea is to add frame type in the tooltip (the tooltip already shows `int`, `c1`, `inln`, but only when the same frame combines multiple types). I'm open to other ideas.

shakuzen:
> One idea is to add frame type in the tooltip

That would help and sounds to me like a good idea regardless of the color issue.

weterock:
[...]

PR #1140, opened 2025-02-17
Add legend to flamegraphs

krk:
Add a legend to the flamegraphs, that is hidden by default.

image

### Related issues
https://github.com/async-profiler/async-profiler/issues/379

### Motivation and context
Colors in the flamegraph have different meanings, which can be learned over time. Instead, provide a legend.

### How has this been tested?
```
java -Djava.library.path=build/test/lib -agentpath:build/lib/libasyncProfiler.so=start,cpu,file=out.html -cp build/test test.cpu.CpuBurner
```

---

### DIFF
--- a/src/res/flame.html
+++ b/src/res/flame.html
@@ -15,12 +15,28 @@
 	#match {right: 0}
 	#reset {cursor: pointer}
 	#canvas {width: 100%; height: /*height:*/300px}
+	#legendDetail div {width: 200px; margin-left: 20px; padding: 2px;}
 </style>
 </head>
 <body style='font: 12px Verdana, sans-serif'>
 <h1>/*title:*/</h1>
-<header style='text-align: left'><button id='reverse' title='Reverse'>&#x1f53b;</button>&nbsp;&nbsp;<button id='search' title='Search'>&#x1f50d;</button></header>
+<header style='text-align: left'>
+	<button id='reverse' title='Reverse'>&#x1f53b;</button>
+	&nbsp;&nbsp;
+	<button id='search' title='Search'>&#x1f50d;</button>
+	&nbsp;&nbsp;
+	<button id='legend' title='Legend'>&#x1f6c8;</button>
+</header>
 <header style='text-align: right'>Produced by <a href='https://github.com/async-profiler/async-profiler'>async-profiler</a></header>
+<div id="legendDetail" style="display: none">Legend:
+	<div>interpreted</div>
+	<div>jitted</div>
+	<div>inlined</div>
+	<div>native</div>
+	<div>cpp</div>
+	<div>kernel</div>
+	<div>c1 compiled</div>
+</div>
 <canvas id='canvas'></canvas>
 <div id='hl'><span></span></div>
 <p id='status'></p>
@@ -33,6 +49,7 @@
 	let level0 = 0, left0 = 0, width0 = 0;
 	let nav = [], navIndex, matchval;
 	let reverse = /*reverse:*/false;
+	let legend = false;
 	const levels = Array(/*depth:*/0);
 	for (let h = 0; h < levels.length; h++) {
 		levels[h] = [];
@@ -64,6 +81,12 @@
 	function getColor(p) {
 		const v = Math.random();
 		return '#' + (p[0] + ((p[1] * v) << 16 | (p[2] * v) << 8 | (p[3] * v))).toString(16);
+	}
+
+	const legendDetail = document.getElementById('legendDetail');
+	const legendItems = legendDetail.getElementsByTagName('div');
+	for(let i = 0; i < palette.length; i++) {
+		legendItems[i].style.backgroundColor = getColor(palette[i]);
 	}
 
 	function f(key, level, left, width, inln, c1, int) {
@@ -256,6 +279,11 @@
 
 	document.getElementById('search').onclick = function() {
 		search(true);
+	}
+
+	document.getElementById('legend').onclick = function() {
+		legend = !legend;
+		legendDetail.style.display = legend ? "block" : "none";
 	}
 
 	document.getElementById('reset').onclick = function() {


src/res/flame.html
```diff
@@ -66,6 +83,12 @@ <h1>/*title:*/</h1>
 		return '#' + (p[0] + ((p[1] * v) << 16 | (p[2] * v) << 8 | (p[3] * v))).toString(16);
 	}
 
+	const legendDetail = document.getElementById('legendDetail');
+	const legendItems = legendDetail.getElementsByTagName('div');
+	for(let i = 0; i < palette.length; i++) {
+		legendItems[i].style.backgroundColor = getColor(palette[i]);
```
apangin, line 89: `getColor` returns random color within palette range. Should we use stable color for the legend?

src/res/flame.html
```diff
@@ -258,6 +281,11 @@ <h1>/*title:*/</h1>
 		search(true);
 	}
 
+	document.getElementById('legend').onclick = function() {
```
apangin, line 284: What if a legend appears automatically when hovering over? Won't it be a better UX?

src/res/flame.html
krk: Yes. Colors will not match exactly to what is in the flamegraph in any case, I don't mind either way.

src/res/flame.html
krk: I imagine accidental appearances would not be helpful. Other buttons in the same area work with a click, I think this is consistent.

simonis:
I love this and I'm really looking forward for this being integrated!

I have two suggestions though: please consider choosing more distinct colors for interpreted frames. Interpreted code is really a performance killer and should clearly stand out in a flame graph.

The second suggestion is about naming. We currently have "jitted" and "c1 compiled" but c1 compiled is also jitted. So consider renaming jitted into "c2 compiled".

Also, I'm not sure how AsyncProfiler is displaying JVMCI compiled frames? Are they currently shown as "jitted"? In that case, replacing "jitted" with "c2 compiled" might be a bad choice and "JIT highest tier" and "JIT low tier" might be a better alternative. To further complicate things, we can have C1, C2 and JVMCI compiled methods if we're running the Graal compiler in hosted mode (for Truffle only) but I'm not sure if C2 and JVMCI compiled methods will be distinguishable in that case?

simonis:
Some more comments/suggestions:
1. @apangin explained to me that only flamegraphs taken with `cstack=vm` have the correct JIT/Inlined/C1 information and coloring on all Java frames. This means that we will have to adapt the legend based on the `cstack` option because all the other flamegraphs only have the correct JIT/Inlined/C1 information only on top frames which are mostly invisible most of the time because of their size. Displaying the same legend on such graphs will be just confusing because it might show frames as JIT-compiled although they are in reality interpreted.
2. The Unicode character you choose for the button (i.e. `&#x1f6c8;`) doesn't display on Ubuntu 22.04, neither in Chrome nor in Firefox. Can we come up with something more portable?
3. I'm not an HTML/CSS pro, but I think it should be possible to make the legend a real pop-up menu which will be display over the existing graph and not shift the whole graph downwards.

simonis:
Thanks @krk, that looks pretty neat now. I like the pop-up version much more than the previous version. The new button now displays fine on Linux.

The only remaining nit-picking from me would be to change the label for "C++" frames from "C++" to something like "C++ (VM)" to make it clear that this is HotSpot code in contrast to all the other native code which might also be C++ code.

krk:
unrelated failure:

```
FAIL [62/63] JfrTests.ttsp took 4.114 s
java.lang.AssertionError
	at test.jfr.JfrTests.ttsp(JfrTests.java:93)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at one.profiler.test.Runner.run(Runner.java:133)
	at one.profiler.test.Runner.main(Runner.java:230)
INFO: Running JfrTests.ttspNostop...
PASS [63/63] JfrTests.ttspNostop took 4.260 s

Failed tests:
JfrTests.ttsp
```

src/res/flame.html
```diff
@@ -258,10 +318,28 @@ <h1>/*title:*/</h1>
...
+		legend = !legend;
```
marcelocenerine, line 322: This variable ^^^ isn’t reset to false when the popup closes, so the user must click the "legend" button twice to reopen it. Since the pop-up overlays the page, it’s probably ok to remove this flag altogether.

krk:
Updated based on latest offline discussions.

.assets/images/flamegraph.png
apangin, line 1: Image became 3830px wide - was it intentional change? IMO, this is too large.

src/res/flame.html
```diff
@@ -3,33 +3,204 @@
...
+	.right-section {
+		display: flex;
+		align-items: center;
+		width: 150px;
+	}
+	p {position: fixed; bottom: 0; margin: 0; padding: 1px 2px 1px 2px; outline: 1px solid #ffc000; display: none; overflow: hidden; white-space: nowrap; background-color: #ffffe0}
 	a {color: #0366d6}
 	#hl {position: absolute; display: none; overflow: hidden; white-space: nowrap; pointer-events: none; background-color: #ffffe0; outline: 1px solid #ffc000; height: 15px}
-	#hl span {padding: 0 3px 0 3px}
+	#hl span {padding: 0 2px 0 2px}
```
apangin: Why are all these changes in margins and paddings?
Now, frame title jumps whenever I hover a pointer over it.

src/res/flame.html
```diff
@@ -128,13 +366,22 @@ <h1>/*title:*/</h1>
 		}
 	}
 
-	function search(r) {
+	function search(r, resetZoom) {
```
apangin: How `resetZoom` is related to search?

src/res/flame.html
```diff
@@ -128,13 +366,22 @@ <h1>/*title:*/</h1>
...
 		if (r === true && (r = prompt('Enter regexp to search:', '')) === null) {
 			return;
 		}
 
 		pattern = r ? RegExp(r) : undefined;
+
+		if (resetZoom) {
+			root = origRoot;
```
apangin: What is `origRoot`? Can it be different from the `all` frame at level 0?

src/res/flame.html
```diff
@@ -221,14 +493,44 @@ <h1>/*title:*/</h1>
 				canvas.title = f.title + '\n(' + samples(f.width) + f.details + ', ' + pct(f.width, levels[0][0].width) + '%)';
 				canvas.style.cursor = 'pointer';
 				canvas.onclick = function() {
-					if (event.altKey && h >= root.level) {
+					if (event.altKey && h > root.level) {
```
apangin: OK, "all" frame should never be removed, but otherwise what's wrong with deleting stack at current level?

src/res/flame.html
```diff
@@ -262,26 +559,76 @@ <h1>/*title:*/</h1>
...
+					event.preventDefault();
+					search(false, true);
+					break;
+			}
 		}
 	}
 
+	function invertMode() {
+		inverted = !inverted;
+		btnInvert.style.rotate = inverted ? '180deg' : '0deg';
```
apangin: nit: the icon is _rotated_, but the graph is _mirrored_, not the same thing :)

apangin: But seriously, I'd prefer a static icon (like up-down triangles or whatever you invent). Two-state toggle buttons are always confusing: whether they show current state or a state to switch to.

.assets/images/flamegraph.png
krk: High dpi screenshot, fixing.

src/res/flame.html
krk: Reverting these paddings, doesn't fix the jumping. I can see that behavior on master as well:

jumping-frame-title

src/res/flame.html
krk: `search` is used as the rendering method. In any case, removing the resetZoom argument.

src/res/flame.html
krk: Fixing by disallowing removal of `all` only.

src/res/flame.html
krk: Changed to a static icon.

src/res/flame.html
krk: `origRoot` is the root of the first `search` call, which should be the `all` frame. Renaming to `allRoot`.

Using `levels[0][0]` now.

src/res/flame.html
apangin: Hm.. might depend on the OS/browser. It does not jump in Chrome.

.assets/images/flamegraph_legend.png
apangin, line 1: I don't think the image in docs is worth replacing with a roughly cropped screenshot, especially since it duplicates information already available on a flamegraph and includes info about shortcuts which is irrelevant here.

src/res/flame.html
```diff
@@ -15,16 +46,156 @@
...
+
+		<div id="invert" class="toolbarIcon" title="Invert Graph [I]"><svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path d="M10 16 L6 12 L14 12 Z" fill="#ff8d40"/><line x1="4" y1="10" x2="16" y2="10" stroke="#004d80" stroke-width="2"/><path d="M10 4 L6 8 L14 8 Z" fill="#004d80" opacity="0.7"/><path d="M2 8 L2 12" stroke="#004d80" stroke-width="1.5" stroke-linecap="round"/><path d="M18 12 L18 8" stroke="#004d80" stroke-width="1.5" stroke-linecap="round"/></svg></div>
```
apangin: This icon looks more like a splitter. IMO it will be better without side vertical lines and with a bit larger triangles.

src/res/flame.html
```diff
@@ -15,16 +46,156 @@
...
-<header style='text-align: left'><button id='inverted' title='Invert'>&#x1f53b;</button>&nbsp;&nbsp;<button id='search' title='Search'>&#x1f50d;</button></header>
-<header style='text-align: right'>Produced by <a href='https://github.com/async-profiler/async-profiler'>async-profiler</a></header>
-<canvas id='canvas'></canvas>
-<div id='hl'><span></span></div>
-<p id='status'></p>
-<p id='match'>Matched: <span id='matchval'></span> <span id='reset' title='Clear'>&#x274c;</span></p>
+
+<div class="title-toolbar-container">
+	<div class="left-toolbar">
+		<div id="graph-info" class="toolbarIcon"><svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><circle cx="10" cy="10" r="8" stroke="#004d80" stroke-width="1" fill="none"/><path d="M10 5.5C8.75 5.5 7.75 6.5 7.75 7.75H9C9 7.06 9.56 6.5 10.25 6.5C10.94 6.5 11.5 7.06 11.5 7.75C11.5 8.4 10.95 8.75 10.5 8.95C9.8 9.3 9.25 9.8 9.25 10.75V11H10.75V10.75C10.75 10.38 11.04 10.1 11.43 9.92C12.16 9.58 12.75 9.05 12.75 7.75C12.75 6.5 11.25 5.5 10 5.5Z" fill="#ff8d40" stroke="#ff8d40" stroke-width="0.6" stroke-linecap="round" stroke-linejoin="round"/><circle cx="10" cy="13.5" r="1" fill="#ff8d40"/></svg></div>
```
apangin: WDYT about changing order of icons to match previous layout? E.g., Invert button first, then Search, and Help in the end.

src/res/flame.html
```diff
@@ -51,24 +230,83 @@ <h1>/*title:*/</h1>
 	if (devicePixelRatio) c.scale(devicePixelRatio, devicePixelRatio);
 	c.font = document.body.style.font;
 
+	function positionTooltip(e) {
```
apangin: Does it even need to float? Static position will be more than enough. It should make code simpler.

src/res/flame.html
```diff
@@ -262,26 +556,74 @@ <h1>/*title:*/</h1>
 		search(false);
 	}
 
+	btnInvert.onclick = function(e) {
+		invertMode();
+		e.preventDefault();
```
apangin: Why is it needed? Is there a default behavior for onclick at all?

src/res/flame.html
```diff
@@ -262,26 +556,74 @@ <h1>/*title:*/</h1>
...
+	}
+
+	btnInvert.dblclick = function(e) {
+		e.preventDefault();
+	}
+
+	btnInvert.onmousedown = function(e) {
+		e.preventDefault();
+		e.stopPropagation();
```
apangin: This repeats in several places. Doesn't `return false` have the same effect?

src/res/flame.html
```diff
@@ -262,26 +556,74 @@ <h1>/*title:*/</h1>
...
+		if (event.key === 'Escape') {
 			search(false);
-		} else if ((event.key === 'n' || event.key === 'N') && nav.length > 0) {
-			navIndex = (navIndex + (event.shiftKey ? nav.length - 1 : 1)) % nav.length;
-			render(nav[navIndex]);
-			document.getElementById('matchval').textContent = matchval + ' (' + (navIndex + 1) + ' of ' + nav.length + ')';
-			window.scroll(0, inverted ? root.level * 16 : canvasHeight - (root.level + 1) * 16);
-			canvas.onmousemove();
+		} else if(event.ctrlKey || event.metaKey) {
+			switch(event.key) {
```
apangin: Rewriting the function with two `switch` statements did not make it any cleaner or more readable. On the contrary, this excessively increased nesting level. IMO, switches do not make sense here, especially with one label only.

src/res/flame.html
krk: "Need" is a strong word for a UI choice. Made it static anyway.

src/res/flame.html
krk: This was to help with Chrome selecting title on double-click, doing some cleanup.

src/res/flame.html
krk: There used to be more cases. switch makes it clearer as opposed to composite if-conditions. Converting to `if` chain.

apangin:
After pressing `0`:

apangin:
It's very unobvious what "Highlight" and "Hide frame" menu items do when popup menu is opened on an empty space of a flamegraph.

I also realized that I feel frustrated when a webpage does not allow me to open the default popup menu by replacing it with a custom popup. Highlighting frame type is a funny feature, but I'm not sure its value outweighs the price

krk:
You could do Shift + Right Click to get the browser's context menu. Additionally, I will fix so only frames will have the custom context menu.

krk:
> After pressing 0:

Fixed too.

krk:
Context menu feature removed based on offline discussions.

krk:
Unrelated test failure:

```
FAIL [105/119] RecoveryTests.suppliers took 8.972 s
java.lang.AssertionError: Expected 0.002007024586051179 < 0.002
	>  test/test/recovery/RecoveryTests.java:80
	>  Assert.isLess(out.ratio("unknown|break_compiled"), 0.002);
	at one.profiler.test.Assert.assertComparison(Assert.java:39)
	at one.profiler.test.Assert.isLess(Assert.java:76)
	at test.recovery.RecoveryTests.suppliers(RecoveryTests.java:80)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at one.profiler.test.Runner.run(Runner.java:124)
```

.assets/html/flamegraph.html
```diff
@@ -3,10 +3,41 @@
...
+		border: none;
+		border-radius: 0;
+	}
+	.title-section {
+		display: flex;
+		flex-direction: column;
+		align-items: center;
+		flex-grow: 1;
+	}
+	.toolbar-section {
```
apangin: Unused?

.assets/html/flamegraph.html
```diff
@@ -15,16 +46,123 @@
...
+	}
+	.left {
+		border-left: .1px solid #000
+	}
+	.bottom {
+		border-bottom: .1px solid #000
+	}
+	.right {
+		border-right: .1px solid #000
+	}
```
apangin: I don't see usages of these classes.

.assets/html/flamegraph.html
```diff
@@ -15,16 +46,123 @@
...
+	.toolbarIcon {
+		width: 24px;
+		height: 24px;
+		margin: 0 2px;
+		display: flex;
+		justify-content: center;
+		align-items: center;
+		cursor: pointer;
+	}
+	.toolbarSelected {
```
apangin: Unused?

.assets/html/flamegraph.html
```diff
@@ -15,16 +46,123 @@
...
+	.left {
+		border-left: .1px solid #000
+	}
+	.bottom {
+		border-bottom: .1px solid #000
+	}
+	.right {
+		border-right: .1px solid #000
+	}
+	.toolbarIcon {
```
apangin: Inconsistent naming style: some classes use  camelCase, but some use kebab-case.

.assets/html/flamegraph.html
```diff
@@ -15,16 +46,123 @@
...
+				Click frame - Zoom in frame<br>
+				Alt + click frame - Remove frame<br>
+				0 - Reset zoom<br>
+				<br>
+				Ctrl + F - Search<br>
+				N - Focus on next match<br>
+				Shift + N - Focus on previous match<br>
+				Esc - Cancel search<br>
+				<br>
+				I - invert graph<br>
```
apangin: Invert (uppercase)

.assets/html/flamegraph.html
```diff
@@ -15,16 +46,123 @@
...
+	<div class="title-section">
+		<h1>CPU profile</h1>
+	</div>
+
+	<div class="right-section">
+		<header>Produced by <a href='https://github.com/async-profiler/async-profiler'>async-profiler</a></header>
+	</div>
+</div>
+
+<div class='bordered' style="display: flex; flex-direction: column;">
```
apangin: What are these styles for?

.assets/html/flamegraph.html
```diff
@@ -262,19 +407,54 @@ <h1>CPU profile</h1>
...
+		infoTooltip.style.display = 'none';
+	};
+
+	btnInfo.onmousedown = function() {
+		return false;
+	};
+
+	btnInfo.ondblclick = function(e) {
+		return false;
+	}
```
apangin, line 412: Somewhere `function()` and somewhere `function(e)`

.assets/html/flamegraph.html
```diff
@@ -262,19 +407,54 @@ <h1>CPU profile</h1>
...
-		if ((event.ctrlKey || event.metaKey) && event.key === 'f') {
-			event.preventDefault();
-			search(true);
-		} else if (event.key === 'Escape') {
+		if (event.key === 'Escape') {
 			search(false);
-		} else if ((event.key === 'n' || event.key === 'N') && nav.length > 0) {
+		} else if((event.ctrlKey || event.metaKey) && event.key === 'f') {
+			search(true);
+		} else if (event.key === 'n' && nav.length > 0) {
```
apangin: `event.key === 'N'` was essential for `Shift+N` to work.

.assets/html/flamegraph.html
```diff
@@ -262,19 +407,54 @@ <h1>CPU profile</h1>
...
 			render(nav[navIndex]);
 			document.getElementById('matchval').textContent = matchval + ' (' + (navIndex + 1) + ' of ' + nav.length + ')';
 			window.scroll(0, inverted ? root.level * 16 : canvasHeight - (root.level + 1) * 16);
 			canvas.onmousemove();
+		} else if (event.key === 'i') {
+			invertMode();
+		} else if (event.key === '0') {
+			resetZoom();
 		}
+		return false;
```
apangin: Why is that? Standard combinations like reload (Ctrl+R), Zoom (Ctrl+, Ctrl-) etc. no longer work.

.assets/images/flamegraph.png
apangin, line 1: How come new image is smaller but its size on disk is 240% bigger?

apangin:
There is something wrong with the layout of "Produced by async-profiler" text (Latest Chrome, Windows 11):

layout

apangin:
I did some modifications: https://profiler.tools/tmp/flamegraph-legend.html
What do you think of this version?

.assets/html/flamegraph.html
krk: Moved `return false` to only handled combinations. For example when Ctrl+F is pressed, search popup is shown. Without `return false`, browser search is also activated.

.assets/images/flamegraph.png
krk: Took a new screenshot and ran it through optipng. Previous image looks like it was scaled up. If I have to hypothesize, I could say previous image had less entropy because of the scaling, hence more compressible.

krk:
> There is something wrong with the layout of "Produced by async-profiler" text (Latest Chrome, Windows 11):

Cannot repro this on latest Chrome + Ubuntu 24. Added `text-wrap-mode: nowrap` to `.right-section` in any case.

apangin:
Integrated with some modifications.

Issue #1125, opened 2025-02-05
Parsing concurrently loaded libraries on Linux is error-prone


apangin:
### Describe the bug

The bug was originally noticed when testing `nativemem` with GraalVM.
Sometimes, when two shared libraries are opened concurrently in different threads, async-profiler sees unadjusted offset (instead of a relocated address) when parsing relocation entries in `parseDynamicSection()`.

### Expected vs. actual behavior

JVM may crash when profiler is started with `nativemem` option.

### Reproduction Steps

```
while true; do
  java -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCIProduct -Djava.library.path=build/test/lib -agentpath:build/lib/libasyncProfiler.so=start,nativemem=1,total,file=out.jfr -cp build/test test.nativemem.CallsAllNoLeak once
done
```

### Async-profiler version

master

### Environment details

- Linux aarch64, glibc
- OpenJDK Runtime Environment GraalVM CE 21.0.5-dev+14.1 (build 21.0.5+14-LTS)

krk:
After running over 40,000 iterations on both `x86_64` and `aarch64` with `GraalVM 21.0.5+9.1` and `21.0.6+8.1`, no crashes were observed. So far, there is no reproducible scenario for this issue.

apangin:
A scenario leading to a crash can be the following:
1. Thread 1 starts loading library `A`.
2. `Profiler::dlopen_hook` is called.
3. `dlopen(A)` completes successfully.
4. Thread 2 starts loading library `B`.
5. `B` is mapped into memory, but not yet relocated. `dlopen(B)` still runs.
6. Thread 1 calls `Profiler::updateSymbols`.
7. It finds library `B` in `/proc/self/maps` and starts parsing it.
8. Since `B` has not been yet relocated, an address for `malloc` in B's GOT is invalid.

Why (5) and (7) can run concurrently? The problem is that `dlopen` acquires `dl_load_lock`, while `dl_iterate_phdr` acquires `dl_load_write_lock`. These are different locks. `dl_iterate_phdr` trick helps to protect from concurrent *unloading*, but not from concurrent *loading*.

To fix this, we should parse only libraries that are known to be completely loaded. We can again appeal to `dl_iterate_phdr` to get a list of such libraries.

krk:
Thanks for the scenario.

I can repro this with Graal CE on `aarch64`, after bisecting I found these commits:

[...]

Issue #689, opened 2022-11-29
Broken symbols when ELF `p_offset != p_vaddr` in the executable segment


markj-db:
`async-profiler` finds the start of the ELF file by [jumping back `p_offset` bytes](https://github.com/jvm-profiling-tools/async-profiler/blob/master/src/symbols_linux.cpp#L608) from the address where it finds the executable segment in `/proc/self/maps`.

Since https://reviews.llvm.org/D64903, `lld` produces libraries (by default) where the start of the executable segment is not necessarily aligned to a page boundary.  As a consequence, it produces a layout where `p_offset <= p_vaddr`.  (IIUC, ELF files with this property can also be generated under other circumstances, it just happens to be this specific trigger is the root cause of my own issues.)

When `p_offset != p_vaddr`, `async-profiler` looks for program headers in the wrong location and fails to load the DWARF data.  It also supplies the wrong base address when loading symbols, causing symbol resolution to be wrong (e.g. nonsense call stacks in native code -- that is, call stacks that include caller-callee relationship that don't exist in practice). Specifically, the `image_base` passed to `ElfParser::parseProgramHeaders` and `ElfParser::parseFile` is off by `p_vaddr - p_offset` bytes.

As an example, if I have a library with the following layout:

```
$ readelf -Wl libfoo.so
...
Entry point 0x64dd180
...
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
...
  LOAD           0x64dc180 0x00000000064dd180 0x00000000064dd180 0x6813120 0x6813120 R E 0x1000
...
```

This library will map into memory (for example) as:

```
# snippet from /proc/self/maps
...
7f7562d7d000-7f756925a000 r--p 00000000 103:01 4162946                   /libfoo.so
7f756925a000-7f756fa6e000 r-xp 064dc000 103:01 4162946                   /libfoo.so
...
```

`async-profiler` looks for an ELF header at `0x7f756925a000 - 0x64dc000 = 0x7f7562d7e000`, which is `0x1000` ahead of where the actual ELF header is located.

I confirmed that I can hack around this by modifying `Symbols::parseLibraries` to do the following:

 1. look for an ELF header at `image_base -= map.offs();`
 2. if the ELF header is not found, try looking a second time at `image_base -= 0x1000;`

For my sample library `libfoo.so` above, condition 1 fails and condition 2 finds the ELF header.

[...]

PR #1156, opened 2025-02-28
Use dlopen/dlclose instead of dl_iterate_phdr.

krk:
### Description

Prevents crashes when parsing libraries that are visible in maps but not fully relocated.

In symbol parsing, replace `dl_iterate_phdr` with a two-step approach: scan `/proc/self/maps` to discover libraries, then explicitly `dlopen`/`dlclose` each library before parsing its symbols.

Using `dlopen`/`dlclose` ensures the library cannot be unloaded in between and is loaded and have its relocations applied.

### Related issues

https://github.com/async-profiler/async-profiler/issues/1125

### Motivation and context

There's a critical race condition in our previous approach: libraries appear in `/proc/self/maps` as soon as their memory is mapped, but the dynamic linker may not have applied relocations yet.

When scanning directly from maps, we might attempt to parse a library that's not ready, causing crashes. 

By using `dlopen` after discovering libraries in maps, we ensure:

* The library is fully initialized by the dynamic linker before we attempt parsing.
* All relocations have been applied, making symbol information reliable.

---


src/symbols_linux.cpp
```diff
@@ -739,35 +741,20 @@ static int parseLibrariesCallback(struct dl_phdr_info* info, size_t size, void*
...
-            } else if ((unsigned long)map_start > map_offs) {
-                // Unlikely case when image_base has not been found.
-                // Be careful: executable file is not always ELF, e.g. classes.jsa
-                ElfParser::parseFile(cc, map_start - map_offs, map.file(), true);
-            }
-        } else if (strcmp(map.file(), "[vdso]") == 0) {
-            ElfParser::parseProgramHeaders(cc, map_start, map_end, true);
-        }
+        // Create CodeCache, do not parse the file yet.
+        cc = new CodeCache(map.file(), count, false, map_start, map_end, map_offs, imageBaseStatus);
```
apangin, line 750: Adding unparsed CodeCache is dangerous: other parts of the code (e.g. `patchLibraries`) assume that if a library presents in CodeCacheArray it is already parsed.

src/symbols_linux.cpp
```diff
@@ -785,10 +772,53 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
-    // Without it, we may access memory of a library that is being unloaded.
-    dl_iterate_phdr(parseLibrariesCallback, array);
+    // Parse vm maps to find bounds.
+    parseMaps(array);
+
+    int lib_count = array->count();
+    for (int i = 0; i < lib_count; i++) {
+        // We only parse libraries that were detected in parseMaps as loaded.
+        CodeCache* cc = (*array)[i];
+        const char* image_base = (const char*)((const char*)cc->minAddress() - (const char*)cc->mapOffset());
```
apangin, line 782: `image_base` is not always equal to `min_address - map_offset`. See #689.

src/symbols_linux.cpp
```diff
@@ -785,10 +772,53 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+
+        if (strchr(cc->name(), ':') != NULL) {
+            // Do not try to parse pseudofiles like anon_inode:name, /memfd:name
+        } else if (strcmp(cc->name(), "[vdso]") == 0) {
+            ElfParser::parseProgramHeaders(cc, (const char*)cc->minAddress(), (const char*)cc->maxAddress(), true);
+        } else {
+            // dlopen while parsing to prevent unloading the library and to make sure it is done loading.
+            // dlopen may fail opening the main executable and dynamic linker, which we ignore.
+            // Using dlopen instead of dlopen_no_hook may cause recursion.
+            void* handle = dlopen_no_hook(cc->name(), RTLD_LAZY);
```
apangin, line 797: `RTLD_NOLOAD` is essential here. Otherwise there is a chance of a race when we load a library that has been just unloaded, and that may have unpredictable consequences (execution of an arbitrary code that we don't control).

apangin: Also, if a library is re-loaded, its new base address may differ from a previous base address.

src/symbols_linux.cpp
```diff
@@ -780,16 +767,65 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+            continue;
+        }
+
+        if (strchr(cc->name(), ':') != NULL) {
+            // Do not try to parse pseudofiles like anon_inode:name, /memfd:name
+        } else if (strcmp(cc->name(), "[vdso]") == 0) {
+            ElfParser::parseProgramHeaders(cc, (const char*)cc->minAddress(), (const char*)cc->maxAddress(), true);
+        } else {
+            // dlopen while parsing to prevent unloading the library and to make sure it is done loading.
+            // dlopen may fail opening the main executable and dynamic linker, which we ignore.
```
apangin, line 797: But we can't ignore failures for other libraries, otherwise we may attempt to parse a library that has been just unloaded.
I'll fix this too as I'm refactoring this code anyway.

Issue #1161, opened 2025-03-06
CpuTests.itimerTotal test times out on mac


parttimenerd:
### Describe the bug

The tests fail on Mac because `p.profile("-d 2 -e itimer -i 1us --total -o collapsed")` frequently runs longer than the 10-second timeout.

We observed the same effect on our internal test queue.

### Expected vs. actual behavior

It should not time out, but it does. A sample for the runtime distribution on a M4 Pro MacBook: 

![Image](https://github.com/user-attachments/assets/3faa9064-5e12-40d7-b18e-caba27127c56)

Other runs observed runtimes up to 273 seconds.

### Reproduction Steps

The following should produce failing tests regularly:
```sh
while true; do make JAVAC_TARGET=8 TESTS=test.cpu.CpuTests test; done
```

### Additional Information/Context

The behaviour can be observed on 

### Async-profiler version

master-branch at the time of writing

### Environment details

MacBook Pro M4, JDKs 21.0.6 and 23 (all versions)

apangin:
Thank you for the report. I also saw this test failing occasionally on macOS. This is a peculiarity of macOS handling itimer signals.
Let me see if there is a simple workaround. If not, I'll just remove this test.

parttimenerd:
Not executing the test part with the tiny profiling interval on Mac OS could be a good solution.

apangin:
I just removed `1us` case in `itimerTotal` test, which is not very useful anyway.

PR #1163, opened 2025-03-06
Check that reading integer from the pos is within limit

tedyu:
### Description
In `readEvent` method, we attempt to read an integer without checking that there are 4 bytes available from `pos`.

This PR adds the check so that we don't get exception reading the integer.

### Motivation and context
I was reading `readEvent` method and found the check missing.

### How has this been tested?

---

### DIFF
--- a/src/converter/one/jfr/JfrReader.java
+++ b/src/converter/one/jfr/JfrReader.java
@@ -158,7 +158,7 @@ public class JfrReader implements Closeable {
             int size = getVarint();
             int type = getVarint();
 
-            if (type == 'L' && buf.getInt(pos) == CHUNK_SIGNATURE) {
+            if (pos+4 <= buf.limit() && type == 'L' && buf.getInt(pos) == CHUNK_SIGNATURE) {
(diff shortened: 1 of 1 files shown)


apangin:
Could you please elaborate on what the _original_ problem you are trying to solve with this patch?
Did you see an exception while parsing a real JFR? What was that JFR?

tedyu:
Thanks for the quick response.

This PR came from static code analysis. If `pos` is close to the limit, reading an integer may produce error.

apangin:
Thank you for the clarification.
We _do_ run static analyzers and fix high severity issues (there were only a couple of them so far). We do not take findings of a static analyzer for granted without proper analysis. If you could point out a real use case when JfrReader fails to parse a valid recording, that would be helpful.

apangin:
`itimerTotal` test failure has been fixed in #1161.

apangin:
I'm closing the PR in the lack of justification for the change.
Feel free to reopen if you have a scenario that demonstrates the problem.

Issue #1164, opened 2025-03-07
Can't profile Cargo with perf_event_paranoid < 2


olivergillespie:
### Describe the bug

When I try to profile `cargo`, the process dies immediately if I have `perf_event_paranoid` less than 2.

```
$ sudo sysctl kernel.perf_event_paranoid=2; sudo sysctl kernel.kptr_restrict=0

LD_PRELOAD=/tmp/libasyncProfiler.so ASPROF_COMMAND=start cargo --version
Profiling started
Profiling started
cargo 1.84.0 (66221abde 2024-11-19)
Profiling stopped after 0 seconds. No dump options specified

$ sudo sysctl kernel.perf_event_paranoid=1

$ LD_PRELOAD=/tmp/libasyncProfiler.so ASPROF_COMMAND=start cargo --version
Profiling started
[1]    22803 profile signal  LD_PRELOAD=/tmp/libasyncProfiler.so
```

With strace I see:

```
--- SIGPROF {si_signo=SIGPROF, si_code=0x6, si_pid=24, si_uid=0, si_value={int=5, ptr=0x5}} ---
+++ killed by SIGPROF +++
[1]    24406 profile signal  LD_PRELOAD=/tmp/libasyncProfiler.so
```

Tested on one x86 host, one aarch64 host, both using Amazon Linux 2 with kernel 5.10, and both async-profiler 3.0 and nightly builds.

It's an easy workaround for me to just profile the built binary directly (I was just profiling my rust program through `cargo run`), but raising this in case there's anything you want to look into. Feel free to close if you suspect it's an issue with cargo itself.

### Async-profiler version

3.0

apangin:
Thank you for the report.
Note two "Profiling started" messages. Apparently, `cargo` binary redirects to another executable by calling `execve`. New process inherits perf descriptors from the parent, but the child process is not prepared to handle perf signals.
A fix is to close perf descriptors on exec. Passing `PERF_FLAG_FD_CLOEXEC` to `perf_event_open` should help.

apangin:
Resolved by #1165.

PR #1165, opened 2025-03-07
Use PERF_FLAG_FD_CLOEXEC with perf_event_open

olivergillespie:
Use PERF_FLAG_FD_CLOEXEC with perf_event_open

If the profiled process spawns another process with execve, the child process may be signalled before the signal handler is installed, which will terminate the process.

This flag was introduced in kernel 3.14, so we just don't use it at runtime in that case.

We don't really need the flag in the check/supported calls, but since that's what we'll be using for the real call it may be helpful for catching unexpected errors.

### Related issues

#1164 

### How has this been tested?

Confirmed that this fixes the issue on my kernel 5.10 host with `LD_PRELOAD=libasyncProfiler.so ASPROF_COMMAND=start cargo --version` (when installed via `rustup`, `cargo` has an indirection which uses `execve` and triggers this issue).

I have not yet tested on other kernels, especially building and running on <3.14, since I don't have a 3.14 host at the moment. I can do that next week, if you don't also have convenient access.

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -45,6 +45,10 @@ struct f_owner_ex {
 };
 #endif // F_SETOWN_EX
 
+// Introduced in kernel 3.14
+#ifndef PERF_FLAG_FD_CLOEXEC
+#define PERF_FLAG_FD_CLOEXEC  8
+#endif // PERF_FLAG_FD_CLOEXEC
 
 enum {
     HW_BREAKPOINT_R  = 1,
@@ -150,6 +154,20 @@ static void adjustFDLimit() {
     }
 }
 
+static bool is_kernel_gte_3_14() {
+    struct utsname uts;
+
+    if (uname(&uts) != 0) {
+        return false;
+    }
+
+    int major, minor;
+    if (sscanf(uts.release, "%d.%d", &major, &minor) != 2) {
+        return false;
+    }
+
+    return (major > 3 || (major == 3 && minor >= 14));
+}
 
 struct FunctionWithCounter {
     const char* name;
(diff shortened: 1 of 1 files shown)


src/perfEvents_linux.cpp
```diff
@@ -150,6 +154,24 @@ static void adjustFDLimit() {
...
+    if (sscanf(uts.release, "%d.%d", &major, &minor) != 2) {
+        return false;
+    }
+
+    return (major > 3 || (major == 3 && minor >= 14));
+}
+
+// Useful to close perf_event fds on exec to avoid issues in child processes.
+// Skip if not supported.
+static int CLOEXEC_FLAG = is_kernel_gte_3_14() ? PERF_FLAG_FD_CLOEXEC : 0;
```
apangin, line 174: I tested perf_events on CentOS 7.9 (Linux 3.10.0-1160.119.1.el7.x86_64) and found that `PERF_FLAG_FD_CLOEXEC` is actually supported. It's not uncommon that some features/options are backported to older kernels.
This means, relying solely on kernel major.minor version is not very accurate. Let's first call `perf_event_open` with the flag, and if the call fails, retry it without the flag. That's also what [perf](https://github.com/torvalds/linux/blob/2a520073e74fbb956b5564818fc5529dcc7e9f0e/tools/perf/util/cloexec.c#L59) and [other profilers](https://github.com/elastic/go-perf/blob/master/perf.go#L210-L211) do.

For the sake of simplicity, `check()` and `supported()` may call `perf_event_open` just once without `flags == 0`.

olivergillespie: Thanks, updated. Would you like me to use fcntl directly on the fd if PERF_FLAG_FD_CLOEXEC is not supported, or just ignore that case?
I also didn't cache the supported/not supported decision - it should be widely supported and if not it just means we'll double the number of perf_event_open calls, which doesn't seem too bad, but let me know if you'd prefer to cache it.

olivergillespie: Thanks!

PR #1167, opened 2025-03-10
Log when no samples are collected

visheshruparelia:
### Description
Adding a log message when there are no samples collected by the profiler.

### Related issues
N/A

### Motivation and context
During some UT failure investigations earlier, I noticed that the issue was because there were no samples being collected. It was not obvious by looking just at the Test output that this was the case. Thus, adding this log line to provide insight for any future investigations.

### How has this been tested?
The CI builds successfully pass thus ensuring that this is not a breaking change.

---

### DIFF
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -133,7 +133,6 @@ void CallTraceStorage::collectTraces(std::map<u32, CallTrace*>& map) {
 }
 
 void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
-    size_t initialSize = samples.size();
     for (LongHashTable* table = _current_table; table != NULL; table = table->prev()) {
         u64* keys = table->keys();
         CallTraceSample* values = table->values();
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -145,13 +145,9 @@ void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
             }
         }
     }
-    if (samples.size() == initialSize) {
-        Log::info("No samples collected");
-    }
 }
 
 void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
-    bool samplesCollected = false;
     for (LongHashTable* table = _current_table; table != NULL; table = table->prev()) {
         u64* keys = table->keys();
         CallTraceSample* values = table->values();
@@ -160,12 +156,8 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
         for (u32 slot = 0; slot < capacity; slot++) {
             if (keys[slot] != 0 && values[slot].acquireTrace() != NULL) {
                 map[keys[slot]] += values[slot];
-                samplesCollected = true;
             }
         }
-    }
-    if (!samplesCollected) {
-        Log::info("No samples collected");
     }
 }
 
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -133,6 +133,7 @@ void CallTraceStorage::collectTraces(std::map<u32, CallTrace*>& map) {
 }
 
 void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
+    size_t initialSize = samples.size();
     for (LongHashTable* table = _current_table; table != NULL; table = table->prev()) {
         u64* keys = table->keys();
         CallTraceSample* values = table->values();
@@ -143,6 +144,9 @@ void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
                 samples.push_back(&values[slot]);
             }
         }
+    }
+    if (samples.size() == initialSize) {
+        Log::warn("No samples collected");
     }
 }
 
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -146,7 +146,7 @@ void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
         }
     }
     if (samples.size() == initialSize) {
-        Log::warn("No samples collected");
+        Log::info("No samples collected");
     }
 }
 
@@ -165,7 +165,7 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
         }
     }
     if (!samplesCollected) {
-        Log::warn("No samples collected");
+        Log::info("No samples collected");
     }
 }
 
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -147,6 +147,7 @@ void CallTraceStorage::collectSamples(std::vector<CallTraceSample*>& samples) {
 }
 
 void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
+    bool samplesCollected = false;
     for (LongHashTable* table = _current_table; table != NULL; table = table->prev()) {
         u64* keys = table->keys();
         CallTraceSample* values = table->values();
@@ -155,8 +156,12 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
         for (u32 slot = 0; slot < capacity; slot++) {
             if (keys[slot] != 0 && values[slot].acquireTrace() != NULL) {
                 map[keys[slot]] += values[slot];
+                samplesCollected = true;
             }
         }
+    }
+    if (!samplesCollected) {
+        Log::warn("No samples collected");
     }
 }
 
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -612,6 +612,23 @@ void Profiler::fillFrameTypes(ASGCT_CallFrame* frames, int num_frames, NMethod* 
     }
 }
 
+void Profiler::logIfNoSamples(Arguments& args, u64 excluded_samples) {
+    // Skip logging if profiling is invoked with --loop mode
+    if (args._loop) {
+        return;
+    }
+    // If total samples are 0 - we log a message
+    if (_total_samples == 0) {
+        Log::info("No samples were collected");
+        return;
+    }
+    // Check if all samples were filtered out by comparing with total samples
+    if (excluded_samples == _total_samples) {
+        Log::info("All %lld samples are filtered out", _total_samples);
+        return;
+    }
+}
+
 u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Event* event) {
     atomicInc(_total_samples);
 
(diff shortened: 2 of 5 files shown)


src/callTraceStorage.cpp
```diff
@@ -155,9 +160,13 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
         for (u32 slot = 0; slot < capacity; slot++) {
             if (keys[slot] != 0 && values[slot].acquireTrace() != NULL) {
                 map[keys[slot]] += values[slot];
+                samplesCollected = true;
             }
         }
     }
+    if (!samplesCollected) {
+        Log::info("No samples collected");
+    }
```
apangin, line 169: `CallTraceStorage` is just a data structure - a kind of specialized hash table. It's not a job of a storage to report UX-related stuff. Furthermore, it does not have enough context to do so: e.g., a JFR recording may have multiple chunks; samples are collected for each chunk, and even though one chunk may contain no samples, other chunks probably do.

It's better to emit a message at the same place where we dump results and where we know for sure that no samples were reported.

Other things to consider:
 - In continuous profiling mode (i.e. with `--loop` option), we keep dumping output periodically with no user interaction involved. A log message will be redundant in this case.
 - If output filter is specified (`-I`/`-X` options), output may be empty even when number of collected samples is non-zero. It makes sense to emit a different message in this case, e.g. `All %d samples were filtered out`.

visheshruparelia: @apangin this makes total sense. Thank you for your insights, I will make the changes.

visheshruparelia: Makes sense @apangin, will make the changes.

Some thoughts/questions:
- The method `dump()` in `Profiler.cpp` class takes care of dumping the output. Now, this method is only invoked once per invocation of async-profiler. So far, this is what I understood by navigating the code. Is this the correct understanding? Or do we have cases where for a single invocation of async profiler, we invoke the `dump()` method multiple times?
- One of the benefits I was getting by adding the log in the `collectSamples()` method is that all the "dump" methods invoke it so one stop place to log our message. If I add it in the individual `dump` methods, the logic of logging needs to be added in all those methods. Are we okay with that or did you have something else in mind? (I am thinking of creating a common method which contains the logging logic and invoke it from different `dump` methods like `dumpCollapsed()`, `dumpText()` etc.

src/profiler.cpp
```diff
@@ -1458,10 +1458,14 @@ void Profiler::dumpCollapsed(Writer& out, Arguments& args) {
...
     std::vector<CallTraceSample*> samples;
     _call_trace_storage.collectSamples(samples);
-
+    u64 filtered_sample_count = 0;
     for (std::vector<CallTraceSample*>::const_iterator it = samples.begin(); it != samples.end(); ++it) {
         CallTrace* trace = (*it)->acquireTrace();
-        if (trace == NULL || excludeTrace(&fn, trace)) continue;
+        if (trace == NULL) continue;
+        if (excludeTrace(&fn, trace)) {
+            filtered_sample_count++;
```
apangin, line 1466: Comparing `filtered_sample_count` to `_total_samples` is not reliable: `filtered_sample_count` does not take into account samples with `trace == NULL` or `counter == 0`.

Why not counting _printed_ samples instead? And simply emit a message if no samples were printed.

src/profiler.cpp
```diff
@@ -1556,9 +1568,14 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
             total_counter += counter;
-            if (trace->num_frames == 0 || excludeTrace(&fn, trace)) continue;
+            if (trace->num_frames == 0) continue;
+            if (excludeTrace(&fn, trace)) {
+                filtered_sample_count++;
+                continue;
+            }
             samples.push_back(it->second);
         }
+        logIfNoSamples(args, filtered_sample_count);
```
apangin, line 1578: Text output already includes the number of collected samples in the summary. I'm not sure if it's worth logging in addition to that. I'd say log matters mostly for `collapsed` and `flamegraph` output. Let me know if you think otherwise.

src/profiler.h
```diff
@@ -134,6 +134,23 @@ class Profiler {
     void stopTimer();
     void timerLoop(void* timer_id);
 
+    void logIfNoSamples(Arguments& args, u64 excluded_samples) {
```
apangin, line 137: I think the function is too big to be inlined in the header file. I'd move implementation to .cpp instead.

src/profiler.cpp
visheshruparelia: @apangin we can do that. I am assuming we count whenever we hit [this LOC](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1474) and compare it with `_total_samples`. However, that means that we won't be able to print statements like `%d samples were filtered out` but rather it will just be blanket statement like `no samples collected`.  Are we fine with that?

visheshruparelia: Effectively we can have a logic like the following:
- Count samples filtered if `excludeTrace` returns `true` to a var called `filtered_sample_count`
- Count samples skipped because of trace being `NULL` or `counter = 0`to a var called `skipped_sample_count`
- Count samples being printed to the output buffer to a var called `printed_sample_count`
Then the `logIfNoSamples()` method can look like this:
```
void logIfNoSamples() {
    if (args._loop) return;
    if (filtered_sample_count > 0 && skipped_sample_count + filtered_sample_count == _total_samples) {
        print("%d samples were filtered out", filtered_sample_count); 
        return;
    }
    if (printed_sample_count == 0) {
        print("No samples collected");
    }
}
```

apangin: Let's stay simple. The reason for a message is to let user know that the output is empty expectedly, not because of some I/O error. We may also print a warning if we know there was an I/O error (`dumpCollapsed` already [does](https://github.com/async-profiler/async-profiler/blob/5ea64a15c362f62f39fab4a7b8ca095b18da3121/src/profiler.cpp#L1477-L1479) this; may be a good idea to merge this into your function).

I propose to identify 3 cases:
1. there was an I/O error: `!out.good()`;
2. no samples were collected: `_total_samples - _failures[-ticks_skipped] == 0`;
3. some samples were collected, but nothing was printed (actual number of filtered samples does not really matter): `printed_samples == 0`. So, all you need is to count `printed_samples`. 

WDYT?

src/profiler.cpp
apangin: That's correct. But this information allows user to identify the cases mentioned above (no collected samples vs. no printed samples).

src/profiler.cpp
visheshruparelia: Yeah makes sense. Quick question though: is it safe to assume that `skipped_sample_count` var if computed based on the above logic, is the same as `_failures[-ticks_skipped]`?

visheshruparelia: @apangin updated the PR btw.

apangin: No, these are different things. A failure of type `ticks_skipped` is counted when profiler was unable to record a sample at all (failed to acquire a lock).

`trace == NULL` is an extremely rare situation when dump is requested while profiler is running. It denotes a race when a new stack trace is being recorded concurrently with dumping.

`counter == 0` is when a stack trace has been found in the CallTraceStorage, but has no counted samples associated with it. For example, when dumping results of live object profiling, a stack trace is recorded on object allocation, but it does not count unless this object survives throughout profiling session.

visheshruparelia: Oh I see, thank you for the explanation. I will dive deep into the code once more now with this information to gain a better understanding of the code. 

PS: In the current diff, we check `!out.good()` for the `dumpFlameGraph()` method as well. It seems redundant to me (you can correct me if I am wrong) as we don't really interact with it. Instead, we interact with the `FlameGraph` class. I did go through the `FlameGraph` class to see if we can do something similar there but could not really come up with something. Do you have any suggestions around that?

src/profiler.cpp
```diff
@@ -1731,6 +1732,26 @@ void Profiler::timerLoop(void* timer_id) {
...
 
+void Profiler::logSampleCollectionStatus(Arguments& args, u64 printed_samples_count, Writer& out) {
+    if (!out.good()) {
+        Log::warn("Output file may be incomplete");
+    }
+    // Skip logging if profiling is invoked with --loop mode
+    if (args._loop) {
+        return;
+    }
+    // If total samples count match the samples which were skipped - we log a message
```
apangin, line 1743: Please avoid comments that just [repeat](https://www.reddit.com/r/ProgrammerHumor/comments/i4v6b2/comments/) what is written in the code.
IMO, in this case, the code is pretty much self-describing. I'd keep it concise like
```
    if (_total_samples == _failures[-ticks_skipped]) {
        Log::info("No samples were collected");
    } else if (printed_samples_count == 0) {
        Log::info("All samples were filtered out");
    }
```

src/profiler.h
```diff
@@ -134,6 +134,8 @@ class Profiler {
     void stopTimer();
     void timerLoop(void* timer_id);
 
+    void logSampleCollectionStatus(Arguments& args, u64 printed_samples_count, Writer& out);
```
apangin, line 137: The name of the function is long but vague. I'd call it something like `logEmptyOutput` which is closer to its original purpose.

apangin (commented):
Thank you for your contribution. I left several minor comments. They may sound scrupulous, but since you are new to the project, this is to calibrate attitude to the coding style.

src/profiler.cpp
apangin: Btw, although I typically prefer simpler expressions, in this particular case
`if (_total_samples - _failures[-ticks_skipped] == 0)` would be more obvious to a reader (i.e. "No samples" means that some value is zero).

src/profiler.cpp
apangin: `flamegraph.dump` interacts with `out`.

src/profiler.cpp
```diff
@@ -1533,7 +1532,9 @@ void Profiler::dumpFlameGraph(Writer& out, Arguments& args, bool tree) {
             }
             f->_total += counter;
             f->_self += counter;
+            printed_sample_count++;
         }
+        logSampleCollectionStatus(args, printed_sample_count, out);
```
apangin, line 1537: Should be called _after_ `flamegraph.dump`

src/profiler.cpp
visheshruparelia: Ohh! I have no idea how I missed it. Thank you.

visheshruparelia:
> Thank you for your contribution. I left several minor comments. They my sound scrupulous, but since you are new to the project, this is to calibrate attitude to the coding style.

Thank you @apangin for the checkstyle suggestions. I have incorporated them in the current PR and will ensure that they are followed in the PRs going forward.

PR #1169, opened 2025-03-11
add asprof_get_thread_local_data

arielb1:
### Description

Add support to `asprof_get_sample_counter` to allow native code to detect when a sample happened.

### Motivation and context

Currently, there is no easy way for code to attach metadata about a sample. This is especially annoying for large multi-tenant applications since knowing the tenant that triggered a sample makes it easier to find out what usage pattern is involved, or to attach [pollcatch] metadata to catch long Tokio poll times.

[pollcatch]: https://github.com/arielb1/pollcatch

Uploading metadata for every request in the system is likely to cause overload. This change allows attaching metadata only for requests that have been sampled.

### How has this been tested?

This has been tested locally using a version of pollcatch.

---


src/profiler.cpp
```diff
@@ -1071,6 +1091,19 @@ Error Profiler::checkJvmCapabilities() {
...
 }
 
+static void initThreadLocalKey() {
+    if (_sample_counter_key != -1) {
+        return; // already initialized
+    }
+
+    pthread_key_t sample_counter;
+    int status;
+    if ((status = pthread_key_create(&sample_counter, NULL)) != 0) {
```
apangin, line 1101: `status` variable seems redundant.

arielb1:
Tested the new version locally

src/profiler.cpp
```diff
@@ -82,11 +85,27 @@ struct MethodSample {
...
 typedef std::pair<std::string, MethodSample> NamedMethodSample;
 
+static void incrementSampleCounter() {
+    // over-paranoia - access the sample key once to avoid weird signal races
+    pthread_key_t sample_counter_key = _sample_counter_key;
+    if (sample_counter_key == -1) return;
+
+    uintptr_t counter = (uintptr_t)pthread_getspecific(sample_counter_key);
+    counter++;
+    pthread_setspecific(sample_counter_key, (void*)counter);
```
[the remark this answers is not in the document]
arielb1: turned out you can do it in `asprof_get_sample_counter` since the API lets you

docs/ProfilingNonJavaApplications.md
```diff
@@ -92,3 +96,15 @@ LD_PRELOAD=/path/to/libasyncProfiler.so ASPROF_COMMAND=start,event=cpu,file=prof
...
+
+The `asprof_get_sample_counter` function increments every time there is a sample. This gives native
+code an easy way to detect when a sample event had occurred, and to log metadata about what the
+program was doing when the event happened.
+
+### Interaction with a JVM
+
+If you are using the native API to use async-profiler on a program that embeds a JVM and calls into it,
+async-profiler will not attach to that JVM, and stack traces involving (JITted) Java code will be mostly
+a single `.unknown` frame.
```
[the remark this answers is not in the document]
arielb1: done

src/profiler.cpp
```diff
@@ -82,11 +92,44 @@ struct MethodSample {
...
+    pthread_key_t sample_counter_key = _sample_counter_key;
+    if (sample_counter_key == -1) return;
+
+    uint64_t *counter = (uint64_t *)pthread_getspecific(sample_counter_key);
+    if (counter != NULL) {
+        (*counter)++;
+    }
+}
+
+DLLEXPORT uint64_t asprof_get_sample_counter(void) {
```
apangin, line 106: Implementation of `asprof_` APIs should be in asprof.cpp.

src/profiler.cpp
```diff
@@ -614,6 +657,7 @@ void Profiler::fillFrameTypes(ASGCT_CallFrame* frames, int num_frames, NMethod*
 
 u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Event* event) {
     atomicInc(_total_samples);
+    incrementSampleCounter();
```
apangin, line 660: I'm thinking where the right place to increment the counter will be.
The counter is only relevant with JFR output, isn't it? Then `recordEvent` (or the places where it is called) seems a better alternative. Or can you imagine a different use case?
Note that some recorded events (e.g. `free` calls in native memory profiling mode) bypass `Profiler::recordSample`.

src/asprof.h
```diff
@@ -37,6 +38,16 @@ typedef const char* (*asprof_error_str_t)(asprof_error_t err);
...
 
+// Gets the thread-local sample counter, which increments (not necessarily by 1) every time a
+// stack profiling sample is taken using a profiling signal.
+//
+// The counter might be initialized lazily, only starting counting from 0 the first time
+// `asprof_get_sample_counter` is called on a given thread. Further calls to
+// `asprof_get_sample_counter` on a given thread will of course not reset the counter.
+//
+// This function is *not* async-signal safe.
+DLLEXPORT uint64_t asprof_get_sample_counter(void);
```
apangin, line 49: Just an idea: what do you think about `asprof_get_tls` function instead? It will return a pointer to the entire thread local area, which besides the counter may include other context-related information.
I'll certainly need something like this for Java API.

arielb1:
waiting for review

src/asprof.h
```diff
@@ -37,6 +38,28 @@ typedef const char* (*asprof_error_str_t)(asprof_error_t err);
...
+// This API is UNSTABLE and might change or be removed in the next version of async-profiler.
+//
+// Gets a pointer to asprof's thread-local data structure, see `asprof_unstable_thread_local_data`'s
+// documentation for the details of each field. This function might lazily initialize that
+// structure.
+//
+// This function can return NULL in case of an allocation failure.
+//
+// This function is *not* async-signal safe.
+DLLEXPORT asprof_unstable_thread_local_data *asprof_unstable_get_thread_local_data(void);
```
apangin, line 61: You've explicitly documented the function as unstable in the code and in the docs, which is totally fine; adding `unstable` in the name is going to be too much ) The name is already too long.

src/asprofPrivate.h
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _ASPROF_PRIVATE_H
+#define _ASPROF_PRIVATE_H
+
+#include "asprof.h"
+
+// Private functions for the native API.
```
apangin, line 11: This "private" header looks somewhat artificial design-wise. There is no real division into public/private headers in the sources.
It's true that `asprof.h` lists publicly exported functions, and it's OK for other parts of the profiler to use these functions, but if some source needs to call an internal function, then this function has likely nothing to do with `asprof*` API.
In your particular case, `FlightRecorder` is the only caller of `asprofIncrementThreadLocalSampleCounter`, so why can't this function be a part of `FlightRecorder` then? Or perhaps a member of `Profiler`? When we have more thread local data in the future, it may be reasonable to extract these functions to a separate `threadLocal.{h,cpp}` or `tls.{h,cpp}`.

src/asprofPrivate.h
```diff
@@ -0,0 +1,19 @@
...
+
+// Increment the thread-local sample counter, if it's already initialized.
+// If the counter is not already initialized, this function is a no-op.
+//
+// This function is signal-safe and safe to call racily with initialization.
+void asprofIncrementThreadLocalSampleCounter(void);
```
apangin, line 17: With rare exceptions, non-local top-level functions are discouraged in async-profiler.
Static classes generally serve as namespaces. Actual C++ namespaces would be probably better, but that was a historical choice similar to the coding style adopted in HotSpot.

src/asprof.cpp
```diff
@@ -4,9 +4,27 @@
...
+// application to detect samples.
+//
+// The per-thread counter is lazily initialized in `asprof_unstable_get_thread_local_data`,
+// and is NULL until `asprof_unstable_get_thread_local_data` is called. This matches the API
+// since the sample counter only needs to start running the first time
+// `asprof_unstable_get_thread_local_data` is called, and allows us to avoid
+// calling `malloc` in weird places.
+//
+// Using std::atomic instead of "native" std::call_once to ensure signal safety.
+static std::atomic<pthread_key_t> _profiler_data_key {(pthread_key_t)-1};
```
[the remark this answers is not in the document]
arielb1: 1. I couldn't find a good non-lazy place to initialize it
2. people might call asprof_get_thread_local_data concurrently with initializing async-profiler, I want that to work.

src/asprof.cpp
```diff
@@ -49,3 +49,7 @@ DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t out
 
     return asprof_error(error.message());
 }
+
+DLLEXPORT asprof_thread_local_data *asprof_get_thread_local_data(void) {
```
apangin, line 53: async-profiler follows `type* identifier` style

src/asprof.h
```diff
@@ -37,6 +38,30 @@ typedef const char* (*asprof_error_str_t)(asprof_error_t err);
...
+// documentation for the details of each field. This function might lazily initialize that
+// structure.
+//
+// This function can return NULL either if the profiler is not yet initializer, or in
+// case of an allocation failure.
+//
+// This function is *not* async-signal-safe. However, it is safe to call concurrently
+// with async-profiler operations, including initialization.
+DLLEXPORT asprof_thread_local_data *asprof_get_thread_local_data(void);
+typedef asprof_thread_local_data *(*asprof_unstable_get_thread_local_data_t)(void);
```
apangin, line 64: `unstable` in the name remained

src/threadLocalData.h
```diff
@@ -0,0 +1,50 @@
...
+
+        asprof_thread_local_data *val = (asprof_thread_local_data*) pthread_getspecific(_profiler_data_key);
+        if (val == NULL) {
+            return initPerThreadThreadLocalData(_profiler_data_key);
+        }
+
+        return val;
+    }
+private:
+    static asprof_thread_local_data *initPerThreadThreadLocalData(pthread_key_t profiler_data_key);
```
apangin, line 45: "PerThreadThreadLocal" is a kind of tautology.
Also, the function does not need to accept an argument, since it's always `_profiler_data_key` accessible from that function.

src/threadLocalData.h
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _ASPROF_THREAD_LOCAL_H
+#define _ASPROF_THREAD_LOCAL_H
+
+#include "asprof.h"
+#include "mutex.h"
```
apangin, line 10: Not needed?

src/threadLocalData.h
```diff
@@ -0,0 +1,50 @@
...
+#include "arguments.h"
+#include <pthread.h>
+#include <atomic>
```
apangin, line 13: Not needed?

PR #1177, opened 2025-03-12
Improve CPUTests.itimer test for systems under load

parttimenerd:
### Description

Change the test to check that the CPU time of the profiled process (via a [MBean](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/OperatingSystemMXBean.html)) is similar to what can be computed from the samples.

Additionally: Make the profiled program not allocate all the time.

### Motivation and context

The test doesn't work when the system is under load.

### How has this been tested?

Ran on multiple platforms.

---

### DIFF
--- a/test/test/cpu/CPUTimeService.java
+++ b/test/test/cpu/CPUTimeService.java
@@ -1,0 +1,45 @@
+package test.cpu;
+
+import com.sun.tools.attach.AttachNotSupportedException;
+import com.sun.tools.attach.VirtualMachine;
+
+import javax.management.MBeanServerConnection;
+import javax.management.ObjectName;
+import javax.management.remote.JMXConnector;
+import javax.management.remote.JMXConnectorFactory;
+import javax.management.remote.JMXServiceURL;
+import java.io.IOException;
+import java.util.Set;
+
+/**
+ * Obtain the CPU time of a process using the Attach API
+ */
+class CPUTimeService implements AutoCloseable {
+
+    private final VirtualMachine vm;
+    private final MBeanServerConnection mbeanServer;
+    private final JMXConnector jmxConnector;
+
+    CPUTimeService(long pid) throws IOException, AttachNotSupportedException {
+        vm = VirtualMachine.attach(pid + "");
+        String connectorAddress = vm.startLocalManagementAgent();
+        JMXServiceURL url = new JMXServiceURL(connectorAddress);
+        jmxConnector = JMXConnectorFactory.connect(url);
+        mbeanServer = jmxConnector.getMBeanServerConnection();
+    }
+
+    long getProcessCPUTimeNanos() throws Exception {
+        Set<ObjectName> mbeans = mbeanServer.queryNames(new ObjectName("java.lang:type=OperatingSystem"), null);
+        if (mbeans.isEmpty()) {
+            throw new Exception("No OperatingSystem MBean found");
+        }
+        ObjectName osMBean = mbeans.iterator().next();
+        return (long) mbeanServer.getAttribute(osMBean, "ProcessCpuTime");
+    }
+
+    @Override
+    public void close() throws IOException {
+        jmxConnector.close();
+        vm.detach();
+    }
+}
--- a/test/test/cpu/CPUTimeService.java
+++ b/test/test/cpu/CPUTimeService.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package test.cpu;
 
 import com.sun.tools.attach.AttachNotSupportedException;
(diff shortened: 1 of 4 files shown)


test/test/cpu/CPUTimeService.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.management.remote.JMXConnector;
+import javax.management.remote.JMXConnectorFactory;
+import javax.management.remote.JMXServiceURL;
+import java.io.IOException;
+import java.util.Set;
+
+/**
+ * Obtain the CPU time of a process using the Attach API
+ */
+class CPUTimeService implements AutoCloseable {
```
apangin, line 22: `CpuTimeService` - for the consistency with other class names.
The same for the method name. Note that MBean attribute is also called `ProcessCpuTime`.

test/test/cpu/CPUTimeService.java
```diff
@@ -0,0 +1,50 @@
...
+    CPUTimeService(long pid) throws IOException, AttachNotSupportedException {
+        vm = VirtualMachine.attach(pid + "");
+        String connectorAddress = vm.startLocalManagementAgent();
+        JMXServiceURL url = new JMXServiceURL(connectorAddress);
+        jmxConnector = JMXConnectorFactory.connect(url);
+        mbeanServer = jmxConnector.getMBeanServerConnection();
+    }
+
+    long getProcessCPUTimeNanos() throws Exception {
+        Set<ObjectName> mbeans = mbeanServer.queryNames(new ObjectName("java.lang:type=OperatingSystem"), null);
```
apangin, line 37: Why queryNames? Can we just call `getAttribute(ObjectName name, String attribute)` for simplicity?

parttimenerd: This seems to lead to errors, so no.

apangin: Sorry it's not clear to me. What errors?

parttimenerd: `getAttribute` is used to get attributes of Means, but to get an MBean, one needs to use `queryNames`.

parttimenerd: Oh, you're right.

apangin: `queryNames` returns a set of `ObjectName`s. But you can as well construct the same `ObjectName` directly. What am I missing here? Can you provide an exact exception?

parttimenerd: I was wrong and fixed it already.

test/test/cpu/CpuTests.java
```diff
@@ -14,7 +14,7 @@
 public class CpuTests {
 
     private static void assertCloseTo(long value, long target, String message) {
-        Assert.isGreaterOrEqual(value, target * 0.75, message);
+        Assert.isGreaterOrEqual(value, target * 0.6, message);
```
apangin, line 17: 40% difference is a lot - it may absorb some real issues.
This is a severe change, I'm reluctant to accept this for all cpu tests on all platforms, especially since we haven't seen problems with this test on GitHub Actions or on EC2 instances.

parttimenerd: Ok, I'll try to reduce it.

parttimenerd: I'm now computing the cpu/wall ratio and use it to get a better expected CPU sample time. Additionally, I skip the test if the cpu/wall ratio is lower than 0.5, to prevent failing on heavily loaded systems.

test/test/cpu/CpuTests.java
```diff
@@ -29,7 +29,24 @@ public void ctimerTotal(TestProcess p) throws Exception {
...
+            cpuTime = timeService.getProcessCpuTimeNanos();
+            out = p.profile("-d 2 -e itimer -i 100ms --total -o collapsed");
+            cpuTime = timeService.getProcessCpuTimeNanos() - cpuTime;
+            wallTime = (System.currentTimeMillis() - wallTime) * 1_000_000;
+        }
+        double ratio = (double)cpuTime / wallTime;
+        if (ratio < 0.5) {
+            System.err.println("Warning: CPU time / wall time ratio is too low: " + ratio);
+            System.err.println("Your system has too high load to run the itimer test properly");
+            System.err.println("Therefore the test is skipped");
```
apangin, line 46: This is not really maintainable, sorry. A simple 2-line test turned into something convoluted.
The purpose of unit tests is to verify basic assumptions (not just to "pass"). But it's no longer clear for a casual reader what this test verifies exactly and what it has to do with the cpu/wall ratio.

Performance related tests may have certain assumptions about environment they run in - this is normal. If assumptions do not hold true in your environment - why don't you just skip this test? There is `SKIP` make variable for this purpose. After all, you already skip this test under load. It's not strictly required for all tests to run everywhere. For example, we already skip PMU tests on systems where there is no access to hardware performance counters.

Issue #82, opened 2018-02-15
Make FlameGraph parameters configurable


apangin:
Add command-line option to tune certain FlameGraph attributes:

- imagewidth
- frameheight
- minwidth
- etc.

apangin:
Added the following command-line options:
```
  --title TITLE
  --width PX
  --height PX
  --minwidth PX
```

retronym:
`--reverse` is handy too.

ceeaspb:
```flamegraph.pl``` ```--reverse``` and ```--inverted``` (ie. icicle graph) work together to keep the stack the right way up while grouping by the top of stack - I use it for the allocation profiling, it can be used elsewhere also.

apangin:
@ceeaspb So what do you suggest?

ceeaspb:
add support for ```--inverted``` if possible

apangin:
While this one is easy to implement, I don't actually plan to replicate and maintain all possible FlameGraph options. It will be enough to support just the most common use case. `collapsed` output provides the possibility to customize flame graphs using external script.

Does it make sense to combine `--reverse` and `--inverted` into one singe option? I have never used icicle graphs myself, but if this is a popular use case, I don't mind adding this option to async-profiler.

PR #1178, opened 2025-03-13
Implement new '--inverted' option to flip graphs vertically

simonis:
### Description

AsyncProfiler currently supports two kind of graphs: 
- The default flame graphs which grow from bottom to top (because flames grow from bottom to top) and show the outermost frames (e.g. the `main()` function) at the bottom and the innermost, leaf frames at the top. They are created from stack traces which are merged starting from the outermost (e.g. `main()`) frames.
- Icicle graphs which grow top-down (because icicles grow to-down) and are created from reverse stack traces, i.e. stack traces which are merged starting with the innermost, leaf frames.

Both kind of graphs can be flipped vertically with the `Revert` button which is located in the upper-left corner of the generated HTML page, when the graph is displayed. 

But normal/reverse stack traces and icicle/flame graphs are orthogonal settings. Both kind of stack traces can be flipped vertically. Unfortunately, the vertical orientation can currently only be changed in the generated HTML page but the default orientation can not be set when the graphs are generated. It always defaults to a bottom-up, flame graph layout for normal stack traces and a top-down, icicle layout for reverse stack traces.

This PR introduces a new command line parameter `--inverted` which makes it possible to change the default vertical orientation for both kind of stack traces. The name of the parameter (i.e. `inverted` instead of `inverse`) is taken from Brendan Greg's original [flamegraph.pl](https://github.com/brendangregg/FlameGraph/blob/41fee1f99f9276008b7cd112fca19dc3ea84ac32/flamegraph.pl#L152-L153) script which also provides this option.

The PR doesn't change the default behavior of AsyncProfiler. Normal stack traces are still rendered as bottom-up flame graphs and reverse stack traces are still rendered as top-down icicle, graphs. But both of these defaults can now be changed with the new `--inverted=<true|false>` command line option.

### Related issues

See the discussion on #82 for a similar request. 

### Motivation and context

See Description section.

### How has this been tested?

`make test` and manual testing of the new parameter for `asprof`,  `jfrconv` and the agent use case.

---


src/converter/one/convert/Arguments.java
```diff
@@ -78,6 +84,9 @@ public Arguments(String... args) {
                 throw new IllegalArgumentException(arg);
             }
         }
+        if (reverse && !invertedSet) {
+            inverted = true;
+        }
```
apangin, line 89: Generally speaking, `Arguments` class is just a holder for user input. I'd prefer it did not endow arguments with semantics.

src/converter/one/convert/FlameGraph.java
```diff
@@ -67,7 +67,7 @@ public void parseHtml(Reader in) throws IOException {
         Frame[] levels = new Frame[128];
         int level = 0;
         long total = 0;
-        boolean needRebuild = args.reverse || args.include != null || args.exclude != null;
+        boolean needRebuild = args.reverse || args.inverted || args.include != null || args.exclude != null;
```
apangin, line 70: `inverted` affects display only - there is no need to rebuild flame graph structure if this option changes.

docs/ConverterUsage.md
```diff
@@ -75,12 +75,15 @@ Flame Graph options:
     --minwidth X       Skip frames smaller than X%
     --grain X          Coarsen Flame Graph to the given grain size
     --skip N           Skip N bottom frames
- -r --reverse          Reverse stack traces (icicle graph)
+ -r --reverse          Reverse stack traces (implies `--inverted true`)
+ -i --inverted BOOL    Display graph top-down instead of bottom-up (icicle graph): true | false
```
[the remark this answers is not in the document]
simonis: I agree that having a mandatory argument for `--inverted` is a little clumsy. So what about removing that argument and making `--inverted` a true toggle switch which changes the default layout for both, flame and icicle graphs:

```diff
- -r --reverse          Reverse stack traces (implies `--inverted true`)
- -i --inverted BOOL    Display graph top-down instead of bottom-up (icicle graph): true | false
+ -r --reverse          Reverse stack traces (defaults to icicle graph)
+ -i --inverted         Toggels the layout for reversed stacktraces from icicle to flamegraph
+                       and for default stacktraces from flamegraph to icicle
```
I think this is what you describe as alternative 2. above.

docs/ConverterUsage.md
```diff
@@ -75,12 +75,16 @@ Flame Graph options:
     --minwidth X       Skip frames smaller than X%
     --grain X          Coarsen Flame Graph to the given grain size
     --skip N           Skip N bottom frames
- -r --reverse          Reverse stack traces (icicle graph)
+ -r --reverse          Reverse stack traces (defaults to icicle graph)
+ -i --inverted         Toggels the layout for reversed stacktraces from icicle to flamegraph
```
apangin, line 79: toggels -> toggles (in multiple places)

src/converter/one/convert/FlameGraph.java
```diff
@@ -181,8 +181,10 @@ public void dump(PrintStream out) {
         tail = printTill(out, tail, "/*title:*/");
         out.print(args.title);
 
-        tail = printTill(out, tail, "/*reverse:*/false");
-        out.print(args.reverse);
+        // inverted toggels the layout for reversed stacktraces from icicle to flamegraph
+        // and for default stacktraces from flamegraphs to icicle.
+        tail = printTill(out, tail, "/*inverted:*/false");
+        out.print(args.reverse ? !args.inverted : args.inverted);
```
apangin, line 187: `args.reverse ^ args.inverted`

src/flameGraph.cpp
```diff
@@ -147,8 +147,10 @@ void FlameGraph::dump(Writer& out, bool tree) {
         tail = printTill(out, tail, "/*title:*/");
         out << _title;
 
-        tail = printTill(out, tail, "/*reverse:*/false");
-        out << (_reverse ? "true" : "false");
+        tail = printTill(out, tail, "/*inverted:*/false");
+        // _inverted toggels the layout for reversed stacktraces from icicle to flamegraph
+        // and for default stacktraces from flamegraphs to icicle.
+        out << (_reverse ? (_inverted ? "false" : "true") : (_inverted ? "true" : "false"));
```
apangin, line 153: Let's simplify this to `(_reverse ^ _inverted ? "true" : "false")`

src/main/main.cpp
```diff
@@ -493,6 +495,9 @@ int main(int argc, const char** argv) {
         } else if (arg == "--reverse" || arg == "--samples" || arg == "--total" || arg == "--sched" || arg == "--live" || arg == "--nofree") {
             format << "," << (arg.str() + 2);
 
+        } else if (arg == "--inverted") {
+            params << ",inverted=" << args.next();
```
apangin, line 499: Remained from the previous revision where `--inverted` had an argument.

docs/ConverterUsage.md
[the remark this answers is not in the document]
apangin: Yes, sounds good.

apangin:
Merged. Thank you for the contribution!

PR #1179, opened 2025-03-13
Add sanity tests for various idling/blocking operations

jbachorik:
### Description
This adds two additional tests based on what we have in [Datadog Java Profiler](https://github.com/DataDog/java-profiler/tree/main/ddprof-test/src/test/java/com/datadoghq/profiler/wallclock).

### Motivation and context
These tests will reliably fail on AARCH64 when running on JDK 8/11 from Temurin or Liberica distribution. On other distributions (Corretto and Zulu - other don't seem to provide up-to-date updates of 8 and 11 so, hopefully, their usage will be phased out) the tests are passing just fine.

The main issue seems to be related to somehow mangled FP/LR when calling to some JIT compiled methods (most frequently this fails for things like 'Thread.sleep()` or  `Object.wait()`) where following the standard rules for obtaining FP and LR as SP[-8] and SP[-16] yield bogus values.
The exact location is here - https://github.com/DataDog/async-profiler/blob/f71c31af7b6972fc5432d0be33ed75a34ab3f710/src/stackWalker.cpp#L292

I checked the instructions for the affected methods and the frame size is correct, so the FP and LR values already arrive mangled.

I have a very experimental and WIP code in [branch](https://github.com/DataDog/async-profiler/tree/jb/stackwalker) where I am trying to wrap my head around what and how is actually mangling the linkage. 
A desperate attempt is to use the standard FP walking to recover in such situation - and, to my surprise, it mostly works. Well, if paired with opportunistic recovery from the thread JavaFrameAnchor (if there is any). But still it can create broken or 'surprising' stacktraces 

I want to add these three tests (well, two tests and one extra configuration of the existing test) to allow perhaps someone more knowledgeable in ways how different toolchain/compilation might mangle FP/LR such that they are not directly usable from a compiled method.

---

### DIFF
--- a/test/one/profiler/test/Output.java
+++ b/test/one/profiler/test/Output.java
@@ -93,7 +93,7 @@ public class Output {
         return new Output(stream(regex).toArray(String[]::new));
     }
 
-    private static long extractSamples(String s) {
+    public static long extractSamples(String s) {
         return Long.parseLong(s.substring(s.lastIndexOf(' ') + 1));
     }
 }
--- a/test/test/wall/PingPongClient.java
+++ b/test/test/wall/PingPongClient.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.wall;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
+
+class PingPongClient {
+    private static final ExecutorService executor = Executors.newCachedThreadPool();
+    public static void main(String[] args) throws Exception {
+        long result = 0;
+        for (int i = 0; i < 10; i++) {
+            result += pingPong();
+        }
+    }
+
+    private static long pingPong() {
+        Object monitor = new Object();
+        AtomicLong counter = new AtomicLong();
+        long startTime = System.currentTimeMillis();
+        List<CompletableFuture<Void>> futures = new ArrayList<>();
+        for (int i = 0; i < 2; i++) {
+            futures.add(CompletableFuture.supplyAsync(() -> {
+                        while (System.currentTimeMillis() - startTime < 500) {
+                            synchronized (monitor) {
+                                counter.addAndGet(busyWork(Duration.ofMillis(100)));
+                            }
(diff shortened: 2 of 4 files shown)


krk:
Thank you for your contribution. 

I tested these changes with Corretto 21.0.7 locally on x86_64 and the `WallTests.waitingWallVM` is failing ~50% of the time, with Corretto 24.0.1, both `WallTests.waitingWallVM` and `pingPongWallVM` fail. 

As it is now, the tests are not stable enough to be executed in the CI.

```
Corretto 21.0.7

FAIL [3/4] WallTests.waitingWallVM took 3.791 s
java.lang.AssertionError: Expected 0.0 == 2.0
        >  test/test/wall/WallTests.java:55
        >  Assert.isEqual(0, unknown);
        at one.profiler.test.Assert.assertComparison(Assert.java:39)
        at one.profiler.test.Assert.isEqual(Assert.java:44)
        at test.wall.WallTests.waitingWallVM(WallTests.java:55)
        at one.profiler.test.Runner.run(Runner.java:133)
        at one.profiler.test.Runner.main(Runner.java:230)
```

```
Corretto 24.0.1

FAIL [4/4] WallTests.pingPongWallVM took 3.790 s
java.lang.AssertionError: Expected 0.0 == 3006.0
        >  test/test/wall/WallTests.java:68
        >  Assert.isEqual(0, unknown);
        at one.profiler.test.Assert.assertComparison(Assert.java:39)
        at one.profiler.test.Assert.isEqual(Assert.java:44)
        at test.wall.WallTests.pingPongWallVM(WallTests.java:68)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at java.base/java.lang.reflect.Method.invoke(Method.java:565)
        at one.profiler.test.Runner.run(Runner.java:133)
        at one.profiler.test.Runner.main(Runner.java:230)
```

jbachorik:
Hi @krk  - were you able to check the profiles to see if the failures are incorrect or we, indeed, have a large number of stacktraces unresolved?

I would guess it is the second case, judging on my experience - and the tests are here to actually show the gaps in the vm structs based stackwalking and then decide whether it is something we want to fix (and how) or we will just go 'meh, it's as good as it gets'.

jbachorik:
An update - I have noticed #1347 and tried running these new tests on top of that patch.
And, lo and behold, the tests are passing ... well, except on alpine linux but this might be addressed in #1329, if I am not mistaken.

I did an experiment when I excluded the new tests on alpine/musl and the tests are nicely passing - https://github.com/DataDog/async-profiler/actions/runs/16221855315

@visheshruparelia or @yanglong1010  - would you be interested in subsuming these tests into your work? They turned out to be a rather reliable way of telling that the vmstructs stackwalking is not working as well as we want it to.

Issue #919, opened 2024-04-22
Introduce filtering by CPU


w-kudla:
There should be an option to collect stack traces only for chosen CPUs. It would drastically reduce the amount of work the profiler has to do and make analysis easier. 
This is not a made up use case; there's a whole industry where pinning a single thread to an isolated core is default optimization strategy.

apangin:
This is the same issue as #481. There was an attempt to implement the feature in #486, however, the pull request was incomplete. If you wish, you may try to revive that PR or create a new one.

Issue #481, opened 2021-10-22
Configure CPUs for profiling


pveentjer:
It is possible to configure particular CPU's for profiling?

With perf I frequently check what a particular CPU is doing by passing an `-C <cpu>`.

Does async-profiler has such an ability? I could not find anything in the documentation.

If it doesn't have this capability, I would not mind picking up this task.

krk:
Implemented in https://github.com/async-profiler/async-profiler/pull/1180

PR #1180, opened 2025-03-14
Allow specifying a target for CPU profiling

fandreuz:
### Description
Allow users to select a specific CPU via `--target-cpu` to sample only threads running on that CPU. See #481 for the original issue.

### Related issues

#481, #919

### Motivation and context

See original issues for use cases.

### How has this been tested?

I introduced some tests in `test.cpu.CpuTests`:
- `perfEventsRightTargetCpu` tests the number of events sampled when the right CPU is selected;
- `perfEventsWrongTargetCpu` tests the number of events sampled when the wrong CPU is selected;

In order to introduce those tests, I had to add a small extension to the test infrastructure in order to allow pinning the newly started process to a specific CPU via `taskset`.

---


test/test/cpu/CpuTests.java
```diff
@@ -32,4 +34,36 @@ public void itimerTotal(TestProcess p) throws Exception {
...
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 1");
+        assertCloseTo(out.total(), 2_000_000_000, "perf_events total should match profiling duration");
+    }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
+    public void itimerDoesNotSupportTargetCpu(TestProcess p) throws Exception {
+        try {
+            Output out = p.profile("-e itimer --target-cpu 1");
+            throw new IllegalStateException("Profiling should have failed");
+        } catch (IOException expectedException) {
```
fandreuz, line 55: Do we have a cleaner way to test that an exception has been thrown? I'm used to JUnit 5 (e.g. `assertThrows)

src/perfEvents_linux.cpp
```diff
@@ -594,10 +594,10 @@ int PerfEvents::createForThread(int tid) {
     if (FdTransferClient::hasPeer()) {
         fd = FdTransferClient::requestPerfFd(&tid, &attr);
```
apangin, line 595: `fdtransfer` path is not covered.
`fdtransfer` option allows to offload creation of perf_events to a separate privileged process, see #397 for details.

test/test/cpu/CpuTests.java
apangin: No, but you can create one if you wish (preferably in a separate PR).

src/perfEvents_linux.cpp
fandreuz: Fixed in 4312d9d48f6906b1c93da8912f4f3641e0c757b4, still need to test it though

src/fdtransfer.h
```diff
@@ -32,6 +32,7 @@ struct fd_request {
 struct perf_fd_request {
     struct fd_request header;
     int tid;
+    int target_cpu;
```
fandreuz, line 35: Do we need to ensure backward compatibility with the protocol here?

apangin: Good point! We don't need it to be backward compatible, though.

src/fdtransferClient_linux.cpp
```diff
@@ -47,10 +47,11 @@ bool FdTransferClient::connectToServer(const char *path) {
     return true;
 }
 
-int FdTransferClient::requestPerfFd(int *tid, struct perf_event_attr *attr) {
+int FdTransferClient::requestPerfFd(int *tid, int target_cpu, struct perf_event_attr *attr) {
     struct perf_fd_request request;
     request.header.type = PERF_FD;
     request.tid = *tid;
+    request.target_cpu;
```
roy-soumadipta, line 54: Is this correct? it's a noop at the moment

visheshruparelia:
@fandreuz @apangin looks like CI builds did not run on this PR. Is it because the pending approvals on the workflow?

src/fdtransferClient_linux.cpp
roy-soumadipta: Also I think it is missed being defined in the struct in fdtransfer.h

```
struct perf_fd_request {
    struct fd_request header;
    int tid;
    struct perf_event_attr attr;
};
```

fandreuz: Not correct indeed, I missed the assignment. Thanks!

808615d1ac150b539bd56ea0e5afcafd9b42e9d5

src/perfEvents_linux.cpp
fandreuz: Tested

src/cpuEngine.cpp
```diff
@@ -21,6 +21,8 @@ CStack CpuEngine::_cstack;
 int CpuEngine::_signal;
 bool CpuEngine::_count_overrun;
 
+int CpuEngine::_target_cpu = -1;
```
apangin, line 24: The option is specific to PerfEvents - I'd move it there.

src/ctimer_linux.cpp
```diff
@@ -92,6 +92,10 @@ Error CTimer::start(Arguments& args) {
         return Error("Could not set pthread hook");
     }
 
+    if (args._target_cpu != -1) {
+        return Error("target-cpu != -1 is not supported by CTimer CPU sampling");
+    }
```
apangin, line 97: Since the option is specific to PerfEvents engine, instead of checking it in every other engine, you may check it globally like [here](https://github.com/async-profiler/async-profiler/blob/5ea64a15c362f62f39fab4a7b8ca095b18da3121/src/profiler.cpp#L1169-L1170). This will also reduce the amount of changes.

src/main/main.cpp
```diff
@@ -87,6 +87,7 @@ static const char USAGE_STRING[] =
     "  --jfrsync config  synchronize profiler with JFR recording\n"
     "  --libpath path    full path to libasyncProfiler.so in the container\n"
     "  --fdtransfer      use fdtransfer to serve perf requests\n"
+    "  --target-cpu      sample threads on a specific CPU (`-e cpu` only, default: -1)\n"
```
apangin, line 90: Not `cpu` only; this can be used with any perf_event.

src/main/main.cpp
```diff
@@ -527,6 +528,9 @@ int main(int argc, const char** argv) {
             fdtransfer = buf;
             params << ",fdtransfer=" << fdtransfer;
 
+        } else if (arg == "--target-cpu") {
+            params << "," << (arg.str() + 2) << "=" << args.next();
```
apangin, line 532: You can merge this branch with other one-arg cases above.

test/one/profiler/test/TestProcess.java
```diff
@@ -93,6 +93,19 @@ public TestProcess(Test test, Os currentOs, String logDir) throws Exception {
             // Give the JVM some time to initialize
             Thread.sleep(700);
         }
+
+        if (test.cpu() > 0) {
+            List<String> pinCpuCmd = new ArrayList<>();
+            pinCpuCmd.add("taskset");
+            pinCpuCmd.add("-acp");
+            pinCpuCmd.add(String.valueOf(test.cpu()));
+            pinCpuCmd.add(String.valueOf(p.pid()));
```
apangin, line 102: Since the number of parameters is static here, you can simplify this to 
```
String[] tasksetCmd = {"taskset", "-acp", ...};
```
Or even simpler, call `Runtime.exec`.

test/test/cpu/CpuTests.java
```diff
@@ -32,4 +35,42 @@ public void itimerTotal(TestProcess p) throws Exception {
         Output out = p.profile("-d 2 -e itimer -i 100ms --total -o collapsed");
         assertCloseTo(out.total(), 2_000_000_000, "itimer total should match profiling duration");
     }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX, cpu = 1)
+    public void perfEventsWrongTargetCpu(TestProcess p) throws Exception {
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 2");
+        assertCloseTo(out.total(), 0, "perf_events total should be 0 when the wrong CPU is targeted");
```
apangin, line 42: Not "close to", but it should be strictly zero in this case, right?

test/one/profiler/test/Test.java
```diff
@@ -36,6 +36,8 @@
 
     Os[] os() default {};
 
+    int cpu() default -1;
```
apangin, line 39: Since the feature is used in one test only (and it's unlikely to be used anywhere else in the tests), may be it's not worth adding as a generic feature of the test framework. Perhaps you can just create an utility method in `CpuTests` to run `taskset`.

src/arguments.cpp
```diff
@@ -107,6 +107,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     title=TITLE      - FlameGraph title
 //     minwidth=PCT     - FlameGraph minimum frame width in percent
 //     reverse          - generate stack-reversed FlameGraph / Call tree
+//     target-cpu=CPU   - sample threads on a specific CPU (`-e cpu` only, default: -1)
```
apangin, line 110: Don't forget to update `ProfilerOptions.md` as well.

apangin, line 110: nit: It would be ideal if an option does not have `-` in its name, but I can't think of a better name unfortunately :)

src/arguments.cpp
```diff
@@ -422,6 +423,10 @@ Error Arguments::parse(const char* args) {
             CASE("reverse")
                 _reverse = true;
 
+            CASE("target-cpu")
+                if (value == NULL || (_target_cpu = atoi(value)) < 0)
+                    _target_cpu = -1;
```
apangin, line 428: Always surround `if` body with `{ }`. Exception is trivial one-liners like `if (!cond) return;`

fandreuz: e6ef966dd053ab35b30f51fdc50cfd7002ee2370

src/arguments.cpp
```diff
@@ -422,6 +423,10 @@ Error Arguments::parse(const char* args) {
             CASE("reverse")
                 _reverse = true;
 
+            CASE("target-cpu")
```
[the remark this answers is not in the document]
fandreuz: e6ef966dd053ab35b30f51fdc50cfd7002ee2370

src/ctimer_linux.cpp
fandreuz: a21d908797d6025ce88177ac337ae9e872308d05

src/cpuEngine.cpp
fandreuz: a21d908797d6025ce88177ac337ae9e872308d05

src/arguments.cpp
fandreuz: How about removing the dash? I guess the name would remain easily interpretable for the user.

Also, what's the problem with the dash?

src/main/main.cpp
fandreuz: a0934a229cbe60e472037c3e707fdc8b2cc6cb5d

src/arguments.cpp
fandreuz: a0934a229cbe60e472037c3e707fdc8b2cc6cb5d

test/test/cpu/CpuTests.java
fandreuz: Yes indeed: bb909cd085144d9d49124f2181832eb150e7c4ec

test/one/profiler/test/Test.java
fandreuz: c290e6f189a21265a019a197d4cc852fd3f68017

src/main/main.cpp
fandreuz: cba56c6

src/arguments.cpp
roy-soumadipta: We want to keep flag names as short as possible for better user experience. Generally different words should ideally be separated by a dash, and hence Andrei mentioned that he too can't think of a better name at the moment.

docs/ProfilerOptions.md
```diff
@@ -56,6 +56,7 @@ The below options are `action`s for async-profiler and common for both `asprof`
...
 | `--fdtransfer`     | `fdtransfer`      | Run a background process that provides access to perf_events to an unprivileged process. `--fdtransfer` is useful for profiling a process in a container (which lacks access to perf_events) from the host.<br>See [Profiling Java in a container](ProfilingInContainer.md).                                                                                                                                                                                                                                                                |
+| `--target-cpu`     | `target-cpu`      | In PerfEvents profiling mode, instruct the profiler to only sample threads running on the specified CPU, defaults to -1.<br>Example: `asprof --target-cpu 3`.                                                                                                                                                                                                                                                                                                                                                                               |
```
apangin, line 59: `PerfEvents` -> `perf_events` everywhere in the docs

src/perfEvents_linux.cpp
```diff
@@ -57,6 +57,7 @@ enum {
     HW_BREAKPOINT_X  = 4
 };
 
+int PerfEvents::_target_cpu = 0;
```
apangin, line 60: Place it next to other PerfEvents members in this file.

test/test/cpu/CpuTests.java
```diff
@@ -32,4 +43,52 @@ public void itimerTotal(TestProcess p) throws Exception {
...
+
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 1");
+        assertCloseTo(out.total(), 2_000_000_000, "perf_events total should match profiling duration");
+    }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
+    public void perfEventsWrongTargetCpuWithFdTransfer(TestProcess p) throws Exception {
+        pinCpu(p, 1);
+
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 2 --fdtransfer");
```
apangin, line 67: I wonder how this will work on a host with 2 CPUs only.
Maybe it's better to use CPUs 0,1 instead of 1,2?

src/perfEvents_linux.cpp
```diff
@@ -592,12 +593,12 @@ int PerfEvents::createForThread(int tid) {
 
     int fd;
     if (FdTransferClient::hasPeer()) {
-        fd = FdTransferClient::requestPerfFd(&tid, &attr);
+        fd = FdTransferClient::requestPerfFd(&tid, _target_cpu, &attr);
     } else {
-        fd = syscall(__NR_perf_event_open, &attr, tid, -1, -1, PERF_FLAG_FD_CLOEXEC);
+        fd = syscall(__NR_perf_event_open, &attr, tid, _target_cpu, -1, PERF_FLAG_FD_CLOEXEC);
```
apangin, line 598: It is worth passing `args._target_cpu` argument to `perf_event_open` call in `check()` function too.
This is to ensure `check` fails whenever `start` fails with the same arguments.

docs/ProfilerOptions.md
fandreuz: 7aaaa17575b5b6be9d2b2e204e650d05f77b7ee6

src/perfEvents_linux.cpp
fandreuz: 6004b68ce96e3585b1dceef606dfd1ce706cfbc6

src/perfEvents_linux.cpp
fandreuz: 52bccf8791352d47a0bd3e01336f412e5aba43c2

test/test/cpu/CpuTests.java
fandreuz: Good point, I reduced all CPU indexes by 1: 07cb40f

test/test/cpu/CpuTests.java
```diff
@@ -32,4 +43,52 @@ public void itimerTotal(TestProcess p) throws Exception {
...
+        Assert.isEqual(out.total(), 0, "perf_events total should be 0 when the wrong CPU is targeted");
+    }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
+    public void perfEventsRightTargetCpu(TestProcess p) throws Exception {
+        pinCpu(p, 0);
+
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 0");
+        assertCloseTo(out.total(), 2_000_000_000, "perf_events total should match profiling duration");
+    }
```
apangin, line 61: You could merge these tests into one to reduce test run time slightly.
The test can run a target process just once and then issue `p.profile` command twice with different `target-cpu`.

test/test/cpu/CpuTests.java
```diff
@@ -32,4 +43,52 @@ public void itimerTotal(TestProcess p) throws Exception {
...
+        Assert.isEqual(out.total(), 0, "perf_events total should be 0 when the wrong CPU is targeted");
+    }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
+    public void perfEventsRightTargetCpuWithFdTransfer(TestProcess p) throws Exception {
+        pinCpu(p, 0);
+
+        Output out = p.profile("-d 2 -e cpu -i 100ms --total -o collapsed --target-cpu 0 --fdtransfer");
+        assertCloseTo(out.total(), 2_000_000_000, "perf_events total should match profiling duration");
+    }
```
apangin, line 77: Same as above for this pair of tests.

test/test/cpu/CpuTests.java
fandreuz: 869b561edfd740e22e67988c617871ef4b2a86a3

test/test/cpu/CpuTests.java
fandreuz: 869b561edfd740e22e67988c617871ef4b2a86a3

test/test/cpu/CpuBurner.class
apangin, line 1: Extra file included in the PR

PR #1188, opened 2025-03-21
Support for OTLP Profile signals

fandreuz:
### Changes

- New output mode (`otlp`) to dump data in OpenTelemetry format
- New Java API to dump Async-Profiler data in OpenTelemetry format
- `protobuf.cpp`: Allow specifying the maximum bit count for the varint encoding the length of a nested message

### Motivation

OpenTelemetry recently added support for [_Profile_ signals](https://opentelemetry.io/blog/2024/profiling/). The end goal of the initiative is to provide a standard, vendor-agnostic format for profiling data. Adopting this protocol as one of our supported output formats will help our users integrating Async-Profiler in their workflow easily.

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1663,7 +1663,7 @@ static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const cha
 
 void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     using namespace Otlp;
-    ProtoBuffer otlp_buffer{OTLP_BUFFER_INITIAL_SIZE};
+    ProtoBuffer otlp_buffer(OTLP_BUFFER_INITIAL_SIZE);
     Index strings;
     Index functions;
 


src/profiler.cpp
```diff
@@ -1632,6 +1642,115 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+
+    int32_t locations_count = 0;
+    
+    FrameName fn(args, args._style & ~STYLE_ANNOTATE, _epoch, _thread_names_lock, _thread_names);
+    for (std::vector<CallTraceSample*>::const_iterator it = samples.begin(); it != samples.end(); ++it) {
+        CallTrace* trace = (*it)->acquireTrace();
+        if (trace == NULL || excludeTrace(&fn, trace)) continue;
+
+        Sample *profileSample = profile.add_sample();
+        profileSample->add_value(1);
```
fandreuz, line 1705: Error here

fandreuz: 1f3eb66058f3763f490a1c73fdac19d08a3e3373

src/javaApi.cpp
```diff
@@ -95,6 +95,33 @@ Java_one_profiler_AsyncProfiler_execute0(JNIEnv* env, jobject unused, jstring co
...
+        throwNew(env, "java/lang/IllegalArgumentException", error.message());
+        return NULL;
+    }
+    if (args.hasOutputFile()) {
+        throwNew(env, "java/lang/IllegalArgumentException", "executeAndGetBuffer calls should not specify an output file argument");
+    }
+
+    Log::open(args);
+
+    BufferTryNoCopyWriter out;
```
fandreuz, line 115: This will break if this API is used to call a `dump*` method whose output buffer points to a location which is not valid after it returns. Some alternatives:
- `switch` on `OUTPUT_*`, if it's `OUTPUT_OTLP` we can use `BufferTryNoCopyWriter`, otherwise we use `BufferWriter`
- Redefine `Writer::write` to have a third argument `bool try_no_copy = false`, which callers (i.e. `dumpOtlp`) can decide to set to `true` if they feel like it's safe to keep using the memory address after the dump method returns

src/profiler.cpp
```diff
@@ -1647,6 +1652,97 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        u64 counter = args._counter == COUNTER_SAMPLES ? (*it)->samples : (*it)->counter;
+        if (counter == 0) continue;
+
+        u64 num_frames = trace->num_frames;
+        for (u64 j = 0; j < num_frames; ++j) {
+            u64 function_idx = maybeAddToIdxMap(function_idx_map, functions_vec, fn.name(trace->frames[j]));
+            _otlp_buffer.field(Otlp::Profile::location_indices, function_idx);
+        }
+
+        protobuf_mark_t sample_mark = _otlp_buffer.startMessage(Otlp::Profile::sample);
```
apangin, line 1712: Your proto writer does not seem to benefit from `repeated` fields - it always writes header for each single message.

src/javaApi.cpp
fandreuz: > DirectByteBuffers are tricky: when large buffers are created often, their interference with GC can be surprising.

Even if they don't manage the underlying memory? [NewDirectByteBuffer](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#NewDirectByteBuffer) gives us a `ByteBuffer` which should be a dummy wrapper around natively managed memory, it should not allocate nor manage any memory.

> I have an idea how to make the API clean and safe.

I'm preparing a document to discuss the Java API signature and implementation, let's discuss your idea tomorrow so we can consider multiple options. In addition to "clean and safe", it would be nice to avoid unnecessary copies.

> But the method should return byte array, IMO.

I considered `SetByteArrayRegion`, but this involves a copy, and the same seems to hold for `GetByteArrayElements`/`ReleaseByteArrayElements`. `GetPrimitiveArrayCritical` seems to have weird semantics and I didn't want to go there. Do you have any insights on these functions?

src/profiler.cpp
fandreuz: Right, I always use the `expanded` mode for both primitive and LEN-type values: https://protobuf.dev/editions/features/#repeated_field_encoding

This won't pose any problem on the reader side, but it's definitely something we could improve in the future. I didn't want to overcomplicate the interface, but I'm keeping track of this and other future improvements (like the optimistic encoding of nested messages which I proposed [here](https://github.com/async-profiler/async-profiler/pull/1292#discussion_r2088484573))

src/javaApi.cpp
fandreuz: Changed to use `byte[]` in 11a5f1fd47592a339660dbd0a9918e105a04bf06

Still, there's two copies, when only one would be needed. I think the options [I mentioned above](https://github.com/async-profiler/async-profiler/pull/1188#discussion_r2098397844) are still relevant, and something like `BufferTryNoCopyWriter` (with better naming) could be helpful

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -271,5 +286,7 @@ private void filterThread(Thread thread, boolean enable) {
 
     private native String execute0(String command) throws IllegalArgumentException, IllegalStateException, IOException;
 
+    private native byte[] executeAndGetByteArray0(String command) throws IllegalArgumentException, IllegalStateException, IOException;
```
fandreuz, line 289: I'm happy to hear alternatives for the name of this method, I could not think of anything better

test/test/api/DumpOtlp.java
```diff
@@ -0,0 +1,28 @@
...
+    public static void main(String[] args) throws Exception {
+        AsyncProfiler.getInstance().start(Events.CPU, 1_000_000);
+
+        for (int i = 0; i < 5; i++) {
+            method1();
+            method2();
+            method3();
+        }
+
+        // TODO: Should we test this further?
```
fandreuz, line 24: https://quip-amazon.com/MG90AaybxemJ/Async-Profiler-OTEL-Java-API-design#temp:C:HbB654d1ef0c5994bae89ffa2434

src/api/one/profiler/AsyncProfiler.java
apangin: `execute1`?
Look, this is a private method, which does the same as `execute0` except for its return type - why should it be called differently? And yes, this is not uncommon: [1](https://github.com/openjdk/jdk/blob/2dfbf41d2a3dbcd44f9ed9a58a1b0932d7536977/src/java.base/share/classes/java/lang/ClassLoader.java#L1057-L1062), [2](https://github.com/openjdk/jdk/blob/2dfbf41d2a3dbcd44f9ed9a58a1b0932d7536977/src/java.management/share/classes/sun/management/ThreadImpl.java#L516-L521)

src/javaApi.cpp
```diff
@@ -95,6 +95,36 @@ Java_one_profiler_AsyncProfiler_execute0(JNIEnv* env, jobject unused, jstring co
...
+    const char* command_str = env->GetStringUTFChars(command, NULL);
+    Error error = args.parse(command_str);
+    env->ReleaseStringUTFChars(command, command_str);
+
+    if (error) {
+        throwNew(env, "java/lang/IllegalArgumentException", error.message());
+        return NULL;
+    }
+    if (args.hasOutputFile()) {
+        throwNew(env, "java/lang/IllegalArgumentException", "executeAndGetBuffer calls should not specify an output file argument");
```
apangin, line 110: `return` is missing.
Throwing an exception from JNI does not automatically bail out of native method.

src/profiler.cpp
```diff
@@ -1647,6 +1652,104 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
     }
 }
 
+/*
+ * If the value has been seen already, return its idx in `values`. If not, update the map,
+ * add the value to `values`, and return the index.
+ */
+static inline u64 getOrSetIndex(
+        std::unordered_map<std::string, u64>& idx_map,
```
apangin, line 1660: u32? It's unrealistic to have more than 4B elements.

src/profiler.cpp
```diff
@@ -1647,6 +1652,104 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        std::vector<std::string>& values,
```
apangin, line 1661: There is no need to fill a vector right here - it's always possible to reconstruct a vector from the map later when it is needed.

fandreuz: Sure, but can we obtain an ordered list of values without sorting the content of the map?

apangin: Yes, just iterate over `{key, value}` pairs in the map and use `value` as an array index for `key`.
Here is [Java equivalent](https://github.com/async-profiler/async-profiler/blob/master/src/converter/one/convert/Index.java#L42).

fandreuz: Ah right, didn't think about that. Thanks

fandreuz: e9837729ca6146eb5c9d4f54f6662af307537b1c

src/profiler.cpp
fandreuz: 3fbdc061c5d4a003bf521a70cf3a7ca332c1fe47

src/javaApi.cpp
fandreuz: 9866f27bdf4e904aa4581e884d7a051aac48da47

test/test/api/DumpOtlp.java
fandreuz, line 24: Yes: #1331

src/profiler.h
```diff
@@ -64,6 +65,7 @@ class Profiler {
     ThreadFilter _thread_filter;
     CallTraceStorage _call_trace_storage;
     FlightRecorder _jfr;
+    ProtoBuffer _otlp_buffer{0};
```
apangin: It's not unsed outside `dumpOtlp`, can we make it local?

src/protobuf.cpp
```diff
@@ -34,18 +34,22 @@ void ProtoBuffer::ensureCapacity(size_t new_data_size) {
     _data = (unsigned char*) realloc(_data, _capacity);
 }
 
-void ProtoBuffer::putVarInt(u64 n) {
-    _offset = putVarInt(_offset, n);
+void ProtoBuffer::putVarInt(u32 n) {
```
apangin: Why did you add an overload for `u32` when it's implementation is exactly the same as for `u64`?

src/protobuf.cpp
```diff
@@ -95,3 +104,15 @@ void ProtoBuffer::commitMessage(protobuf_mark_t mark) {
...
+    putVarInt((u32) b);
+}
+
+void ProtoBuffer::appendRepeated(u32 n) {
+    putVarInt(n);
+}
+
+void ProtoBuffer::appendRepeated(u64 n) {
+    putVarInt(n);
+}
```
apangin: I see no sense in all these extra functions. Why not just call `putVarInt` directly where needed?

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+/*
+ * If the value has been seen already, return its idx in `values`.
+ * If not, update the map and return the index.
+ */
+static inline u32 getOrSetIndex(
+        std::unordered_map<std::string, u32>& idx_map,
+        const std::string& value
+) {
+    return idx_map.insert({value, idx_map.size()}).first->second;
+}
```
apangin: Can we have a class (let's say, `Index`, similarly to `one.convert.Index` in Java) that encapsulates this logic?
This class could be probably reused for Element names in `jfrMetadata`.

And as I mentioned in a previous review, `getOrSet` name is not good. What the function does is unambiguous: it always obtains an index of a string, there is no "or" in its behavior. The fact that the function updates internal map as a side effect is an implementation detail.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+
+void Profiler::dumpOtlp(Writer& out, Arguments& args) {
+    // _otlp_buffer contains a ProfileData message
+    _otlp_buffer.reset();
+
+    std::unordered_map<std::string, u32> string_idx_map;
+    std::unordered_map<std::string, u32> function_idx_map;
+
+    protobuf_mark_t resource_profiles_mark = _otlp_buffer.startMessage(Otlp::ProfilesData::resource_profiles);
```
apangin: Namespaces here make sense, but they make identifiers way too long. Can we shorten them somehow? Perhaps we can get rid of `Otlp::` at least.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    protobuf_mark_t scope_profiles_mark = _otlp_buffer.startMessage(Otlp::ResourceProfiles::scope_profiles);
+    protobuf_mark_t profile_mark = _otlp_buffer.startMessage(Otlp::ScopeProfiles::profiles);
+
+    _otlp_buffer.field(Otlp::Profile::period, (u64) args._interval);
+
+    protobuf_mark_t sample_type_mark = _otlp_buffer.startMessage(Otlp::Profile::sample_type);
+    _otlp_buffer.field(Otlp::ValueType::type_strindex, getOrSetIndex(string_idx_map, _engine->type()));
+    _otlp_buffer.field(Otlp::ValueType::unit_strindex, getOrSetIndex(string_idx_map, _engine->units()));
```
apangin: Units are valid only for `--total` option. Otherwise it's just a counter of samples.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    u32 samples_num_frames_cache[samples.size()];
+    size_t samples_idx = 0;
+
+    FrameName fn(args, args._style & ~STYLE_ANNOTATE, _epoch, _thread_names_lock, _thread_names);
+    protobuf_mark_t location_indices_mark = _otlp_buffer.startMessage(Otlp::Profile::location_indices);
+    for (auto it = samples.begin(); it != samples.end(); ++it) {
+        CallTrace* trace = (*it)->acquireTrace();
+        if (trace == NULL || excludeTrace(&fn, trace)) continue;
+
+        samples_counter_cache[samples_idx] = args._counter == COUNTER_SAMPLES ? (*it)->samples : (*it)->counter;
```
apangin: Why choose? Can we record both? I think OTLP format allows multiple values per sample.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    _otlp_buffer.field(Otlp::ValueType::type_strindex, getOrSetIndex(string_idx_map, _engine->type()));
+    _otlp_buffer.field(Otlp::ValueType::unit_strindex, getOrSetIndex(string_idx_map, _engine->units()));
+    _otlp_buffer.field(Otlp::ValueType::aggregation_temporality, Otlp::AggregationTemporality::cumulative);
+    _otlp_buffer.commitMessage(sample_type_mark);
+
+    std::vector<CallTraceSample*> samples;
+    _call_trace_storage.collectSamples(samples);
+
+    u32 samples_counter_cache[samples.size()];
+    u32 samples_num_frames_cache[samples.size()];
```
apangin: It's not "cache", this is just a temporary holder. `samples_counter` and `samples_num_frames` will be good enough.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        for (u64 j = 0; j < samples_num_frames_cache[samples_idx]; ++j) {
+            u64 function_idx = getOrSetIndex(function_idx_map, fn.name(trace->frames[j]));
+            _otlp_buffer.appendRepeated(function_idx);
+        }
+        ++samples_idx;
+    }
+    _otlp_buffer.commitMessage(location_indices_mark);
+
+    u64 frames_seen = 0;
+    for (samples_idx = 0; samples_idx < samples.size() && samples_counter_cache[samples_idx] > 0; samples_idx++) {
```
apangin: This will break the loop on the very first skipped sample.

src/profiler.cpp
```diff
@@ -1647,6 +1652,124 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    // Write mapping_table
+    // TODO: Include per-binary debug info
+    protobuf_mark_t mapping_mark = _otlp_buffer.startMessage(Otlp::ProfilesDictionary::mapping_table);
+    _otlp_buffer.commitMessage(mapping_mark);
+
+    // Convert function map (name to index) to an array (index to name mapping) for fast ordered access
+    const std::string* function_arr[function_idx_map.size()];
+    for (const auto& it : function_idx_map) {
+        function_arr[it.second] = &it.first;
+    }
```
apangin: Can we have this logic in the `Index` class mentioned above?

src/profiler.h
fandreuz: The idea would be to reuse it for multiple calls to `dumpOtlp` to save some time. Do you think it's not worth it? The buffer inside may be resized multiple times during `dumpOtlp` if it's too small.

src/protobuf.cpp
fandreuz: The idea was to save some stack memory by not having a `u64` when a `u32` is enough, if it's not worth the extra effort I'll remove it

src/protobuf.cpp
fandreuz: See comment above, I'll remove it

src/profiler.cpp
fandreuz: Maybe we could have a `using namespace Otlp` inside `dumpOtlp`?

src/profiler.cpp
fandreuz: Good point, will do

src/profiler.cpp
fandreuz: 93b4189c57119791275cf6bebb17571be243e911

src/profiler.cpp
fandreuz: 93b4189c57119791275cf6bebb17571be243e911

src/protobuf.cpp
fandreuz: 93b4189c57119791275cf6bebb17571be243e911

src/protobuf.cpp
fandreuz: 93b4189c57119791275cf6bebb17571be243e911

src/profiler.cpp
fandreuz: What do you mean by `skipped sample`? If you check the code a few lines above, I set `samples_counter[samples_idx]` to the counter for the current sample, and `continue` if the value is zero. Thus samples with counter zero are not going to be recorded, and the same index will be reused for the next sample. Do you disagree with this approach?

src/profiler.cpp
fandreuz: 266db578fe7005429b00b595081d2f230e7501c8

src/profiler.cpp
fandreuz: > And as I mentioned in a previous review, getOrSet name is not good. What the function does is unambiguous: it always obtains an index of a string, there is no "or" in its behavior. The fact that the function updates internal map as a side effect is an implementation detail.

I don't agree with this comment, the `Set` part of the method name is relevant because it informs the caller about a potential side effect (which affects all the next calls with the same value). I agree this is not visible from outside (as you say, it's an implementation detail how we do it), but I would be surprised to see a `getIndex` which is not `const` or thread safe. I found other examples of such a pattern (e.g. Kotlin's [thread local](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent/get-or-set.html)).

I'll rename it to something like `recordIndex` to remove any ambiguity.

fandreuz: Can you check 0ea647b581b0b7be0f0155459fa9bf62c08b741a @apangin ?

src/profiler.cpp
fandreuz: 0ea647b581b0b7be0f0155459fa9bf62c08b741a

src/profiler.cpp
fandreuz: 266db578fe7005429b00b595081d2f230e7501c8

src/index.h
```diff
@@ -0,0 +1,34 @@
...
+  public:
+    u32 recordIndex(std::string value) {
+      return idx_map.insert({value, idx_map.size()}).first->second;
+    }
+
+    u32 size() const {
+      return idx_map.size();
+    }
+
+    void forEachInOrder(std::function<void(const std::string&)> consumer) const {
```
fandreuz: I thought about returning an `std::vector<std::string>` here instead of getting an `std::function` from the user. But this means a copy of the content of the whole map.

Another option could have been `std::vector<std::string*>`, pointing to the strings inside `idx_map`. But then we have an unsafe API if `IndexContainer` is destroyed before the output.

src/index.h
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <functional>
+#include <string>
+#include <unordered_map>
+#include "arch.h"
+
+// Keeps track of values seen and their index of occurrence
+class IndexContainer {
```
apangin: Why `Container`? Try to avoid [enterprisy names](https://projects.haykranen.nl/java/).
It can be `StringIndex` or simply `Index`.

src/index.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
apangin, line 4: All include files should have `#ifndef` double guard against double inclusion.

src/index.h
```diff
@@ -0,0 +1,34 @@
...
+#include "arch.h"
+
+// Keeps track of values seen and their index of occurrence
+class IndexContainer {
+  private:
+    std::unordered_map<std::string, u32> idx_map;
+  
+  public:
+    u32 recordIndex(std::string value) {
+      return idx_map.insert({value, idx_map.size()}).first->second;
```
apangin: Indent size

src/index.h
apangin: forEach approach is fine, but I'd call it `forEachOrdered` similar to existing Java Stream API.

src/profiler.h
apangin: I think it's acceptable. Let's make a buffer local with a reasonable initial size.

src/protobuf.cpp
apangin: The function does not take stack memory, all computation is in registers.

src/profiler.h
fandreuz: cbb5604f3ccd1e6df8e44f98ecf526ef2ee271c4

src/index.h
fandreuz: 6629adcfc479fb3816a9c7574892a013787498f3

src/index.h
fandreuz: Right, missed that: 25fe78f766d033ad0afac6c08577b1c12f369ba6

src/index.h
fandreuz: Fixed: 362320c50570f534b8f47852baeb5e7de895f75f

Thanks for the link, it'll be much easier to decide class names now

src/index.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <functional>
+#include <string>
+#include <unordered_map>
+#include "arch.h"
+
+// Keeps track of values seen and their index of occurrence
+class Index {
+  private:
+    std::unordered_map<std::string, u32> idx_map;
```
apangin: private class fields start with `_`

src/profiler.cpp
```diff
@@ -11,6 +11,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/param.h>
+#include <unordered_map>
```
apangin: not used

src/profiler.cpp
```diff
@@ -1647,6 +1653,125 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    otlp_buffer.field(Profile::period, (u64) args._interval);
+
+    // TODO: Test this once it gets reviewed
+    if (args._counter == COUNTER_TOTAL) {
+        recordSampleType(otlp_buffer, _engine, string_idx_container, _engine->units());
+        recordSampleType(otlp_buffer, _engine, string_idx_container, "count");
+    } else if (args._counter == COUNTER_SAMPLES) {
+        recordSampleType(otlp_buffer, _engine, string_idx_container, "count");
+        recordSampleType(otlp_buffer, _engine, string_idx_container, _engine->units());
+    }
```
apangin: Why different order? What does it affect?

src/profiler.cpp
```diff
@@ -1647,6 +1653,125 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+
+    std::vector<CallTraceSample*> samples;
+    _call_trace_storage.collectSamples(samples);
+
+    u32 samples_counter[samples.size()][2];
```
apangin: Number of samples can be large; do not do unbounded stack allocations.

src/protobuf.cpp
```diff
@@ -95,3 +95,7 @@ void ProtoBuffer::commitMessage(protobuf_mark_t mark) {
     }
     _data[mark - 1] = (unsigned char) message_length;
 }
+
+void ProtoBuffer::appendRepeated(u64 n) {
+    putVarInt(n);
+}
```
apangin: I meant `appendRepeated` is useless. Just call `putVarInt` directly.

src/profiler.cpp
```diff
@@ -1647,6 +1653,125 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        for (u64 j = 0; j < samples_num_frames[samples_idx]; ++j) {
+            u64 function_idx = function_idx_container.recordIndex(fn.name(trace->frames[j]));
+            otlp_buffer.appendRepeated(function_idx);
+        }
+        ++samples_idx;
+    }
+    otlp_buffer.commitMessage(location_indices_mark);
+
+    u64 frames_seen = 0;
+    for (samples_idx = 0; samples_idx < samples.size() && samples_counter[samples_idx][0] > 0; samples_idx++) {
```
apangin: This `samples_counter > 0` condition confused me, cross out my previous comment.
What about having a simple for-loop from `0` to `samples_idx`? (probably giving a better name for `samples_idx`)
Or, considering my comment about stack allocations, `samples_counter` could be a vector instead.

src/profiler.cpp
```diff
@@ -1647,6 +1653,125 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative);
+    otlp_buffer.commitMessage(sample_type_mark);
+}
+
+void Profiler::dumpOtlp(Writer& out, Arguments& args) {
+    using namespace Otlp;
+    ProtoBuffer otlp_buffer{OTLP_BUFFER_INITIAL_SIZE};
+
+    Index string_idx_container;
+    Index function_idx_container;
```
apangin: Containers again?

src/profiler.cpp
```diff
@@ -1647,6 +1653,125 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    // Write mapping_table
+    // TODO: Include per-binary debug info
+    protobuf_mark_t mapping_mark = otlp_buffer.startMessage(ProfilesDictionary::mapping_table);
+    otlp_buffer.commitMessage(mapping_mark);
+
+    // Write function_table
+    function_idx_container.forEachOrdered([&] (const std::string& function_name) {
+        protobuf_mark_t function_mark = otlp_buffer.startMessage(ProfilesDictionary::function_table);
+        u32 function_name_strindex = string_idx_container.recordIndex(function_name);
+        otlp_buffer.field(Function::name_strindex, function_name_strindex);
```
apangin: IMO, thoughtfully chosen short names will make code more readable. Compare two lines above with:
```
proto.field(Function::name_strindex, strings.indexOf(function_name));
```

src/profiler.cpp
fandreuz: The reasoning here is that if the user uses the flag `--total`, it would be weird to have samples count at index zero, which is usually the default sample type.

src/profiler.cpp
fandreuz: I'm a bit puzzled by the class name still. `Index` looks like a simple index to me, not a "container of indexes". `Index::recordIndex` feels weird.

I like `Index strings` and `Index functions` though, I'll rename them

src/profiler.cpp
apangin: `--total` is meaningless for output formats that support multiple counters per sample. It has no effect for JFR, and I don't think it should have for OTLP.

src/profiler.cpp
fandreuz: bf717034fb615ac130e0f11d5eb07df7b549d5e4

src/index.h
fandreuz: d1fe7dcf0a7e1d56b1bd33877a9d3552805a69bd

src/protobuf.cpp
fandreuz: 9166b3515632c39e3d62f761aefd08ef57417879

src/profiler.cpp
fandreuz: dde4a1a75c086869cae15d79a498a3059a343a09

src/profiler.cpp
fandreuz: dde4a1a75c086869cae15d79a498a3059a343a09

src/profiler.cpp
fandreuz: Good point: a47f6b2e53b259edb55c52f2611c89271905b87c

src/profiler.cpp
fandreuz: Right, thanks: a47f6b2e53b259edb55c52f2611c89271905b87c

src/profiler.cpp
fandreuz: 5ac9ac6fc9389d75a46f520bd43e7153647fcb11

src/index.h
```diff
@@ -0,0 +1,39 @@
...
+    u32 indexOf(std::string value) {
+        return _idx_map.insert({value, _idx_map.size()}).first->second;
+    }
+
+    u32 size() const {
+        return _idx_map.size();
+    }
+
+    void forEachOrdered(std::function<void(const std::string&)> consumer) const {
+        const std::string* arr[_idx_map.size()];
```
apangin: This again can be too large for on-stack allocation.

src/profiler.cpp
```diff
@@ -1647,6 +1652,108 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
     }
 }
 
+static void recordSampleType(ProtoBuffer& otlp_buffer, Engine* engine, Index& strings, const char* unit) {
+    using namespace Otlp;
+    protobuf_mark_t sample_type_mark = otlp_buffer.startMessage(Profile::sample_type);
+    otlp_buffer.field(ValueType::type_strindex, strings.indexOf(engine->type()));
```
apangin: Pass `const char* type` instead.
`engine` does not look natural in this function.

src/profiler.cpp
```diff
@@ -1647,6 +1652,108 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    Index functions;
+
+    protobuf_mark_t resource_profiles_mark = otlp_buffer.startMessage(ProfilesData::resource_profiles);
+    protobuf_mark_t scope_profiles_mark = otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
+    protobuf_mark_t profile_mark = otlp_buffer.startMessage(ScopeProfiles::profiles);
+
+    recordSampleType(otlp_buffer, _engine, strings, "count");
+    recordSampleType(otlp_buffer, _engine, strings, _engine->units());
+
+    std::vector<CallTraceSample*> callTraceSamples;
```
apangin: call_trace_samples

src/profiler.cpp
```diff
@@ -1647,6 +1652,108 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    _call_trace_storage.collectSamples(callTraceSamples);
+
+    struct SampleInfo {
+        u64 counter;
+        u64 samples;
+        u32 num_frames;
+    };
+    std::vector<SampleInfo> samplesInfo;
```
apangin: `samples_info`

src/profiler.cpp
```diff
@@ -1647,6 +1652,108 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        protobuf_mark_t function_mark = otlp_buffer.startMessage(ProfilesDictionary::function_table);
+        otlp_buffer.field(Function::name_strindex, strings.indexOf(function_name));
+        otlp_buffer.commitMessage(function_mark);
+    });
+
+    // Write location_table
+    for (u64 function_idx = 0; function_idx < functions.size(); ++function_idx) {
+        protobuf_mark_t location_mark = otlp_buffer.startMessage(ProfilesDictionary::location_table);
+        // TODO: Fix me when more Mappings are added
+        otlp_buffer.field(Location::mapping_index, (u64) 0);
```
apangin: The field is optional. If you don't use it, don't include it.

src/profiler.cpp
```diff
@@ -1647,6 +1652,108 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    }
+    otlp_buffer.commitMessage(location_indices_mark);
+
+    u64 frames_seen = 0;
+    for (const SampleInfo& sampleInfo : samplesInfo) {
+        protobuf_mark_t sample_mark = otlp_buffer.startMessage(Profile::sample);
+        otlp_buffer.field(Sample::locations_start_index, frames_seen);
+        otlp_buffer.field(Sample::locations_length, sampleInfo.num_frames);
+
+        protobuf_mark_t sample_value_mark = otlp_buffer.startMessage(Sample::value);
```
apangin: There are many tiny messages whose length can be encoded in 1 byte.
Do you want to provide a version of `startMessage` (or add an optional argument) to specify maximum length?

fandreuz: Yes, the idea is to have an optional argument eventually. Do you want this in this PR?

src/profiler.cpp
fandreuz: 267d13bb44bdfca139e1262d8e69546a44a04552

src/profiler.cpp
fandreuz: be5f41394176d7d101805a3fbb023784759b8104

src/profiler.cpp
fandreuz: 45951727b82f5220103edba7a2b702a7cd4eed88

src/profiler.cpp
fandreuz: 45951727b82f5220103edba7a2b702a7cd4eed88

src/index.h
fandreuz: 45951727b82f5220103edba7a2b702a7cd4eed88

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -271,5 +288,7 @@ private void filterThread(Thread thread, boolean enable) {
 
     private native String execute0(String command) throws IllegalArgumentException, IllegalStateException, IOException;
 
+    private native byte[] execute1(String command) throws IllegalArgumentException, IllegalStateException, IOException;
```
apangin, line 290: Don't forget to include new method in `profiler_natives` list. This is necessary to make AsyncProfiler shadable.

src/index.h
```diff
@@ -0,0 +1,40 @@
...
+    u32 indexOf(std::string value) {
+        return _idx_map.insert({value, _idx_map.size()}).first->second;
+    }
+
+    u32 size() const {
+        return _idx_map.size();
+    }
+
+    void forEachOrdered(std::function<void(const std::string&)> consumer) const {
+        std::vector<const std::string*> arr{_idx_map.size()};
```
apangin: There is a subtle difference between `{}` and `()` initializers. In this case, `()` is more appropriate. For example, if it was `std::vector<u64>`, your code would break. I generally prefer `()` everywhere.

src/profiler.cpp
```diff
@@ -1647,6 +1652,106 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        samples_info.push_back(SampleInfo{call_trace_sample->samples, call_trace_sample->counter, (u32) trace->num_frames});
+        for (u32 j = 0; j < trace->num_frames; ++j) {
+            u32 function_idx = functions.indexOf(fn.name(trace->frames[j]));
+            otlp_buffer.putVarInt(function_idx);
+        }
+    }
+    otlp_buffer.commitMessage(location_indices_mark);
+
+    u64 frames_seen = 0;
+    for (const SampleInfo& sampleInfo : samples_info) {
```
apangin: `sample_info` or maybe `info` or even `si`.
Don't hesitate to shorten local variable names when they are unambiguous and used in a narrow scope.

src/profiler.cpp
apangin: Yeah, that would be a more valuable optimization than reusing a buffer :)

src/profiler.cpp
```diff
@@ -1647,6 +1652,106 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    otlp_buffer.commitMessage(profile_mark);
+    otlp_buffer.commitMessage(scope_profiles_mark);
+    otlp_buffer.commitMessage(resource_profiles_mark);
+
+    protobuf_mark_t dictionary_mark = otlp_buffer.startMessage(ProfilesData::dictionary);
+
+    // Write mapping_table
+    // TODO: Include per-binary debug info
+    protobuf_mark_t mapping_mark = otlp_buffer.startMessage(ProfilesDictionary::mapping_table);
+    otlp_buffer.commitMessage(mapping_mark);
```
apangin: Not needed right now?
You may leave `// TODO: Write mapping_table` if you wish

src/profiler.h
```diff
@@ -9,6 +9,7 @@
 #include <map>
 #include <string>
 #include <time.h>
+#include "protobuf.h"
```
apangin: Should be in .cpp instead

src/profiler.cpp
fandreuz: It is needed, some parsers (e.g. [Pyroscope](https://github.com/grafana/pyroscope/blob/main/pkg/ingester/otlp/convert.go#L192)) will complain because they need at least one default mapping to be present. But indeed it's not enforced by the spec, so maybe it's something to be improved on their side. I'll report it there.

I'd leave it as it is since for now until this is clarified, unless you have a strong opinion against this

apangin: OK, that makes sense. Leave a comment then explaining why there is an empty table.

src/index.h
fandreuz: Ouch I got confused here, thanks for catching it: 975bd6fef52eae095865ce570c8d80d97530ab9a

src/profiler.h
fandreuz: 975bd6fef52eae095865ce570c8d80d97530ab9a

src/profiler.cpp
fandreuz: 975bd6fef52eae095865ce570c8d80d97530ab9a

src/api/one/profiler/AsyncProfiler.java
fandreuz: a552fde6919a8a5cf4460df55f3a762f5705d537

fandreuz: what do you mean by shadable?

src/profiler.cpp
fandreuz: Check 84d0ce1c178dc083983bdafad03675e73c5d1eac and e52f0bed1ae621624a4fc0344a36ad61661d6f6a

src/api/one/profiler/AsyncProfiler.java
apangin: Shading a library means renaming/moving classes to other package in order to embed dependencies and to avoid potential version conflicts. libasyncProfiler automatically detects if it was loaded from a renamed Java class and binds its native methods to that class.

src/protobuf.cpp
```diff
@@ -78,20 +69,35 @@ void ProtoBuffer::field(protobuf_index_t index, const unsigned char* s, size_t l
...
-void ProtoBuffer::commitMessage(protobuf_mark_t mark) {
-    size_t message_length = _offset - mark;
-    for (size_t i = 0; i < NESTED_FIELD_BYTE_COUNT - 1; ++i) {
-        size_t idx = mark - NESTED_FIELD_BYTE_COUNT + i;
-        _data[idx] = (unsigned char) (0x80 | (message_length & 0x7f));
-        message_length >>= 7;
+void ProtoBuffer::commitMessage(const protobuf_mark_t& mark) {
+    size_t actual_len = _offset - (mark.message_start + mark.expected_len_byte_count);
+    size_t actual_len_byte_count = varIntSize(actual_len);
+    if (actual_len_byte_count > mark.expected_len_byte_count) {
```
apangin: This should be an assertion. If reserved space was smaller than required, treat it as error.
Otherwise we may end up moving 100MB blocks without knowing there was a computation mistake.

src/protobuf.h
```diff
@@ -14,7 +14,10 @@ protobuf_t VARINT = 0;
 protobuf_t LEN = 2;
 
 typedef u32 protobuf_index_t;
-typedef u32 protobuf_mark_t;
+struct protobuf_mark_t {
+    size_t message_start;
+    size_t expected_len_byte_count;
+};
```
apangin: It's not a good practice to return a structure larger than machine word from a function.
64-bit field for a value that needs at most 3 is too much anyway.
It would be natural (in system programming world) to fit `protobuf_mark_t` in a machine word or in `u64` and pass it by value: 3 bits for `len_bytes` and remaining 61 bits for `message_start`. You may use either bit-field struct or just `u64` and do manipulations by yourself.

src/protobuf.h
```diff
@@ -41,18 +41,18 @@ class ProtoBuffer {
...
-    void field(protobuf_index_t index, bool b);
     void field(protobuf_index_t index, u64 n);
     // LEN
     void field(protobuf_index_t index, const char* s);
     void field(protobuf_index_t index, const char* s, size_t len);
     void field(protobuf_index_t index, const unsigned char* s, size_t len);
 
-    protobuf_mark_t startMessage(protobuf_index_t index);
-    void commitMessage(protobuf_mark_t mark);
+    protobuf_mark_t startMessage(protobuf_index_t index, size_t len_byte_count = NESTED_FIELD_BYTE_COUNT);
```
apangin: 3 bytes limit us to 2MiB profiles, which is likely not enough for long profiling sessions.
I'd make it 4 or even 5 bytes by default to be absolutely safe, but then carefully revise all uses of `startMessage` to have a proper estimation.

src/profiler.cpp
```diff
@@ -1647,6 +1653,110 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
     }
 }
 
+static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const char* type, const char* units) {
+    using namespace Otlp;
+    protobuf_mark_t sample_type_mark = otlp_buffer.startMessage(Profile::sample_type);
```
apangin: 1 byte

src/profiler.cpp
```diff
@@ -1647,6 +1653,110 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+    protobuf_mark_t dictionary_mark = otlp_buffer.startMessage(ProfilesData::dictionary);
+
+    // Write mapping_table
+    // TODO: Include per-binary debug info
+    protobuf_mark_t mapping_mark = otlp_buffer.startMessage(ProfilesDictionary::mapping_table, 1);
+    otlp_buffer.commitMessage(mapping_mark);
+
+    // Write function_table
+    functions.forEachOrdered([&] (const std::string& function_name) {
+        protobuf_mark_t function_mark = otlp_buffer.startMessage(ProfilesDictionary::function_table);
```
apangin: 1 byte

src/profiler.cpp
```diff
@@ -1647,6 +1653,110 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
+        otlp_buffer.field(Function::name_strindex, strings.indexOf(function_name));
+        otlp_buffer.commitMessage(function_mark);
+    });
+
+    // Write location_table
+    for (u64 function_idx = 0; function_idx < functions.size(); ++function_idx) {
+        protobuf_mark_t location_mark = otlp_buffer.startMessage(ProfilesDictionary::location_table);
```
apangin: 1 byte

src/protobuf.cpp
fandreuz: I don't agree with this. We cannot estimate the size of the message without low level (and error prone) computations at runtime. And any such estimation would be fragile and hard to change later. Any fixed size we decide now may be not enough for long profiling sessions. But either case, I don't think it's a "computation mistake". 

Are you implying we should have an exact estimate of the size of the message, before we start writing it to the buffer?

src/protobuf.h
fandreuz: Check 17084feb82647693e3b5f35c0e326e1fbe1933a8

src/protobuf.h
fandreuz: 2c6bd7253a5aa7f51c1480112fcd4b88bda08f2a

src/profiler.cpp
fandreuz: 2d8648ad807494093125c7dc72f6d29e7a795689

src/profiler.cpp
fandreuz: 2d8648ad807494093125c7dc72f6d29e7a795689

src/profiler.cpp
fandreuz: 2d8648ad807494093125c7dc72f6d29e7a795689

src/protobuf.cpp
apangin: OK, let's call it "underestimation" rather than "computation mistake". I'm saying that `startMessage` should accept the _maximum_ size of a message instead of an _estimation_. 5 bytes (which can encode up to 32 GiB messages) is sufficient for all use cases. So, let's use 5 bytes default for the top-level messages and provide an appropriate upper bound for inner ones. Exact estimate is not required.

fandreuz: 21d82118fb2f9407043c009715cd060b056dd2e5

src/protobuf.h
```diff
@@ -41,18 +39,18 @@ class ProtoBuffer {
...
     void field(protobuf_index_t index, u64 n);
     // LEN
     void field(protobuf_index_t index, const char* s);
     void field(protobuf_index_t index, const char* s, size_t len);
     void field(protobuf_index_t index, const unsigned char* s, size_t len);
 
-    protobuf_mark_t startMessage(protobuf_index_t index);
-    void commitMessage(protobuf_mark_t mark);
+    protobuf_mark_t startMessage(protobuf_index_t index, size_t max_len_byte_count = NESTED_FIELD_BYTE_COUNT);
+    void commitMessage(const protobuf_mark_t& mark);
```
apangin: It does not need to be a reference, right?

src/arguments.cpp
```diff
@@ -200,6 +200,9 @@ Error Arguments::parse(const char* args) {
                 _output = OUTPUT_TEXT;
                 _dump_flat = value == NULL ? INT_MAX : atoi(value);
 
+            CASE("otlp")
```
apangin, line 204: Do you mind updating all relevant docs and examples with new option?

src/profiler.cpp
```diff
@@ -29,6 +30,8 @@
 #include "fdtransferClient.h"
 #include "frameName.h"
 #include "os.h"
+#include "otlp.h"
+#include "protobuf.h"
```
apangin: I guess protobuf.h is redundant here, since otlp.h already includes it.

src/protobuf.cpp
```diff
@@ -78,20 +70,27 @@ void ProtoBuffer::field(protobuf_index_t index, const unsigned char* s, size_t l
...
+void ProtoBuffer::commitMessage(const protobuf_mark_t& mark) {
+    size_t max_len_byte_count = mark & 7;
+    size_t message_start = mark >> 3;
+
+    size_t actual_len = _offset - (message_start + max_len_byte_count);
+    assert(varIntSize(actual_len) <= max_len_byte_count);
+
+    for (size_t i = 0; i < max_len_byte_count - 1; ++i) {
+        size_t idx = message_start + i;
+        _data[idx] = (unsigned char) (0x80 | (actual_len & 0x7f));
```
apangin: `& 0x7f` is redundant

src/arguments.cpp
fandreuz: Sure, let me know if I forgot something: a90b3471de557f674963511eb3292fe4203c2d9d

src/profiler.cpp
fandreuz: e5e8d3312e1359bc093e7e5562020723cd30dacc

src/protobuf.cpp
fandreuz: e5e8d3312e1359bc093e7e5562020723cd30dacc

src/protobuf.h
fandreuz: Right, e5e8d3312e1359bc093e7e5562020723cd30dacc

apangin:
I already merged PR, but can I ask you to update its description with background, motivation, summary of changes, etc., so we could refer it in the future. Thanks.

fandreuz:
> I already merged PR, but can I ask you to update its description with background, motivation, summary of changes, etc., so we could refer it in the future. Thanks.

Done, let me know if you'd like to have any changes or additions.

apangin:
Looks good, thanks.

Issue #1174, opened 2025-03-12
async-profiler native mode does not attach to an embedded JVM


arielb1:
### Describe the bug

When native API (`asprof_init`/`asprof_execute`) async-profiler runs on code that embeds a JVM and calls from Rust/C/C++ into Java, async-profiler will not attach to the embedded JVM. Therefore, it won't have visibility into Java, and code that includes Java stack frames will appear to be `.unknown`.

Since this breaks unwinding, you won't even see the Rust/C/C++ that called into Java, just `.unknown`.

### Expected vs. actual behavior

expected: Java stack traces that are useful
actual: Java stack traces that are `.unknown`

### Reproduction Steps

I don't have a small reproducer, but you can have Rust/C/C++ code that embeds a JVM.

### Async-profiler version

latest

### Environment details

Linux. Both x86-64 and aarch64

apangin:
Resolved by #1192.
Now, when async-profiler is started from a native application, it automatically detects JVM running in the same process and attaches to it.

PR #1192, opened 2025-03-25
#1174: Detect JVM in non-Java application & attach it to the async-profiler

Baraa-Hasheesh:
### Description
Dynamically detecting & linking JVM instances in none Java applications to the async-profiler.

### Related issues
#1174

### Motivation and context
When async profiler is used on none Java applications (C/CPP/Rust) that contain a JVM instance (Created via JNI for example), the async-profiler won't sample Java stack correctly as the JVM instance isn't linked to the profiler.

By dynamically detecting & linking the JVM instance, it would provide better sampling for any made calls to the Java layer. 
Please refer to the following before & after screen shots

image
image

### How has this been tested?
Two tests have been added to handle the two major use cases where the order of the JVM & async-profiler loading happens differently, in both tests logs are checked for certain calls made into the JVM layer.
1. In the first test the async-profiler is loaded first followed by the JVM instance.
2. In the second test the JVM instance is loaded second followed by the async-profiler.

---


src/vmEntry.cpp
```diff
@@ -467,3 +473,56 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+        return;
+    }
+
+    jint result = _JNI_GetCreatedJavaVMs_func(&jvm, 1, &nVMs);
+    // No JVM is detected
+    if (result != JNI_OK || nVMs != 1) {
+        Log::debug("No JVM is yet detected in manual load");
+        return;
+    }
+    VM::init(jvm, true);
```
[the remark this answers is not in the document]
apangin: I mean, this code is supposed to handle the case when JVM is just being started. So can find an earlier hook where we are able to call `VM::init` with `attach=false`?

src/codeCache.cpp
```diff
@@ -222,6 +222,10 @@ void CodeCache::addImport(void** entry, const char* name) {
                 saveImport(im_realloc, entry);
             }
             break;
+        case 'J':
+            if (strcmp(name, "JVM_StartThread") == 0) {
+                saveImport(im_JVM_StartThread, entry);
```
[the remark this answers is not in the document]
Baraa-Hasheesh: After some experimentation the `pthread` hooks don't seem to be a good option currently 
However I do agree that using the `JVM_StartThread` is sub-optimal .

**pthread_create**
All threads that go through the `pthread_create` step were observed & found that only one thread has the JVM in a good enough state to attach to it (JVM is visible via JNI method) but using that thread to initiate the operation would result in a deadlock condition on the JNI GetEnv, please refer to the following stack dump for reference 

```
#0  0x00007f0732f61377 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007f073228e8fb in PlatformMonitor::wait(unsigned long) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#2  0x00007f073222abc5 in Monitor::wait(unsigned long) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#3  0x00007f07325ab642 in VMThread::wait_until_executed(VM_Operation*) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#4  0x00007f07325ac003 in VMThread::execute(VM_Operation*) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#5  0x00007f07320858eb in JvmtiEnvBase::enable_virtual_threads_notify_jvmti() () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#6  0x00007f073208e42b in JvmtiExport::get_jvmti_interface(JavaVM_*, void**, int) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
#7  0x00007f07306ba1e5 in VM::init(JavaVM_*, bool) () from build/lib/libasyncProfiler.so
#8  0x00007f07306ba28f in VM::VMManualLoad(bool) () from build/lib/libasyncProfiler.so
#9  0x00007f07306ba337 in pthread_create_hook(unsigned long*, pthread_attr_t const*, void* (*)(void*), void*) () from build/lib/libasyncProfiler.so
#10 0x00007f0732282c4f in os::create_thread(Thread*, os::ThreadType, unsigned long) () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
[#11](https://amzn-aws.slack.com/archives/C03K52T0BV2) 0x00007f0731f7bca9 in JVM_StartThread () from /usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so
``` 

** pthread_start**
Using this method will result `segmentation fault` when trying to initialize the JVM with the async profiler.
=> A deeper investigation will be needed on why this happens & how to prevent it.

**pthread_exit**
Using this method is too late in many case as it could happen at the end of the life cycle of the application being profiled.

src/vmEntry.cpp
```diff
@@ -467,3 +473,56 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+}
+
+void VM::JVM_StartThread_hook(JNIEnv *env, jobject thread) {
+    if (!VM::loaded()) {
+        VM::VMManualLoad(false);
```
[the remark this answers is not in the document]
Baraa-Hasheesh: That is by intention as the JVM internally start multiple threads.
Not all threads will be started when the JVM is fully initialized and fully available via the JNI APIs.

However I agree that we should add a counter for example to prevent trying too many times.

src/vmEntry.cpp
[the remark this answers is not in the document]
Baraa-Hasheesh: When implementing the Thread solution which handles then Java after Async profiler case
It was observed that the that the JVM will already be in a running state when it's first detected via the Thread

Baraa-Hasheesh:
@apangin thanks for the review
I will be doing as you suggested & remove all logic related to the JVM_StartThread hook
Let's discuss on how to better handle this in depth, The JVM_StartThread was the best hook I could find

a simple solution would be starting a separate pthread that simply does a relatively long sleep (1-5seconds) & then attempt to load the JVM into async profiler, this thread could be started only if a dlopen is detected on the libjava.so & the VM isn't loaded library 

I.E. shouldn't have any effect on native Java applications & in cases where the VM starts first

src/vmEntry.cpp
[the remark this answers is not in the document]
Baraa-Hasheesh: As suggested this PR currently only handles the JVM starting first case
So in this case the attache will always be true as the VM is running

Baraa-Hasheesh:
@apangin PR has been updated
Please feel fine to check the updated code

src/vmEntry.cpp
```diff
@@ -467,3 +466,23 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
         profiler->stop();
     }
 }
+
+void VM::VMManualLoad() {
+    JavaVM* jvm;
+    jsize nVMs;
+
+    jint (*handle)(JavaVM **, jsize, jsize *) = (jint (*)(JavaVM **, jsize, jsize *))dlsym(RTLD_DEFAULT, "JNI_GetCreatedJavaVMs");
```
apangin, line 474: I wonder why `RTLD_DEFAULT` works. What if libjvm is loaded with `dlopen(..., RTLD_LOCAL)`?

src/vmEntry.cpp
```diff
@@ -467,3 +466,23 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
         profiler->stop();
     }
 }
+
+void VM::VMManualLoad() {
```
apangin, line 470: Since this function will be called on every profiling session of a native app, it should use something lightweight to detect if libjvm is loaded. Check what is the most efficient: dlopen("libjvm.so", RTLD_NOLOAD), dlsym, dl_iterate_phdr or something else.

test/test/nonenative/NonenativeTests.java
```diff
@@ -0,0 +1,23 @@
...
+package test.nonenative;
+
+import one.profiler.test.Os;
+import one.profiler.test.Output;
+import one.profiler.test.Test;
+import one.profiler.test.TestProcess;
+
+public class NonenativeTests {
+
+    @Test(sh = "%testbin/java_first_none_native_app %s.html", output = true)
```
apangin, line 15: What is "none native"? You mean "non-native"? There are Java applications and Native (or Non-Java) applications, but "non-native" sounds unusual.

src/vmEntry.cpp
Baraa-Hasheesh: I did some checking here, it's due to the compile command I'm using which will use the `RTLD_GLOBAL` 
You're right this needs to change to account for the fact it's possible that JVM will be loaded in a `RTLD_LOCAL` manner

src/vmEntry.cpp
Baraa-Hasheesh: Problems that could happen:
1. `[libjvm.so](http://libjvm.so/)` is initially loaded with `RTLD_LOCAL` option `void* libHandle = dlopen("/usr/lib/jvm/java-21-amazon-corretto/lib/server/libjvm.so", RTLD_NOW | RTLD_LOCAL);`
2. `dlsym` won’t detect the `JNI_GetCreatedJavaVMs` function as the `[libjvm.so](http://libjvm.so/)` isn’t visible in the async-profiler scope => check first point
3. `dl_iterate_phdr` will allow us to detect if `[libjvm.so](http://libjvm.so/)` was indeed loaded but won’t guarantee the loading options => check first point

Suggested Solution Steps:
1. Limit the amount of `dlopen` & `dlsym` operation to at most to operations of each.
The point above can be handled via simply adding a Flag that is initially `TRUE`.
If the Flag is turned on the async-profiler on the next profiling session will attempt to load the `JNI_GetCreatedJavaVMs` function.
This however will require first loading the `[libjvm.so](http://libjvm.so/)` first due the first problem mentioned above.

If the load operation is successful the function reference can be stored & used in future profiling session starts to try & detect the JVM to load it into async-profiler.
If the operation is a failure the Flag is turned off (LIB or function wasn’t found via the respective `dlopen` & `dlsym` to prevent this loading from happening again  in future profiling sessions.

2. If a `dlopen` was detected on the `libjvm` between profiling sessions the flag can be turned back on for a reattempt of Step 1

Known issues & fixes if possible:
1. If `libjvm` is detected mid profiling session it won’t attach until a new profiling session starts, so the JVM symbols won’t be visible for the initial profiling session 
Suggested fix is to find a method that could be hooked into to detect that the JVM is ready for use & attach it mid profiling session to the async-profiler
=> TODO: this should be studied in depth & handled in a different PR
2. `libjvm` is decide but a JVM is never started.
To prevent constant calls to try to load the JVM (JNI_GetCreatedJavaVMs was detected successfully) over many profiling sessions, a simple stopgap solution of a counter could be introduced to limit the attempts to detect & load the JVM to maximum amount `X` -> I would suggest 5 attempts here at maximum

@apangin what do you think?

Baraa-Hasheesh:
@apangin code has been updated 
JNI_GetCreatedJavaVMs is cached to avoid unneeded calls dlopen & dlsym 
No need to cache the `dlopen` call as is if fails it will return NULL & if success dlsym will get the JNI_GetCreatedJavaVMs correctly

Make file updated & test refactored to use dlopen & dlsym to create the VM

multiple more tests are added to handle JVM starting between session cases

Known issues: JVM won't attach for an already run session

Solution: This will be automatically be solved once we have the JVM hook. solution online to detect a JVM start & attach it the profiler so it won't be handled here

Please feel free to check the updates made here

src/profiler.cpp
```diff
@@ -1077,6 +1077,11 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("Profiler already started");
     }
 
+    // Try loading JVM if not already loaded
```
apangin, line 1080: The comment is misleading: `tryAttach` does not load a JVM, it rather checks if the JVM is already loaded.

src/vmEntry.cpp
```diff
@@ -467,3 +467,26 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
         profiler->stop();
     }
 }
+
+void VM::tryAttach() {
+    JavaVM* jvm;
+    jsize nVMs;
+
+    if (_getJvm == NULL) {
+        void* libHandle = dlopen(OS::isLinux() ? "libjvm.so" : "libjvm.dylib", RTLD_LAZY | RTLD_NOLOAD);
```
apangin, line 476: Local variables follow snake_case naming convention.

src/vmEntry.cpp
```diff
@@ -467,3 +467,26 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+        _getJvm = (GetJvm)dlsym(libHandle, "JNI_GetCreatedJavaVMs");
```
apangin, line 477: This should be called only if the above `dlopen` succeeds.

src/vmEntry.cpp
```diff
@@ -467,3 +467,26 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+    }
+    
+    if (_getJvm == NULL) {
+        Log::debug("JNI_GetCreatedJavaVMs is not loaded");
+        return;
+    }
+
+    // Try actually loading the JVM
```
apangin, line 485: Again, the comment is misleading.

src/vmEntry.h
```diff
@@ -177,6 +180,13 @@ class VM {
 
     static jvmtiError JNICALL RedefineClassesHook(jvmtiEnv* jvmti, jint class_count, const jvmtiClassDefinition* class_definitions);
     static jvmtiError JNICALL RetransformClassesHook(jvmtiEnv* jvmti, jint class_count, const jclass* classes);
+
+    /**
```
apangin, line 184: Javadoc-style comments do not make sense for C++ code.

test/test/nonjava/JavaClass.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import java.lang.Math;
```
apangin, line 6: Redundant import

test/test/nonjava/JavaClass.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class JavaClass {
+    public void cpuHeavyTask() {
+        for (int i = 0; i < 100000; i++) {
+            Math.sqrt(Math.random());
+            Math.pow(Math.random(), Math.random());
```
apangin, line 12: Since computation results are nowhere used, JIT compiler can eliminate calls altogether.

test/test/nonjava/NonjavaTests.java
```diff
@@ -0,0 +1,44 @@
...
+    public void jvmLoadedAfterSessionStartTest(TestProcess p) throws Exception {
+        Output out = p.waitForExit(TestProcess.STDOUT);
+        assert p.exitCode() == 0;
+
+        out = p.readFile("%s");
+        assert !out.contains(".cpuHeavyTask");
+    }
+
+    @Test(sh = "%testbin/non_java_app 3 %f.html %s.html", output = true)
+    public void jvmLoadedAfterFirstSessionButBeforeSecondSessionTest(TestProcess p) throws Exception {
```
apangin, line 34: Method name is unreadable. Also, individual tests can be selected/skipped by name, and specifying such a long name on a command line is not convenient. Keep method names simple. When description is needed, put it in comments.

test/test/nonjava/common_non_java.cpp
```diff
@@ -0,0 +1,124 @@
...
+JNIEnv* CommonNonJava::env;
+
+void* CommonNonJava::jvmLib;
+
+void CommonNonJava::outputCallback(const char* buffer, size_t size) {
+    fwrite(buffer, sizeof(char), size, stderr);
+}
+
+void CommonNonJava::loadProfiler() {  
+    #ifdef __linux__
```
apangin, line 22: IIRC, this won't compile with some older gcc. There should not be whilespace before `#`.
Btw, inline #ifdefs in a method implementation are discouraged anyway.

test/test/nonjava/common_non_java.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void CommonNonJava::loadJvmLib() {
+    char* java_home = getenv("JAVA_HOME");
+    if (java_home == NULL) {
+        fprintf(stderr, "JAVA_HOME is not set\n");
+        exit(1);
+    }
+    char libPath[4096];
+    #ifdef __linux__
```
apangin, line 64: `#ifdef` inside a function again.
Since you use this for choosing between .so/.dylib in multiple places, it may be better to declare a platform-specific constant or a macro at the top of the file.

test/test/nonjava/common_non_java.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+void CommonNonJava::executeJvmTask() {
+    for (int i = 0; i < 100; ++i) {
+        jclass customClass = env->FindClass("JavaClass");
+        if (customClass == nullptr) {
+            fprintf(stderr, "Can't find JavaClass\n");
+            exit(1);
+        }
+        jmethodID constructor = env->GetMethodID(customClass, "<init>", "()V");
+        jobject customObject = env->NewObject(customClass, constructor);
```
apangin, line 105: Why creating an instance? These lines won't be needed if you declare the method `static`.

test/test/nonjava/common_non_java.cpp
```diff
@@ -0,0 +1,124 @@
...
+        jmethodID cpuHeavyTask = env->GetMethodID(customClass, "cpuHeavyTask", "()V");
+        if (cpuHeavyTask == nullptr) {
+            fprintf(stderr, "Can't find cpuHeavyTask");
+            exit(1);
+        }
+        env->CallVoidMethod(customObject, cpuHeavyTask);
```
apangin, line 111: Exception must be checked after JNI-to-Java upcall.
You may run program with `-Xcheck:jni` to find potential issues with JNI code.

test/test/nonjava/common_non_java.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <dlfcn.h>
+#include <jni.h>
+#include <unistd.h>
+#include <iostream>
+#include "asprof.h"
+
+typedef jint (*CreateJvm)(JavaVM **, void **, void *);
+
+class CommonNonJava {
```
apangin, line 17: In what sense it is "common"?
Having 3 source files for a couple of tests seems an overkill. Can it be just one .cpp file?

Baraa-Hasheesh:
Thanks @apangin for the feedback 
I restructured the tests into a single file, adjusted the commends & handled the other details
Please feel to check the updates made

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,283 @@
...
+    snprintf(cmd, sizeof(cmd), "stop,file=%s", outputFile);
+    asprof_error_t err = _asprof_execute(cmd, outputCallback);
+    if (err != NULL) {
+        std::cerr << _asprof_error_str(err) << std::endl;
+        exit(1);
+    }
+}
+
+void loadJvmLib() {
+    char* java_home = getenv("JAVA_HOME");
```
apangin, line 68: This may break test, since `JAVA_HOME` is not always set.

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,283 @@
...
+    if (java_home == NULL) {
+        std::cerr << "JAVA_HOME is not set" << std::endl;
+        exit(1);
+    }
+    char lib_path[4096];
+    snprintf(lib_path, sizeof(lib_path), "%s/%s", java_home, jvm_lib_path);
```
apangin, line 74: Will not work with JDK 8, where there is additional `jre` directory on the path.

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,283 @@
...
+    // Start JVM
+    JavaVMInitArgs vm_args;
+    JavaVMOption options[2];
+    int pid = getpid();
+
+    options[0].optionString = const_cast<char*>("-Djava.class.path=build/test");
+    options[1].optionString = const_cast<char*>("-Xcheck:jni");
+
+    // Configure JVM
+    vm_args.version = JNI_VERSION_10;
```
apangin, line 92: Why 10?

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,283 @@
...
+        std::cerr << "Can't find cpuHeavyTask" << std::endl;
+        exit(1);
+    }
+
+    for (int i = 0; i < 100; ++i) {
+        jdouble result = _env->CallStaticDoubleMethod(customClass, cpuHeavyTask);
+        if (_env->ExceptionCheck()) {
+            jthrowable exception = _env->ExceptionOccurred();
+            _env->ExceptionDescribe();
+            _env->ExceptionClear();
```
apangin, line 129: You've listed all possible JNI exception functions :)
`ExceptionDescribe` is enough. It does everything: prints exception message with a stack trace and clears current exception.

src/vmEntry.cpp
```diff
@@ -467,3 +467,26 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+        return;
+    }
+
+    jint result = _getJvm(&jvm, 1, &nVMs);
+    if (result != JNI_OK || nVMs != 1) {
+        Log::debug("No JVM is yet detected in manual load");
+        return;
+    }
+    VM::init(jvm, true);
+}
```
apangin, line 492: New line character at EOF is missing

Makefile
```diff
@@ -210,6 +210,7 @@ build-test-bins:
 	@mkdir -p $(TEST_BIN_DIR)
 	gcc -o $(TEST_BIN_DIR)/malloc_plt_dyn test/test/nativemem/malloc_plt_dyn.c
 	gcc -o $(TEST_BIN_DIR)/native_api -Isrc test/test/c/native_api.c -ldl
+	$(CXX) -o $(TEST_BIN_DIR)/non_java_app $(CPP_TEST_INCLUDES) $(INCLUDES) $(LIBS) test/test/nonjava/non_java_app.cpp -ldl
```
apangin, line 213: Why is `$(LIBS)` needed here?

src/vmEntry.cpp
```diff
@@ -467,3 +467,26 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
+    }
+    
+    if (_getJvm == NULL) {
+        Log::debug("JNI_GetCreatedJavaVMs is not loaded");
+        return;
+    }
+
+    jint result = _getJvm(&jvm, 1, &nVMs);
+    if (result != JNI_OK || nVMs != 1) {
+        Log::debug("No JVM is yet detected in manual load");
```
apangin, line 488: "manual load" is unclear.
Anyways, I'd remove debug printout altogether.

test/test/nonjava/non_java_app.cpp
Baraa-Hasheesh: Hello @apangin 
I tried using JDK 8 to run the tests & found that the async-profiler won't compile at all using java 8
This make testing on java 8 not doable as it seems that additional changes on the async profiler will be required before it is possible to be able to correctly support testing using java 8

image

I will make the needed changes on the test to address the comment for future enable of Java 8, but further work will be required to enable proper testing for Java 8

test/test/nonjava/non_java_app.cpp
Baraa-Hasheesh: I used it as it's the version available for all currently used JDK versions
I will make it dynamic based on JDK version being used

Baraa-Hasheesh: Please note in `JNI_OnLoad` currently is using `JNI_VERSION_1_6`
This is because `JNI_VERSION` is depends on the JDK version 
In my case `JNI_VERSION_10` was used as it the highest common version between JDK11 -> JDK21

I will make the value more dynamic depending the JDK version 
But note this will require redefining some constants to avoid compile errors 

image

test/test/nonjava/non_java_app.cpp
Baraa-Hasheesh: That's correct.
I will make do the changes to prepare to make it work with JDK 8 in the future
However as if now Async-profiler doesn't compile using JDK-8, so as if now JDK-8 can't be used to run the tests

Some changes will be needed on the Async-profiler as a whole to support compiling & running the tests in JDK 8

image

apangin: It's true that async-profiler requires JDK 11+ to compile. This is because it uses JDK 11 APIs.
However, this does not mean tests cannot run with JDK 8. async-profiler compiled with JDK 11 can perfectly run on JDK 8.

Baraa-Hasheesh: @apangin 
It's indeed true that async-profiler will run fine on JDK8

however what I'm referring to is the automation of the `Makefile`

when you use `make test`, a recompile will happen which means even if you initially compiled with JDK11+ using `make`, the `make test` will fail as it will recompile the profiler 

A Change will be needed to enable running the tests without recompiling the profiler which isn't currently the case

Please correct me if I'm wrong.

Baraa-Hasheesh:
@apangin 
Code has been updated to work with JDK 8 & additional feedback has been addressed
please feel free to check it

src/vmEntry.cpp
```diff
@@ -107,14 +107,21 @@ static void* resolveMethodIdEnd() {
...
     if (_jvmti != NULL) return true;
 
     _vm = vm;
     if (_vm->GetEnv((void**)&_jvmti, JVMTI_VERSION_1_0) != 0) {
         return false;
     }
 
+    if (attach_thread) {
+        JNIEnv* env = nullptr;
+        if (_vm->AttachCurrentThreadAsDaemon((void**)&env, NULL) != JNI_OK) {
```
apangin, line 120: Attaching thread to a JVM has side effects: e.g. it becomes visible in Java thread dumps, JDK Flight Recorder starts to report this thread, etc. If this is needed only to call `VM::init`, make sure the thread is detached as soon as initialization completes. But note that a thread may be already owned by the JVM - in this case, it must not be detached.

test/one/profiler/test/TestProcess.java
```diff
@@ -33,6 +33,19 @@ public class TestProcess implements Closeable {
     public static final String LIBPROF = "%lib";
     public static final String TESTBIN = "%testbin";
 
+    private static final String JAVA_HOME = System.getProperty("java.home");
+    private static final String JAVA_VERSION;
+
+    static {
+        String[] javaVersionParts = System.getProperty("java.version").split("\\.");
```
apangin, line 40: This is not fully correct way to get the JDK version.
`Runner` class already gets the JVM version [here](https://github.com/async-profiler/async-profiler/blob/ee75d80622a54e8919052cdd195fd49c8977c755/test/one/profiler/test/Runner.java#L94-L100).
However, it's likely that you don't need this at all.

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,348 @@
...
+#include <jni.h>
+#include <unistd.h>
+#include <iostream>
+#include <limits.h>
+#include "asprof.h"
+ 
+#ifdef __linux__
+const char profiler_lib_path[] = "build/lib/libasyncProfiler.so";
+const char jvm_lib_path[] = "lib/server/libjvm.so";
+const char jvm8_lib_path[] = "lib/amd64/server/libjvm.so";
```
apangin, line 16: It's not good to have the architecture hard-coded.

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,348 @@
...
+#define JNI_VERSION_20  0x00140000
+#endif
+
+#ifndef JNI_VERSION_21
+#define JNI_VERSION_21  0x00150000
+#endif
+
+#ifndef JNI_VERSION_24
+#define JNI_VERSION_24  0x00180000
+#endif
```
apangin, line 45: Why is that? Can't you just always use `JNI_VERSION_1_6`?

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,348 @@
...
+    // Start JVM
+    JavaVMInitArgs vm_args;
+    JavaVMOption options[1];
+    int pid = getpid();
+
+    options[0].optionString = const_cast<char*>("-Djava.class.path=build/test");
+    //options[1].optionString = const_cast<char*>("-Xcheck:jni");
+
+    // Configure JVM
+    vm_args.version = getJniVersion();;
```
apangin, line 159: Can be a constant. Also, double `;` at the end.

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: I will need it as to load the libjvm.so lib dynamically 
The path is different between JDK8 & other versions on linux 

=> it includes the arch as amd64 or i386 (x86) As well as aarch64 or arm (arm)

src/vmEntry.cpp
Baraa-Hasheesh: After checking JNI doc
I will remove this 
The above `_vm->GetEnv((void**)&_jvmti, JVMTI_VERSION_1_0) != 0` condition guarantees that the thread is already associated with the JVM

image

Baraa-Hasheesh:
Hey @apangin 
I've updated the files here
Also regarding the #1203 issue, I was able to reproduce on the async-profiler `master` branch (Changes here aren't included)

please feel free to check the made updates

apangin:
I realized that in all the tests async-profiler starts in the very same thread that creates a JVM. It will not this way in real Rust applications. Can you create a tests that starts async-profiler in a different thread not attached to the JVM?

Baraa-Hasheesh:
@apangin the code has been updated to handle JVM starting in another thread & a test was added to check for this flow
A work around was needed due to the [JDK-8308341](https://bugs.openjdk.org/browse/JDK-8308341) openjdk bug

Please feel free to check it when you can

Please note that creating a solid reproducer for the Bug will be hard as it depends on multiple factors that are hard to control & it will be different from one machine to another

As it's hard to exactly control the start of the profiler to be after the VM is created but not fully initialized 

So I dependent on a local reproduction that works reliably on my machine (for the JDK bug) & made sure that the work around is working as expected

src/vmEntry.cpp
```diff
@@ -106,6 +106,49 @@ static void* resolveMethodIdEnd() {
     return NULL;
 }
 
+void VM::initWrapper(JavaVM* vm) {
```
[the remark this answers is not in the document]
Baraa-Hasheesh: Doing that would effect the performance & readability of the normal `init` method
As we need to use `vm->GetEnv((void**)&env, JNI_VERSION_1_6);`

Using the `GetEnv` to get `JVM TI` isn't safe in this specific case

So moving this extra read & other thread logic outside the logic of original `init` is more readable IMO than adding a flag to only do this flow if the flag is enabled (Not having this wrapping will most likely result in more nesting in the `init` method)

I however agree that the `initWrapper` logic is somewhat vague

Baraa-Hasheesh:
@apangin Applied the latest changes, please feel free to check it when possible

src/vmEntry.cpp
```diff
@@ -106,6 +106,33 @@ static void* resolveMethodIdEnd() {
...
+        if (!OS::threadName(list->next(), thread_name, sizeof(thread_name))) {
+            continue;
+        }
+
+        // For MacOs Thread name starts with "Java: "
+        if (strncmp(thread_name, "Java: ", 6) == 0) {
+            thread_name_offset = 6;
+        }
+
+        if (strcmp(thread_name + thread_name_offset, "VM Thread") == 0) {
```
apangin, line 127: May cause out-of-bounds access, when thread_name_offset == 6, but thread_name does not start with "Java: "

src/vmEntry.cpp
```diff
@@ -467,3 +494,39 @@ JNI_OnUnload(JavaVM* vm, void* reserved) {
...
 }
+
+// Try to find a running JVM instance & attach it to the profiler
+void VM::tryAttach() {
+    JavaVM* jvm;
+    JNIEnv* env;
+    jsize nVMs;
+
+    if (_getJvm == NULL) {
+        void* lib_handle = dlopen(OS::isLinux() ? "libjvm.so" : "libjvm.dylib", RTLD_LAZY | RTLD_NOLOAD);
```
apangin, line 505: Possible handle leak: need `dlclose(lib_handle)` after use.

test/test/nonjava/JavaClass.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+public class JavaClass {
```
apangin, line 6: Should be in `test.nonjava` package.

test/test/nonjava/NonjavaTests.java
```diff
@@ -0,0 +1,56 @@
...
+package test.nonjava;
+
+import one.profiler.test.Output;
+import one.profiler.test.Test;
+import one.profiler.test.TestProcess;
+
+public class NonjavaTests {
+
+    // jvm is loaded before the profiling session is started
+    @Test(sh = "%testbin/non_java_app 1 %s.html", output = true)
```
apangin, line 15: Why `output = true` if stdout is never read?

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,368 @@
...
+    struct timespec wait_time = {(time_t)(2), 0L};
+
+    validateArgsCount(argc, 3, "Minimum Arguments is 2");
+
+    loadProfiler();
+
+    pthread_t thread;
+    pthread_create(&thread, NULL, jvmThreadWrapper, NULL);
+   
+    // busy wait for JVM to start on the thread
```
apangin, line 333: It's not _busy_ wait

test/test/nonjava/non_java_app.cpp
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    char java_lib_home[PATH_MAX];
+    struct dirent* entry;
+    DIR* dir;
+
+    // libjvm wasn't found under standard path, this could happen in JDK 8 where the path is formated like:
+    // ${TEST_JAVA_HOME}/lib/${ARCH}/server/libjvm.(so|dylib)
+    snprintf(java_lib_home, sizeof(java_lib_home), "%s/lib", java_home);
+    dir = opendir(java_lib_home);
```
apangin, line 95: `closedir` is missing

apangin:
I fixed above issues together with minor typos and aligned the code with the project style.
Thanks for the contribution.

PR #1196, opened 2025-03-27
Replace `std::map` with `std::unordered_map`

fandreuz:
### Description
In this PR I replace usages of `std::map` with `std::unordered_map`, since no ordering is required in most cases.

### Motivation and context
Modernization and performance improvements.

### How has this been tested?
`make test`

---

### DIFF
--- a/src/flameGraph.h
+++ b/src/flameGraph.h
@@ -6,7 +6,7 @@
 #ifndef _FLAMEGRAPH_H
 #define _FLAMEGRAPH_H
 
-#include <map>
+#include <unordered_map>
 #include <string>
 #include "arch.h"
 #include "arguments.h"
@@ -16,7 +16,7 @@
 
 class Trie {
   public:
-    std::map<u32, Trie> _children;
+    std::unordered_map<u32, Trie> _children;
     u64 _total;
     u64 _self;
     u64 _inlined, _c1_compiled, _interpreted;
@@ -46,7 +46,7 @@ class Trie {
 
     int depth(u64 cutoff, u32* name_order) const {
         int max_depth = 0;
-        for (std::map<u32, Trie>::const_iterator it = _children.begin(); it != _children.end(); ++it) {
+        for (std::unordered_map<u32, Trie>::const_iterator it = _children.begin(); it != _children.end(); ++it) {
             if (it->second._total >= cutoff) {
                 name_order[nameIndex(it->first)] = 1;
                 int d = it->second.depth(cutoff, name_order);
@@ -61,7 +61,7 @@ class Trie {
 class FlameGraph {
   private:
     Trie _root;
-    std::map<std::string, u32> _cpool;
+    std::unordered_map<std::string, u32> _cpool;
     u32* _name_order;
     u64 _mintotal;
     char _buf[4096];
--- a/test/native/testRunner.hpp
+++ b/test/native/testRunner.hpp
@@ -17,7 +17,7 @@ struct TestCase;
 
 class TestRunner {
   private:
-    std::map<std::string, TestCase> _test_cases;
+    std::unordered_map<std::string, TestCase> _test_cases;
 
     TestRunner(const TestRunner&) = delete;
     TestRunner& operator=(const TestRunner&) = delete;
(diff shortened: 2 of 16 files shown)


fandreuz:
This needs #1197

test/native/testRunner.hpp
```diff
@@ -13,11 +13,23 @@
...
+    bool skipped = false;
+
+    TestCase(const std::string& name, std::function<void()> test_function, bool only, const std::string& filename, int line_no)
+        : name(name), test_function(test_function), only(only), filename(filename), line_no(line_no) {}
+};
 
 class TestRunner {
   private:
-    std::map<std::string, TestCase> _test_cases;
+    std::unordered_map<std::string, TestCase> _test_cases;
```
apangin, line 32: In this case, running tests in a consistent order is useful.

apangin (commented):
Why didn'y you replace `Trie::_children`?

test/native/testRunner.hpp
fandreuz: 6a640e0dc29a52ced53bec0c9a2346ad19c59d80

src/flameGraph.h
```diff
@@ -71,7 +71,7 @@ class Trie {
 class FlameGraph {
   private:
     Trie _root;
-    std::map<std::string, u32> _cpool;
+    std::unordered_map<std::string, u32> _cpool;
```
apangin, line 74: Not that simple :) This task requires thoughtful decision what occurrences of `map` should be changed and what should not.

apangin: Keeping strings in a constant pool sorted alphabetically pursues two goals:
1. Consistent order of frames on a flamegraph.
2. Smaller html size due to [string compression](https://github.com/async-profiler/async-profiler/blob/ee75d80622a54e8919052cdd195fd49c8977c755/src/res/flame.html#L204-L208).

fandreuz: 👍 I plan to profile the profiler (I got some resources from @krk) to check whether any paths became less/more efficient. That should give some guidance on the access patterns for each occurrent of `std::map`.

apangin: Sounds good! Also check how changes affect static size of the binary.

fandreuz:
- `jfrMetadata.h`
   - `Element::getId` got ~30% faster 
- `Dictionary::collect` (considering all call paths) got ~25% faster
- `profiler.h`
   - No significant change for `Profiler::updateNativeThreadNames`
   - `Profiler::convertNativeTrace` got slower
   - `Profiler::recordSample` got slower
   - `Profiler::dumpCollapsed` got 20% faster
   - `Profiler::dumpFlamegraph` got 15% faster

The results above are what I got "on average" (there's some variance) running async-profiler with `-i 10ns -e cpu` and various output formats. I think the only optimization which may really be worth it is `Element::getId` and `Dictionary::collect`, the others can be dropped since I believe optimizing `Profiler::recordSample` is much more important than optimizing `dump*`.

I don't have a clear insight on why `Profiler::recordSample` got slower, it feels a bit counterintuitive to me and I'd need to spend some more time on this.

fandreuz:
@apangin I finally managed to _almost_ fix my code to profile Async-Profiler (there's still a sporadic deadlock when a malloc within `__cyg_profile_func_enter` is interrupted by an Async-Profiler signal, but that's good enough I guess). Code [here](https://github.com/async-profiler/async-profiler/compare/master...fandreuz:async-profiler:instrumenting-asprof).

These are the results I got with `ASPROF_CMD=start,event=cpu,interval=100ns,file=/dev/null,timeout=40`. You can download the SVGs for deeper inspection. Note that there's definitely some noise since I'm comparing two different runs, one based on `master` and one on `238-unordered-map`.

I believe we should drop this PR, I don't see clear wins anywhere. 

### `collapsed`
diff_master_238-unordered-map_collapsed

### `flamegraph`
diff_master_238-unordered-map_flamegraph

### `jfr`
diff_master_238-unordered-map_jfr

apangin:
Thank you for your analysis: it's very useful despite not showing obvious benefits from the replacement. I'm inclined to agree with your conclusion: let's not fix things that are not broken. We will, however, prefer `unordered_map` in the future in the cases where key order is not important.

Issue #872, opened 2024-01-03
Get rid of libstdc++ dependency


apangin:
Currently, async-profiler uses standard C++ library classes such as:
 - IO streams
 - std::string
 - std::map
 - std::vector

This makes async-profiler depend on `libstdc++.so`, which causes certain compatibility issues.
 - async-profiler built on a newer Linux distribution does not work with older distributions;
 - libstdc++ is not preinstalled on Alpine Linux;
 - a statically linked async-profiler binary becomes too large.

I suggest rewriting relevant parts of async-profiler without use of libstdc++ classes. This will allow building a compact async-profiler binary that works both on glibc and musl libc.

korniltsev:
Is someone already working on it? 


The reason Im asking is because Im linking libstdc++  statically on centos and when I try to use this library on alpine, dynamic linker is unable to load async-profiler which built this way. There are two reasons for this failure:

1. I found this set of function dependencies and `__strftime_l` seem to not exist in musl
```
__strftime_l => 
std::__timepunct<char>::_M_put <=
std::time_put<char,std::ostreambuf_iterator<char>>::do_put <=
vtable for'std::time_put<char, std::ostreambuf_iterator<char>> 
```
2.    async-profiler end up `DT_NEED` `ld-linux-x86-64.so.2`  which is likely there because `libstdc++` has it

When I rewrite io streams without libstdc++  , the resulting size of the binary goes from 1.4M to 500k and alpine is able to load the library (with one extra little hacky change just for POC - I manually binary patched to remove DT_NEED ld-linux-x86-64.so.2)

I would like to submit a PR rewriting  io streams. 
Is this something that can be accepted as an external contribution?
This could be a good first step towards removing libstdc++.

cc @apangin

apangin:
Hi @korniltsev,
I haven't started implementing this yet.

Your proposal makes total sense to me. Did not expect that replacing streams only could make such a huge difference in binary size. With that in mind, getting rid of stdc++ classes completely may not be even needed: I think 500 KB binary with statically linked libstdc++ (comparing to 380K as of today) would be an acceptable trade-off.

richardstartin:
FWIW we link our fork against libstdc++ statically and applied some compiler and linker flags (we build with gradle so we can't offer a direct patch if this information helps):

[...]

PR #1197, opened 2025-03-28
Use `std::shared_ptr<Trie>` to represent children in a Trie node

fandreuz:
### Description
In this PR I apply this change:
```diff
class Trie {
   public:
-      std::map<u32, Trie> _children;
+      std::map<u32, std::shared_ptr<Trie>> _children;
...
```

### Related issues
This will fix the compilation errors in #1196.

### Motivation and context
`std::unordered_map` does not support incomplete types: https://www.reddit.com/r/cpp_questions/comments/utaxnb/forward_type_declaration_in_unordered_map/

Thus, by wrapping the value with `std::shared_ptr` we can avoid the problem. I could have used a naked pointer, but I think we could benefit from automatic memory management.

I could not use `std::unique_ptr` because `Trie` objects need to be passed around while traversing a stack trace in `Profiler::dumpFlameGraph`.

### How has this been tested?
`make test`

---

### DIFF
--- a/src/flameGraph.h
+++ b/src/flameGraph.h
@@ -12,11 +12,12 @@
 #include "arguments.h"
 #include "vmEntry.h"
 #include "writer.h"
+#include <memory>
 
 
 class Trie {
   public:
-    std::map<u32, Trie> _children;
+    std::map<u32, std::shared_ptr<Trie>> _children;
     u64 _total;
     u64 _self;
     u64 _inlined, _c1_compiled, _interpreted;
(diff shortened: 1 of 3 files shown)


apangin:
No smart pointers, please. They make object lifetime unobvious, they introduce overhead and make code more verbose. A rule of thumb is no new c++ _library_ features should be brought to the project (not to confuse with _language_ features). Ideally, we should _reduce_ the amount of dependency on libstdc++, not _increase_ it, see #872 for more context.

fandreuz:
> No smart pointers, please. They make object lifetime unobvious, they introduce overhead and make code more verbose. A rule of thumb is no new c++ _library_ features should be brought to the project (not to confuse with _language_ features). Ideally, we should _reduce_ the amount of dependency on libstdc++, not _increase_ it, see #872 for more context.

Got it @apangin, I adapted the PR to use standard pointers

src/flameGraph.h
```diff
@@ -41,15 +47,19 @@ class Trie {
     }
 
     Trie* child(u32 name_index, FrameTypeId type) {
-        return &_children[name_index | type << 28];
+        auto ptr = &_children[name_index | type << 28];
```
apangin, line 50: It's perfectly fine to use `auto` for iterators for the sake of brevity, but please do not overuse it.
Types are very important for understanding code, even more than names.

fandreuz: Sure, fixed in 0885f458a42744238df5d2f3408042837d1f4e2c

PR #1201, opened 2025-03-30
Profiling tail latency events with async-profiler

RainM:
### Description

Patch for profiling tail latency events with async-profiler.

### Motivation and context

(1):
To determine a tail latency event, there must be some interaction with the application so that it can indicate whether a given event represents average latency or tail latency.

(2):
Such an interface between the application and the profiler should be as lightweight as possible, making it easy to integrate into a wide range of different software.

(3):
A conceptually correct fix: async-profiler should retain samples only if a tail latency event occurs; otherwise, for median cases, it should discard them. However, implementing this fix is complex, as it requires modifications to many internal components of async-profiler.

(4) - Important:
Instead of capturing the entire profile for a tail latency event, we can profile only what happens after exceeding a certain SLA threshold.
For example, if the 0.999 quantile latency is 1ms, we can capture only the samples that occur after the first 1ms of response processing. This approach would provide a reasonably good profile of 0.999 quantile events.

Implementation (1):
A simple fix: Ignore a sample if insufficient time has passed since processing started. This can be implemented with just one if statement and an early exit in the signal handler.

Implementation (2):
If the application follows an event loop architecture, the start time of processing can be determined from the beginning of the poll() call.

Implementation (3):
A separate file in /dev/shm can be used to store timestamps of poll() start times from the profiled application.

The profiler will read the timestamp of the current poll() start time.

If not enough time has passed (i.e., it's not a tail latency event), the sample is discarded.

### How has this been tested?

It works on our production servers. We are able to get profile for tail latency events and get to know what worsen our tail latency.

---


src/filter.h
```diff
@@ -0,0 +1,40 @@
...
+class HeartBeatFilter {
+private:
+    int _fd;
+    u64 *_region_ptr;
+    int _clock_id;
+    u64 _delay_ns;
+    const char* _file_path;
+public:
+    HeartBeatFilter(const char* heartbeat_file, u64 delay_ns, bool use_unix_clock, bool use_realtime_clock);
+  ~HeartBeatFilter();
```
incubos, line 20:
```suggestion
    ~HeartBeatFilter();
```

src/filter_linux.cpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
incubos, line 5:
```suggestion
```

src/main/main.cpp
```diff
@@ -537,7 +541,14 @@ int main(int argc, const char** argv) {
...
             pid = jps("jps -J-XX:+PerfDisableSharedMem", arg.str());
-
+        } else if (arg == "--heartbeat-file") {
+            params << ",hb-file=" << args.next();
+        } else if (arg == "--heartbeat-interval-ns") {
+            params << ",delayns=" << args.next();
+        } else if (arg == "--heartbeat-clock-realtime") {
+            params << ",clock-rt";
+        } else if (arg == "--heartbeat-clock-unix") {
+            params << ",clock-unix";
```
incubos, line 551: What do you think about unifying param and option naming?

src/profiler.cpp
```diff
@@ -1101,6 +1105,13 @@ Error Profiler::start(Arguments& args, bool reset) {
             return Error("Failed to initialize FdTransferClient");
         }
     }
+    if (args._heartbeat_file) {
```
incubos, line 1108:
```suggestion

    if (args._heartbeat_file) {
```

incubos:
Looks quite interesting! Could you please add some test (if possible) and/or describe use case in README too?

src/filter_linux.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (res) {
+	Log::error("Can't get time. error: %s", strerror(errno));
+    }
+
+    u64 full_ts = (nano_time.tv_sec * NS_IN_1_SEC) + nano_time.tv_nsec;
+
+    return full_ts >= profile_after;
+}
+
+#endif // __linix__
```
fandreuz, line 62:
```suggestion
#endif // __linux__
```

src/arguments.cpp
```diff
@@ -419,6 +419,22 @@ Error Arguments::parse(const char* args) {
...
                 _reverse = true;
 
+            CASE("hb-file")
+                _heartbeat_file = value;
+
+            CASE("delayns")
+                _heartbeat_delay_ns = strtol(value, NULL, 0);
+
+            CASE("clock-unix") {
+                _heartbeat_unix_clock = true;
```
fandreuz, line 429: We could directly store either `CLOCK_REALTIME` or `CLOCK_MONOTONIC` instead of two boolean flags here, and pass it directly to the constructor of `HeartBeatFilter` instead of having to call `clock_id(bool, bool)`

src/filter_linux.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+bool HeartBeatFilter::shouldProcess() {
+    u64 last_heartbeat = *(u64*)this->_region_ptr;
+    u64 profile_after = last_heartbeat + this->_delay_ns;
+
+    struct timespec nano_time = {};
+    int res = clock_gettime(this->_clock_id, &nano_time);
+    if (res) {
+	Log::error("Can't get time. error: %s", strerror(errno));
+    }
```
fandreuz, line 55: If we cannot get the time, we should probably reject the profile since there's un underlying issue anyway

src/filter_linux.cpp
```diff
@@ -0,0 +1,62 @@
...
+  	return -1;
+    }
+}
+
+HeartBeatFilter::~HeartBeatFilter() {
+    munmap(this->_region_ptr, sizeof(u64));
+    close(this->_fd);  
+}
+
+HeartBeatFilter::HeartBeatFilter(const char* heartbeat_file, u64 interval_ns, bool use_unix_clock, bool use_realtime_clock) {
```
fandreuz, line 33: Can we unify the naming of the variable holding threshold to accept the profile? It's `interval_ns` in some files, and `delay_ns` in some others. I'd use `threshold_ns` for all of them, what do you think?

apangin:
Thank you for your proposal. We are currently working on a similar problem - to profile causes of long-running requests. To do this, we need a cheap way to coordinate application-specific metadata about current request/transaction with profiler's sampling engine. This was implemented in #1169 for native applications, and the related Java API is [on the way](https://github.com/async-profiler/async-profiler/compare/master...context-events).

There are two key differences with your approach:
 - profiler runs sampling constantly, and it's an application/framework that reads last profiler timestamp to decide whether to emit latency metadata or not;
 - context is thread-local vs. one global timestamp in your proposed solution.

I admit this may not be ideal for your specific use case, but I believe our approach has wider applicability and provides base for more features such as context-aware profiling with an ability to filter profiles by app-specific tags.

I'm happy to discuss how we can improve this solution to address your needs too.

apangin:
Closing in favor of #1755, which will be a more generic way to do latency profiling.

Issue #1199, opened 2025-03-28
Amend `Locktests#raceToLocks` to test profiling results of contended locks


fandreuz:
### Describe the bug

> `Locktests#raceToLocks` is supposed to check that a lock that is being contended more often, is sampled more frequently by the profiler.

### Async-profiler version

Nightly

PR #1204, opened 2025-03-31
Amend `LockTests#raceToLocks` to test profiling results of contended locks

fandreuz:
### Description
In this PR, I amend the test `LockTests#raceToLocks` to add a new check to assert that threads locking with strong contention are sampled more frequently than threads locking with light contention. 

The idea is that `semiShared` threads always lock an uncontended (thread-local) lock except for a minority of the cases, i.e. once every `SHARED_LOCK_INTERVAL`. `shared` threads always lock on the contended lock.

### Related issues
#1199 

### Motivation and context
The test needs to be fixed to make sure profiling locks works properly.

### How has this been tested?
`make test` both locally and in CI.

---

### DIFF
--- a/test/test/lock/LockTests.java
+++ b/test/test/lock/LockTests.java
@@ -27,12 +27,15 @@ public class LockTests {
     public void raceToLocks(TestProcess p) throws Exception {
         int interval = Integer.parseInt(p.inputs()[0]);
         Output out = p.profile("--lock " + interval + " --threads -o collapsed");
-        Output stdout = p.readFile(TestProcess.STDOUT);
 
         Assert.isGreater(out.samples("\\[random1"), 0, "sampled all threads 1/4");
         Assert.isGreater(out.samples("\\[random2"), 0, "sampled all threads 2/4");
         Assert.isGreater(out.samples("\\[shared1"), 0, "sampled all threads 3/4");
         Assert.isGreater(out.samples("\\[shared2"), 0, "sampled all threads 4/4");
+
+        long maxSamplesRandom = Math.min(out.samples("\\[random1"), out.samples("\\[random2"));
+        long minSamplesShared = Math.min(out.samples("\\[shared1"), out.samples("\\[shared2"));
+        Assert.isGreater(minSamplesShared, maxSamplesRandom, "threads with shared lock are sampled more frequently");
 
         if (interval == 0) {
             long shared1 = out.samples("\\[shared1");
--- a/test/test/lock/LockTests.java
+++ b/test/test/lock/LockTests.java
@@ -28,14 +28,14 @@ public class LockTests {
         int interval = Integer.parseInt(p.inputs()[0]);
         Output out = p.profile("--lock " + interval + " --threads -o collapsed");
 
-        Assert.isGreater(out.samples("\\[random1"), 0, "sampled all threads 1/4");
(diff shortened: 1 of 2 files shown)


test/test/lock/LockTests.java
```diff
@@ -27,12 +27,15 @@ public void datagramSocketLock(TestProcess p) throws Exception {
     public void raceToLocks(TestProcess p) throws Exception {
         int interval = Integer.parseInt(p.inputs()[0]);
         Output out = p.profile("--lock " + interval + " --threads -o collapsed");
-        Output stdout = p.readFile(TestProcess.STDOUT);
```
apangin, line 30: If `stdout` is no longer read, then `output = true` attribute is not necessary?

test/test/lock/RaceToLock.java
```diff
@@ -8,81 +8,65 @@
...
-    private final Lock[] locks = {
-            new ReentrantLock(true),
-            new ReentrantLock(true),
-            new ReentrantLock(true)
-    };
-    private volatile long randomWaitTime;
+    private static final Lock sharedLock = new ReentrantLock(true);
 
-    private volatile boolean exitRequested;
+    private static volatile boolean exitRequested = false;
```
apangin, line 20: It's a bad practice to initialize `volatile` fields with a default value.
Although it does not make difference here, this is known to have performance impact in other context.

test/test/lock/RaceToLock.java
```diff
@@ -8,81 +8,65 @@
...
+        Lock nonsharedLock = new ReentrantLock(true);
         while (!exitRequested) {
-            Lock lock = locks[random.nextInt(locks.length)];
+            Lock lock = count % SHARED_LOCK_INTERVAL == 0 ? sharedLock : nonsharedLock;
 
-            long start = System.nanoTime();
-            lock.lock();
-            randomWaitTime += System.nanoTime() - start;
             try {
+                lock.lock();
```
apangin, line 44: Correct ReentrantLock pattern is to call `lock()` outside `try-finally`. Otherwise `unlock` may be called on an object that is not locked.

test/test/lock/RaceToLock.java
fandreuz: Where can I read more about this?

fandreuz: 578f470ebf4ea08a0a828eaae2fbeddc33e6c9ef

test/test/lock/RaceToLock.java
fandreuz: Thanks, a3234fc49b7442dcca0f0049f8189a7afabe2d0a

test/test/lock/LockTests.java
fandreuz: Not needed indeed, a38c091828e95da58709967b9d7b8e706e386436

test/test/lock/RaceToLock.java
apangin: @fandreuz Some context in https://bugs.openjdk.org/browse/JDK-6736490 and related issues. In short, semantics of the "volatile" keyword does not allow JIT compiler to easily remove redundant write, and this write is accompanied by the memory barrier which can be expensive on performance critical code paths.

apangin:
Unfortunately, it still fails sometimes:
https://github.com/async-profiler/async-profiler/actions/runs/14276432594/job/40020127880?pr=1218#step:6:277

Issue #1208, opened 2025-04-02
Tests in GHA are not configured to run with x64 JDK as expected


fandreuz:
### Describe the bug

https://github.com/async-profiler/async-profiler/actions/runs/14202343679/job/39791972502#step:9:1

### Expected vs. actual behavior

Expected: 

Run with JDK `X64` as per the value of `JAVA_HOME`: `JAVA_HOME=$JAVA_HOME__X64 make test`

Actual:
cc -shared -fPIC -o build/test/lib/libreladyn.dylib test/native/libs/reladyn.c
cc -shared -fPIC -I/Users/runner/hostedtoolcache/Java_Corretto_jdk/11.0.26-4.1/**arm64**/Contents/Home/include -Isrc/helper -I/Users/runner/hostedtoolcache/Java_Corretto_jdk/11.0.26-4.1/**arm64**/Contents/Home/include/darwin -Isrc -o build/test/lib/libjnimalloc.dylib test/native/libs/jnimalloc.c
echo "Running cpp tests..."
Running cpp tests...
...

### Reproduction Steps

GHA

### Async-profiler version

Latest

### Environment details

MacOS tests in GHA

PR #1209, opened 2025-04-02
Wait for warmup phase completion in time-to-safepoint tests

fandreuz:
### Description
In this PR I add some logic to `JfrTests` to make sure the warmup is over when we start time-to-safepoint tests.

### Related issues
#1208 

### Motivation and context
I noticed that in some special conditions (i.e. M1 MacOS with x64 JDK) the warmup phase can take as much as 20 seconds. We have some hardcoded wait time in `TestProcess` to let the Java process under test prepare for the run, but that's not enough (`700ms`) and does not accomodate platform differences.

### How has this been tested?
`make test`

---

### DIFF
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -74,8 +74,14 @@ public class TestProcess implements Closeable {
 
         ProcessBuilder pb = new ProcessBuilder(cmd).inheritIO();
         if (test.output()) {
+            if (!test.waitWarmupOutput().isEmpty()) {
+                throw new IllegalArgumentException("TODO");
+            }
             pb.redirectOutput(createTempFile(STDOUT));
+        } else if (!test.waitWarmupOutput().isEmpty()) {
+            pb.redirectOutput(ProcessBuilder.Redirect.PIPE);
         }
+
         if (test.error()) {
             pb.redirectError(createTempFile(STDERR));
         }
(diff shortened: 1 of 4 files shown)


test/test/jfr/JfrTests.java
```diff
@@ -78,14 +79,35 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
         Assert.isGreater(eventsCount.get("jdk.ObjectAllocationInNewTLAB"), 50);
     }
 
+    private static void waitTtspWarmupOutput(TestProcess p) throws Exception {
```
fandreuz, line 82: This was initially in `TestProcess`, but I moved it here since at the moment this is only used by this class. However, there might be other occurrences following the same pattern which could benefit from the construct

test/one/profiler/test/TestProcess.java
```diff
@@ -279,6 +279,10 @@ public int exitCode() {
         return p.exitValue();
     }
 
+    public boolean processIsAlive() {
```
apangin, line 282: The class is already named `TestProcess`, another `process` in the name is redundant.

test/test/jfr/JfrTests.java
apangin: I agree it may be useful to have a generic method that waits for the target process to spin up, especially on macOS where launching new process may take a long time. But for a solution to be generic, ideally it should not rely on the application doing something special. I was thinking of using JDK Attach Mechanism (like `jstack`) to verify if the JVM is up and running the main class. Anyhow, this deserves a separate discussion, not for this PR.

test/test/jfr/JfrTests.java
```diff
@@ -78,14 +79,35 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
...
+        while (true) {
+            try (FileReader isr = new FileReader(p.getFile(TestProcess.STDOUT));
```
apangin, line 84: Is it necessary to reopen file on each iteration?

apangin:
> I noticed that in some special conditions (i.e. M1 MacOS with x64 JDK) the warmup phase can take as much as 20 seconds.

This is not normal. We certainly don't want tests to run for extra 20 seconds. Instead of waiting for warmup to complete, it's better to fix the root cause.

test/test/jfr/JfrTests.java
fandreuz: We could take an `InputStream` from the running process and wait for `read` to return something, but this would result in quite some code (with an additional thread) if we also want to fully support `output = true`. I can do it, I just figured we could keep it simple since it's test code

fandreuz:
Closing as discussed with @apangin

PR #1212, opened 2025-04-03
Run tests on MacOS with X64 JDK

fandreuz:
Description
Fixes https://github.com/async-profiler/async-profiler/issues/1208

Related issues
https://github.com/async-profiler/async-profiler/issues/1208

Motivation and context
See related issue

How has this been tested?
GHA

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -4,9 +4,6 @@ on:  # We are very liberal in terms of triggering builds. This should be revisit
   - push
   - pull_request
 
-env:
-  RELEASE_FROM_JAVA_VERSION: '11'
-
 jobs:
   build-jars:
     runs-on: ubuntu-latest
@@ -36,11 +33,18 @@ jobs:
         container: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
         include:
           - runson:
-              display: macos-14-arm64
+              display: macos-arm64
               name: macos-14
             java-version: 11
             java-distribution: corretto
             container: ""  # Not using container for mac-os as we have images only for linux
+          - runson:
+              display: macos-x64
+              name: macos-13
+            java-version: 11
+            java-distribution: corretto
+            architecture: x64
+            container: ""
           - runson:
               display: linux-arm64
               name: ubuntu-24.04-arm # There is no "latest" tag available (yet) as ARM runners are still in public preview
@@ -53,19 +57,12 @@ jobs:
       options: --privileged
     name: "test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }})"
     steps:
-      - name: Setup Java for macOS x64
-        uses: actions/setup-java@v4
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        with:
-          distribution: ${{ matrix.java-distribution }}
-          java-version: ${{ matrix.java-version }}
-          architecture: x64 # set up for x64, as the default arm one will override this later
-      - name: Setup Java for default architecture
+      - name: Setup Java
         uses: actions/setup-java@v4
         with:
           distribution: ${{ matrix.java-distribution }}
           java-version: ${{ matrix.java-version }}
-          # architecture: not specifying this defaults to architecture of the runner
+          architecture: ${{ matrix.architecture }}
       - name: Checkout sources
         uses: actions/checkout@v4
       - name: Build and test
@@ -94,7 +91,7 @@ jobs:
               make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc coverage -j
             ;;
           esac
-      - name: Upload test logs for default architecture
+      - name: Upload test logs
         uses: actions/upload-artifact@v4
 
         if: always() # we always want to upload test logs, especially when tests fail
@@ -109,26 +106,19 @@ jobs:
         with:
           name: test-coverage-${{ matrix.runson.display }}-${{ matrix.java-version }}
           path: build/test/coverage/
-      - name: Test macOS x64
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        run: JAVA_HOME=$JAVA_HOME_${{ env.JAVA_VERSION }}_X64 make test
-
       - name: Upload async-profiler binaries to workflow
         uses: actions/upload-artifact@v4
 
-        if: env.RELEASE_FROM_JAVA_VERSION == matrix.java-version
+        if: |
+          matrix.java-version == 11 &&
+          (
+            matrix.runson.display == 'linux-x64' ||
+            matrix.runson.display == 'linux-arm64' ||
+            matrix.runson.display == 'macos-arm64'
+          )
         with:
           name: ${{ steps.build.outputs.archive }}
           path: ${{ steps.build.outputs.archive }}
-      - name: Upload test logs for macOS x64
-        uses: actions/upload-artifact@v4
-
-        if: matrix.runson.name == 'macos-14' && always()
-        with:
-          name: test-logs-macos-14-x64-${{ matrix.java-version }}
-          path: |
-            build/test/logs/
-            hs_err*.log
   publish-only-on-push:
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     name: publish (nightly)


.github/workflows/test-and-publish-nightly.yml
```diff
@@ -36,10 +36,18 @@ jobs:
         container: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
         include:
           - runson:
-              display: macos-14-arm64
+              display: macos-13
```
apangin, line 39: Why not including architecture in display name similarly to how it's done for linux?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,21 +59,14 @@ jobs:
     container:
       image: ${{ matrix.container }}
       options: --privileged
-    name: "test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }})"
+    name: "test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }}${{ matrix.architecture && ' ' || '' }}${{ matrix.architecture }}"
```
apangin, line 62: I guess this change won't be needed if `matrix.runson.display` includes architecture.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -36,10 +36,18 @@ jobs:
...
+              display: macos-13
+              name: macos-13
+            java-version: 11
+            java-distribution: corretto
+            architecture: x64
+            container: ""  # Not using container for mac-os as we have images only for linux
+          - runson:
+              display: macos-14
               name: macos-14
             java-version: 11
```
apangin, line 48: Maybe use different JDK versions on x64 and arm64 for wider coverage? E.g. can be 21 on macos-arm64 and 11 on macos-x86.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -94,41 +95,28 @@ jobs:
...
           path: build/test/coverage/
-      - name: Test macOS x64
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        run: JAVA_HOME=$JAVA_HOME_${{ env.JAVA_VERSION }}_X64 make test
-
       - name: Upload async-profiler binaries to workflow
         uses: actions/upload-artifact@v4
 
-        if: env.RELEASE_FROM_JAVA_VERSION == matrix.java-version
+        if: ${{ env.RELEASE_FROM_JAVA_VERSION == matrix.java-version && matrix.runson.name != 'macos-13' }}
```
apangin, line 116: With this addition, `RELEASE_FROM_JAVA_VERSION` becomes less meaningful.
What about removing `RELEASE_FROM_JAVA_VERSION` and checking explicit values of matrix.runson/matrix.java-version instead?

.github/workflows/test-and-publish-nightly.yml
fandreuz: We could include it here, but we would be duplicating information which we have already in another variable (`matrix.architecture`). I'd personally like to remove `display` alltogether and only keep only `name` and `architecture`, what do you think?

.github/workflows/test-and-publish-nightly.yml
fandreuz: See comment above

.github/workflows/test-and-publish-nightly.yml
fandreuz: 42a2a2f87576bab42bbafbee859c0d924b7f0b59

.github/workflows/test-and-publish-nightly.yml
fandreuz: 7b3581d0f16c2893e77d66e1c9f337d6667e0dee

.github/workflows/test-and-publish-nightly.yml
apangin: Sounds reasonable. Please go ahead and do the change. Thanks.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -94,41 +92,28 @@ jobs:
...
           path: build/test/coverage/
-      - name: Test macOS x64
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        run: JAVA_HOME=$JAVA_HOME_${{ env.JAVA_VERSION }}_X64 make test
-
       - name: Upload async-profiler binaries to workflow
         uses: actions/upload-artifact@v4
 
-        if: env.RELEASE_FROM_JAVA_VERSION == matrix.java-version
+        if: ${{ matrix.java-version == '11' && matrix.runson.name != 'macos-13' }}
```
apangin, line 113: This will exclude macOS from releases entirely.
I suggest we release macOS binaries from builds on `macos-14`.

.github/workflows/test-and-publish-nightly.yml
fandreuz: Done

.github/workflows/test-and-publish-nightly.yml
fandreuz: Should we release only binaries built with JDK 11? `macos-14` now uses JDK 21 due to [this](https://github.com/async-profiler/async-profiler/pull/1212#discussion_r2027448323) previous comment

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -28,51 +25,46 @@ jobs:
...
             java-distribution: corretto
+            architecture: arm64
             container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
-    runs-on: ${{ matrix.runson.name }}
+    runs-on: ${{ matrix.name }}
     container:
       image: ${{ matrix.container }}
       options: --privileged
-    name: "test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }})"
+    name: "test (${{ matrix.name }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }}${{ matrix.architecture && ' ' || '' }}${{ matrix.architecture }}"
```
apangin, line 53: `architecture` is now always defined, right?

fandreuz: Yes, should I enforce it with a check?

.github/workflows/test-and-publish-nightly.yml
fandreuz: Please check now: https://github.com/async-profiler/async-profiler/actions/runs/14266840749

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -28,64 +25,53 @@ jobs:
...
           HASH=$(echo ${{ github.sha }} | cut -c-7)
-          case "${{ matrix.runson.name }}" in
-            macos*)
-              make COMMIT_TAG=$HASH FAT_BINARY=true release test -j
-            ;;
-            *)
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release test -j
-            ;;
-          esac
+          PACKAGE_NAME_SUFFIX="jdk${{ matrix.java-version }}"
```
apangin, line 67: What is it for and why is it needed?

fandreuz: See [here](https://github.com/async-profiler/async-profiler/actions/runs/14266840749)

image

Due to how we select the OS name in `Makefile`, we don't have a way to discriminate between artifacts produced on `macos-13` or `macos-14`. Thus, without this suffix, GitHub actions complains because two artifacts with the same name are being published (e.g. [here](https://github.com/async-profiler/async-profiler/actions/runs/14266362853)).

This happens to work because we use two different JDKs for `macos-13` and `macos-14`, but it would be probably cleaner to allow injecting OS info to the `Makefile` via a variable.

apangin: OK, I understand about architecture suffix, but why having jdk suffix? Ideally, contents of a package produced with different JDKs should be identical. Previously we had only one artifact per platform, and that was enough.

fandreuz: If each job publishes an artifact, we need to differentiate the name somehow. If the only difference between two jobs is the JDK version, we need to have it in the artifact name. 

Another solution is not to publish from all jobs, maybe we should keep the variable `RELEASE_FROM_JAVA_VERSION`?

apangin: Right, we don't need to publish from all jobs. My point on `RELEASE_FROM_JAVA_VERSION` was to get rid of the variable tied to a Java version in favor of explicit condition that specifies jobs to publish from.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -28,64 +25,53 @@ jobs:
   build-binaries-and-test:
     strategy:
       matrix:
-        runson:
-          - display: linux-x64
-            name: ubuntu-latest # Using "latest" here as the build and test will any ways run inside a container which we control
```
krk, line 33: Having `ubuntu-latest` in the gha / artifact name instead of `linux-x64` is confusing. It doesn't matter where it was built, especially when it is built in docker on `ubuntu-latest`:

```
Test and Publish Nightly Builds / test (ubuntu-24.04-arm, corretto 11 arm64 (pull_request)
Test and Publish Nightly Builds / test (ubuntu-latest, corretto 11 x64 (pull_request)
```

apangin: Good point!

fandreuz: I guess we need to keep the display name then

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -109,26 +110,13 @@ jobs:
...
           name: test-coverage-${{ matrix.runson.display }}-${{ matrix.java-version }}
           path: build/test/coverage/
-      - name: Test macOS x64
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        run: JAVA_HOME=$JAVA_HOME_${{ env.JAVA_VERSION }}_X64 make test
-
       - name: Upload async-profiler binaries to workflow
         uses: actions/upload-artifact@v4
 
         if: env.RELEASE_FROM_JAVA_VERSION == matrix.java-version
```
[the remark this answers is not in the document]
fandreuz: Could you check now @apangin ? I've expanded the matrix, it's not the cleanest solution but it gives us more control, thus we can have a per-element variable `publish`. We can make it cleaner when we get [YAML anchors](https://github.com/actions/runner/issues/1182#issuecomment-2722005293) in GHA

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -28,44 +28,73 @@ jobs:
   build-binaries-and-test:
     strategy:
       matrix:
-        runson:
-          - display: linux-x64
-            name: ubuntu-latest # Using "latest" here as the build and test will any ways run inside a container which we control
-        java-version: [11, 17, 21, 24]
-        java-distribution: [corretto]
-        container: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
```
krk, line 36: Is there a way to keep this matrix simplification and not repeat these in each configuration below?

fandreuz: Unfortunately the matrix is a cartesian product, thus we cannot put `published` here. I see only a few solutions:
- YAML anchors (not supported in GitHub Actions yet)
- Flattening the matrix as I did in the most recent revision
- Having a set of filtering variables like `RELEASE_FROM_JAVA_VERSION`
- Preparing a JSON (can be static, or dynamic as the output of another job) containing a list of configurations (OS, JDK, JDK version, publish? ), something similar to what they do [here](https://stackoverflow.com/a/65434401)

Let me know which one you prefer @krk @apangin

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
fandreuz: Reverted to an hardcoded condition in the publication step

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -109,26 +109,19 @@ jobs:
...
           path: build/test/coverage/
-      - name: Test macOS x64
-        if: ${{ matrix.runson.name == 'macos-14' }}
-        run: JAVA_HOME=$JAVA_HOME_${{ env.JAVA_VERSION }}_X64 make test
-
       - name: Upload async-profiler binaries to workflow
         uses: actions/upload-artifact@v4
 
-        if: env.RELEASE_FROM_JAVA_VERSION == matrix.java-version
+        if: |
```
apangin, line 115:
```
env:
  RELEASE_FROM_JAVA_VERSION: '11'
```
is no longer needed and can be removed then?

.github/workflows/test-and-publish-nightly.yml
krk: Do we ever want to publish from more than 1 configuration?

krk: Also, we could add more versions and distributions other than corretto for testing in the future - sounds like we actually need a cartesian product here, rather than duplicating entries.

fandreuz: Sure, last revision restored the matrix

Issue #1176, opened 2025-03-12
Crash on invoking parsed symbol


krk:
### Describe the bug

Symbols are parsed without segfault, but the calculated symbol address is incorrect, causing segfault when invoked.

```
Program received signal SIGILL, Illegal instruction.
0x0000000000011000 in ?? () from build/test/lib/libvirtaddroffset.so

(correct address to the function is 0x10000)
```

### Expected vs. actual behavior

No crash.

### virtaddroffset.c

```
int hello_viadoff(int x) {
    return x * 2;
}
```

### linker script virtaddroffset.ld

```
SECTIONS
{
  . = ALIGN(0x1000); /* Ensure alignment */
  
  .text 0x2000 : AT(0x1000)
  {
    *(.text)
    *(.text.*)
  }
  
  . = ALIGN(0x1000); /* Align next section to page boundary */
  .rodata : ALIGN(0x1000) 
  { 
    *(.rodata) 
    *(.rodata.*) 
  }
  
  . = ALIGN(0x1000);
  .data : ALIGN(0x1000) 
  { 
    *(.data) 
    *(.data.*) 
  }
  
  .got : ALIGN(0x1000) 
  { 
    *(.got) 
    *(.got.plt) 
  }
  
  . = ALIGN(0x1000);
  .dynamic : ALIGN(0x1000) 
  { 
    *(.dynamic) 
  }
  
  . = ALIGN(0x1000);
  .bss : ALIGN(0x1000) 
  { 
    *(.bss) 
    *(.bss.*) 
    *(COMMON) 
  }
}
```

### Test case

```
 ONLY_TEST_CASE(ViAdOf) {
    void* result = dlopen("libvirtaddroffset.so", RTLD_NOW);           
    if(!result) {
        printf("%s\n", dlerror());
    }
    ASSERT(result);                                                                
    Profiler::instance()->updateSymbols(false);

    PerfEventType* event_type = PerfEventType::forName("hello_viadoff");
    ASSERT(event_type);

    // cast event_type->config1 to int f()
    int (*hello_viadoff)(int) = (int (*)(int))event_type->config1;

    // should print "hello_viadoff: 4"
    printf("hello_viadoff: %d\n", hello_viadoff(2));
}
```

### Building

```
$(CC) -c -shared -fPIC -o $(TEST_LIB_DIR)/virtaddroffset.o test/native/libs/virtaddroffset.c
[...]

PR #1216, opened 2025-04-04
Add test for having different offset and virtaddr for LOAD segment

krk:
Add test for https://github.com/async-profiler/async-profiler/issues/1176

### How has this been tested?
`make test-cpp`

---

### DIFF
--- a/test/native/libs/vaddrdif.ld
+++ b/test/native/libs/vaddrdif.ld
@@ -1,0 +1,27 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+SECTIONS
+{
+  . = ALIGN(0x1000);
+  .text 0x2000 : AT(0x1000)
+  {
+    *(.text)
+    *(.text.*)
+  }
+
+  . = ALIGN(0x1000);
+  .rodata : ALIGN(0x1000)
+  {
+    *(.rodata)
+    *(.rodata.*)
+  }
+
+  . = ALIGN(0x1000);
+  .dynamic : ALIGN(0x1000)
+  {
+    *(.dynamic)
+  }
+}
--- a/test/native/libs/vaddrdif.ld
+++ b/test/native/libs/vaddrdif.ld
@@ -6,7 +6,7 @@
 SECTIONS
 {
   . = ALIGN(0x1000);
-  .text 0x2000 : AT(0xbeef0000)
+  .text 0xfe000 : AT(0x1000)
   {
     *(.text)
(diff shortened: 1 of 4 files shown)


test/native/libs/vaddrdif.ld
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+SECTIONS
+{
+  . = ALIGN(0x1000);
+  .text 0x2000 : AT(0x1000)
```
apangin, line 9: Thank you for adding the test, it's very helpful!
I'd make virtual load address completely different (e.g. `0xab000`) rather than 0x1000 increment to 0x1000 offset. Otherwise, since 0x1000 is overloaded with different meanings, it's easy to mix up things (I almost implemented a wrong fix that accidentally worked).

test/native/symbolsLinuxTest.cpp
```diff
@@ -33,4 +33,19 @@ TEST_CASE(ResolveFromRela_dyn_R_ABS64) {
...
 
+TEST_CASE(VirtAddrDifferentLoadAddr) {
+    void* result = dlopen("libvaddrdif.so", RTLD_NOW);
+    if (!result) {
+        printf("%s\n", dlerror());
+    }
+    ASSERT(result);
+    Profiler::instance()->updateSymbols(false);
+
+    PerfEventType* event_type = PerfEventType::forName("vaddrdif_square");
```
apangin, line 44: `Profiler::resolveSymbol()` will be more reliable.
First, because `PerfEventType::forName` attempts to use `dlsym`, which will return a correct address even when profiler's parser breaks. Second, because `PerfEventType::forName` returns an address of a breakpoint instead of an entry address of a function. They [differ](https://github.com/async-profiler/async-profiler/blob/49cca40e472b5156de4b2ab112429991d42088ac/src/arch.h#L118-L120) on PPC.

test/native/libs/vaddrdif.ld
```diff
@@ -6,7 +6,7 @@
 SECTIONS
 {
   . = ALIGN(0x1000);
-  .text 0x2000 : AT(0x1000)
+  .text 0x2000 : AT(0xbeef0000)
```
apangin, line 9: This has no effect; I meant
```suggestion
  .text 0xbeef0000 : AT(0x1000)
```

apangin: Not sure if `0xbeef0000` will work on 32-bit system.

Issue #1214, opened 2025-04-04
Crash during `Symbols::parseLibraries` when Async-Profiler is compiled with `MERGE=false`


fandreuz:
### Describe the bug

```
Thread 2 "java" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7fe2700 (LWP 20221)]
0x00007ffff52aba1a in std::local_Rb_tree_decrement(std::_Rb_tree_node_base*) () from ./build/lib/libasyncProfiler.so
(gdb) where
#0  0x00007ffff52aba1a in std::local_Rb_tree_decrement(std::_Rb_tree_node_base*) () from ./build/lib/libasyncProfiler.so
#1  0x00007ffff529f003 in std::_Rb_tree_iterator<void const*>::operator-- (this=0x7ffff7fdfd40) at /usr/include/c++/7/bits/stl_tree.h:302
#2  0x00007ffff529e8c6 in std::_Rb_tree<void const*, void const*, std::_Identity<void const*>, std::less<void const*>, std::allocator<void const*> >::_M_get_insert_unique_pos (this=0x7ffff54ece40 <_parsed_libraries>, __k=@0x7ffff7fdff10: 0x555555400000)
    at /usr/include/c++/7/bits/stl_tree.h:2047
#3  0x00007ffff529e4e1 in std::_Rb_tree<void const*, void const*, std::_Identity<void const*>, std::less<void const*>, std::allocator<void const*> >::_M_insert_unique<void const*> (this=0x7ffff54ece40 <_parsed_libraries>,
    __v=@0x7ffff7fdff10: 0x555555400000) at /usr/include/c++/7/bits/stl_tree.h:2090
#4  0x00007ffff529e314 in std::set<void const*, std::less<void const*>, std::allocator<void const*> >::insert (this=0x7ffff54ece40 <_parsed_libraries>, __x=@0x7ffff7fdff10: 0x555555400000) at /usr/include/c++/7/bits/stl_set.h:510
#5  0x00007ffff529de50 in parseLibrariesCallback (info=0x7ffff7fdffb0, size=64, data=0x8428) at src/symbols_linux.cpp:723
#6  0x00007ffff730c58b in __GI___dl_iterate_phdr (callback=0x7ffff529dc91 <parseLibrariesCallback(dl_phdr_info*, size_t, void*)>, data=0x8428) at dl-iteratephdr.c:76
[...]

PR #1219, opened 2025-04-04
Always sort sources so zInit is last.

krk:
### Description
Sort sources to zInit is last, ensuring its statics are initialized last - this works in gcc versions after 4.7.0 (related https://gcc.gnu.org/bugzilla/show_bug.cgi?id=46770).

### Related issues
https://github.com/async-profiler/async-profiler/issues/1214

### How has this been tested?
```
make MERGE=false
```

and starting the profiler with any target, as in https://github.com/async-profiler/async-profiler/issues/1214.

---


apangin:
Wow, that is brilliant! :)

Issue #1151, opened 2025-02-26
AP can deadlock the JVM is used alongside tcmalloc


MattAlp:
### Describe the bug

It appears that there's a deadlock that's reliably reproduced if one uses tcmalloc (i.e. via `LD_PRELOAD`) and AP together.

![Image](https://github.com/user-attachments/assets/deb65da2-3ffd-42bf-9f48-d3c67c12ce42)

## AP side

The profiler goes through the glibc’s function [dl_iterate_phdr()](https://man7.org/linux/man-pages/man3/dl_iterate_phdr.3.html) which performs a lock on glibc’s side to ensure the library is not [dlclose()](https://man7.org/linux/man-pages/man3/dlclose.3.html) at the same time we iterate through the symbols, i.e.

- locks `dl_load_write_lock`

- performs some [malloc()](https://man7.org/linux/man-pages/man3/free.3.html) in its callback, which may grow TCMalloc’s local cache and then reserve some memory from the OS, locking `tcmalloc::Static::pageheap_lock_`

## TCMalloc side

When TCMalloc actually reserves some memory from the OS:

- it locks `tcmalloc::Static::pageheap_lock_` to perform the memory allocations

- when this happens it keeps the current stack trace for the heap profiler

- this heap profiler then fetches the stack trace, calling glibc’s ELF symbols lookups and locking `dl_load_write_lock` (this can perform some mallocs but they do not re-trigger the heap profiler thanks to a [recursive call check](https://github.com/gperftools/gperftools/blob/gperftools-2.9.1/src/stacktrace_libunwind-inl.h#L61), thus tcmalloc cannot deadlock itself)

### Reproduction Steps

Using `LD_PRELOAD` with tcmalloc on x86_64, specifically in order to rely on the deadlocking `libunwind`-based stacktrace mechanism, should yield a fairly high defect rate. See https://github.com/gperftools/gperftools/wiki/gperftools'-stacktrace-capturing-methods-and-their-issues

### Async-profiler version

latest + whenever dl_iterate locks were introduced, in theory

### Environment details

x86_64 machine, `libunwind` based ST mechanism for tcmalloc, tcmalloc via `LD_PRELOAD`

Thanks to @trazfr for identifying & root-causing this internally. For our Datadog fork of AP, we are looking at writing a custom allocator for use within the callback function + eliminating the explicit `malloc` calls used by the CodeCache.

apangin:
[...]

Issue #1147, opened 2025-02-21
Async Profiler causes VM to become unresponsive


j-bahr:
### Describe the bug

When running async profiler normally from the CLI we asprof hangs and the VM eventually becomes unresponsive. This happens when attaching with the following command:

```
$ ./bin/asprof collect -d 10s <pid>
```

When running asprof, it never returns, we never see "profiling started" statement. The command just hangs indefinitely. This happens with 3.0 and all the latest nightly builds.

This may be circumstantial but in *some* instances running the `load` command before running a `collect` allows us to get samples, however we get the following error when attempt to load the profiler:

```
$ ./bin/asprof load <pid>
Connected to remote JVM
JVM response code = -1
 was not loaded.
lib.so: cannot open shared object file: No such file or directory
```

Again i'm not sure if this is luck or an actual workaround but it has worked a handful of times.

In some cases, even when the VM becomes unresponsive, I do see that `libasyncProfiler.so` is mapped into the programs address space. However running any commands from the asprof binary simply hang. 

```
7fc4ef787000-7fc4ef78d000 r--p 00000000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
7fc4ef78d000-7fc4ef7dd000 r-xp 00006000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
7fc4ef7dd000-7fc4ef7f1000 r--p 00056000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
7fc4ef7f1000-7fc4ef7f2000 ---p 0006a000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
7fc4ef7f2000-7fc4ef7f4000 r--p 0006a000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
7fc4ef7f4000-7fc4ef7f5000 rw-p 0006c000 00:1d 43533                      /tmp/async-profiler-3.0-e6e0494-linux-x64/lib/libasyncProfiler.so
```

A note on the setup. Our JVM is prefixed with a custom `ld.so` for locating native libs that are deployed for JNI usage and not part of the program's jar. 

```
/path/to/ld.so  /path/to/java <args>
```
[...]

PR #1220, opened 2025-04-06
#1125: Use dlopen instead of dl_iterate_phdr for parsing libraries

apangin:
### Description

Rework of #1156.

This PR fixes two issues:
1. Parsing libraries that have not been fully loaded/relocated.
2. Deadlock due to allocations under `dl_iterate_phdr`.

The idea is the same as in #1156: parse libraries in two steps: first, collect a list of new libraries from `/proc/self/maps`, and second, freeze each new library by dlopen'ing it with `RTLD_NOLOAD`, parse symbols and program headers, and finally add it to `CodeCacheArray`.

Key differences from the previous work:
1. Minimize changes outside `symbols_linux.cpp`.
2. Keep track of parsed libraries in a single set of inodes.
3. Do not parse program headers if dlopen fails.
4. Handle shared objects whose backing file was removed from disk.
5. Do not patch `dlopen` inside libasyncProfiler.

### Related issues

#1125, #1147, #1151

### How has this been tested?

`make test`, test from #1125, runnling IntelliJ IDEA with profiler enabled.

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -5,13 +5,15 @@
 
 #ifdef __linux__
 
-#include <set>
+#include <unordered_map>
+#include <unordered_set>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/mman.h>
+#include <dlfcn.h>
 #include <elf.h>
 #include <errno.h>
 #include <unistd.h>
@@ -117,6 +119,13 @@ class MemoryMapDesc {
           unsigned long minor = strtoul(colon + 1, NULL, 16);
           return major << 8 | minor;
       }
+};
+
+struct SharedLibrary {
+    char* file;
+    const char* map_start;
+    const char* map_end;
+    const char* image_base;
 };
 
 
@@ -642,8 +651,7 @@ void ElfParser::addRelocationSymbols(ElfSection* reltab, const char* plt) {
 
 Mutex Symbols::_parse_lock;
 bool Symbols::_have_kernel_symbols = false;
-static std::set<const void*> _parsed_libraries;
-static std::set<u64> _parsed_inodes;
+static std::unordered_set<u64> _parsed_inodes;
 
 void Symbols::parseKernelSymbols(CodeCache* cc) {
     int fd;
@@ -690,22 +698,19 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
     fclose(f);
 }
 
-static int parseLibrariesCallback(struct dl_phdr_info* info, size_t size, void* data) {
+static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count) {
     FILE* f = fopen("/proc/self/maps", "r");
     if (f == NULL) {
-        return 1;
+        return;
     }
 
-    CodeCacheArray* array = (CodeCacheArray*)data;
-    CodeCache* cc = NULL;
     const char* image_base = NULL;
     u64 last_inode = 0;
-    u64 cc_inode = 0;
     char* str = NULL;
     size_t str_size = 0;
     ssize_t len;
 
-    while ((len = getline(&str, &str_size, f)) > 0) {
+    while (max_count > 0 && (len = getline(&str, &str_size, f)) > 0) {
         str[len - 1] = 0;
 
         MemoryMapDesc map(str);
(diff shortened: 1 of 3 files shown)


src/symbols_linux.cpp
```diff
@@ -690,91 +698,67 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
...
     fclose(f);
-
-    return 1;
 }
 
 void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     MutexLocker ml(_parse_lock);
 
+    if (array->count() >= MAX_NATIVE_LIBS) {
+        return;
```
fandreuz, line 759: Shouldn't we log something here instead of skipping quietly?

src/symbols_linux.cpp
```diff
@@ -787,10 +771,52 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+            ElfParser::parseFile(cc, lib.map_start, lib.file, true);
+        } else {
+            // Parse debug symbols first
+            ElfParser::parseFile(cc, lib.image_base, lib.file, true);
+
+            // Protect library from unloading while parsing in-memory ELF program headers.
+            // Also, dlopen() ensures the library is fully loaded.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them.
+            void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
+            if (handle != NULL || dlerror() == NULL) {
```
fandreuz, line 806: https://man7.org/linux/man-pages/man3/dlerror.3p.html
```
The dlerror() function shall return a null-terminated character
string (with no trailing <newline>) that describes the last error
that occurred during dynamic linking processing. If no dynamic
linking errors have occurred since the last invocation of
dlerror(), dlerror() shall return NULL.  Thus, invoking dlerror()
a second time, immediately following a prior invocation, shall
result in NULL being returned.
```
Should we call `dlerror` before `dlopen` to make sure a non-null error refers to our call to `dlopen`, and not to something which happened before?

src/symbols_linux.cpp
apangin: Good idea. Added one-time warning.

src/symbols_linux.cpp
apangin: Good catch. Added `dlerror` call.

krk (approved):
thanks!

src/symbols_linux.cpp
```diff
@@ -787,10 +772,59 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+            ElfParser::parseFile(cc, lib.image_base, lib.file, true);
+
+            dlerror();  // reset any error from previous dl function calls
+
+            // Protect library from unloading while parsing in-memory ELF program headers.
+            // Also, dlopen() ensures the library is fully loaded.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them.
+            void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
+            if (handle != NULL || dlerror() == NULL) {
+                ElfParser::parseProgramHeaders(cc, lib.image_base, lib.map_end, OS::isMusl());
```
r1viollet, line 810: One thing that I liked in async-profiler, was your ability to parse deleted files.
Too bad dlopen can not work with file descriptors (or we could use fds).
With our out of process profilers, we noticed this with `libzstd-jni` (and we could not unwind stack frames within these binaries).
It feels like we will now have a similar limitation in `async-profiler` ?

r1viollet:
Overall a very smart strategy. Thanks for looking into this :bow:

src/symbols_linux.cpp
krk: Yes, as in the current version we cannot determine if such a library is "in the middle of being unloaded and unmapped" or not.

apangin: No, this change does not affect async-profiler ability to parse symbols of deleted libraries. I mean, deleting a file imposes some limitations, of course, but what async-profiler could parse before this change, it can still parse with dlopen approach.

The reason is, `dlopen(..., RTLD_NOLOAD)` returns a valid handle to an already loaded library even if the underlying file has been deleted, at least with glibc (with musl, it's not relevant anyway, since musl does not support library unloading).

r1viollet: Thanks! `RTLD_NOLOAD` and checking that handles are the same is good thinking. :bow:

PR #1223, opened 2025-04-08
add support for user JFR events

arielb1:
This adds an (unstable) `profiler.UserEvent` JFR event, that can be emitted by user native code and contain custom data.

Using T_STRING for the data instead of T_BYTE/F_ARRAY due to a bug in `jfr print`:

```
java.lang.NullPointerException
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printValue(PrettyWriter.java:357)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printArray(PrettyWriter.java:281)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printValue(PrettyWriter.java:325)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printFieldValue(PrettyWriter.java:273)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.print(PrettyWriter.java:213)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.print(PrettyWriter.java:76)
        at jdk.jfr/jdk.jfr.internal.tool.EventPrintWriter.print(EventPrintWriter.java:79)
        at jdk.jfr/jdk.jfr.internal.tool.Print.execute(Print.java:165)
        at jdk.jfr/jdk.jfr.internal.tool.Main.main(Main.java:84)
```
The relevant code is at
https://github.com/openjdk/jdk/blob/965e330cb4a972fd96ff04e149029d1ee0561940/src/jdk.jfr/share/classes/jdk/jfr/internal/tool/PrettyWriter.java#L357

`printArray` calls `printValue` with a `null` `field`, and `printValue` only supports a `null` `field` with a value of type `Byte`.

### How has this been tested?

---

### DIFF
--- a/src/asprof.cpp
+++ b/src/asprof.cpp
@@ -63,8 +63,8 @@ DLLEXPORT asprof_jfr_event_key asprof_register_jfr_event(const char* name) {
 #define asprof_str(s) #s
 
 DLLEXPORT asprof_error_t asprof_emit_jfr_event(asprof_jfr_event_key type, const uint8_t* data, size_t len) {
-    if (len > ASPROF_MAX_USER_JFR_LENGTH) {
-        return asprof_error("unable to emit JFR event larger than " asprof_str(ASPROF_MAX_USER_JFR_LENGTH) " bytes");
+    if (len > ASPROF_MAX_JFR_EVENT_LENGTH) {
+        return asprof_error("Unable to emit JFR event larger than " asprof_str(ASPROF_MAX_JFR_EVENT_LENGTH) " bytes");
     }
 
     UserEvent event;
--- a/src/asprof.cpp
+++ b/src/asprof.cpp
@@ -6,7 +6,9 @@
 #include "asprof.h"
 #include "hooks.h"
 #include "profiler.h"
+#include "tsc.h"
 #include "threadLocalData.h"
+#include "userEvents.h"
 
 static asprof_error_t asprof_error(const char* msg) {
     return (asprof_error_t)msg;
@@ -52,4 +54,18 @@ DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t out
 
 DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void) {
     return ThreadLocalData::getThreadLocalData();
+}
+
+DLLEXPORT asprof_user_jfr_key asprof_create_user_jfr_key(const char *name) {
+    return UserEvents::registerEvent(name);
+}
+
+DLLEXPORT int asprof_emit_user_jfr(asprof_user_jfr_key key, const uint8_t *data, size_t len) {
+    UserEvent event;
+    event._start_time = TSC::ticks();
+    event._key = key;
+    event._data = data;
+    event._len = len;
+    Profiler::instance()->recordEventOnly(USER_EVENT, &event);
+    return 0;
 }
--- a/src/asprof.cpp
+++ b/src/asprof.cpp
@@ -56,16 +56,22 @@ DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void) {
     return ThreadLocalData::getThreadLocalData();
 }
 
-DLLEXPORT asprof_user_jfr_key asprof_create_user_jfr_key(const char* name) {
+DLLEXPORT asprof_jfr_event_key asprof_register_jfr_event(const char* name) {
     return UserEvents::registerEvent(name);
 }
 
-DLLEXPORT int asprof_emit_user_jfr(asprof_user_jfr_key type, const uint8_t* data, size_t len, uint64_t _flags) {
+#define asprof_str(s) #s
+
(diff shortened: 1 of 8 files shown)


src/asprof.cpp
```diff
@@ -53,3 +55,17 @@ DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t out
 DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void) {
     return ThreadLocalData::getThreadLocalData();
 }
+
+DLLEXPORT asprof_user_jfr_key asprof_create_user_jfr_key(const char *name) {
+    return UserEvents::registerEvent(name);
+}
+
+DLLEXPORT int asprof_emit_user_jfr(asprof_user_jfr_key key, const uint8_t *data, size_t len) {
```
apangin, line 63: Will it ever be needed to record events with stack traces?
I initially thought event will have a bitmask of standard fields that should be recorded along with user-provided data. These fields may include `threadId` and `stackTrace`.

(It's more like a question for discussion, not necessarily for this PR).

src/asprof.h
```diff
@@ -63,6 +63,30 @@ typedef struct {
 DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void);
 typedef asprof_thread_local_data* (*asprof_get_thread_local_data_t)(void);
 
+
+typedef int asprof_user_jfr_key;
```
apangin, line 67: Maybe `unsigned int`?

Also, I'm not fully happy with "user" in this and other identifiers in this API (who is user?)
WDYT about `asprof_event_tag`? Alternative suggestions are welcome.

src/flightRecorder.cpp
```diff
@@ -1006,6 +1014,7 @@ class Recording {
         writePackages(buf, &lookup);
         writeSymbols(buf, &lookup);
         writeLogLevels(buf);
+        writeUserEventTypes(buf);
```
apangin, line 1017: Write them before log levels. There is no difference for parsing, but it will be easier to spot visually if JFR file is complete or truncated, since valid JFR will always end with predefined recognizable list of log levels.

src/flightRecorder.cpp
```diff
@@ -1193,6 +1202,18 @@ class Recording {
...
     }
 
+    void writeUserEventTypes(Buffer* buf) {
+        std::map<u32, const char*> events;
+        UserEvents::collect(events);
+
+        writePoolHeader(buf, T_USER_EVENT_TYPE, events.size());
+        for (std::map<u32, const char*>::const_iterator it = events.begin(); it != events.end(); ++it) {
+            flushIfNeeded(buf, RECORDING_BUFFER_LIMIT - MAX_STRING_LENGTH);
+            buf->putVar64(it->first);
```
apangin, line 1212: putVar32?

src/flightRecorder.cpp
```diff
@@ -1250,6 +1271,16 @@ class Recording {
         buf->put8(start, buf->offset() - start);
     }
 
+    void recordUserEvent(Buffer* buf, int tid, UserEvent *event) {
+        int start = buf->skip(5);
+        buf->put8(T_USER_EVENT);
+        buf->putVar64(event->_start_time);
+        buf->putVar32(tid);
+        buf->putVar64(event->_key);
```
apangin, line 1279: Var32?

src/asprof.cpp
```diff
@@ -53,3 +55,17 @@ DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t out
 DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void) {
     return ThreadLocalData::getThreadLocalData();
 }
+
+DLLEXPORT asprof_user_jfr_key asprof_create_user_jfr_key(const char *name) {
```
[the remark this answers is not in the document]
arielb1: fixed

src/asprof.h
arielb1: It can be -1 in case of failure, like in "normal" C functions.

arielb1: user = the user of async-profiler (as opposed to the developer of async profiler = you).

src/asprof.cpp
```diff
@@ -53,3 +55,17 @@ DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t out
 DLLEXPORT asprof_thread_local_data* asprof_get_thread_local_data(void) {
     return ThreadLocalData::getThreadLocalData();
 }
+
+DLLEXPORT asprof_user_jfr_key asprof_create_user_jfr_key(const char* name) {
```
apangin, line 59: As mentioned previously, I'm not in favor of having "user" in the names of public API functions. Whoever calls an API is already a user of this API.
What about `asprof_register_jfr_event`, `asprof_emit_jfr_event` or something like that?

Issue #1229, opened 2025-04-14
8.5% of stacks are incomplete if profiling with cstack=vm and -Xint


yanglong1010:
### Describe the bug

I used cstack=vm to profile java with -Xint option, found about 8.5% of the stacks only have 2 frames, the top is a Java frame, and the 2nd frame is [unknown].

This bug mainly affect x64, including JDK 8/11/17/21, because `rsp` is updated only by push, so more instructions needs to be executed before `is_plausible_interpreter_frame` becomes `true`. On aarch64, `sp` is updated in advance, this results in `is_plausible_interpreter_frame` becomes true in almost all cases, but there are still very few incomplete stacks, a possible reason is the instruction that pushes `sp` or `fp` onto the stack has not yet been executed.

see: [discussion](https://github.com/async-profiler/async-profiler/discussions/1225)

### Expected vs. actual behavior

- Expected: Incomplete stacks is very few
- Actual: 8.5% of stacks are incomplete

### Reproduction Steps

[simpledemo-0.0.1-SNAPSHOT.jar.zip](https://github.com/user-attachments/files/19732328/simpledemo-0.0.1-SNAPSHOT.jar.zip)

java -Xint -agentpath:/root/ap/async-profiler/build/lib/libasyncProfiler.so=start,event=cpu,file=profile.html,cstack=vm -jar simpledemo-0.0.1-SNAPSHOT.jar

### Async-profiler version

Latest

### Environment details

x64:
- Linux version 3.10.0-1160.119.1.el7.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) ) #1 SMP Tue Jun 4 14:43:51 UTC 2024
- openjdk version "1.8.0_412"; OpenJDK Runtime Environment (build 1.8.0_412-b08); OpenJDK 64-Bit Server VM (build 25.412-b08, mixed mode)
- openjdk version "11.0.23" 2024-04-16 LTS; OpenJDK Runtime Environment (Red_Hat-11.0.23.0.9-2.el7_9) (build 11.0.23+9-LTS); OpenJDK 64-Bit Server VM (Red_Hat-11.0.23.0.9-2.el7_9) (build 11.0.23+9-LTS, mixed mode, sharing)
[...]

PR #1234, opened 2025-04-15
Fix incomplete interpreter stacks produced by VM structs stack walker

yanglong1010:
Hi,

Could I have a review of this patch ? Thanks !

### Description

The instructions of `method entry point` can be viewed as 3 parts.

- The first part is the first instruction to the `epilogue` instruction which pushed `rsp` into `rbp`, i.e. [0x00007fffe4b4a220~ 0x00007fffe4b4a27b). In this part, the `rbp` is not changed, pointing to the caller of the current method.`is_plausible_interpreter_frame` is true, because `rsp` is far enough from `rbp`. Use the code bellow to unwind, we get the caller of the `Method*` in `rbx` as the current method, and we get the caller of the caller of the `Method*` in `rbx` as the sender. The correct way is to use `rbx` to get the current `Method*`, and unwind to the caller of the current method, but it's  a little complex, the place which saved `sender pc` changed multiple times (see the comments in the instructions bellow), so we trade correctness (missing the top frame) for simplicity, I think this is acceptable.

https://github.com/async-profiler/async-profiler/blob/b034e4c3141d0632fd60bc452d7532d84adad572/src/stackWalker.cpp#L310-L322

- The second part is from the end of the first part to the instruction which pushed `bcp` onto stack, i.e. [0x00007fffe4b4a27b~0x00007fffe4b4a2be). In this part, `is_plausible_interpreter_frame` is false, because before `bcp` is push onto stack, `rsp` is not far enough from `rbp`. Though the frame is not complete, fortunately we can use `rbp` to get the `caller sp`, `caller pc` and `caller fp`. The first instruction of this part, i.e. 0x00007fffe4b4a27b, needs to be treated specially, because the `sender sp` is pushed onto stack by it, so at this instruction, the value in `rbp[-1]` is random , we have to use register `r13` to get the `sender sp`.

- The third part is the remaining instructions, i.e. from 0x00007fffe4b4a2be to the end. In this part, `is_plausible_interpreter_frame` is true, and  we can use `rbp` to get the `caller sp`, `caller pc` and `caller fp`. 

There are some differences between the instructions generated with or without `-Xint`. Fortunately, the rules above can be applied to the Compiler (without -Xint) version too.

```
----------------------------------------------------------------------
method entry point (kind = zerolocals)  [0x00007fffe4b4a220, 0x00007fffe4b4a3c8]  424 bytes

--------------------------------------------------------------------------------      
  0x00007fffe4b4a220:   mov    0x8(%rbx),%rdx                               is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=*sp,             sender fp=fp
  0x00007fffe4b4a224:   movzwl 0x2e(%rdx),%ecx          
  0x00007fffe4b4a228:   movzwl 0x2c(%rdx),%edx          
  0x00007fffe4b4a22c:   sub    %ecx,%edx          
          
  0x00007fffe4b4a22e:   cmp    $0x1f5,%edx          
  0x00007fffe4b4a234:   jbe    0x00007fffe4b4a25f          
  0x00007fffe4b4a23a:   mov    %rdx,%rax          
  0x00007fffe4b4a23d:   shl    $0x3,%rax          
  0x00007fffe4b4a241:   add    $0x58,%rax          
  0x00007fffe4b4a245:   add    0x4d0(%r15),%rax                 
  0x00007fffe4b4a24c:   cmp    %rax,%rsp                                              
  0x00007fffe4b4a24f:   ja     0x00007fffe4b4a25f                       
  0x00007fffe4b4a255:   pop    %rax                                                                    
  0x00007fffe4b4a256:   mov    %r13,%rsp                                    is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=rax,             sender fp=fp
  0x00007fffe4b4a259:   push   %rax                                         is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13/*sp,        sender pc=rax,             sender fp=fp   
  0x00007fffe4b4a25a:   jmpq   0x00007fffe4b44f80                           is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=*sp/rax,         sender fp=fp                          
                               
  0x00007fffe4b4a25f:   pop    %rax                                         is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=*sp,             sender fp=fp
  0x00007fffe4b4a260:   lea    -0x8(%rsp,%rcx,8),%r14                       is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=rax,             sender fp=fp
              
  0x00007fffe4b4a265:   test   %edx,%edx                                                               
  0x00007fffe4b4a267:   jle    0x00007fffe4b4a276                           
  0x00007fffe4b4a26d:   pushq  $0x0                           
  0x00007fffe4b4a272:   dec    %edx                           
  0x00007fffe4b4a274:   jg     0x00007fffe4b4a26d                           
                          
generate_fixed_frame               
rbx: Method*                           
r13: sender sp                           
rax: return address                           
                           
  0x00007fffe4b4a276:   push   %rax                  <- return address                    
  0x00007fffe4b4a277:   push   %rbp                  <- old rbp             is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=rax/*sp,         sender fp=fp
  0x00007fffe4b4a278:   mov    %rsp,%rbp             <- new rbp             is_plausible_interpreter_frame=true,  Method*=rbx,           sender sp=r13,            sender pc=sp[1]/rax,       sender fp=fp/*sp
  0x00007fffe4b4a27b:   push   %r13                  <- sender sp           is_plausible_interpreter_frame=false, Method*=rbx,           sender sp=r13,            sender pc=fp[1]/sp[1]/rax, sender fp=*fp/*sp
                  
  0x00007fffe4b4a27d:   pushq  $0x0                                         is_plausible_interpreter_frame=false, Method*=rbx,           sender sp=fp[-1]/r13/*sp, sender pc=fp[1]/rax/sp[2], sender fp=*fp/sp[1]
  0x00007fffe4b4a282:   mov    0x8(%rbx),%r13                  
  0x00007fffe4b4a286:   lea    0x38(%r13),%r13                              is_plausible_interpreter_frame=false, Method*=rbx,           sender sp=fp[-1]/*sp,     sender pc=fp[1]/rax/sp[2], sender fp=*fp/sp[1]
  0x00007fffe4b4a28a:   push   %rbx                  <- push Method*            
  0x00007fffe4b4a28b:   mov    0x8(%rbx),%rdx                               is_plausible_interpreter_frame=false, Method*=fp[-3]/rbx,    sender sp=fp[-1],         sender pc=fp[1]/rax,       sender fp=*fp
  0x00007fffe4b4a28f:   mov    0x8(%rdx),%rdx    
  0x00007fffe4b4a293:   mov    0x18(%rdx),%rdx    
  0x00007fffe4b4a297:   mov    0x70(%rdx),%rdx    
  0x00007fffe4b4a29b:   mov    (%rdx),%rdx    
  0x00007fffe4b4a29e:   push   %rdx    
  0x00007fffe4b4a29f:   pushq  $0x0    
  0x00007fffe4b4a2a4:   mov    0x8(%rbx),%rdx    
  0x00007fffe4b4a2a8:   mov    0x8(%rdx),%rdx    
  0x00007fffe4b4a2ac:   mov    0x10(%rdx),%rdx    
  0x00007fffe4b4a2b0:   push   %rdx    
  0x00007fffe4b4a2b1:   mov    %r14,%rax    
  0x00007fffe4b4a2b4:   sub    %rbp,%rax                                   is_plausible_interpreter_frame=false, Method*=fp[-3]/rbx,    sender sp=fp[-1],         sender pc=fp[1],           sender fp=*fp
  0x00007fffe4b4a2b7:   shr    $0x3,%rax    
  0x00007fffe4b4a2bb:   push   %rax    
  0x00007fffe4b4a2bc:   push   %r13                  <- bcp                

  0x00007fffe4b4a2be:   pushq  $0x0                                        is_plausible_interpreter_frame=true,  Method*=fp[-3]/rbx,    sender sp=fp[-1],         sender pc=fp[1],           sender fp=*fp
  0x00007fffe4b4a2c3:   mov    %rsp,(%rsp)    

  0x00007fffe4b4a2c7:   movb   $0x1,0x479(%r15)
  0x00007fffe4b4a2cf:   cmp    0x4e8(%r15),%rsp
  0x00007fffe4b4a2d6:   ja     0x00007fffe4b4a378
  0x00007fffe4b4a2dc:   mov    %eax,-0x1000(%rsp)
  0x00007fffe4b4a2e3:   mov    %eax,-0x2000(%rsp)
  0x00007fffe4b4a2ea:   mov    %eax,-0x3000(%rsp)
  0x00007fffe4b4a2f1:   mov    %eax,-0x4000(%rsp)
  0x00007fffe4b4a2f8:   mov    %eax,-0x5000(%rsp)
  0x00007fffe4b4a2ff:   mov    %eax,-0x6000(%rsp)
  0x00007fffe4b4a306:   mov    %eax,-0x7000(%rsp)
  0x00007fffe4b4a30d:   mov    %eax,-0x8000(%rsp)
  0x00007fffe4b4a314:   mov    %eax,-0x9000(%rsp)
  0x00007fffe4b4a31b:   mov    %eax,-0xa000(%rsp)
  0x00007fffe4b4a322:   mov    %eax,-0xb000(%rsp)
  0x00007fffe4b4a329:   mov    %eax,-0xc000(%rsp)
  0x00007fffe4b4a330:   mov    %eax,-0xd000(%rsp)
  0x00007fffe4b4a337:   mov    %eax,-0xe000(%rsp)
  0x00007fffe4b4a33e:   mov    %eax,-0xf000(%rsp)
  0x00007fffe4b4a345:   mov    %eax,-0x10000(%rsp)
  0x00007fffe4b4a34c:   mov    %eax,-0x11000(%rsp)
  0x00007fffe4b4a353:   mov    %eax,-0x12000(%rsp)
  0x00007fffe4b4a35a:   mov    %eax,-0x13000(%rsp)
  0x00007fffe4b4a361:   mov    %eax,-0x14000(%rsp)
  0x00007fffe4b4a368:   cmp    0x4e0(%r15),%rsp
  0x00007fffe4b4a36f:   jbe    0x00007fffe4b4a378
  0x00007fffe4b4a371:   mov    %rsp,0x4e8(%r15)
  0x00007fffe4b4a378:   movb   $0x0,0x479(%r15)
  0x00007fffe4b4a380:   cmpb   $0x0,0x1283e68e(%rip)
  0x00007fffe4b4a387:   je     0x00007fffe4b4a3b5
  0x00007fffe4b4a38d:   mov    -0x18(%rbp),%rsi
  0x00007fffe4b4a391:   mov    %r15,%rdi
  0x00007fffe4b4a394:   test   $0xf,%spl
  0x00007fffe4b4a398:   je     0x00007fffe4b4a3b0
  0x00007fffe4b4a39e:   sub    $0x8,%rsp
  0x00007fffe4b4a3a2:   callq  0x00007ffff6ce2480 = SharedRuntime::dtrace_method_entry(JavaThread*, Method*)
  0x00007fffe4b4a3a7:   add    $0x8,%rsp
  0x00007fffe4b4a3ab:   jmpq   0x00007fffe4b4a3b5
  0x00007fffe4b4a3b0:   callq  0x00007ffff6ce2480 = SharedRuntime::dtrace_method_entry(JavaThread*, Method*)
  0x00007fffe4b4a3b5:   movzbl 0x0(%r13),%ebx
  0x00007fffe4b4a3ba:   movabs $0x7ffff73c4a20,%r10
  0x00007fffe4b4a3c4:   jmpq   *(%r10,%rbx,8)
```

### Related issues
#1229 

### Motivation and context
#1225 

### How has this been tested?
Before the fix, incomplete stacks are obvious
image
image
image
image

With the fix, incomplete stacks disappeared
image
image
image
image

On aarch64, the results are the same as before:
Run without the fix
image

Run with the fix
image

---

### DIFF
--- a/src/stackFrame.h
+++ b/src/stackFrame.h
@@ -74,8 +74,11 @@ class StackFrame {
 
     bool checkInterruptedSyscall();
 
+    void unwindIncompleteFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
+
     // Check if PC points to a syscall instruction
     static bool isSyscall(instruction_t* pc);
+    static bool isSenderSPOnStack(instruction_t* pc, bool is_plausible_interpreter_frame);
 };
 
 #endif // _STACKFRAME_H
--- a/src/stackFrame.h
+++ b/src/stackFrame.h
@@ -78,7 +78,6 @@ class StackFrame {
 
     // Check if PC points to a syscall instruction
     static bool isSyscall(instruction_t* pc);
-    static bool isSenderSPOnStack(instruction_t* pc, bool is_plausible_interpreter_frame);
 };
 
 #endif // _STACKFRAME_H
--- a/src/stackFrame.h
+++ b/src/stackFrame.h
@@ -67,14 +67,13 @@ class StackFrame {
     bool unwindStub(instruction_t* entry, const char* name, uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
     bool unwindCompiled(NMethod* nm, uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
     bool unwindAtomicStub(const void*& pc);
+    void unwindIncompleteIntFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
 
     void adjustSP(const void* entry, const void* pc, uintptr_t& sp);
 
     bool skipFaultInstruction();
 
     bool checkInterruptedSyscall();
-
-    void unwindIncompleteFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
 
     // Check if PC points to a syscall instruction
     static bool isSyscall(instruction_t* pc);
--- a/src/stackFrame_aarch64.cpp
+++ b/src/stackFrame_aarch64.cpp
@@ -65,7 +65,16 @@ uintptr_t StackFrame::method() {
 }
 
 uintptr_t StackFrame::senderSP() {
-    return (uintptr_t)REG(regs[19], x[19]);
+    // On JDK 8/11/17, sender sp is stored in x13, and on JDK 20 and above, sender sp is stored in x19
+    // https://bugs.openjdk.org/browse/JDK-8288971
+    // https://github.com/openjdk/jdk8u-dev/blob/55273f7267b95cf38743bb32ea61a513fbafb06e/hotspot/src/cpu/aarch64/vm/templateInterpreter_aarch64.cpp#L633
+    // https://github.com/openjdk/jdk17u-dev/blob/85d0ab55d6bae2aea4368e6668e361db8a9cbd1c/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp#L841
+    // https://github.com/openjdk/jdk21u-dev/blob/4daaffcd9dae87b5b51f9277e7f407a7d31a1eb9/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp#L870
+    if (VM::hotspot_version() >= 20) {
+        return (uintptr_t)REG(regs[19], x[19]);
+    } else {
+        return (uintptr_t)REG(regs[13], x[13]);
+    }
 }
 
 void StackFrame::ret() {
@@ -230,9 +239,48 @@ bool StackFrame::checkInterruptedSyscall() {
 #endif
 }
 
+void StackFrame::unwindIncompleteFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp) {
+    pc = (uintptr_t)stripPointer(*(void**)sp);
+    sp = senderSP();
+}
+
 bool StackFrame::isSyscall(instruction_t* pc) {
     // svc #0 or svc #80
     return (*pc & 0xffffefff) == 0xd4000001;
 }
 
+// On aarch64, for [method entry point], is_plausible_interpreter_frame is always true,
+// so we need to parse the instructions to determine whether sender sp is on the stack.
+// If the instruction is in the range of (set fp = sp, push sender sp], then sender sp is not on the stack.
+bool StackFrame::isSenderSPOnStack(instruction_t* pc, bool is_plausible_interpreter_frame) {
+    // Bellow OpenJDK 20
+    // 0x0000ffff873184e4:   stp	x29, x30, [sp, #80]     <- push fp, lr    // 0xa9057bfd
+    // 0x0000ffff873184e8:   add	x29, sp, #0x50          <- set fp = sp    // 0x910143fd
+    // 0x0000ffff873184ec:   stp	xzr, x13, [sp, #64]     <- push sender sp // 0xa90437ff
+
+    // OpenJDK 20 and above
+    // 0x0000ffffac4e76f4:   stp	x20, x22, [sp, #-96]! <- push java expression stack pointer, bcp
+    // 0x0000ffffac4e76f8:   stp	xzr, x12, [sp, #48]   <- push Method*
+    // 0x0000ffffac4e76fc:   stp	x29, x30, [sp, #80]   <- push fp, lr  // 0xa9057bfd
+    // 0x0000ffffac4e7700:   add	x29, sp, #0x50        <- set fp = sp  // 0x910143fd
+    // 0x0000ffffac4e7704:   ldr	x26, [x12, #8]                        // 0xf940059a
+    // 0x0000ffffac4e7708:   ldr	x26, [x26, #8]                        // 0xf940075a
+    // 0x0000ffffac4e770c:   ldr	x26, [x26, #16]                       // 0xf9400b5a
+    // 0x0000ffffac4e7710:   sub	x8, x24, x29                          // 0xcb1d0308
+    // 0x0000ffffac4e7714:   lsr	x8, x8, #3                            // 0xd343fd08
+    // 0x0000ffffac4e7718:   stp	x8, x26, [sp, #16]                    // 0xa9016be8
+    // 0x0000ffffac4e771c:   stp	xzr, x19, [sp, #64] <- push sender sp // 0xa9044fff
+
+    instruction_t v = *pc;
+    return !(
+        (v == 0xf940059a && *(pc - 1) == 0x910143fd) || // check the current and the [set fp = sp] instruction
+        (v == 0xf940075a && *(pc - 2) == 0x910143fd) ||
+        (v == 0xf9400b5a && *(pc - 3) == 0x910143fd) ||
+        (v == 0xcb1d0308 && *(pc - 4) == 0x910143fd) ||
+        (v == 0xd343fd08 && *(pc - 5) == 0x910143fd) ||
+        (v == 0xa9016be8 && *(pc - 6) == 0x910143fd) ||
+        (v == 0xa9044fff && *(pc - 7) == 0x910143fd) ||
+        (v == 0xa90437ff && *(pc - 1) == 0x910143fd)
+        );
+}
 #endif // __aarch64__
--- a/src/stackFrame_aarch64.cpp
+++ b/src/stackFrame_aarch64.cpp
@@ -240,7 +240,7 @@ bool StackFrame::checkInterruptedSyscall() {
 #endif
 }
 
-void StackFrame::unwindIncompleteFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp) {
+void StackFrame::unwindIncompleteIntFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp) {
     pc = (uintptr_t)stripPointer(*(void**)sp);
     sp = senderSP();
 }
--- a/src/stackFrame_aarch64.cpp
+++ b/src/stackFrame_aarch64.cpp
@@ -249,39 +249,4 @@ bool StackFrame::isSyscall(instruction_t* pc) {
     // svc #0 or svc #80
     return (*pc & 0xffffefff) == 0xd4000001;
 }
-
-// On aarch64, for [method entry point], is_plausible_interpreter_frame is always true,
-// so we need to parse the instructions to determine whether sender sp is on the stack.
-// If the instruction is in the range of (set fp = sp, push sender sp], then sender sp is not on the stack.
-bool StackFrame::isSenderSPOnStack(instruction_t* pc, bool is_plausible_interpreter_frame) {
-    // Bellow OpenJDK 20
-    // 0x0000ffff873184e4:   stp	x29, x30, [sp, #80]     <- push fp, lr    // 0xa9057bfd
-    // 0x0000ffff873184e8:   add	x29, sp, #0x50          <- set fp = sp    // 0x910143fd
-    // 0x0000ffff873184ec:   stp	xzr, x13, [sp, #64]     <- push sender sp // 0xa90437ff
-
-    // OpenJDK 20 and above
-    // 0x0000ffffac4e76f4:   stp	x20, x22, [sp, #-96]! <- push java expression stack pointer, bcp
-    // 0x0000ffffac4e76f8:   stp	xzr, x12, [sp, #48]   <- push Method*
-    // 0x0000ffffac4e76fc:   stp	x29, x30, [sp, #80]   <- push fp, lr  // 0xa9057bfd
-    // 0x0000ffffac4e7700:   add	x29, sp, #0x50        <- set fp = sp  // 0x910143fd
-    // 0x0000ffffac4e7704:   ldr	x26, [x12, #8]                        // 0xf940059a
-    // 0x0000ffffac4e7708:   ldr	x26, [x26, #8]                        // 0xf940075a
-    // 0x0000ffffac4e770c:   ldr	x26, [x26, #16]                       // 0xf9400b5a
-    // 0x0000ffffac4e7710:   sub	x8, x24, x29                          // 0xcb1d0308
-    // 0x0000ffffac4e7714:   lsr	x8, x8, #3                            // 0xd343fd08
-    // 0x0000ffffac4e7718:   stp	x8, x26, [sp, #16]                    // 0xa9016be8
-    // 0x0000ffffac4e771c:   stp	xzr, x19, [sp, #64] <- push sender sp // 0xa9044fff
-
-    instruction_t v = *pc;
-    return !(
-        (v == 0xf940059a && *(pc - 1) == 0x910143fd) || // check the current and the [set fp = sp] instruction
-        (v == 0xf940075a && *(pc - 2) == 0x910143fd) ||
-        (v == 0xf9400b5a && *(pc - 3) == 0x910143fd) ||
-        (v == 0xcb1d0308 && *(pc - 4) == 0x910143fd) ||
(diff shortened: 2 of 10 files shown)


src/stackWalker.cpp
```diff
@@ -333,8 +338,19 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+                                    sp = frame.senderSP();
+                                } else {
+                                    sp = ((uintptr_t*)fp)[InterpreterFrame::sender_sp_offset];
+                                }
+                                pc = stripPointer(((void**)fp)[FRAME_PC_SLOT]);
+                                fp = *(uintptr_t*)fp;
+                            } else {
+                                pc = stripPointer(*(void**)sp);
+                                sp = frame.senderSP();
+                            }
```
fandreuz, line 353: How about removing the flag `X64`?
```suggestion
#ifdef __x86_64__
                                char* ptr = (char*)pc;
                                if (*ptr == 0x41 && *(ptr + 1)== 0x55) { // push %r13
                                    sp = frame.senderSP();
                                } else {
                                    sp = ((uintptr_t*)fp)[InterpreterFrame::sender_sp_offset];
                                }
                                pc = stripPointer(((void**)fp)[FRAME_PC_SLOT]);
                                fp = *(uintptr_t*)fp;
#else
                                pc = stripPointer(*(void**)sp);
                                sp = frame.senderSP();
#endif
```

yanglong1010:
I edited the description of the first part:


- ~~Use the code bellow to unwind, we get the caller of caller. If we want to unwind to the caller of the current method, it's a little complex (see the comments in the instructions bellow), so we trade correctness (missing a frame) for simplicity.~~

- Use the code bellow to unwind, we get the caller of the Method* in rbx as the current method, and we get the caller of the caller of the Method* in rbx as the sender. The correct way is to use rbx to get the current Method*, and unwind to the caller of the current method, but it's a little complex, the place which saved sender pc changed multiple times (see the comments in the instructions bellow), so we trade correctness (missing the top frame) for simplicity, I think this is acceptable.

src/stackWalker.cpp
yanglong1010: Thanks, pushed

fandreuz:
Hi @yanglong1010, thanks for the PR and for looking into the issue! Do you think it would be possible to write a self-contained test to check that stacktraces are now complete when profiling with `cstack=vm` a simple program started with `-Xint`?

yanglong1010:
> Hi @yanglong1010, thanks for the PR and for looking into the issue! Do you think it would be possible to write a self-contained test to check that stacktraces are complete after your change, when profiling with `cstack=vm` a simple program started with `-Xint`?

I think it's totally feasible. Let me try ~

yanglong1010:
Hi, the test is added. The test fails on the async-profiler without the fix, and it passes on the async-profiler with this fix.

```
TESTS=test.cstack.CstackTests make test-java
```

test/test/cstack/CstackTests.java
```diff
@@ -41,4 +42,10 @@ public void vmStructs(TestProcess p) throws Exception {
         assert out.contains("call_stub");
         assert out.contains("JavaMain");
     }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX, jvmArgs = "-Xint", jvm = Jvm.HOTSPOT)
+    public void interpreter(TestProcess p) throws Exception {
```
fandreuz, line 47:
```suggestion
    public void interpretedOnlyModeHasLowUnknownRatio(TestProcess p) throws Exception {
```

src/stackWalker.cpp
```diff
@@ -333,8 +332,19 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                             sp = frame.senderSP();
                             fp = *(uintptr_t*)fp;
                         } else {
+#if defined(__x86_64__)
+                            char* ptr = (char*)pc;
+                            if (*ptr == 0x41 && *(ptr + 1)== 0x55) { // push %r13
```
apangin, line 337: Architecture-specific unwinding, especially with instruction stream parsing, typically sits in `stackFrame_<arch>.cpp`
But I wonder if this can be unified? Isn't AArch64 interpreter frame layout similar?

test/test/cstack/CstackTests.java
```diff
@@ -41,4 +42,10 @@ public void vmStructs(TestProcess p) throws Exception {
         assert out.contains("call_stub");
         assert out.contains("JavaMain");
     }
+
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX, jvmArgs = "-Xint", jvm = Jvm.HOTSPOT)
+    public void interpretedOnlyModeHasLowUnknownRatio(TestProcess p) throws Exception {
```
apangin, line 47: No story-long method names, please. Test method names can be specified in a command line to run only selected tests; it's inconvenient to type such a name. If you want to clarify what the test is doing, add a comment instead.

src/stackWalker.cpp
apangin: Also, is this instruction a reliable indicator for using `frame.senderSP()`? I assume `push %r13` can occur in many more places in the interpreter.

apangin (commented):
Thank you for the detailed analysis and the proposed fix.

src/stackWalker.cpp
apangin: Finally, it may be useful to add a brief explanation of the logic in the code, or to leave a reference to the description in this PR at least.

yanglong1010: > Architecture-specific unwinding, especially with instruction stream parsing, typically sits in `stackFrame_<arch>.cpp` But I wonder if this can be unified? Isn't AArch64 interpreter frame layout similar?

I'm checking and will reply later.

yanglong1010:
Hi, the stackWalker.cpp now is architecture unified. 

On aarch64, `method entry point` always sub `sp` first (make room for the current frame), and then set `fp` = `sp`, so for `method entry point`, `is_plausible_interpreter_frame` is true almost all the time.  So the execution path almost goes through the line bellow. But before `sender sp` is push onto stask, we could only read it from register (and the register changed from `x13` to `x19` on JDK 20).
https://github.com/async-profiler/async-profiler/blob/387dee13b8838a8775e890d8ae269872645b8b2f/src/stackWalker.cpp#L318

Ran on JDK 21 with the patch, On aarch64, there is a visible decrease in the `break_interpreted` stacks.
before (the left most part)
image
after (the left most part)
image

yanglong1010:
Please review again, thanks ~

src/stackFrame_aarch64.cpp
```diff
@@ -230,9 +240,48 @@ bool StackFrame::checkInterruptedSyscall() {
...
+    instruction_t v = *pc;
+    return !(
+        (v == 0xf940059a && *(pc - 1) == 0x910143fd) || // check the current and the [set fp = sp] instruction
+        (v == 0xf940075a && *(pc - 2) == 0x910143fd) ||
+        (v == 0xf9400b5a && *(pc - 3) == 0x910143fd) ||
+        (v == 0xcb1d0308 && *(pc - 4) == 0x910143fd) ||
+        (v == 0xd343fd08 && *(pc - 5) == 0x910143fd) ||
+        (v == 0xa9016be8 && *(pc - 6) == 0x910143fd) ||
+        (v == 0xa9044fff && *(pc - 7) == 0x910143fd) ||
+        (v == 0xa90437ff && *(pc - 1) == 0x910143fd)
```
apangin, line 284: This is not maintainable, sorry. These checks are very intricate and fragile. Moreover, accessing memory before current instruction is generally unsafe.

You previously mentioned there was no problem on ARM64, so what is it all about?

yanglong1010: Let me clarify.

In [#1229](https://github.com/async-profiler/async-profiler/issues/1229), I said:
> This bug mainly affects x64, including JDK 8/11/17/21, because rsp is updated only by push, so more instructions needs to be executed before is_plausible_interpreter_frame becomes true. On aarch64, sp is updated in advance, this results in is_plausible_interpreter_frame becomes true in almost all cases, but there are still very few incomplete stacks, a possible reason is the instruction that pushes sp or fp onto the stack has not yet been executed.

On x64 and any JDK version, the proportion of incomplete stacks is over 5% (on JDK 21 is 8.5%), while on aarch64 JDK 21, the proportion is less than 1%, on aarch64 JDK 17 and below the proportion is very few and ignorable. Because on x64 JDK, there are more than 20 instructions which unwind out incomplete stacks, while on JDK 21 aarch64, the number of incomplete stack instruction is about 7, and on JDK 17 aarch64 and bellow, the number is 1.

On aarch64, If use my patch, most incomplete stacks described above will disappear. If you think this is dangerous and not maintainable, I can remove the aarch64 part and leave some comments.

src/stackFrame.h
```diff
@@ -74,6 +74,8 @@ class StackFrame {
 
     bool checkInterruptedSyscall();
 
+    void unwindIncompleteFrame(uintptr_t& pc, uintptr_t& sp, uintptr_t& fp);
```
apangin, line 77: Please place it next to other unwind* functions and name it to make clear the function is relevant to interpreter only (something like `unwindIncompleteInterpretedFrame` or preferably shorter).

src/stackFrame_x64.cpp
```diff
@@ -18,6 +18,8 @@
 #  define REG(l, m)  _ucontext->uc_mcontext.gregs[REG_##l]
 #endif
 
+const u32 EPILOGUE = 0xec8b4855; // push %rbp, mov %rsp,%rbp
```
apangin, line 21: It's a _prologue_, not _epilogue_, right?
Anyways, this change is not related to the issue, I propose to exclude it.

src/stackFrame_x64.cpp
```diff
@@ -221,8 +223,33 @@ bool StackFrame::checkInterruptedSyscall() {
...
+    // ...
+    //  0x00007fffe4b4a276:   push   %rax
+    //  0x00007fffe4b4a277:   push   %rbp
+    //  0x00007fffe4b4a278:   mov    %rsp,%rbp
+    //  0x00007fffe4b4a27b:   push   %r13            <- push sender sp
+    //  0x00007fffe4b4a27d:   pushq  $0x0
+    //  0x00007fffe4b4a282:   mov    0x8(%rbx),%r13
+    //  0x00007fffe4b4a286:   lea    0x38(%r13),%r13
+    // ...
+    sp = ((uintptr_t*)fp)[InterpreterFrame::sender_sp_offset];
```
apangin, line 247: But this branch is not only called for method entry point?
How can you be sure that `fp` is valid?

src/stackFrame_x64.cpp
yanglong1010: Excluded

src/stackFrame.h
yanglong1010: Renamed to unwindIncompleteIntFrame, similar to -Xint in JDK

src/stackFrame_x64.cpp
yanglong1010: Yes, I am not sure. I will investigate this and com back later.

PR #1226, opened 2025-04-11
Build and run tests on Alpine

fandreuz:
### Description
In this PR I add a new Alpine build/test target.

### Related issues
Fixes #1136 

### Motivation and context
Find musl-specific issues early.

### How has this been tested?
GHA

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

Issue #1136, opened 2025-02-14
Run tests on Alpine Linux in CI


apangin:
### Describe the feature

Include Alpine Linux in test matrix.

### Use Case

Find musl-specific issues early.
For example, this bug - e6407ee3494030825485240faca5b966c0a42aac was only found during manual testing.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1235, opened 2025-04-16
Fix musl-specific problems

fandreuz:
### Description
Fixing several issues which I encountered while working on #1226:
- https://github.com/async-profiler/async-profiler/pull/1226#issuecomment-2797294532
- `malloc` is not properly hooked sometimes, thus native memory profiling does not work properly. The problem is not malloc-specific, other similar functions are affected as well

### Related issues
#1136 

### Motivation and context
We want to fully support Alpine and other musl-based environment.

### How has this been tested?
`docker run --privileged -it --rm -v $(pwd):/async-profiler amazoncorretto:11-alpine-jdk sh -c 'apk add --no-cache musl-dev util-linux make gcc g++ linux-headers && cd /async-profiler && rm -rf build && make -j test'`

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ endif
 CFLAGS_EXTRA ?=
 CXXFLAGS_EXTRA ?=
 CFLAGS=-O3 -fno-exceptions $(CFLAGS_EXTRA)
-CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fno-optimize-sibling-calls -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
+CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
 CPPFLAGS=
 DEFS=-DPROFILER_VERSION=\"$(PROFILER_VERSION)\"
 INCLUDES=-I$(JAVA_HOME)/include -Isrc/helper
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ endif
 CFLAGS_EXTRA ?=
 CXXFLAGS_EXTRA ?=
 CFLAGS=-O3 -fno-exceptions $(CFLAGS_EXTRA)
-CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
+CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fno-optimize-sibling-calls -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
 CPPFLAGS=
 DEFS=-DPROFILER_VERSION=\"$(PROFILER_VERSION)\"
 INCLUDES=-I$(JAVA_HOME)/include -Isrc/helper
--- a/src/mallocTracer.cpp
+++ b/src/mallocTracer.cpp
@@ -62,6 +62,7 @@ extern "C" void* calloc_hook(size_t num, size_t size) {
     return ret;
 }
 
+// Make sure this is not optimized away (function-scoped -fno-optimize-sibling-calls)
 extern "C" __attribute__((optimize("O1")))
 void* calloc_hook_dummy(size_t num, size_t size) {
     return _orig_calloc(num, size);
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ endif
 CFLAGS_EXTRA ?=
 CXXFLAGS_EXTRA ?=
 CFLAGS=-O3 -fno-exceptions $(CFLAGS_EXTRA)
-CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
+CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fno-optimize-sibling-calls -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
```
fandreuz, line 32: This is needed to preserve `calloc_hook_dummy` from being optimized away in `-O2+`

src/symbols_linux.cpp
```diff
@@ -804,14 +804,27 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
-                    dlclose(handle);
+            if (handle == NULL) {
+                char *dlerror_output = dlerror();
+                // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror()
+                // returns NULL for them.
+                if (dlerror_output != NULL) {
+                    char buffer[1024];
+                    // https://git.musl-libc.org/cgit/musl/tree/ldso/dynlink.c?h=v1.2.5#n2167
+                    sprintf(buffer, "Library %s is not already loaded", lib.file);
+                    if (strcmp(dlerror_output, buffer) != 0) {
```
fandreuz, line 816: I would very much prefer to check an integer error code here, but I couldn't find anything like that

Makefile
apangin: You essentially disable a useful optimization for the whole profiler on all platforms just to satisfy a couple of tests on musl. Doesn't look right to me. Can you fine-tune optimizations for the particular function instead? (using compiler-specific attributes)

src/symbols_linux.cpp
apangin: This looks a bit ugly. For which libraries is this error returned?
Can we just ignore any dlopen errors on musl, since musl does not support library unloading?

fandreuz: This is for the main executable, `dlerror()` returns `Library ... is not already loaded` instead of `NULL`.
> Can we just ignore any dlopen errors on musl, since musl does not support library unloading?
Sure

Makefile
fandreuz: 2abe0fdb19e92248fa39061e4936c4a6f09a18c8

src/symbols_linux.cpp
fandreuz: 383c29dd221f37c08bcb2258aada68fea5a25108

fandreuz:
@apangin ready for II review round

src/symbols_linux.cpp
```diff
@@ -804,14 +804,22 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
-                ElfParser::parseProgramHeaders(cc, lib.image_base, lib.map_end, OS::isMusl());
-                if (handle != NULL) {
-                    dlclose(handle);
+            if (handle == NULL && !OS::isMusl()) {
+                char *dlerror_output = dlerror();
+                // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror()
+                // returns NULL for them. musl does not support library unloading, thus we can
+                // safely ignore dlerror().
+                if (dlerror_output != NULL) {
+                    Log::info("dlerror '%s': '%s'", lib.file, dlerror_output);
```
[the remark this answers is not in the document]
fandreuz: My reasoning here is that it took me some time to debug this issue because the error is quietly ignored if non-null. Reverted in 5b26ed38eb4e9a5e5bce33ed91e2e036afbd8bbe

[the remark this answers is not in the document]
fandreuz: But looking at the new implementation, is it ok to ignore any error on musl? @apangin

src/mallocTracer.cpp
```diff
@@ -62,6 +62,11 @@ extern "C" void* calloc_hook(size_t num, size_t size) {
     return ret;
 }
 
+extern "C" __attribute__((optimize("O1")))
```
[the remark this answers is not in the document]
fandreuz: 63e51a9e0adca5fada3593509af763e177b8c0ce

Issue #1230, opened 2025-04-15
Save information about truncated stacktraces


jbachorik:
### Describe the feature

Let's save the information about incompletely unwound stacktraces where we had to bail out before reaching the root.

### Use Case

The 'truncated' flag is already present in the `Stacktrace` datatype but we are not using it. 

However, having the information that a particular stacktrace is not completely unwound can be useful to have - to let the user know that the information is not complete or to help efficiently store such truncated stacktraces in a long-term aggregated format.

### Proposed Solution

Extend the internal call-trace storage with the 'truncated' flag.
Embellish the stackwalking code to set the 'truncated' flag where we know we are aborting the stackwalk.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1237, opened 2025-04-16
Add support for truncated stacktraces

jbachorik:
### Description
This change adds the ability to set the 'truncated' flag of a stacktrace when we bail out of the stackwalking before reaching the root.

### Related issues
Save information about truncated stacktraces #1230 

### Motivation and context
The 'truncated' flag is already present in the `Stacktrace` datatype but we are not using it. 
However, having the information that a particular stacktrace is not completely unwound can be useful to have - to let the user know that the information is not complete or to help efficiently store such truncated stacktraces in a long-term aggregated format.

The 'traces' output has been extended to show `[truncated]` as the root frame if the trace was truncated. For JFR files we are using the already existing flag. Other outputs have not been modified yet - mostly because the 'traces' was necessary for running tests and the rest is more or less optional. I can add the modifications, if it is desired, before finishing this PR.

### How has this been tested?
Added tests to validate the truncation on various modes of stackwalking.

The existing tests are passing fine. 

---

### DIFF
--- a/src/callTraceStorage.cpp
+++ b/src/callTraceStorage.cpp
@@ -79,7 +79,7 @@ class LongHashTable {
     }
 };
 
-CallTrace CallTraceStorage::_overflow_trace = {1, {BCI_ERROR, LP64_ONLY(0 COMMA) (jmethodID)"storage_overflow"}};
+CallTrace CallTraceStorage::_overflow_trace = {false, 1, {BCI_ERROR, LP64_ONLY(0 COMMA) (jmethodID)"storage_overflow"}};
 
 CallTraceStorage::CallTraceStorage() : _allocator(CALL_TRACE_CHUNK) {
     _current_table = LongHashTable::allocate(NULL, INITIAL_CAPACITY);
@@ -167,12 +167,12 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
 }
 
 // Adaptation of MurmurHash64A by Austin Appleby
-u64 CallTraceStorage::calcHash(int num_frames, ASGCT_CallFrame* frames) {
+u64 CallTraceStorage::calcHash(int num_frames, ASGCT_CallFrame* frames, bool truncated) {
     const u64 M = 0xc6a4a7935bd1e995ULL;
     const int R = 47;
 
     int len = num_frames * sizeof(ASGCT_CallFrame);
-    u64 h = len * M;
+    u64 h = len * M * (truncated ? 1 : 2);
 
     const u64* data = (const u64*)frames;
     const u64* end = data + len / 8;
@@ -198,7 +198,7 @@ u64 CallTraceStorage::calcHash(int num_frames, ASGCT_CallFrame* frames) {
     return h;
 }
 
-CallTrace* CallTraceStorage::storeCallTrace(int num_frames, ASGCT_CallFrame* frames) {
+CallTrace* CallTraceStorage::storeCallTrace(int num_frames, ASGCT_CallFrame* frames, bool truncated) {
     const size_t header_size = sizeof(CallTrace) - sizeof(ASGCT_CallFrame);
     CallTrace* buf = (CallTrace*)_allocator.alloc(header_size + num_frames * sizeof(ASGCT_CallFrame));
     if (buf != NULL) {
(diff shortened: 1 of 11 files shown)


src/callTraceStorage.cpp
```diff
@@ -167,12 +167,12 @@ void CallTraceStorage::collectSamples(std::map<u64, CallTraceSample>& map) {
...
 
 // Adaptation of MurmurHash64A by Austin Appleby
-u64 CallTraceStorage::calcHash(int num_frames, ASGCT_CallFrame* frames) {
+u64 CallTraceStorage::calcHash(int num_frames, ASGCT_CallFrame* frames, bool truncated) {
     const u64 M = 0xc6a4a7935bd1e995ULL;
     const int R = 47;
 
     int len = num_frames * sizeof(ASGCT_CallFrame);
-    u64 h = len * M;
+    u64 h = len * M * (truncated ? 1 : 2);
```
apangin, line 175: I think `+` makes more sense. Otherwise, a truncated stacktrace will have the same initial value as a full stack trace two times shorter. As an edge case, it will not be possible to distinguish truncated and non-truncated zero-length stack traces.

src/perfEvents_linux.cpp
```diff
@@ -937,9 +938,9 @@ int PerfEvents::walk(int tid, void* ucontext, const void** callchain, int max_de
     event->unlock();
 
     if (_cstack == CSTACK_FP) {
-        depth += StackWalker::walkFP(ucontext, callchain + depth, max_depth - depth, java_ctx);
+        depth += StackWalker::walkFP(ucontext, callchain + depth, max_depth - depth, java_ctx, truncated);
     } else if (_cstack == CSTACK_DWARF) {
-        depth += StackWalker::walkDwarf(ucontext, callchain + depth, max_depth - depth, java_ctx);
+        depth += StackWalker::walkDwarf(ucontext, callchain + depth, max_depth - depth, java_ctx, truncated);
```
apangin, line 943: `truncated` flag is not really meaningful with `fp` and `dwarf` stack walking. It can be even misleading, since it refers to the native part of a stack only. `AsyncGetCallTrace` may return complete or truncated stacktrace, and profiler has no means to find this out.

Whether a stacktrace is truncated or not, can be only determined with VMStructs-based stack walking.

src/profiler.cpp
```diff
@@ -1604,11 +1620,16 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
             CallTrace* trace = it->trace;
-            for (int j = 0; j < trace->num_frames; j++) {
+            int j = 0;
+            for (; j < trace->num_frames; j++) {
                 const char* frame_name = fn.name(trace->frames[j]);
                 snprintf(buf, sizeof(buf) - 1, "  [%2d] %s\n", j, frame_name);
                 out << buf;
             }
+            if (trace->truncated) {
+                snprintf(buf, sizeof(buf) - 1, "  [%2d] [truncated]\n", j);
```
apangin, line 1630: All truncated stack traces obtained with cstack=vm/vmx should ideally end with an error frame.
These are `[break_compiled]`, `[break_interpreted]` etc. in the current implementation, but we can change names if needed. If an error frame is missed in some cases, let's fix it.

src/callTraceStorage.cpp
jbachorik: Good point.
Would it then make sense to do something like `u64 h = 2 * len * M + (truncated ? 0 : 1)` so we have clear disambiguation between truncated and non-truncated?

src/perfEvents_linux.cpp
jbachorik: Ok, the idea is to have an indication when we didn't walk up to root. Either because some error or because we reached the maximum number of frames we are allowed to walk (eg. because of the stackdepth argument).
Does it make sense?

src/profiler.cpp
jbachorik: This may not be the case if you set the maximum stackdepth - as I do in the tests. Without outputting the synthetic root frame you have no idea that a stacktrace is not complete.

src/perfEvents_linux.cpp
apangin: But that's not what the current implementation does. In this PR, `truncated` flag can be set for stacks fully unwound to the bottommost `Thread.run` or vice versa: a partial stack can be returned with `truncated = false`, because all stack unwinding modes except `vm` rely on `AsyncGetCallTrace`. I'd suggest not to confuse people and to leave `fp` and `dwarf` modes intact.

jbachorik:
Closing. Having just a partial information about when we didn't walk the full stack depth is not very useful.
Also, there is another PR in progress which is adding a synthetic frame instead of having a flag.

Issue #1233, opened 2025-04-15
`NativememTests.malloc_plt_dyn` failing "sometimes"


krk:
### Describe the bug

`malloc_plt_dyn` test started to fail:

```
FAIL [15/67] NativememTests.malloc_plt_dyn took 0.013 s
java.lang.AssertionError: Expected 0.0 > 0.0
        >  test/test/nativemem/NativememTests.java:133
        >  Assert.isGreater(totalAllocated, 0);
        at one.profiler.test.Assert.assertComparison(Assert.java:39)
        at one.profiler.test.Assert.isGreater(Assert.java:60)
        at test.nativemem.NativememTests.assertNoLeaks(NativememTests.java:133)
        at test.nativemem.NativememTests.malloc_plt_dyn(NativememTests.java:176)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at java.base/java.lang.reflect.Method.invoke(Method.java:565)
        at one.profiler.test.Runner.run(Runner.java:133)
        at one.profiler.test.Runner.main(Runner.java:230)
```

### Expected vs. actual behavior

Expected test to pass.

### Reproduction Steps

`make test-java TESTS=nativemem`

### Async-profiler version

master

### Environment details

tested on AL2 x86_64 `OpenJDK 64-Bit Server VM Corretto-23.0.2.7.1 (build 23.0.2+7-FR, mixed mode, sharing)`.

krk:
Bisected to https://github.com/async-profiler/async-profiler/pull/1220

fandreuz:
Spotted that too in some of my GHA runs: https://github.com/fandreuz/async-profiler/actions/runs/14477851268/job/40608527415

Issue #1241, opened 2025-04-17
Symbols of main executable are not parsed


Baraa-Hasheesh:
### Describe the bug

symbol patching isn't working on libraries that are dlopen-ed on the fly in a non java process

After investigation the problem it was found the dlopen hook in the async-profiler isn't functioning as expected (Not being called) 

I.E. if you dlopen a custom library & call a function from that library such as a `malloc` call then this won't be logged correctly by the async profiler when doing native memory profiling

### Expected vs. actual behavior

dlopen hook should be called normally to batch libraries that are loaded on the fly

### Reproduction Steps

Create the following `H` file (`lib.h`)

```
#ifndef MYLIB_H
#define MYLIB_H

#define EXPORT __attribute__((visibility("default")))

#include "stdlib.h"


EXPORT void* my_malloc(size_t size);

EXPORT void my_free(void* ptr);

#endif
```

Create the following `C` file (`lib.c`)

```
#include "lib.h"

EXPORT void* my_malloc(size_t size) {
    return malloc(size);
}

EXPORT void my_free(void* ptr) {
    free(ptr);
}
```

You will need to create a shared lib => **`gcc -fPIC  lib.c  -shared -o my_lib.so`**

Create the following small `C` application (`main.cpp`)

```
#include <iostream>
#include <dlfcn.h>
#include <time.h>
#include <stdlib.h>
#include "asprof.h"


typedef void* (*my_malloc_t)(size_t);
typedef void (*my_free_t)(void*);

asprof_init_t _asprof_init;
asprof_execute_t _asprof_execute;
asprof_error_str_t _asprof_error_str;

void outputCallback(const char* buffer, size_t size) {
    fwrite(buffer, sizeof(char), size, stderr);
}

void loadProfiler() {
    void* lib = dlopen("./libasyncProfiler.so", RTLD_NOW | RTLD_GLOBAL);
    if (lib == NULL) {
        std::cout << "Async profiler lib not found: " << dlerror() << std::endl;
        exit(1);
    }

    _asprof_init = (asprof_init_t)dlsym(lib, "asprof_init");
    _asprof_execute = (asprof_execute_t)dlsym(lib, "asprof_execute");
    _asprof_error_str = (asprof_error_str_t)dlsym(lib, "asprof_error_str");

    _asprof_init();
}

void startProfiler() {
    asprof_error_t err = _asprof_execute("start,nativemem=1,total,cstack=dwarf", outputCallback);
    if (err != NULL) {
[...]

PR #1243, opened 2025-04-17
Add nativemem test with dlopen after profiler start.

krk:
### Description
Parse symbols and hook functions of the main executable, when profiler is *not* loaded via `LD_PRELOAD`.

Without this fix, profiler is unable to hook non-java executables if `dlopen` is called after starting the profiler. 

### Related issues
Issue was introduced in https://github.com/async-profiler/async-profiler/pull/1220 as `dlerror` may *not* return `NULL` for the main executable.

Thanks for the repro in https://github.com/async-profiler/async-profiler/issues/1241 @Baraa-Hasheesh  and for discussions @fandreuz and @Baraa-Hasheesh.

Fixes https://github.com/async-profiler/async-profiler/issues/1233 and https://github.com/async-profiler/async-profiler/issues/1241.

### How has this been tested?
`make test-java TESTS=nativemem`

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -810,11 +810,10 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
 
             // Protect library from unloading while parsing in-memory ELF program headers.
             // Also, dlopen() ensures the library is fully loaded.
-            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() "in some cases" returns NULL for them.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them on some systems.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
 
-            // Main executable will return NULL from handle, we need to parse it anyway.
-            // dlopen is not required in this case as the main exe cannot be unloaded.
+            // Parse main executable regardless of dlopen result, since it cannot be unloaded.
             bool is_main_exe = main_phdr >= lib.image_base && main_phdr < lib.map_end;
             if (handle != NULL || dlerror() == NULL || is_main_exe) {
                 ElfParser::parseProgramHeaders(cc, lib.image_base, lib.map_end, OS::isMusl());
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -812,13 +812,11 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
             // Also, dlopen() ensures the library is fully loaded.
             // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() "in some cases" returns NULL for them.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
-            const char* dlerror_str = dlerror();
 
             // Main executable will return NULL from handle, we need to parse it anyway.
             // dlopen is not required in this case as the main exe cannot be unloaded.
-            bool isMainExe = handle == NULL && main_phdr >= lib.image_base && main_phdr < lib.map_end;
-            bool dlopenSuccess = handle != NULL || dlerror_str == NULL;
-            if (isMainExe || dlopenSuccess) {
+            bool is_main_exe = main_phdr >= lib.image_base && main_phdr < lib.map_end;
+            if (handle != NULL || dlerror() == NULL || is_main_exe) {
                 ElfParser::parseProgramHeaders(cc, lib.image_base, lib.map_end, OS::isMusl());
             }
 
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -775,6 +775,19 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     std::unordered_map<u64, SharedLibrary> libs;
     collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count());
 
+    char* exePath = realpath("/proc/self/exe", NULL);
+    if (exePath == NULL) {
+        char buf[PATH_MAX];
+
+        // realpath() may fail for a path like /proc/[pid]/root/bin/asprof
+        // In this case, resolve the link as is.
+        ssize_t size = readlink("/proc/self/exe", buf, sizeof(buf) - 1);
+        if (size >= 0) {
+            buf[size] = 0;
+            exePath = buf;
+        }
+    }
+
     for (auto& it : libs) {
         u64 inode = it.first;
         _parsed_inodes.insert(inode);
@@ -804,13 +817,20 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
 
             // Protect library from unloading while parsing in-memory ELF program headers.
             // Also, dlopen() ensures the library is fully loaded.
-            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() "in some cases" returns NULL for them.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
-            if (handle != NULL || dlerror() == NULL || OS::isMusl()) {
+            const char* dlerror_str = dlerror();
+
+            // Main executable will return NULL from handle, we need to parse it anyway.
+            // dlopen is not required in this case as the main exe cannot be unloaded.
+            bool isMainExe = handle == NULL && exePath != NULL && strcmp(lib.file, exePath) == 0;
+            bool dlopenSuccess = handle != NULL || dlerror_str == NULL || OS::isMusl();
+            if (isMainExe || dlopenSuccess) {
                 ElfParser::parseProgramHeaders(cc, lib.image_base, lib.map_end, OS::isMusl());
-                if (handle != NULL) {
-                    dlclose(handle);
-                }
+            }
+
+            if (handle != NULL) {
+                dlclose(handle);
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/symbols_linux.cpp
```diff
@@ -804,13 +817,20 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
-            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() "in some cases" returns NULL for them.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
-            if (handle != NULL || dlerror() == NULL || OS::isMusl()) {
+            const char* dlerror_str = dlerror();
+
+            // Main executable will return NULL from handle, we need to parse it anyway.
+            // dlopen is not required in this case as the main exe cannot be unloaded.
+            bool isMainExe = handle == NULL && exePath != NULL && strcmp(lib.file, exePath) == 0;
+            bool dlopenSuccess = handle != NULL || dlerror_str == NULL || OS::isMusl();
```
fandreuz, line 827: Maybe we don't need `OS::isMusl()` anymore? The only situation when this was appropriate is for the main executable, in which case `isMainExe` is `true` anyways

test/native/libs/callsmalloc.c
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdlib.h>
+
+__attribute__((visibility("default"))) void* call_malloc(size_t size) {
```
fandreuz, line 8: Maybe you could reuse `Java_test_nativemem_Native_malloc` from `jnimalloc.c` instead of creating a new file?

test/one/profiler/test/RunnableTest.java
```diff
@@ -25,7 +25,9 @@ public Test test() {
     }
 
     public String testName() {
-        return className() + '.' + m.getName();
+        return test.nameSuffix().isEmpty()
+            ? className() + '.' + m.getName()
+            : className() + '.' + m.getName() + ' ' + test.nameSuffix();
```
fandreuz, line 30: return String.format("%s.%s %s", className(), m.getName(), test.nameSuffix()).trim();

test/test/nativemem/profile_with_dlopen.c
```diff
@@ -0,0 +1,83 @@
...
+#include "asprof.h"
+#include <dlfcn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define ASSERT_NO_DLERROR()  \
+    err = dlerror();         \
+    if (err != NULL) {       \
+        printf("%s\n", err); \
```
fandreuz, line 15: `stderr`? Same for L21

test/test/nativemem/profile_with_dlopen.c
```diff
@@ -0,0 +1,83 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "asprof.h"
+#include <dlfcn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define ASSERT_NO_DLERROR()  \
```
fandreuz, line 12: What's the advantage of `#define` over a function?

test/one/profiler/test/RunnableTest.java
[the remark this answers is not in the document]
apangin: Kerem's version is better: it's more obvious and far more efficient (although the latter is not important here, let's not support bad practice).

src/symbols_linux.cpp
```diff
@@ -775,6 +775,19 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+        char buf[PATH_MAX];
+
+        // realpath() may fail for a path like /proc/[pid]/root/bin/asprof
+        // In this case, resolve the link as is.
+        ssize_t size = readlink("/proc/self/exe", buf, sizeof(buf) - 1);
+        if (size >= 0) {
+            buf[size] = 0;
+            exePath = buf;
+        }
+    }
```
apangin, line 789: There is a simpler and more efficient way to find main executable:
```
    const void* main_phdr = NULL;
    dl_iterate_phdr([](struct dl_phdr_info* info, size_t size, void* data) {
        *(const void**)data = info->dlpi_phdr;
        return 1;
    }, &main_phdr);
```
Then check if `main_phdr >= lib.image_base && main_phdr < lib.map_end`

Calling `dl_iterate_phdr` should be fine as long as there are no allocations inside.

src/symbols_linux.cpp
```diff
@@ -804,13 +817,20 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
             // Also, dlopen() ensures the library is fully loaded.
-            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them.
+            // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() "in some cases" returns NULL for them.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
-            if (handle != NULL || dlerror() == NULL || OS::isMusl()) {
+            const char* dlerror_str = dlerror();
+
+            // Main executable will return NULL from handle, we need to parse it anyway.
+            // dlopen is not required in this case as the main exe cannot be unloaded.
+            bool isMainExe = handle == NULL && exePath != NULL && strcmp(lib.file, exePath) == 0;
```
apangin, line 826: `snake_case` for variable names.

apangin (commented):
Thank you for root causing the issue.

test/test/nativemem/NativememTests.java
```diff
@@ -178,4 +178,14 @@ public void malloc_plt_dyn(TestProcess p) throws Exception {
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_SIZE, 0L), 1);
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_DYN_SIZE, 0L), 1);
     }
+
+    @Test(os = Os.LINUX, sh = "%testbin/profile_with_dlopen dlopen_first %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "dlopen_first")
+    @Test(os = Os.LINUX, sh = "%testbin/profile_with_dlopen profile_first %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "profile_first")
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=%lib %testbin/profile_with_dlopen dlopen_first %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "dlopen_first with LD_PRELOAD")
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=%lib %testbin/profile_with_dlopen profile_first %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "profile_first with LD_PRELOAD")
```
Baraa-Hasheesh, line 185: We should add a new test that checks the dlopen patching on both MacOs & Linux?
As this issue of not patching the main executable is also relevant to MacOs

src/symbols_linux.cpp
```diff
@@ -821,6 +841,8 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
         array->add(cc);
     }
 
+    free(exePath);
```
Baraa-Hasheesh, line 844: is this safe? as in some cases it points toward the `buf[PATH_MAX]` which is defined in the stack not the heap

test/test/nativemem/profile_with_dlopen.c
```diff
@@ -0,0 +1,83 @@
...
+        exit(1);             \
+    }
+
+#define ASSERT_NO_ASPROF_ERR(err)               \
+    if (err != NULL) {                          \
+        printf("%s\n", _asprof_error_str(err)); \
+        exit(1);                                \
+    }
+
+asprof_error_str_t _asprof_error_str;
```
Baraa-Hasheesh, line 25: no need for this global variable, you can define it in the main function

Baraa-Hasheesh:
@krk @apangin @fandreuz 

The above fix doesn't fully fix the issue of patching libs,
Due to how the dynamic linker work & common optimization it does -> if you do the following steps

```
void* lib1 = dlopen("libcallsmalloc.so", RTLD_NOW);
dlclose(lib1);
void2 lib2 = dlopen("libcallsmalloc.so", RTLD_NOW);
```

You will actually find that the `lib1` & `lib2` point to the same address even though the shared objects were deleted & recreated

=> you can try this in the new test by doing the following
image

This means due to the Async-profiler built in optimization of skipping already patched libs the second dlopen will never be patched as it will be seen as already patched from the first dlopen 

This is quite an old issue (Even the old logic before #1220 has the issue) that could require significant changes to address
* Hooking the dlclose calls & actually removing the patches from the async profiler cache if reference counts reaches 0 for a lib.
* We could need locking logic as it's possible race conditions may happen when dlclose-ing & dlopen-ing from multiple threads.

Interesting edge cases will be handling the `RTLD_NODELETE` flag as even if the reference counter reaches 0, there is no need to cleanup the async-profiler local cache as the shared objects are never unloaded.

test/test/nativemem/profile_with_dlopen.c
krk: We can exit directly, without checking the return value of a new function. I used macros here to reduce repetition.

test/test/nativemem/NativememTests.java
krk: Macos support is "best effort", will leave it out from the current PR.

src/symbols_linux.cpp
krk: Good catch! This will be deleted in the new version.

PR #1246, opened 2025-04-18
Build and run tests on `amazonlinux:2023` and `amazonlinux:2`

fandreuz:
### Description
In this PR I add `amazonlinux:2023` and `amazonlinux:2` to the GHA build/test matrix. For the latter, we need some hacks to deal with the outdated glibc, which doesn't work well with node20:
- https://repost.aws/questions/QUrXOioL46RcCnFGyELJWKLw/glibc-2-27-on-amazon-linux-2
- https://github.com/dawidd6/action-download-artifact/issues/262
- https://github.com/aws-actions/configure-aws-credentials/issues/1106

Basically any action does not work on `amazonlinux:2`, thus I got the following solutions:
- `Setup Java`: replaced with plain old `yum install ...`
- `Checkout sources`: replaced with `git clone ...`
- `Coverage`, `Upload test logs`, `Upload coverage report`: they all rely on `actions/upload-artifact`, which we cannot use. Thus, these are disabled for `amazonlinux:2`. I believe we can live with it. If that's not the case, we _could_ do some manual HTTP calls to upload the relevant artifacts.

### Related issues
#1241, #1243

### How has this been tested?
GHA

---


fandreuz:
@krk `NativememTests.malloc_plt_dyn` from `master` consistently fails on `amazonlinux:2`, and is consistently green after #1243.

`docker run -it --privileged --rm -v $(pwd):/async-profiler amazonlinux:2023 bash -c "yum update -y && yum install -y make gcc-c++ libstdc++-devel libstdc++-static util-linux tar patchelf git java-11-amazon-corretto && cd async-profiler && rm -rf build && make -j test-java TESTS=nativemem"`

krk:
> NativememTests.malloc_plt_dyn from master consistently fails on amazonlinux:2, and is consistently green after https://github.com/async-profiler/async-profiler/pull/1243.

Thanks, this is what I saw on the AL2 test machine locally.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,64 +51,100 @@ jobs:
...
+      - name: Setup Java on amazonlinux-2
+        if: matrix.runson.display == 'amazonlinux-2'
+        run: yum install -y ${{ matrix.java-package }}
       - name: Checkout sources
+        if: matrix.runson.display != 'amazonlinux-2'
         uses: actions/checkout@v4
+      - name: Checkout sources on amazonlinux-2
+        if: matrix.runson.display == 'amazonlinux-2'
+        run: |
+          git clone https://${{ github.token }}@github.com/${{ github.repository }}.git .
```
krk, line 96: We can use `--depth=1` for faster cloning, as we don't need the repo history.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,64 +51,100 @@ jobs:
             java-version: 11
             java-distribution: corretto
             container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
+          - runson:
+              display: amazonlinux-2023 # Use dash instead of colon, because GHA artifacts cannot contain colons
+              name: ubuntu-latest
```
krk, line 56: Do we need `ubuntu` in the name, when `container` is set to amazonlinux?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,64 +51,100 @@ jobs:
...
+      - name: Checkout sources on amazonlinux-2
+        if: matrix.runson.display == 'amazonlinux-2'
+        run: |
+          git clone https://${{ github.token }}@github.com/${{ github.repository }}.git .
       - name: Build and test
         id: build
         run: |
           HASH=$(echo ${{ github.sha }} | cut -c-7)
-          case "${{ matrix.runson.name }}" in
+          case "${{ matrix.runson.display }}" in
```
krk, line 101: Why do we need to switch to `display` from `name` in all these places?

.github/workflows/test-and-publish-nightly.yml
fandreuz: Actually the reason why I've done it this way is that I foreseen a subsequent `git checkout ${{ github.sha }}` which I forgot to put in the diff. We cannot use `--branch` because that could result in some nasty race conditions

.github/workflows/test-and-publish-nightly.yml
fandreuz: `name` is the name of the runner. For e.g. `alpine`, `amazonlinux`, and other entries relying on Docker containers to run on a specific OS, we're gonna have something like:
```
           -runson:
              display: amazonlinux-2
              name: ubuntu-latest
            ...
            container: amazonlinux:2
```

.github/workflows/test-and-publish-nightly.yml
fandreuz: Yes, this is the name of the runner, which then runs a Docker container from an `amazonlinux` image

.github/workflows/test-and-publish-nightly.yml
krk: Let's use this to avoid checking out the entire history:

```
git init
git remote add origin https://github.com/async-profiler/async-profiler
git fetch --depth 1 origin b9f3456f892088e9aac8cadbb3a9f4a86a7f3654
git checkout b9f3456f892088e9aac8cadbb3a9f4a86a7f3654
```

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,64 +51,101 @@ jobs:
...
     container:
       image: ${{ matrix.container }}
       options: --privileged
     name: "test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }})"
     steps:
+      - name: Setup
+        run: |
+          if [[ "${{ matrix.container }}" == "amazonlinux"* ]]; then
+              yum update -y
+              yum install -y make gcc-c++ libstdc++-devel libstdc++-static util-linux tar git patchelf
```
apangin, line 78: Does it make sense to prepare a container similar to `asprof-builder` so that you don't need to install all this stuff every time? In theory, you could also try to build node.js in this container that will be able to run standard github actions, but this is optional - I'm not sure how easy it will be,

krk: Node.js build is not too hard but will take a lot of time, having separate builder images sound better.

fandreuz: Do we really want to use an outdated Node version? What if some any of the Actions we use (setup-java, checkout, download-artifacts, ...) breaks on older Node.js versions?

apangin: I meant including new manually built version of Node.js in a container image

fandreuz: If we want to have a new Docker image based on `amazonlinux:2` running a new Node.js version (probably `>20`) we'll need to change the environment a bit. The glibc version which comes out of the box with `amazonlinux:2` is not compatible for sure, thus we must change that at least.

Tweaking the environment this way means that we are also not going to test Async-Profiler on the same environment which is being used by `amazonlinux:2` users, which breaks a bit the motivation behind this PR. Am I missing something here @krk @apangin ?

.github/workflows/test-and-publish-nightly.yml
fandreuz: Thanks!

.github/workflows/test-and-publish-nightly.yml
krk: Node 20 (probably 22 as well) can be built with the libc in AL2, no update to the glibc version is necessary.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -51,64 +51,104 @@ jobs:
...
+        if: matrix.runson.display != 'amazonlinux-2'
         uses: actions/checkout@v4
+      - name: Checkout sources on amazonlinux-2
+        if: matrix.runson.display == 'amazonlinux-2'
+        run: |
+          git config --global --add safe.directory '*'
+          git init
+          git remote add origin https://${{ github.token }}@github.com/${{ github.repository }}.git
+          git fetch --depth 1 origin ${{ github.sha }}
+          git checkout ${{ github.sha }}
```
apangin, line 100: LLM suggests me a shorter alternative:
```
git clone --no-checkout --depth 1 <repo-url>
git fetch --depth=1 origin <commit-hash>
git checkout <commit-hash>
```

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -54,11 +61,16 @@ jobs:
...
           set -x
-          case "${{ matrix.runson.name }}" in
+          case "${{ matrix.runson.display }}" in
             macos*)
               brew install gcovr
               make COMMIT_TAG=${GITHUB_SHA:0:7} FAT_BINARY=true release coverage -j
             ;;
+            amazonlinux*)
+              python3 -m ensurepip
+              python3 -m pip install gcovr
```
apangin, line 71: Can these be included in the image?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -43,6 +47,9 @@ jobs:
       image: ${{ matrix.image }}
     name: "Build and unit test (${{ matrix.runson.display }})"
     steps:
+      - name: Update PATH
+        if: matrix.runson.display == 'amazonlinux2'
+        run: echo "/usr/local/bin" >> $GITHUB_PATH
```
apangin, line 52: Can paths be configured in the image? (or tools be installed on a path accessible by default?)
The idea is to minimize image-specific configuration as much as possible.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -35,6 +35,10 @@ jobs:
               display: linux-x64
               name: ubuntu-latest
             image: public.ecr.aws/async-profiler/asprof-builder-x86:latest
+          - runson:
+              display: amazonlinux2
+              name: ubuntu-latest
+            image: public.ecr.aws/async-profiler/asprof-builder-amazonlinux:al2
```
apangin, line 41: "al" both in the name and in the version looks a bit clumsy.
What about `asprof-builder-al2:latest` or `asprof-builder-amazonlinux:2` ?

krk: `asprof-builder-amazonlinux:2` :+1:

krk (approved):

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -38,11 +38,21 @@ jobs:
...
+            # '/__e/node20', and it's not writable unless we override it via 'container.volumes'.
+            volumes: '["/tmp/node20:/__e/node20"]'
     runs-on: ${{ matrix.runson.name }}
     container:
       image: ${{ matrix.image }}
+      volumes: ${{ fromJSON(matrix.volumes || '[]') }}
     name: "Build and unit test (${{ matrix.runson.display }})"
     steps:
+      - name: Run container setup
+        run: "[ -f /root/setup.sh ] && { /root/setup.sh; exit $?; }; exit 0"
```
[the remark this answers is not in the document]
apangin: Alternatively, can we just call `ln` here ignoring any error?

docker/amazonlinux2.Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN amazon-linux-extras enable python3.8
 
 RUN yum update -y && yum install -y git make python38 gcc10 gcc10-c++ binutils tar
 
-ARG node_version=22.15.0
-ARG node_sha256=4f2515e143ffd73f069916ecc5daf503e7a05166c0ae4f1c1f8afdc8ab2f8a82
+ARG node_version=20.19.1
```
apangin, line 7: Didn't version 22 work?

fandreuz: GitHub Actions runners start the Docker container mounting Node.js to `/__e/node20`. This directory is something we cannot change, thus by having Node.js 22 in our Docker container we would have to symlink `/__e/node20` to a directory containing a Node.js build. It just feels cleaner to me to use Node.js 20, especially since we don't have any special requirement, and GHA uses Node.js 20 anyway

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
fandreuz: I like having an optional `/root/setup.sh` script because we can execute any non-standard setup here, specialized for any image we may want to add in the future.

Also, the reason why it's this intricate is that I want to propagate any error if the file exists (we don't want to proceed if the setup did not complete), and exit cleanly otherwise

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -38,11 +38,21 @@ jobs:
           - runson:
               display: macos
               name: macos-15
+          - runson:
+              display: amazonlinux2
```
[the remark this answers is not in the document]
fandreuz: I'd expect al2 to complain about the binaries having been compiled with a newer glibc, I'll check this tomorrow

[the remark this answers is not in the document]
apangin: That would be a fault. Published binaries must work on all reasonably old Linux distributions such as CentOS 7.

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
apangin: > I want to propagate any error if the file exists

So does my version, doesn't it?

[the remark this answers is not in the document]
fandreuz: Right, I misread it, thanks! 33bc4d41d28d110f103cbb16d56a562f3d34820c

Issue #1213, opened 2025-04-04
JDK8 isn't being covered by GHA


Baraa-Hasheesh:
As title mentions, GHA aren't covering JDK8.

apangin:
Another task related to CI #1136

PR #1238, opened 2025-04-16
Enable running integration tests against a pre-built artifact

fandreuz:
### Description
In this PR I introduce an optional parameter `BINARIES_DIRECTORY` to enable running integration tests against a pre-built AP artifact.

### Related issues
Changes in GHA to simplify build/testing.

### Motivation and context
This is the first step towards testing AP on the binaries released via GHA.

### How has this been tested?
`make release -j && rm -rf build && tar xvf async-profiler-4.0-linux-x64.tar.gz && make test-java BINARIES_DIRECTORY=async-profiler-4.0-linux-x64`

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #1253, opened 2025-04-22
Separate jobs for build/unit tests and integration tests

fandreuz:
### Description
In this PR I split the GHA job `build-binaries-and-test` in two jobs:
- `build-and-upload-binaries`: build release, run unit tests, upload artifacts
- `integration-tests`: run integration tests

<details>

### Old
image

### New
image
</details>

## Artifacts

I changed a bit how we publish publication:
- Removed the confusing `.tar.gz` and `.zip` in the *artifact name* (not *artifact path*, which stays the same), since anyway GHA would create an archive named `{artifact_name}.zip`, thus resulting in a file named `{artifact_name}.zip.zip` or `{artifact_name}.tar.gz.zip`
   - The tree stays the same though
   - *Question*: do we really need to publish a `.tar.gz`/`.zip` here? We could simply upload the artifact as a directory, and GHA will take care of zipping it, this would also simplify the workflow in this PR 
- JAR artifacts are now published as a single `async-profiler-jars` artifact. We can discuss this, it felt more symmetric to me

<details>

### Old
image

### New
image
</details>

### Related issues
- #1213 will benefit from this
- #1226 and #1246 will need to change a bit
- we may need to adapt the publication workflow, let's see how these changes are received first

### Motivation and context
- We now have two different matrixes for build/upload artifact and integration tests, which means we can remove the hardcoded condition for artifact upload
- We don't need to setup Java two times when eventually we add JDK 8 to the integration test matrix
- Binaries and JARs which are published as artifacts are the same which are used for integration tests

### How has this been tested?
GHA

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -94,14 +94,11 @@ jobs:
             # Using "latest" here as the build and test will any ways run inside a container which we control
             name: ubuntu-latest
         java-version: [11, 17, 21, 24]
-        java-distribution: [corretto]
         image: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
         include:
           - runson:
               display: macos-arm64
               name: macos-15
-            java-version: 11
-            java-distribution: corretto
             # Not using container for mac-os as we have images only for linux
             image: ""
             # ARM MacOS should take fat binaries built on ARM
@@ -109,8 +106,6 @@ jobs:
           - runson:
               display: macos-x64
               name: macos-13
-            java-version: 11
-            java-distribution: corretto
             architecture: x64
             image: ""
             # x64 MacOS should take fat binaries built on ARM
@@ -119,8 +114,6 @@ jobs:
               display: linux-arm64
               # There is no "latest" tag available (yet) as ARM runners are still in public preview
               name: ubuntu-24.04-arm
-            java-version: 11
-            java-distribution: corretto
             image: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
     runs-on: ${{ matrix.runson.name }}
     container:
@@ -131,8 +124,8 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
-          distribution: ${{ matrix.java-distribution }}
-          java-version: ${{ matrix.java-version }}
+          distribution: ${{ matrix.java-distribution || env.java_default_distribution }}
+          java-version: ${{ matrix.java-version || env.java_default_version }}
           architecture: ${{ matrix.architecture }}
       - name: Checkout sources
         uses: actions/checkout@v4
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -91,7 +91,8 @@ jobs:
       matrix:
         runson:
           - display: linux-x64
-            name: ubuntu-latest # Using "latest" here as the build and test will any ways run inside a container which we control
+            # Using "latest" here as the build and test will any ways run inside a container which we control
+            name: ubuntu-latest
         java-version: [11, 17, 21, 24]
         java-distribution: [corretto]
         container: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
@@ -101,7 +102,8 @@ jobs:
               name: macos-14
             java-version: 11
             java-distribution: corretto
-            container: ""  # Not using container for mac-os as we have images only for linux
+            # Not using container for mac-os as we have images only for linux
+            container: ""
             # ARM MacOS should take fat binaries built on ARM
             asprof-binaries-job: macos
           - runson:
@@ -115,7 +117,8 @@ jobs:
             asprof-binaries-job: macos
           - runson:
               display: linux-arm64
-              name: ubuntu-24.04-arm # There is no "latest" tag available (yet) as ARM runners are still in public preview
+              # There is no "latest" tag available (yet) as ARM runners are still in public preview
+              name: ubuntu-24.04-arm
             java-version: 11
             java-distribution: corretto
             container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
@@ -173,7 +176,8 @@ jobs:
           make test-java -j
       - name: Upload integration test logs
         uses: actions/upload-artifact@v4
-        if: always() # we always want to upload test logs, especially when tests fail
+        # Always upload, especially after test failure
+        if: always()
         with:
           name: integration-test-logs-${{ matrix.runson.display }}-${{ matrix.java-version }}
           path: |
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -69,7 +69,7 @@ jobs:
           GITHUB_SHA: ${{ github.sha }}
       - name: Set artifact name
         id: set_artifact_name
-        run: echo "artifact_name=async-profiler-${GITHUB_SHA:0:7}-${{ matrix.runson.display }}" >> $GITHUB_OUTPUT
+        run: echo "artifact_name=async-profiler-${{ matrix.runson.display }}-${GITHUB_SHA:0:7}" >> $GITHUB_OUTPUT
         shell: bash
         env:
           GITHUB_SHA: ${{ github.sha }}
@@ -131,7 +131,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Set artifact name
         id: set_artifact_name
-        run: echo "artifact_name=async-profiler-${GITHUB_SHA:0:7}-${{ matrix.asprof-binaries-job || matrix.runson.display }}" >> $GITHUB_OUTPUT
+        run: echo "artifact_name=async-profiler-${{ matrix.asprof-binaries-job || matrix.runson.display }}-${GITHUB_SHA:0:7}" >> $GITHUB_OUTPUT
         shell: bash
         env:
           GITHUB_SHA: ${{ github.sha }}
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -36,11 +36,8 @@ jobs:
               name: ubuntu-latest
             container: public.ecr.aws/async-profiler/asprof-builder-x86:latest
           - runson:
-              display: macos-arm64
+              display: macos
               name: macos-14
-          - runson:
-              display: macos-x64
-              name: macos-13
     runs-on: ${{ matrix.runson.name }}
     container:
       image: ${{ matrix.container }}
@@ -105,6 +102,8 @@ jobs:
             java-version: 11
             java-distribution: corretto
             container: ""  # Not using container for mac-os as we have images only for linux
+            # ARM MacOS should take fat binaries built on ARM
+            asprof-binaries-job: macos
           - runson:
               display: macos-x64
               name: macos-13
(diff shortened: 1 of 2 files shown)


fandreuz:
You can expand the `Details` dropdown in the PR description to see some images which depict how this is going to look like

Makefile
```diff
@@ -232,6 +232,14 @@ test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
 	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/test.jar:build/jar/*:build/lib/*" one.profiler.test.Runner $(TESTS)
 
+test-java-from-release: build-test-libs build-test-bins
```
fandreuz, line 235: This is coming from #1238

apangin (approved):

apangin:
LGTM. Let @krk have a look too

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -4,25 +4,92 @@ on:  # We are very liberal in terms of triggering builds. This should be revisit
...
+            container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
+          - runson:
+              display: linux-x64
+              name: ubuntu-latest
+            container: public.ecr.aws/async-profiler/asprof-builder-x86:latest
+          - runson:
+              display: macos-arm64
+              name: macos-14
+          - runson:
+              display: macos-x64
```
apangin, line 42: Why do we build on macos-x64?
`macos-arm64` produces universal (aka "fat") binaries that can run both on ARM64 and x64.

fandreuz: 6fb94f7

apangin (approved):

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -65,65 +136,58 @@ jobs:
...
+      - name: Download async-profiler JAR artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: async-profiler-jars
+          path: jar_artifacts
+      - name: Extract async-profiler artifact
+        id: extract_artifact
         run: |
-          HASH=$(echo ${{ github.sha }} | cut -c-7)
+          release_archive=$(basename $(find async_profiler_release -type f -iname "async-profiler-*" ))
```
krk, line 158: Can we build the name instead of using `find`, to ensure that we are getting the intended file?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -65,65 +136,58 @@ jobs:
...
-              make COMMIT_TAG=$HASH FAT_BINARY=true coverage -j
-            ;;
-            *)
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc coverage -j
-            ;;
-          esac
-      - name: Upload test logs
+          mkdir -p build/jar
+          cp ${{ steps.extract_artifact.outputs.jars_directory }}/* build/jar
+          make build/test.jar
```
krk, line 173: Why are we building the jars here, shouldn't we take the ones built in the "build" steps?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -65,65 +136,58 @@ jobs:
...
+          make test-java -j
+      - name: Upload integration test logs
         uses: actions/upload-artifact@v4
-
-        if: always() # we always want to upload test logs, especially when tests fail
+        # Always upload, especially after test failure
+        if: always()
         with:
-          name: test-logs-${{ matrix.runson.display }}-${{ matrix.java-version }}
+          name: integration-test-logs-${{ matrix.runson.display }}-${{ matrix.java-version }}
```
krk, line 182: Let's add <sha> to the logs zips, e.g. `integration-test-logs-linux-arm64-11.zip` -> `integ-logs-linux-arm64-11-<sha>.zip`.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -65,65 +136,58 @@ jobs:
...
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release test -j
+              tar xvf async_profiler_release/$release_archive
             ;;
           esac
-          echo "archive=$(basename $(find . -type f -iname "async-profiler-*" ))" >> $GITHUB_OUTPUT
-      - name: Coverage
-        id: coverage
+          echo "jars_directory=jar_artifacts" >> $GITHUB_OUTPUT
+          echo "release_directory=$(basename $(find . -type d -iname "async-profiler-*" ))" >> $GITHUB_OUTPUT
+      - name: Run integration tests
```
krk, line 169: We could move this to the `Makefile`, as `make integ ASPROF_BUILD=<asprof build folder>`, to enable "integration testing" local async-profiler binaries from other folders.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -65,65 +136,58 @@ jobs:
...
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc coverage -j
-            ;;
-          esac
-      - name: Upload test logs
+          mkdir -p build/jar
+          cp ${{ steps.extract_artifact.outputs.jars_directory }}/* build/jar
+          make build/test.jar
+          cp -r ${{ steps.extract_artifact.outputs.release_directory }}/bin build
+          cp -r ${{ steps.extract_artifact.outputs.release_directory }}/lib build
+          make test-java -j
```
krk, line 176: Good effort, this should parallelize some parts of the build but the test runner itself is not parallel. We may consider running tests in parallel to `Runner.java` - not in the current PR though.

krk (changes requested):
Great simplification!

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -4,30 +4,95 @@ on:  # We are very liberal in terms of triggering builds. This should be revisit
...
+              display: linux-arm64
+              name: ubuntu-24.04-arm
+            container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
+          - runson:
+              display: linux-x64
+              name: ubuntu-latest
+            container: public.ecr.aws/async-profiler/asprof-builder-x86:latest
+          - runson:
+              display: macos
+              name: macos-14
```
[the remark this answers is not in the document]
fandreuz: We need macos-13 to run the test suite on x64 MacOS, we can replace macos-14 with macos-15

.github/workflows/test-and-publish-nightly.yml
fandreuz: https://github.com/async-profiler/async-profiler/pull/1238

.github/workflows/test-and-publish-nightly.yml
fandreuz: #1258

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -4,30 +4,95 @@ on:  # We are very liberal in terms of triggering builds. This should be revisit
...
   - pull_request
 
+env:
+  java_default_distribution: corretto
+  java_default_version: 11
+
 jobs:
   build-jars:
     runs-on: ubuntu-latest
+    name: Build JARs
```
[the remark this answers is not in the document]
fandreuz: I don't particularly like bending human-readable naming to the limitations of a specific UI (GitHub Actions UI is quite bad indeed). For instance, this is not a problem in GitHub CLI.

I'll make this change if you have a strong opinion

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
fandreuz: 254d89f9f11633b07a4945925234dcb32740dbae

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -4,30 +4,95 @@ on:  # We are very liberal in terms of triggering builds. This should be revisit
...
+          path: build/jar/*
+          if-no-files-found: error
+  build-and-upload-binaries:
+    strategy:
+      matrix:
+        include:
+          - runson:
+              display: linux-arm64
+              name: ubuntu-24.04-arm
+            container: "public.ecr.aws/async-profiler/asprof-builder-arm:latest"
```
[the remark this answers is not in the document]
fandreuz: f1042343583d55efc70541cd111e5621ed7dc5c1

.github/workflows/test-and-publish-nightly.yml
fandreuz: The reasoning here is that the name of the `.tar.gz` is decided by `Makefile`, which we do not fully control in GitHub Actions. So, we would need one of the following:
- impose the full name via some variable set in GHA
- figure what the name is going to look like, and make sure changes in `Makefile` are reflected in GHA when the logic to create the name changes in the former

Both are less clean than using `find` in my opinion, especially because the directory `async_profiler_release` should only contain one file (the `.tar.gz`)

.github/workflows/test-and-publish-nightly.yml
fandreuz: dcebf8d21886ca4a8ffbe37a080dc041f0feb678

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -37,25 +102,31 @@ jobs:
               name: macos-14
             java-version: 11
             java-distribution: corretto
```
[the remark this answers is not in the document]
fandreuz: Actually no, this cannot be done: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#env

GHA's `env` variables are not available during matrix definition. We could omit the `java-version` here and default to `default_java_version` during the steps, but we also need it for the job name definition:
```
name: "Integration test (${{ matrix.runson.display }}, ${{ matrix.java-distribution }} ${{ matrix.java-version }})"
```
which, again, cannot use GHA's `env` variables

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
krk: I am good if we can shorten one of these or `Test and Publish Nightly Builds` -> `ci` as the name does not add value:

image

.github/workflows/test-and-publish-nightly.yml
krk: Let's keep it this way, makes sense.

.github/workflows/test-and-publish-nightly.yml
krk: Can't we do both, add a makefile target that does not do any copy but changes paths to read from `ASPROF_BUILD` and call this from GHA?

The use case I have in mind is running same tests in different locally checked-out async-profiler folders.

krk:
Thanks, replied to some.

.github/workflows/test-and-publish-nightly.yml
fandreuz: Does it need to be another target? Maybe it's enough to parametrize `build/`

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
fandreuz: 38f13b63379d299a7051ef3d3542e60aa0e6bbc8

.github/workflows/test-and-publish-nightly.yml
krk: When `ASPROF_BUILD` is passed, we do not want to build makefile dependencies of `test-java`, otherwise they should be built as is today.

.github/workflows/test-and-publish-nightly.yml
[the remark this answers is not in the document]
krk: Had a discussion with apangin, we need to keep macos-13 as it is the only free x64 macos image. 14 can be replaced with 15.

.github/workflows/test-and-publish-nightly.yml
krk: Let's consider this separately and not block the PR.

krk (approved):

PR #1254, opened 2025-04-23
Fix `NativememTests#dlopenCustomLib` on Alpine

fandreuz:
### Description
This PR fixes `NativememTests#dlopenCustomLib` on Alpine.

`Hooks::init` includes the following three lines:
```cpp
_orig_pthread_create = ADDRESS_OF(pthread_create);
_orig_pthread_exit = ADDRESS_OF(pthread_exit);
_orig_dlopen = ADDRESS_OF(dlopen);
```

`ADDRESS_OF` is defined as follows:
```cpp
#define ADDRESS_OF(sym) ({ \
    void* addr = dlsym(RTLD_NEXT, #sym); \
    addr != NULL ? (sym##_t)addr : sym;  \
})
```
and ignores `dlerror`, which could then leak in subsequent parts of the code, even though `sym` is eventually patched using the original symbol rather than the one resolved by `dlsym`. This is why the `ASSERT_NO_DLERROR` at L48 fails with the following error:
```
Symbol not found: dlopen
```

A similar problem happens a bit below, in the `ASSERT_NO_DLERROR` after `(call_malloc_t)dlsym(lib, "call_malloc")`:
```
Library libjvm.so is not already loaded
```
This is coming from `Profiler::start -> VM::tryAttach`, which does not find any `libjvm.so` (as expected) and does not reset the `dlerror`.

### Related issues
#1226 

### Motivation and context
Tests introduced in #1243 do not work properly on Alpine, presumably because `dlerror` is not used in the same way in glibc and musl.

### How has this been tested?
```bash
docker run \
   -it --privileged --rm \
   -v $(pwd):/async-profiler amazoncorretto:11-alpine-jdk 
   sh -c 'apk add --no-cache musl-dev util-linux make gcc g++ linux-headers && \
      cd /async-profiler && \
      rm -rf build && \
      make -j test-java TESTS=jfr'
```

---

### DIFF
--- a/test/test/nativemem/profile_with_dlopen.c
+++ b/test/test/nativemem/profile_with_dlopen.c
@@ -44,8 +44,10 @@ int main(int argc, char** argv) {
     void* libprof = dlopen("libasyncProfiler.so", RTLD_NOW);
     ASSERT_NO_DLERROR();
 
-    ((asprof_init_t)dlsym(libprof, "asprof_init"))();
+    asprof_init_t init_func = (asprof_init_t) dlsym(libprof, "asprof_init");
     ASSERT_NO_DLERROR();
+    init_func();
+    dlerror();
 
     asprof_execute_t asprof_execute = (asprof_execute_t)dlsym(libprof, "asprof_execute");
     ASSERT_NO_DLERROR();
@@ -69,6 +71,8 @@ int main(int argc, char** argv) {
     if (!dlopen_first) {
         lib = dlopen("libcallsmalloc.so", RTLD_NOW);
         ASSERT_NO_DLERROR();
+    } else {
+        dlerror();
     }
 
     call_malloc_t call_malloc = (call_malloc_t)dlsym(lib, "call_malloc");


fandreuz:
Should we instead make sure `dlerror()` is reset appropriately by Async-Profiler? I believe this kind of failures could also happen in different parts of the code, and is not specific to this test

fandreuz:
Failure here: https://github.com/async-profiler/async-profiler/actions/runs/14613212540/job/40995410369

fandreuz:
> I think a better fix (and a more canonical way to check for errors) will be not to call `dlerror()` unless the previous function fails, i.e. returns NULL.

If you mean that we should not call `dlerror()` in the test code, the dlerror buffer can be full even with no detectable failure for the code calling Async-Profiler API, or even with no failure at all, e.g. when calling `asprof_execute` for a non-JVM process. Am I missing something?

krk:
> Should we instead make sure dlerror() is reset appropriately by Async-Profiler?

I would think we can log such errors, or ignore them purposefully by calling `dlerror` from async-profiler. Ideally, returning from `asprof_execute` should not leave a `dlerror` that was already handled or ignored.

fandreuz:
@krk @apangin moved the fix in Async-Profiler main source set: f5ae5ce1600a9da40b960ad6a6ea4a004fba46d7

krk (approved):

apangin:
I mean, calling `dlerror()` alone is not a proper way of checking errors. `dlerror()` should be only called if the function that you are testing fails. So I'd modify `ASSERT_NO_DLERROR()` and its usages instead.

krk:
How do you determine if a dl function is failed, without calling `dlerror`? The man example always calls `dlerror`: https://linux.die.net/man/3/dlopen

apangin:
> How do you determine if a dl function is failed, without calling dlerror?

By checking return value. `dlopen` and `dlsym` return `NULL` when fail. If they return non-null, they've succeeded, even though `dlerror` may return junk.

krk:
That sounds good, but in the example we can see that `dlerror` is called regardless of dlopen return value:

```
   *(void **) (&cosine) = dlsym(handle, "cos");

   if ((error = dlerror()) != NULL)  {
        fprintf(stderr, "%s\n", error);
        exit(EXIT_FAILURE);
    }
```

fandreuz:
@krk @apangin what's the conclusion here?

krk:
We can change the tests not to be too strict about `dlerror` return values, as long as dlopen/dlsym functions do not return null.

krk (approved):

Issue #1250, opened 2025-04-22
Async profiler crashing during symbol parsing & hook installation


Baraa-Hasheesh:
### Describe the bug

While investigating the issue found here #1245 

It was found it's possible to trigger a race condition in the async profiler, which could result in the profiler crashing during parsing library symbols or during installation of hooks 

This happens in the provided examples happens due to having two different distinct flows for `dlopen`,
1. Some `dlopen` calls will go through the profiler hooked methods
2. Some `dlopen` calls will never be redirected to the profiler hooked methods

This in turn makes it possible for the profiler to have multiple possible race conditions that include the following:
1. During `/proc/self/maps` the lib could be found in a non ready state (Not fully loaded) which could result in a crash during `ElfParser::parseFile` call
2. During `/proc/self/maps` the lib could be found in a fully loaded state but is unloaded before `ElfParser::parseFile` which results in a crash
3. Lib is detected & parsed correctly during the `Symbols::parseLibraries` call, but is unloaded before the `MallocTracer::installHooks` call which results in a crash

All of the above could happen due to race conditions between hooked & unhooked `dlopen` calls.

**Important Note:** After discussing with @apangin it was agreed pm that `dlopen` hook shortcomings could be used to reproduce the issue but it should not be considered a root cause as it's not a goal to hook everything & trying to do that is a dead end as calls to absolute addresses of methods will bypass installed hooks.

### Expected vs. actual behavior

Async profiler should not crash while parsing & patching libs 

### Reproduction Steps

Reproducer to cause Seg Fault from race condition due to unpatched dlopen

Note: 
1. Make sure to replace ${PWD} with current directory where files are created
2. Create all files into the same directory
3. Create all files under a directory called `unpatched`

Create the following Java file `Main.java`

```
package unpatched;

import java.io.FileInputStream;
import java.io.InputStream;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.StandardCopyOption;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.CompletableFuture;
import java.util.concurrent.ForkJoinPool;
import java.util.stream.IntStream;

public class Main {

    private static final ForkJoinPool THREAD_POOL = new ForkJoinPool(10);

    public static native void loadLibUnpatched();

    public static native void execute(String libPath);

    private static Void executeFlow() {
        try {
            // Create a temp library file
            Path libraryPath = Files.createTempFile("temp", "my_lib.so");
[...]

PR #1261, opened 2025-04-24
#1250 validate that the obtained dlopen handle is same as the originally discovered lib in maps

Baraa-Hasheesh:
### Description
This fixes fixes one of the potential problems discovered in the #1250 
This change detects if the library was reloaded to a different location & marks the handle as invalid

Note: This solution isn't complete & needs to change handling of `dlerror() == null` to only be valid for `ld` library 

### Related issues
#1250 

### Motivation and context
Reduce effect of race conditions to async profiler

### How has this been tested?
`make test`

---


src/symbols_linux.cpp
```diff
@@ -812,10 +812,16 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
             // Also, dlopen() ensures the library is fully loaded.
             // Main executable and ld-linux interpreter cannot be dlopen'ed, but dlerror() returns NULL for them on some systems.
             void* handle = dlopen(lib.file, RTLD_LAZY | RTLD_NOLOAD);
+            bool valid_handle = false;
```
krk, line 815: If handle is not null, it is valid. I would rename the variable to reflect intent: "handle is to a different instance of the shared object, compared to what was parsed from `/maps`".

krk (changes requested):

src/symbols_linux.cpp
Baraa-Hasheesh: renamed to `original_handle`

src/symbols_linux.cpp
```diff
@@ -812,10 +812,16 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+            struct link_map* map;
+
+            if (handle != NULL && dlinfo(handle, RTLD_DI_LINKMAP, &map) == 0) {
+                original_handle = lib.image_base == (const char*)map->l_addr; 
+            }
 
             // Parse main executable regardless of dlopen result, since it cannot be unloaded.
             bool is_main_exe = main_phdr >= lib.image_base && main_phdr < lib.map_end;
-            if (handle != NULL || dlerror() == NULL || is_main_exe) {
+            if (original_handle || dlerror() == NULL || is_main_exe) {
```
apangin, line 824: This is not correct: because of "dlerror() == NULL", condition will be still true, even if base address differs.
Fortunately, `dlerror()` should go away after Kerem's fix.

Baraa-Hasheesh: That's correct 
@krk change is required here

I will sync & resolve conflicts

Baraa-Hasheesh: @apangin branch synced & conflicts handled

krk (approved):

Baraa-Hasheesh:
included here #1264

PR #1263, opened 2025-04-24
Detect and parse the loader in parseLibraries.

krk:
Use `AT_BASE` from `auxv` to get loader base address and parse it regardless of dlopen/dlerror results.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1250

### Motivation and context
The loader itself cannot be unloaded, so what dlopen/dlerror returns is not relevant. This simplifies and corrects the condition of calling `parseProgramHeaders`.

### How has this been tested?
`make test`

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -778,7 +778,7 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
 
     const char* ld_base = (const char*)getauxval(AT_BASE);
     if (!ld_base) {
-        Log::warn("Cannot determine base address of the loader.");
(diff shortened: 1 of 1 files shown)


src/symbols_linux.cpp
```diff
@@ -775,6 +776,11 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     std::unordered_map<u64, SharedLibrary> libs;
     collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count());
 
+    const char* ld_base = (const char*)getauxval(AT_BASE);
```
apangin, line 779: Nice! TIL

Issue #1256, opened 2025-04-24
Native memory profiling could lead to crashes


Baraa-Hasheesh:
### Describe the bug

Async-Profiler only patches libraries to track `malloc` calls if & only if it's started in `nativemem` profiling mode
Furthermore async-profiler keeps record of all found libs in it's own local cache of `CodeCacheArray`

This behaviour makes it possible to have a library inside `CodeCacheArray` but that library isn't patched to track `malloc` calls

Due to these optimization choices it's possible for the async-profiler to try to patch for `nativemem` profiling after it's unloaded from memory

### Expected vs. actual behavior

async-profiler shouldn't crash on native memory profiling

### Reproduction Steps

Create the following C/C++ file

* main.cpp

```
#include <dlfcn.h>
#include <iostream>
#include <pthread.h>
#include "asprof.h"

#define ASSERT_NO_DLERROR()           \
    err = dlerror();                  \
    if (err != NULL) {                \
        printf("ERROR => %s\n", err); \
        exit(1);                      \
    }

#define ASSERT_NO_ASPROF_ERR(err)                       \
    if (err != NULL) {                                  \
        fprintf(stderr, "%s\n", _asprof_error_str(err)); \
        exit(1);                                        \
    }

    
typedef void* (*my_malloc_t)(size_t);
typedef void (*my_free_t)(void*);

void outputCallback(const char* buffer, size_t size) {
    fwrite(buffer, sizeof(char), size, stdout);
}


int main() {
    void* lib;
    char* err;

    lib = dlopen("my_lib.so", RTLD_NOW | RTLD_GLOBAL);
    ASSERT_NO_DLERROR();
    printf("dlopen => %p\n", lib);

    my_malloc_t local_malloc = (my_malloc_t) dlsym(lib, "my_malloc");
    ASSERT_NO_DLERROR();

    my_free_t local_free = (my_free_t) dlsym(lib, "my_free");
    ASSERT_NO_DLERROR();

    void* libprof = dlopen("libasyncProfiler.so", RTLD_NOW);
    ASSERT_NO_DLERROR();

    ((asprof_init_t)dlsym(libprof, "asprof_init"))();
    ASSERT_NO_DLERROR();

    asprof_execute_t _asprof_execute = (asprof_execute_t)dlsym(libprof, "asprof_execute");
    ASSERT_NO_DLERROR();

[...]

PR #1264, opened 2025-04-25
Protect patching with an open handle

fandreuz:
### Description
In this PR I introduce additional protection during patches. The main change is a new function `isSafeToPatch`, which combines the checks introduced in #1261, #1263, #1243, and can be used in other parts of the code like `MallocTracer`.

Thanks @Baraa-Hasheesh for the discussion and feedback on the code.

### Related issues
#1250, #1256

### Motivation and context
Keeping an open handle makes sure we the library cannot be unloaded while the patch is being made, thus providing additional protection against the failures in the linked issues.

### How has this been tested?
`make test`

---


fandreuz:
Also tested the crash in the [stress test](https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/unpatched) by @Baraa-Hasheesh, and could not reproduce the issue anymore after this PR

src/symbols.h
```diff
@@ -8,13 +8,24 @@
...
     static bool _libs_limit_reported;
+    static const void* _main_phdr;
+    static const char* _ld_base;
+
+    static bool isMainExecutable(const char* image_base, const void* map_end) {
+      return _main_phdr != NULL && _main_phdr >= image_base && _main_phdr < map_end;
+    }
+
+    static bool isLoader(const char* image_base) {
+      return image_base != NULL && _ld_base == image_base;
```
Baraa-Hasheesh, line 27: I think the `!= NULL` check should be against _ld_base

```suggestion
      return _ld_base != NULL && _ld_base == image_base;
```

krk:
Let's add minimal reproducers as cpp unit tests.

src/codeCache.h
```diff
@@ -201,6 +213,19 @@ class CodeCache {
     void setDwarfTable(FrameDesc* table, int length);
     FrameDesc* findFrameDesc(const void* pc);
 
+    bool isValidHandle(void* handle) const {
```
roy-soumadipta, line 216: codeCache has no os specific impl. Why not just declare here and implement in codeCache.cpp?

fandreuz: 56539688585770d7e0213fae6a79171b122d023f

src/symbols.h
```diff
@@ -23,6 +34,32 @@ class Symbols {
     static bool haveKernelSymbols() {
         return _have_kernel_symbols;
     }
+
+    static bool isSafeToPatch(CodeCache* cc, void **handle_ptr) {
```
roy-soumadipta, line 38: I get that symbols.h has os specific impls. But unless you want inlining this is not a good fit for a simple header file like symbols. Maybe something like a reusable library_patcher may be a better choice. Would like @apangin and @krk to weigh in on this.

src/codeCache.h
```diff
@@ -7,6 +7,11 @@
 #define _CODECACHE_H
 
 #include <jvmti.h>
+#include <dlfcn.h>
```
Baraa-Hasheesh, line 10: nitpick: let's also move these imports imports into cpp as they are no longer needed here in header file

src/symbols.h
Baraa-Hasheesh: @roy-soumadipta do you mean creating a shared symbols.cpp that has the shared complex logic between Macos & Linux?

src/codeCache.h
fandreuz: 03d8f7917983a99a74b0a15d7c8e939a8525231a

src/symbols.h
roy-soumadipta: @Baraa-Hasheesh Something on those lines but maybe not shared symbols.cpp. I feel a lib_patcher would address the responsibility better and enable future additions which is a possibility with lib patching.

fandreuz:
> Let's add minimal reproducers as cpp unit tests.

Adapted test from @Baraa-Hasheesh's [repo](https://github.com/Baraa-Hasheesh/asprof-reproducers/blob/main/native_mem_crash)

test/test/nativemem/NativememTests.java
```diff
@@ -188,4 +188,11 @@ public void dlopenCustomLib(TestProcess p) throws Exception {
 
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_SIZE, 0L), 1);
     }
+
+    @Test(os = Os.LINUX, sh = "%testbin/nativemem_known_lib_crash %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "dlopen_first")
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=%lib %testbin/nativemem_known_lib_crash %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/test/lib:build/lib"}, nameSuffix = "LD_PRELOAD")
```
Baraa-Hasheesh, line 193: why do we have 2 test here?

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,77 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "asprof.h"
+#include <pthread.h>
+#include <dlfcn.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+typedef void *(*my_malloc_t)(size_t);
```
Baraa-Hasheesh, line 12: definition not used

fandreuz: Thanks, 1feaf129ebf90bc90e4fe496cfc4fc1ee54caaee

test/test/nativemem/NativememTests.java
fandreuz: Does not really make any difference indeed, removed the second: efae0893ca5939fb73bc1706b7bebfa427ddcd6a

src/mallocTracer.cpp
```diff
@@ -153,6 +154,10 @@ void MallocTracer::patchLibraries() {
 
     while (_patched_libs < native_lib_count) {
         CodeCache* cc = (*native_libs)[_patched_libs++];
+        void* handle;
+        if (!Symbols::isSafeToPatch(cc, &handle)) {
+            continue;
+        }
```
apangin, line 160: Maybe follow RAII idiom with a stack-allocated class?

fandreuz: Nice idea, will do

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,73 @@
...
+        fprintf(stderr, "%s\n", asprof_error_str(err)); \
+        exit(1);                                        \
+    }
+
+void outputCallback(const char* buffer, size_t size) {
+    fwrite(buffer, sizeof(char), size, stdout);
+}
+
+/*
+Idea of the test (the behavior defined above applies as of 0c72a8d):
```
fandreuz, line 29:
```suggestion
Idea of the test (the behavior applies as of 0c72a8d):
```

src/mallocTracer.cpp
fandreuz: 79d93a0c0df7b60d69cfa2b4630b8878900c9406

src/codeCache.cpp
```diff
@@ -234,26 +243,12 @@ void CodeCache::addImport(void** entry, const char* name) {
...
-    for (int ty = 0; ty < NUM_IMPORT_TYPES; ty++) {
-        void** entry = _imports[id][ty];
-        if (entry != NULL) {
-            *entry = hook_func;
-        }
-    }
-}
-
 void CodeCache::makeImportsPatchable() {
+#ifdef __linux__
```
Baraa-Hasheesh, line 251: why was this definition added?

src/codeCache.cpp
```diff
@@ -311,3 +308,55 @@ size_t CodeCache::usedMemory() {
...
+#endif
+}
+
+PatchingHandle::~PatchingHandle() {
+    if (_lib_handle) {
+        dlclose(_lib_handle);
+    }
+}
+
+PatchingHandle CodeCache::makePatchingHandle() {
```
Baraa-Hasheesh, line 331: I think the following names would be better

`CodeCachePatcher CodeCache::getPatcher()`

what do you think?

src/codeCache.cpp
```diff
@@ -311,3 +308,55 @@ size_t CodeCache::usedMemory() {
...
+
+    if (Symbols::isMainExecutable(_image_base, _max_address) || Symbols::isLoader(_image_base)) {
+      return PatchingHandle(this);
+    }
+
+    // Protect library from unloading while parsing in-memory ELF program headers.
+    // Also, dlopen() ensures the library is fully loaded.
+    void* handle_ptr = dlopen(_name, RTLD_LAZY | RTLD_NOLOAD);
+    if (isValidHandle(handle_ptr)) {
+      return PatchingHandle(this, handle_ptr);
```
Baraa-Hasheesh, line 344: Wouldn't this potentially call the copy constructor & cause the de-constructor to be called prematurely hence losing the dlopen protection we just obtained?

src/codeCache.cpp
fandreuz: Which definition?

fandreuz: If you refer to the `ifdef` check, I noticed that imports are always assumed to be patchable in MacOS, thus this code block does not apply in MacOS.

I added a `_imports_patchable=false` in `saveImport` because I noticed a new import could be saved after `makeImportsPatchable` was called (in fact I was getting a SEGFAULT due to this). The side-effect of this new assignment though is that we are going to call `makeImportsPatchable` also on MacOS, which is why I had to adapt it.

src/codeCache.cpp
fandreuz: I kinda prefer the current naming, any other opinions? @async-profiler/core-developers

src/codeCache.cpp
fandreuz: My understanding here is that there are two cases:
- When copy elision gets involved (which is guaranteed after c++17 and optional before) the variable is directly transferred to the caller. In this case the compiler is even allowed to potentially alter the behavior of the program by not calling the destructor at all (see notes [here](https://en.cppreference.com/w/cpp/language/copy_elision))
- When copy elision is not involved, the returned value is casted into an rvalue, which is eligible for initialization on the caller side via move constructor

Ref:
- https://stackoverflow.com/a/17473874
- https://en.cppreference.com/w/cpp/language/copy_elision

apangin: To avoid such a confusion and to make the intent clearer, why not construct an object right in the place where the protection is needed? (E.g. similar to how `MutexLocker` is constructed).
Something like this:
```
ProtectedLibrary lib(cc);
if (lib.isValid()) {
    // use cc
}
```
Btw, `PatchingHandle` sounds vague.
Some alternative names from the top of my head: LibraryHolder / ProtectedLibrary / UnloadProtection

fandreuz: - Moved `makePatchingHandle` into the handle ctor: 5378fcf9bd45465b24a039adff281a222e5734af
- Renamed `PatchingHandle`: 3dd17e6ab909373e9f6a8a5b7d65e15124c4c086

src/codeCache.cpp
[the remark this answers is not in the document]
fandreuz: 2c694850c645e61db97f76597735419afe451cd3

fandreuz: 2c694850c645e61db97f76597735419afe451cd3

src/codeCache.cpp
```diff
@@ -311,3 +308,57 @@ size_t CodeCache::usedMemory() {
     }
     return bytes;
 }
+
+bool CodeCache::isValidHandle(void* handle) const {
```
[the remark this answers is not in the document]
fandreuz: I think this rather belongs to `CodeCache`, it's just a check against a private field of `CodeCache`. Also, it's called in the `UnloadProtection` ctor, thus the object is not fully constructed when we need it

src/codeCache.cpp
```diff
@@ -355,6 +347,10 @@ void UnloadProtection::patchImport(ImportId id, void* hook_func) const {
         return;
     }
 
+    if (!_protected_cc->_imports_patchable) {
```
Baraa-Hasheesh, line 350: _imports_patchable should be set to true after patching to avoid calling makeImportsPatchable multiple times in linux

src/codeCache.cpp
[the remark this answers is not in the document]
krk: We could eliminate this inversion by introducing a factory method:

```diff
-UnloadProtection handle = UnloadProtection(cc);
+UnloadProtection handle = cc->protectUnload();
```

Then, `cc` is free to access its private fields before constructing the `UnloadProtection`.

src/codeCache.h
```diff
@@ -100,13 +100,38 @@ class CodeBlob {
...
+};
+
 class CodeCache {
   private:
     char* _name;
     short _lib_index;
     const void* _min_address;
     const void* _max_address;
     const char* _text_base;
+    // TODO: Cleanup usages of lib.image_base
```
krk, line 133: Is this leftover?

src/mallocTracer.cpp
```diff
@@ -153,21 +154,22 @@ void MallocTracer::patchLibraries() {
 
     while (_patched_libs < native_lib_count) {
         CodeCache* cc = (*native_libs)[_patched_libs++];
+        UnloadProtection handle = UnloadProtection(cc);
 
-        cc->patchImport(im_malloc, (void*)malloc_hook);
-        cc->patchImport(im_realloc, (void*)realloc_hook);
-        cc->patchImport(im_free, (void*)free_hook);
-        cc->patchImport(im_aligned_alloc, (void*)aligned_alloc_hook);
+        handle.patchImport(im_malloc, (void*)malloc_hook);
```
krk, line 159: We should still use `cc->patchImport`. The fact that `UnloadProtection` is alive, should give us the needed protection.

src/symbols.h
```diff
@@ -15,14 +15,28 @@ class Symbols {
...
+    }
+
     static bool haveKernelSymbols() {
         return _have_kernel_symbols;
     }
+
+    Symbols();
 };
 
+static Symbols symbols;
```
krk, line 40: All members seem to be static, is this variable required?

krk: I see the static ctor, disregard.

src/mallocTracer.cpp
fandreuz: What I'm trying to achieve here is that you cannot patch without having an active `UnloadProtection`

src/codeCache.cpp
fandreuz: See implementation of `makeImportsPatchable`

src/codeCache.cpp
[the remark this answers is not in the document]
fandreuz: This was exactly the previous implementation, before I got [this comment](https://github.com/async-profiler/async-profiler/pull/1264#discussion_r2064094730): 5378fcf9bd45465b24a039adff281a222e5734af

src/codeCache.h
fandreuz: This is something we should do, but could pollute this PR. Basically there are some cases where we pass both `cc` and `lib.image_base` to some function, which we could now simplify to passing only `cc`. What's your preference?

src/codeCache.cpp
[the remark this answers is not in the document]
fandreuz: 20f5f81cead8dc17a485ae41c056bd49bbbcde57

src/codeCache.h
```diff
@@ -100,13 +100,38 @@ class CodeBlob {
...
+    CodeCache* _protected_cc;
+    void* _lib_handle;
+
+  public:
+    UnloadProtection(CodeCache *cc);
+    UnloadProtection(UnloadProtection&& other) {
+        _protected_cc = other._protected_cc;
+        _lib_handle = other._lib_handle;
+        other._lib_handle = nullptr;
+    }
```
apangin, line 116: This boilerplate is not needed anymore.

fandreuz: 79da179bcf966e9afcf92643f0c04a59946f6749

src/codeCache.cpp
```diff
@@ -311,3 +306,57 @@ size_t CodeCache::usedMemory() {
...
+
+    if (Symbols::isMainExecutable(cc->_image_base, cc->_max_address) || Symbols::isLoader(cc->_image_base)) {
+        _protected_cc = cc;
+        return;
+    }
+
+    // Protect library from unloading while parsing in-memory ELF program headers.
+    // Also, dlopen() ensures the library is fully loaded.
+    void* handle_ptr = dlopen(cc->name(), RTLD_LAZY | RTLD_NOLOAD);
+    if (cc->isValidHandle(handle_ptr)) {
```
krk, line 341: Should we log a warning if dlopen fails here, except for the main executable and the loader?

src/codeCache.cpp
```diff
@@ -311,3 +306,57 @@ size_t CodeCache::usedMemory() {
...
+        _protected_cc = cc;
+        _lib_handle = handle_ptr;
+    }
+}
+
+void UnloadProtection::patchImport(ImportId id, void* hook_func) const {
+    if (!isValid()) {
+        return;
```
krk, line 349: Is this an error? Should we log a warning?

src/codeCache.h
krk: Since `image_base` is added to the `CodeCache` in the current PR, we could include this instead of keeping a TODO comment.

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,75 @@
...
+void outputCallback(const char* buffer, size_t size) {
+    fwrite(buffer, sizeof(char), size, stdout);
+}
+
+/*
+Idea of the test (the behavior applies as of 0c72a8d):
+- We load libcallsmalloc.so
+- We start AP without nativemem mode
+    => 'malloc' is not hooked
+- We dlclose libcallsmalloc.so
```
krk, line 35: A step missing before this: Stop AP.

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,75 @@
...
+    => The previously accessible memory location dedicated to libcallsmalloc.so becomes inaccessible
```
krk, line 36:
```suggestion
    => Memory region(s) for libcallsmalloc.so are now unmapped.
```

src/codeCache.h
fandreuz: 15eeade44555c0ef6381f9ed4bfe62f651a83a87

src/codeCache.cpp
fandreuz: I think we should rather log one time in `patchImport`

src/codeCache.cpp
fandreuz: 7b0511c853709db33c92eea5c460301570d07658

test/test/nativemem/nativemem_known_lib_crash.c
fandreuz: cda6051cebf575254e2e1d5a56afd4ba11cd11e2

src/codeCache.h
```diff
@@ -168,6 +168,10 @@ class CodeCache {
         return _max_address;
     }
 
+    const char* imageBase() const {
+        return _image_base != NULL ? _image_base : (char*) minAddress();
```
krk, line 173: Does this cover all cases in `parseLibraries`?

fandreuz: The only case when this may not behave exactly like the previous implementation is for `[vdso]`:
```
(gdb) print lib.file
$1 = 0x55555555ea60 "[vdso]"
(gdb) print lib.image_base
$2 = 0x7ffff7fc3000 "\177ELF\002\001\001"
(gdb) print lib.map_start
$3 = 0x7ffff7fc3000 "\177ELF\002\001\001"
```
Luckily in this case `lib.image_base == lib.map_start`, is this something we can rely on @krk ?

apangin: I don't think so.
But why is this change needed in the first place? A function called `imageBase()` that may return something other than image base looks suspicious to me.

fandreuz: So maybe we can't just use the `image_base` we store in `CodeCache` as a full replacement of the second argument of the second argument of `parseFile` and `parseProgramHeaders`. I reverted the commit and removed the TODO which triggered this change: bc306667edf00872fe74766694c27da3e1e05068

src/codeCache.cpp
```diff
@@ -311,3 +307,60 @@ size_t CodeCache::usedMemory() {
...
+    _valid = false;
+
+    if (Symbols::isMainExecutable(cc->_image_base, cc->_max_address) || Symbols::isLoader(cc->_image_base)) {
+        _valid = true;
+        return;
+    }
+
+    // Protect library from unloading while parsing in-memory ELF program headers.
+    // Also, dlopen() ensures the library is fully loaded.
+    void* handle_ptr = dlopen(cc->name(), RTLD_LAZY | RTLD_NOLOAD);
```
apangin, line 342: `cc->name()` may be incorrect in case of deleted underlying file.
Previous implementation chopped ` (deleted)` suffix for `dlopen`.

src/codeCache.cpp
```diff
@@ -311,3 +307,60 @@ size_t CodeCache::usedMemory() {
     }
     return bytes;
 }
+
+bool CodeCache::isValidHandle(void* handle) const {
+#ifdef __linux__
```
apangin, line 312: CodeCache is just a data structure, it's not good to have platform-dependent code here.
`dlopen`, `dlinfo` and even more so, mapping-related stuff are more suitable for `Symbols` class, whose implementation is platform-dependent. Moreover, in your implementation, that responsibility is already split across CodeCache and Symbols: the code calls `Symbols::isMainExecutable` and `Symbols::isLoader` which should not be public functions in the first place, since these are platform-specific implementation details.

src/codeCache.cpp
```diff
@@ -311,3 +307,60 @@ size_t CodeCache::usedMemory() {
...
+    if (cc->isValidHandle(handle_ptr)) {
+        _protected_cc = cc;
+        _lib_handle = handle_ptr;
+        _valid = true;
+    }
+}
+
+void UnloadProtection::patchImport(ImportId id, void* hook_func) const {
+    if (!isValid()) {
+        Log::info("Could not patch '%d' for '%s', unload protection failed", id, _protected_cc->name());
```
apangin, line 352: I don't think this should be logged as INFO. This is not an unusual case, and as a rule, async-profiler should not alter application output in a normal situation.

src/symbols.h
```diff
@@ -15,14 +15,28 @@ class Symbols {
     static Mutex _parse_lock;
     static bool _have_kernel_symbols;
     static bool _libs_limit_reported;
+    static const void* _main_phdr;
+    static const char* _ld_base;
```
apangin, line 19: Those are Linux-specific and not part of Symbols public interface.
Can be just static variables in `symbols_linux.cpp`.

src/symbols.h
```diff
@@ -15,14 +15,28 @@ class Symbols {
...
+
+    static bool isLoader(const char* image_base) {
+      return _ld_base != NULL && _ld_base == image_base;
+    }
+
     static bool haveKernelSymbols() {
         return _have_kernel_symbols;
     }
+
+    Symbols();
```
apangin, line 37: Constructor for all-static class?

src/codeCache.cpp
```diff
@@ -311,3 +307,60 @@ size_t CodeCache::usedMemory() {
...
+        dlclose(_lib_handle);
+    }
+}
+
+UnloadProtection::UnloadProtection(CodeCache *cc) {
+    _protected_cc = cc;
+    _lib_handle = nullptr;
+    _valid = false;
+
+    if (Symbols::isMainExecutable(cc->_image_base, cc->_max_address) || Symbols::isLoader(cc->_image_base)) {
```
apangin, line 335: `OS::isMusl()` check is needed here, too. `dlopen` protection is not applicable on Alpine.

src/symbols.h
fandreuz: 8f57df7def2cb55da274cd498c91b7d6aa228de0

src/symbols.h
fandreuz: This is to initialize `_ld_base` and `_main_phdr`, I kinda prefer this over mutex+lazy initialization

fandreuz: Do we need a mutex here? If not, I'll just initialize them lazily

src/codeCache.cpp
fandreuz: Made it debug: 8de5e01e7df48d9d25565512a830d1b14d6cf7d8

src/codeCache.cpp
fandreuz: cf1f79b96ef66b7385de307dfef56631e3d4c7ec

src/symbols.h
fandreuz: c17d592e63691f44e0fb8cb59fe4a8afb7ec1dd2

fandreuz:
All comments addressed @apangin

src/codeCache.cpp
apangin: The check is required, otherwise `dlopen()` fails on musl if the underlying fail has been deleted. `dlopen` protection is useless on musl anyway, since libraries cannot be unloaded.

fandreuz: 45e962273959e117891b4b7d4adb73ade8d063fd

src/symbols_linux.cpp
```diff
@@ -845,4 +842,35 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     }
 }
 
+static bool isValidHandle(const CodeCache* cc, void* handle) {
+    struct link_map* map;
+    // validate that the current loaded library is the same library that was observed during the /proc/self/maps processing
+    if (handle != NULL && dlinfo(handle, RTLD_DI_LINKMAP, &map) == 0) {
+        return cc->imageBase() == (const char*)map->l_addr;
```
apangin, line 849: Does this apply to shared objects whose virtual load address is non-zero?

src/symbols_macos.cpp
```diff
@@ -14,6 +14,19 @@
 #include "symbols.h"
 #include "log.h"
 
+UnloadProtection::UnloadProtection(CodeCache *cc) {
```
apangin, line 17: I don't see its usage in `Symbols::parseLibraries` on macOS

src/codeCache.h
```diff
@@ -100,13 +100,20 @@ class CodeBlob {
...
 
+class UnloadProtection;
+
 class CodeCache {
+  friend UnloadProtection;
+
   private:
     char* _name;
+    // E.g. remove (deleted) so that removed library can be reopened
+    char* _clean_name;
```
apangin, line 111: It doesn't need to be stored here - it can be materialized where needed.
Having two name fields is confusing. Furthermore, " (deleted)" is Linux-specific thing.

src/symbols.cpp
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "log.h"
+#include "symbols.h"
+
+void UnloadProtection::patchImport(ImportId id, void* hook_func) const {
```
apangin, line 9: On one hand, I like your idea to protect `patchImport` calls on the interface level.
But at the same time, I'm not fan of blurring logic between CodeCache and Symbols. Things being equal, I'd prefer to minimize changes in existing code.

src/symbols_linux.cpp
```diff
@@ -64,6 +65,33 @@ static void applyPatch(CodeCache* cc) {}
...
+static const char* getLdBase() {
+    const char* ld_base = (const char*)getauxval(AT_BASE);
+    if (!ld_base) {
+        Log::warn("Cannot determine base address of the loader");
+    }
+    return ld_base;
+}
+
+static const void* _main_phdr = getMainPhdr();
+static const char* _ld_base = getLdBase();
```
apangin, line 86: I'd just inline `getauxval` call here.
Log might not be initialized at this point. I can't see `getauxval` failing anyway.

src/codeCache.h
fandreuz: It can be materialized when needed, but then somebody has to free it. If we return a simple `char*` that's going to make things more complicated. How about returning `std::string`?

src/symbols_macos.cpp
fandreuz: 5e10cdd462942f2fda16317a8e729b6a19207150

src/symbols_linux.cpp
fandreuz: b5d4e19224fba4fca4a9367276959207ec952dbb

src/symbols.cpp
fandreuz: 70c3800772e488db3d9845853514fe6f1778e398

src/symbols_linux.cpp
apangin: I'm interested in a situation where a shared object has `PT_LOAD` program header with `p_offset == 0 ` and `p_vaddr != 0`. Can you provide me with an evidence that this case is covered or that it never happens?

src/symbols_macos.cpp
```diff
@@ -14,6 +14,19 @@
...
+    _protected_cc = cc;
```
apangin, line 18: `_protected_cc` is no longer used.

src/codeCache.h
apangin: I mean it's better for CodeCache not to deal with `cleanName()`. Having both `name()` and `cleanName()` in the same interface is not really clean (pun intended).
AFAICS, there is only one place in Linux-specific code where you need a name stripped off "(deleted)" suffix, when re-opening a library in UnloadProtection. Why not process the name there? It can be even stored in a stack allocated buffer, so no free is required. Though dynamic allocation is also fine.

krk: >  (deleted)" is Linux-specific thing.

This shouldn't be part of `CodeCache`.

src/symbols_macos.cpp
fandreuz: Removed all occurrences, thanks: 2abe2aa238474e648137c8eca8fcfcf1499fa430

src/symbols_linux.cpp
Baraa-Hasheesh: @apangin I will investigate this

Baraa-Hasheesh: @apangin I managed to confirm that it's indeed a problem 

`$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libmalloc.$(SOEXT) test/native/libs/malloc.c -Wl,--Ttext-segment=0x10000000`

I'm investigating how to solve

Baraa-Hasheesh: @apangin My recommendation to unblock this is to simply remove the `dlinfo` checks for now & use standard dlopen protection 

The `dlinfo` protection is only really needed if library is re-opened using a different thread after close 
I.E. after Lib is detected -> `dlclose` -> `dlopen` back again  on a different thread

After this is solved/merged, I can try to create a reproducer for that case in particular & track it as a different issue

Baraa-Hasheesh: @apangin 

I would consider this & the above unrelated issues & just go with just the dlopen protections without the extra protection (**to check if the current opened handle is the same as the original**) to prevent obvious crashes  such as #1256

Note the lack of these extra checks will cause crashes if you do the correct flows => I can think of one way to crash it using reloading the same library across multiple threads (`dlopen -> dlclose -> dlopen` chains across multiple threads)

From what I can observe the biggest risk factor we have is the fact we parse & patch it two different locations

My recommendation which should make crashing quite unlikely (I think it will be guaranteed) is the following (will not have the best performance):
1. Use `/proc/self/maps` to detect libs
2. Use `dlopen` to prevent libs from unloading that were discovered in step 1
3. Use `/proc/self/maps` to make sure the library wasn't re-loaded at a different address between steps 1 & 2
4. Only parse libs that were found in both step 1 & 3 (dlopen protection is guaranteed & absolute in this case)
5. Patch in place (move patching to be part of the Symbol parsing logic) 
6. use `dlclose` to do final cleanup

fandreuz: How about we check if this situation applies, and if it does we simply skip the check and assume the handle is valid? The situation highlighted by @Baraa-Hasheesh seems to be an edge case, and I think it's still valuable to have this check until we find a proper fix. 

Having this will be give us more protection than what we currently have, and won't make things worse than they are now

apangin: > The dlinfo protection is only really needed if library is re-opened using a different thread after close

No, the problem can happen even in a single-threaded cases, for instance, if we attempt to do native memory profiling long after a library was first parsed.

I do think we need some level of protection against this situation, even if it is not 100% accurate. If comparing `l_addr` does not work, maybe we can check `l_ld` instead.

src/symbols_linux.cpp
```diff
@@ -845,4 +845,39 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+UnloadProtection::UnloadProtection(const CodeCache *cc) {
+    _lib_handle = NULL;
+    _valid = false;
+
+    if (OS::isMusl() || isMainExecutable(cc->imageBase(), cc->maxAddress()) || isLoader(cc->imageBase())) {
+        _valid = true;
+        return;
+    }
+
+    char* nameCopy = strdup(cc->name());
```
apangin, line 866: Local variables follow `snake_case` naming convention.
Can we avoid `strdup` if there is no `deleted` suffix? Alternatively, we can use a stack-allocated buffer.

src/symbols_linux.cpp
```diff
@@ -776,29 +806,13 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
         SharedLibrary& lib = it.second;
-        CodeCache* cc = new CodeCache(lib.file, array->count(), false, lib.map_start, lib.map_end);
-
-        // Strip " (deleted)" suffix so that removed library can be reopened
-        size_t len = strlen(lib.file);
-        if (len > 10 && strcmp(lib.file + len - 10, " (deleted)") == 0) {
-            lib.file[len - 10] = 0;
-        }
+        CodeCache* cc = new CodeCache(lib.file, array->count(), false, lib.map_start, lib.map_end, lib.image_base);
+        stripDeletedSuffix(lib.file);
```
apangin, line 815: If the original file has been deleted, I don't think we should attempt to parse a file with the similar name, because it can be a different file with different symbol layout.

src/codeCache.h
apangin: Yes, it will be helpful to let users know that the orignal file has been deleted and therefore symbol infromation can be incomplete.

src/symbols_linux.cpp
fandreuz: 7a1895a034e241192175b1b5be90c70b83118df6

src/codeCache.cpp
```diff
@@ -29,12 +29,18 @@ size_t NativeFunc::usedMemory(const char* name) {
 
 
 CodeCache::CodeCache(const char* name, short lib_index, bool imports_patchable,
-                     const void* min_address, const void* max_address) {
+                     const void* min_address, const void* max_address,
+                     const char* image_base) {
     _name = NativeFunc::create(name, -1);
+
+    // Strip " (deleted)" suffix so that removed library can be reopened
+    size_t len = strlen(name);
```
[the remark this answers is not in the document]
fandreuz: 1d049c5937847ebdcf460693cf6926cf52ad087e

src/symbols_linux.cpp
fandreuz: I'm not sure I understand, this seems to be replicating the original logic. Do you propose to skip parsing the library if the original file was deleted?

src/symbols_linux.cpp
Baraa-Hasheesh: @apangin technically that's true but it will require that the library is reloaded into a different location in memory (could happen in JIT from what I'm reading)

But this raises the concern that the library will not be patched (skipped) or loose it's patching during these long profile sessions (profiling accuracy degrades with time) while being a valid behavior of the profiled process (Different issue that can be tracked separately)

Baraa-Hasheesh: Just note that this is only valid on the patching side, while the mentioned case [here](https://github.com/async-profiler/async-profiler/pull/1264#discussion_r2079211092) will still be possible if lib was relocated after it was detected but before it was parsed

I think it's best if we have the dladdr side on the patching side as it should also protect MacOs

For the parsing side, for now I think it should be an acceptable risk & we can create a different thread to track & reproduce the possible crash

fandreuz: Bara's fix was implemented in 42879e3426bbe39da44220a70ef718264aa5c09c

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -209,8 +209,20 @@ jobs:
...
+        run: |
+          case "${{ matrix.runson.name }}" in
+            macos*)
+              sudo sysctl -w kern.corefile=core.%P
+              sudo sysctl kern.coredump=1
+            ;;
+            *)
+              echo "$GITHUB_WORKSPACE/core" | sudo tee /proc/sys/kernel/core_pattern
+            ;;
+          esac
```
apangin, line 222: Please exclude non-relevant changes from this PR.

fandreuz: I added that for debugging: ffdeb52faf34b3c5ca8cc2f07443dd01183a7fb5

src/symbols_linux.cpp
```diff
@@ -64,6 +67,25 @@ static void applyPatch(CodeCache* cc) {}
...
+static const void* _main_phdr = getMainPhdr();
+static const char* _ld_base = (const char*)getauxval(AT_BASE);
+
+static bool isMainExecutable(const char* image_base, const void* map_end) {
+    return _main_phdr != NULL && _main_phdr >= image_base && _main_phdr < map_end;
+}
+
+static bool isLoader(const char* image_base) {
+    return _ld_base != NULL && _ld_base == image_base;
+}
```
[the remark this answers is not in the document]
fandreuz: 01fc8ebff74cc91550274af6760c683f10692919

src/symbols_linux.cpp
```diff
@@ -845,4 +840,48 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+    _valid = false;
+
+    if (OS::isMusl() || isMainExecutable(cc->imageBase(), cc->maxAddress()) || isLoader(cc->imageBase())) {
+        _valid = true;
+        return;
+    }
+
+    const char* stripped_name;
+    size_t name_len = strlen(cc->name());
+    if (hasDeletedSuffix(cc->name(), name_len)) {
```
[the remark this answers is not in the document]
fandreuz: 01fc8ebff74cc91550274af6760c683f10692919

src/symbols_linux.cpp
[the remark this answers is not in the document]
fandreuz: Ah I see. 017ff0d7915d9c5cb805c550433312eb2a5c7ca6

I'll keep the check above because `_main_phdr` might be `NULL`, and if `image_base` is also `NULL` we want to return `false` I guess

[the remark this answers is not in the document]
apangin: `image_base` should never be NULL, is it?
But I won't insist on removing the check, there is nothing wrong with it.

src/symbols_linux.cpp
```diff
@@ -845,4 +836,48 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+    if (handle == NULL || dlinfo(handle, RTLD_DI_LINKMAP, &map) != 0 || dladdr(map->l_ld, &dl_info) == 0) {
+        return false;
+    }
+
+    return cc->imageBase() == (const char*)dl_info.dli_fbase;
+}
+
+UnloadProtection::UnloadProtection(const CodeCache *cc) {
+    _lib_handle = NULL;
+    _valid = false;
```
apangin, line 853: These are overwritten below. Perhaps initialize them once where needed?

src/symbols_linux.cpp
```diff
@@ -845,4 +836,48 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+        buf[name_len - 10] = 0;
+        stripped_name = buf;
+    } else {
+        stripped_name = cc->name();
+    }
+
+    // Protect library from unloading while parsing in-memory ELF program headers.
+    // Also, dlopen() ensures the library is fully loaded.
+    _lib_handle = dlopen(stripped_name, RTLD_LAZY | RTLD_NOLOAD);
+    _valid = isValidHandle(cc, _lib_handle);
```
apangin, line 874: Not the best name, IMO.
Any non-null handle is valid, but what we really check here is whether it is the same shared library that was seen at CodeCache creation time.

src/symbols_linux.cpp
```diff
@@ -845,4 +836,48 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+    const char* stripped_name;
+    size_t name_len = strlen(cc->name());
+    if (name_len > 10 && strcmp(cc->name() + name_len - 10, " (deleted)") == 0) {
+        char* buf = (char*) alloca(name_len);
+        memcpy(buf, cc->name(), name_len);
+        buf[name_len - 10] = 0;
+        stripped_name = buf;
+    } else {
+        stripped_name = cc->name();
+    }
```
apangin, line 869: I'm trying to write the same in a shorter way. WDYT about the following?
```c
    const char* stripped_name = cc->name();
    size_t name_len = strlen(stripped_name);
    if (name_len > 10 && strcmp(stripped_name + name_len - 10, " (deleted)") == 0) {
        char* buf = (char*) alloca(name_len);
        *stpncpy(buf, stripped_name, name_len - 10) = 0;
        stripped_name = buf;
    }
```

fandreuz: Looks good, `buf` can also be `name_len - 9`

src/symbols_linux.cpp
fandreuz: - `checkHandleConsistency`
- `isHandleConsistent`

I'll think about it, let me know if you have any preference

src/symbols_linux.cpp
fandreuz: 9ae2c0f512356df2c4441841aaf3a28a010f088f

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,76 @@
...
+void outputCallback(const char* buffer, size_t size) {
+    fwrite(buffer, sizeof(char), size, stdout);
+}
+
+/*
+Idea of the test (the behavior applies as of 0c72a8d):
+- We load libcallsmalloc.so
+- We start AP without nativemem mode
+    => 'malloc' is not hooked
+- We stop AP
```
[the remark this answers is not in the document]
fandreuz: Right, thanks: d93698b4ab897a269a7b4dcd44922b8f7a27afd0

test/test/nativemem/nativemem_known_lib_crash.c
```diff
@@ -0,0 +1,76 @@
...
+    asprof_execute_t asprof_execute = (asprof_execute_t)dlsym(libprof, "asprof_execute");
+    ASSERT_NO_DLERROR(asprof_execute);
+
+    asprof_error_str_t asprof_error_str = (asprof_error_str_t)dlsym(libprof, "asprof_error_str");
+    ASSERT_NO_DLERROR(asprof_error_str);
+
+    void* lib = dlopen("libcallsmalloc.so", RTLD_NOW | RTLD_GLOBAL);
+    ASSERT_NO_DLERROR(lib);
+
+    asprof_error_t asprof_err = asprof_execute("start,cstack=dwarf,collapsed", outputCallback);
```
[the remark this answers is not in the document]
fandreuz: Not needed indeed: 7acb726d93cbbcabadcff7bd903ce6f1d111a494

src/symbols_linux.cpp
fandreuz: 41d89d1a4c6b4a80b4f2a2591d051b39eea72a84

apangin:
100 commits, 140 comments, 5 contributors. That was an epic PR. Thanks everyone involved!

Issue #1266, opened 2025-04-25
async-profiler incorrect installation of hooks


Baraa-Hasheesh:
### Describe the bug

It's possible for async-profiler to incorrectly install hooks while doing it's profiling
This leads the profiler to call the wrong method which changes the behaviour of the program when compared to running with & without profiler

This leads to having different outputs for certain method calls depending if profiler is being used or not

### Expected vs. actual behavior

profiler shouldn't change the fundamental behaviour of the program being profiled  

### Reproduction Steps

You can find the test case here https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/malloc 

You just need to do the following
* set PATH_TO_ASYNC_PROFILER_LIB_DIR to point to the directory that contains the profiler `so`
* execute `make`
* execute `./runner.sh`
* execute `./runner.sh 1`

You can observe that malloc behavior changes here

### Additional Information/Context

In the mentioned test case originally the malloc will be called from the preloaded Library

However when profiler is started it incorrectly redirect malloc to the original malloc call rather than the custom malloc declaration in the preloaded Library

### Async-profiler version

latest

apangin:
> profiler shouldn't change the fundamental behaviour of the program being profiled

Can you add more specifics? What is the original "fundamental" behavior, and how does async-profiler change it?

Baraa-Hasheesh:
@apangin 
Originally the malloc will be called from the preloaded Library 

However when profiler is started it incorrectly redirect malloc to the original malloc call rather than the custom malloc declaration in the preloaded Library

apangin:
Got it, thanks. That's what should be in the bug description then. Current description is too vague.

Baraa-Hasheesh:
Fixed in #1269

Baraa-Hasheesh:
Fixed in #1269

PR #1269, opened 2025-04-28
Fix memory hook installation

Baraa-Hasheesh:
### Description
async-profiler patches dynamic references to memory allocation calls to monitor these calls to help detect possible memory leaks.
However it's possible to create custom memory allocation functions to replace the existing ones.

```
#include <dlfcn.h>
#include "stdlib.h"
#include "string.h"

void* malloc(size_t size) {
    void* ptr = calloc(1, size);
    memset(ptr, 0xFF, size);
    return ptr;
}
```

If this wrapper is defined within the scope of the main executable no problems will occur as the async-profiler only patches dynamic references as such the `malloc` won't be patched while the internal `calloc` would be patched, so everything will work as expected.

However if the above wrapper was preloaded into the executable via `LD_PRELOAD` then the dynamic reference of the main will originally resolve to the custom `malloc`, however when async-profiler is started it will resolve the `malloc` via a `dlsym` call which returns the original reference of `malloc` rather than the custom declaration.

This in turn means the async profiler will potentially change the behaviour (which methods are called) of a running program if it's used to memory profile the executable which is risky & could potentially lead to crashes & undefined behaviour.

### Related issues
#1266

### Motivation and context
Avoid changing the behaviour of running processes when attaching the profiler to them.

### How has this been tested?
make test
=> A new test case has been added to check possible combination of the async-profiler loading with the library 

---


test/native/libs/mallocwrapper.c
```diff
@@ -0,0 +1,15 @@
...
+ #include <stdlib.h>
+ #include <string.h>
+
+ __attribute__((visibility("default")))
+ void* malloc(size_t size) {
+     void* ptr = calloc(1, size);
+     memset(ptr, 0xFF, size);
+     return ptr;
+ }
+
```
apangin, line 15: There should be no empty lines in the end, but the last line should terminate with EOL character (check other files too).

test/test/nativemem/malloc_wrapper.c
```diff
@@ -0,0 +1,96 @@
...
+    _asprof_execute = (asprof_execute_t)getSymbol(libprof, "asprof_execute");
+    _asprof_error_str = (asprof_error_str_t)getSymbol(libprof, "asprof_error_str");
+}
+
+void preloadOrderTest() {
+    char* ptr = (char*)malloc(1999993);
+
+    for (int i = 0; i < 1999993; i++) {
+        // On Linux ARMx64 the char is dereferenced differently (-1 = 0xFF = 255 unsigned)
+        if (*ptr != -1 && *ptr != 255) {
```
apangin, line 62: Use `unsigned char` then.

src/mallocTracer.cpp
```diff
@@ -154,6 +121,11 @@ void MallocTracer::patchLibraries() {
     while (_patched_libs < native_lib_count) {
         CodeCache* cc = (*native_libs)[_patched_libs++];
 
+        if (cc->contains((const void*)MallocTracer::initialize)) {
+            // Let libasyncProfiler always use original allocation methods
```
apangin, line 125: Async-profiler will not be able to trace its own allocations. That's unfortunate, although acceptable.

src/mallocTracer.cpp
```diff
@@ -13,55 +13,22 @@
...
-    _orig_calloc = ADDRESS_OF(calloc);
-    _orig_realloc = ADDRESS_OF(realloc);
-    _orig_free = ADDRESS_OF(free);
-    _orig_posix_memalign = ADDRESS_OF(posix_memalign);
-    _orig_aligned_alloc = ADDRESS_OF(aligned_alloc);
-}
-
 extern "C" void* malloc_hook(size_t size) {
-    void* ret = _orig_malloc(size);
+    void* ret = malloc(size);
```
apangin, line 23: Have you tested your solution on Alpine Linux and on macOS?

Baraa-Hasheesh: The fix is working on alpine, however there are some problems in the created test 
I will update the test to work on alpine

For MacOs it's not relevant as nativememory profiling isn't supported on MacOs

test/test/nativemem/malloc_wrapper.c
Baraa-Hasheesh: handled here [421d617](https://github.com/async-profiler/async-profiler/pull/1269/commits/421d617c81687dc2cd640a0ecc3a26c46385272d)

src/mallocTracer.cpp
Baraa-Hasheesh: tested & fixed here [22e8737](https://github.com/async-profiler/async-profiler/pull/1269/commits/22e87370c74621974a83c0e88215b47614776f17) 

Using calloc inside malloc was causing a stack overflow in alpine as calloc internally call malloc :) 
(malloc -> calloc -> malloc -> calloc .....)

Baraa-Hasheesh:
@apangin code updated & tests pass locally on alpine

src/mallocTracer.cpp
apangin: > nativememory profiling isn't supported on MacOs

It _is_ supported to some extent. At least, allocations in JDK should be handled correctly.
My ask here is to verify that it did not get worse with this PR.

test/native/libs/mallocwrapper.c
```diff
@@ -0,0 +1,21 @@
...
+#include <string.h>
+#include <dlfcn.h>
+
+typedef void* (*malloc_t)(size_t);
+static malloc_t _orig_malloc = NULL;
+
+__attribute__((visibility("default")))
+void* malloc(size_t size) {
+    if (_orig_malloc == NULL) {
+        _orig_malloc = (malloc_t) dlsym(((void *) -1l), "malloc");
```
apangin, line 16: What is `(void *) -1l`? Why not a named constant?

Baraa-Hasheesh: It's the value of RLTD_NEXT

The compile command used to compile the test isn't making the definitions needed to bundle it

Baraa-Hasheesh: I will check & update the compile command as needed

apangin: Yes, `_GNU_SOURCE` must be defined to use this constant.

src/mallocTracer.cpp
Baraa-Hasheesh: @apangin I've confirmed that the java libs are still being patched

Just as a note here, the patching logic is still the same as before, the only real change we created is that we used to call `malloc` using it's evaluated address via `dlsym` & now we just let the compiler & linker do that for us which less error prone

test/native/libs/mallocwrapper.c
Baraa-Hasheesh: solved here [fb003af](https://github.com/async-profiler/async-profiler/pull/1269/commits/fb003af766fa7c1fe6601925061004d27e95d67c)

test/test/nativemem/malloc_wrapper.c
```diff
@@ -0,0 +1,95 @@
...
+
+    _asprof_execute = (asprof_execute_t)getSymbol(libprof, "asprof_execute");
+    _asprof_error_str = (asprof_error_str_t)getSymbol(libprof, "asprof_error_str");
+}
+
+void preloadOrderTest() {
+    unsigned char* ptr = (unsigned char*)malloc(1999993);
+
+    for (int i = 0; i < 1999993; i++) {
+        if (*ptr != 255) {
```
apangin, line 61: You check the same byte every time.

apangin: There is also a slight inconsistency between what you set (0xFF) and what you test (255). Try keeping the same style (async-profiler typically uses lowercase for hex numbers).

test/test/nativemem/malloc_wrapper.c
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+    free(ptr);
+}
+
+void apiTest(char* filename) {
+    initAsyncProfiler();
+    
+    char start_cmd[2048] = {0};
+    snprintf(start_cmd, sizeof(start_cmd), "start,nativemem,cstack=dwarf,file=%s", filename);
```
apangin, line 73: Is `cstack=dwarf` essential here? Async-profiler does not supported it on all platforms (e.g. PPC64), so it's preferable not to use this option unnecessarily in order to avoid excluding test on these platforms.

test/test/nativemem/malloc_wrapper.c
```diff
@@ -0,0 +1,95 @@
+/*
```
apangin, line 1: Two similarly named files: `malloc_wrapper.c` and `mallocwrapper.c`
This is confusing, not to say that "wrapper" is a weasel word.

Baraa-Hasheesh: renamed here [f270377](https://github.com/async-profiler/async-profiler/pull/1269/commits/f27037757764c5542d5b193629124ceebff82882)

test/test/nativemem/malloc_wrapper.c
Baraa-Hasheesh: remvoed [f270377](https://github.com/async-profiler/async-profiler/pull/1269/commits/f27037757764c5542d5b193629124ceebff82882)

test/test/nativemem/malloc_wrapper.c
Baraa-Hasheesh: updated [f270377](https://github.com/async-profiler/async-profiler/pull/1269/commits/f27037757764c5542d5b193629124ceebff82882) 

Thanks for the catch about pointer value check :)

Makefile
```diff
@@ -214,6 +214,7 @@ ifeq ($(OS_TAG),linux)
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libreladyn.$(SOEXT) test/native/libs/reladyn.c
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libcallsmalloc.$(SOEXT) test/native/libs/callsmalloc.c
 	$(CC) -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjnimalloc.$(SOEXT) test/native/libs/jnimalloc.c
+	$(CXX) -shared -fPIC -o $(TEST_LIB_DIR)/libmalloc.$(SOEXT) test/native/libs/malloc.cpp
```
apangin, line 217: This simple file does not use any of C++ features, let's compile it as plain C.
To make `RTLD_NEXT` declaration available, simply add `#define _GNU_SOURCE` before includes.

test/test/nativemem/NativememTests.java
```diff
@@ -188,4 +188,15 @@ public void dlopenCustomLib(TestProcess p) throws Exception {
 
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_SIZE, 0L), 1);
     }
+
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=\"%lib %testlib/libmalloc.so\" ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_reference preload %f.jfr",
+            output = true, env = {"LD_LIBRARY_PATH=build/lib"}, nameSuffix="LD_PRELOAD+profiler_first")
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=\"%testlib/libmalloc.so %lib\" ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_reference preload %f.jfr",
+            output = true, env = {"LD_LIBRARY_PATH=build/lib"}, nameSuffix="LD_PRELOAD+profiler_second")
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=%testlib/libmalloc.so %testbin/malloc_reference api %f.jfr", output = true, env = {"LD_LIBRARY_PATH=build/lib"}, nameSuffix="api_test")
+    public void mallocWrapper(TestProcess p) throws Exception {
```
apangin, line 197: Let's call it `preloadMalloc` as it will better reflect the essence of the test. Similarly, `malloc_reference.c` can be renamed to `preload_malloc.c`

test/test/nativemem/NativememTests.java
```diff
@@ -188,4 +188,15 @@ public void dlopenCustomLib(TestProcess p) throws Exception {
 
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_SIZE, 0L), 1);
     }
+
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=\"%lib %testlib/libmalloc.so\" ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_reference preload %f.jfr",
+            output = true, env = {"LD_LIBRARY_PATH=build/lib"}, nameSuffix="LD_PRELOAD+profiler_first")
```
apangin, line 193: `output = true` is typically set when a test consumes output of the target program, which does not seem to be the case here.

test/test/nativemem/NativememTests.java
```diff
@@ -188,4 +188,15 @@ public void dlopenCustomLib(TestProcess p) throws Exception {
...
+    @Test(os = Os.LINUX, sh = "LD_PRELOAD=\"%testlib/libmalloc.so %lib\" ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_reference preload %f.jfr",
+            output = true, env = {"LD_LIBRARY_PATH=build/lib"}, nameSuffix="LD_PRELOAD+profiler_second")
```
apangin, line 195: Spaces are missing: `nameSuffix_=_`

test/test/nativemem/malloc_reference.c
```diff
@@ -0,0 +1,96 @@
...
+    _asprof_init = (asprof_init_t)getSymbol(libprof, "asprof_init");
+    _asprof_init();
+
+    _asprof_execute = (asprof_execute_t)getSymbol(libprof, "asprof_execute");
+    _asprof_error_str = (asprof_error_str_t)getSymbol(libprof, "asprof_error_str");
+}
+
+void preloadOrderTest() {
+    unsigned char* ptr = (unsigned char*)malloc(1999993);
+    int size = 1999993 / sizeof(unsigned char);
```
apangin, line 59: Please don't. Size of `unsigned char` is known to be 1.

test/test/nativemem/NativememTests.java
Baraa-Hasheesh: applied [c8f2638](https://github.com/async-profiler/async-profiler/pull/1269/commits/c8f263866128472f79184e46f28c4b2e21907a84)

test/test/nativemem/NativememTests.java
Baraa-Hasheesh: removed [c8f2638](https://github.com/async-profiler/async-profiler/pull/1269/commits/c8f263866128472f79184e46f28c4b2e21907a84)

test/test/nativemem/malloc_reference.c
Baraa-Hasheesh: applied [c8f2638](https://github.com/async-profiler/async-profiler/pull/1269/commits/c8f263866128472f79184e46f28c4b2e21907a84)

apangin:
Thanks! Merged.

PR #1270, opened 2025-04-28
Mention JVM flags in "Getting Started"

parttimenerd:
### Description

Mention the `+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints` in the "Getting Started" section of the documentation.

### Motivation and context

These are important options to improve accuracy (and reduce safepoint bias), but they are currently hidden in
the Troubleshooting section of the documentation, where people are unlikely to look.

---

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -23,6 +23,7 @@ For example, to list all Java process IDs with their full command lines, run
 async-profiler works in the context of the target Java application,
 i.e. it runs as an agent in the process being profiled.
 `asprof` is a tool to attach and control the agent.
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for your PR. Getting started guide has "Before profiling" section in the beginning, would not it be a better place for this notice? Also, a message should be clear on recommending `DebugNonSafepoints` for `asprof` use case, whereas this flag is redundant when using `-agentpath`.

parttimenerd:
You're correct.

parttimenerd:
I tried to phrase such a sentence, but I ran in to the problem that the reader doesn't know that async-profiler is in agent and that it can be attached in two ways.

PR #1274, opened 2025-04-30
Add corretto-8 to test matrix 

### LINKED ISSUE #1213: JDK8 isn't being covered by GHA

fandreuz:
### Description
In this PR I extend the test matrix to run Java integration tests on JDK 8.

### Related issues
Fixes #1213 

### Motivation and context
We should test JDK 8 as well.

### How has this been tested?
GHA

---


test/test/jfr/JfrTests.java
```diff
@@ -86,7 +86,7 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
      */
     @Test(mainClass = Ttsp.class)
     public void ttsp(TestProcess p) throws Exception {
-        p.profile("-d 3 -i 1ms --ttsp -f %f.jfr");
+        p.profile("-d 3 -i 100ns --ttsp -f %f.jfr");
```
fandreuz, line 89: The test would consistently fail in GHA for JDK 8 only. I could not reproduce the failure locally, but I managed to get some more info with some additional logging.

In [JDK 8](https://github.com/user-attachments/files/19978169/9_Run.integration.tests.txt), the safepoint pauses are apparently very short in GHA:
```
         vmop                    [threads: total initially_running wait_to_block]    [time: spin block sync cleanup vmop] page_trap_count
6.371: ThreadDump                       [      11          1              1    ]      [     0     0     0     0     0    ]  1   
```
The numbers in `time:` are in milliseconds, thus apparently the VM spent very little time between `SafepointSynchronize::begin` and `RuntimeService::record_safepoint_synchronized`.

In [JDK11](https://github.com/user-attachments/files/19978179/9_Run.integration.tests.txt):
```
          vmop                            [ threads:    total initially_running wait_to_block ][ time:    spin   block    sync cleanup    vmop ] page_trap_count
   7.400: ThreadDump                      [                12                 1             1 ][          3067       0    3067       0       0 ]               1
```

Thus, the hypothesis here is that the sampling is not frequent enough to have a sample happen in a moment when the engine is enabled (i.e. during a safepoint pause). Indeed, I could see in the logs of a previous run that `signalHandler` aborted multiple times because the engine was not enabled. 

Increasing the sampling frequency fixes the problem.

roy-soumadipta: @fandreuz can you check the test by either enforcing parallel gc for both or G1GC for both?

fandreuz: Unfortunately forcing G1GC for JDK8 does not solve the problem, apparently short safepoint pauses are not related to the Parallel GC

test/test/recovery/RecoveryTests.java
```diff
@@ -43,7 +43,7 @@ public void numbers(TestProcess p) throws Exception {
         Assert.isGreater(out.ratio("Numbers.loop"), 0.8);
 
         out = p.profile("-d 3 -e cpu -o collapsed --safe-mode 31");
-        Assert.isGreater(out.ratio("unknown_Java"), 0.1);
+        Assert.isGreater(out.ratio("unknown_Java"), 0.05);
```
fandreuz, line 46: Does `0.1` has any specific meaning? The smallest value I've seen so far is `0.08`, but this feels like it should be a simple `> 0` unless we have a specific value in mind

test/test/jfr/JfrTests.java
apangin: Hmm. Perhaps, `String.indexOf` is not intrinsified. Can you double check with `-XX:+UnlockDiagnosticVMOptions -XX:+PrintCompilation -XX:+PrintInlining` ?

test/test/recovery/RecoveryTests.java
apangin: Can you show profile when the test fails?

fandreuz:
```
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 4
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 1
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg;java/lang/Integer.doubleValue 5
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg;java/lang/Float.doubleValue 1
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 53
[unknown_Java] 27
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg;java/lang/Integer.doubleValue 6
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg;java/lang/Long.doubleValue 23
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 54
test/recovery/Numbers.main;test/recovery/Numbers.loop 47
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 52
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg;java/lang/Long.doubleValue 15
test/recovery/Numbers.main;test/recovery/Numbers.loop 7
test/recovery/Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg 4
```

test/test/jfr/JfrTests.java
fandreuz: It looks like it is intrinsified:
```
% grep -n "intrinsic" jdk8.txt | grep "indexOf "
14415:2025-05-01T08:53:21.2738753Z                               @ 57   java.lang.String::indexOf (7 bytes)   (intrinsic)
15501:2025-05-01T08:53:25.2538126Z                               @ 57   java.lang.String::indexOf (7 bytes)   (intrinsic)

% grep -n "ttsp" jdk8.txt
14133:2025-05-01T08:53:21.1854780Z INFO: Running JfrTests.ttsp...
15215:2025-05-01T08:53:25.1679154Z PASS [28/74] JfrTests.ttsp took 3.982 s
```

fandreuz:
@apangin do you have any other input on the changes in this PR?

test/test/jfr/JfrTests.java
apangin: What is current status of this?
`100ns` is not a realistic profiling interval. Time-to-safepoint pauses need to be at least 1ms long, or they cannot be considered "pauses".

fandreuz: I'm not sure about how we should fix this, the numbers I got seem to suggest that the pause, as you said, is not really a pause.

But what are we really testing here? `--ttsp` is based on `--begin` and `--end`, how about we drop all Ttsp tests (which are flaky anyway) and we replace them with `begin=Timer::start_timer,end=Waiter::on_timer_end`, where `Timer` is some class we can control deterministically?

apangin: Testing `--ttsp` is essential - this is the feature customers are interested in. `begin`/`end` are not that useful on their own. E.g., if in some future JDK version safepoint functions are renamed, TTSP profiling will break, while an artificial begin/end test will keep passing.

fandreuz: 399d167

src/arguments.cpp
```diff
@@ -444,6 +449,13 @@ Error Arguments::parse(const char* args) {
             CASE("nostop")
                 _nostop = true;
 
+            CASE("ttsp")
```
apangin, line 452: Double `CASE("ttsp")`

fandreuz: 0b57c6cb7ddf57bb93feb22811d41cd96e25d421

test/test/jfr/JfrTests.java
```diff
@@ -88,10 +88,10 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
      * @param p The test process to profile with.
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,file=%f.jfr", os = Os.LINUX)
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,alloc=100,file=%f.jfr", os = Os.LINUX)
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,file=%f.jfr", os = Os.MACOS, arch = Arch.ARM64)
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,lock=10ms,file=%f.jfr", os = Os.MACOS, arch = Arch.ARM64)
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,file=%f.jfr", os = Os.LINUX, jvmVer = {11, Integer.MAX_VALUE})
```
apangin, line 91: Why 11+? Is it because of `live` profiling not supported on JDK 8?
I think `--all` should still work on JDK 8, although without live object profiling.

test/test/jfr/Ttsp.java
```diff
@@ -6,41 +6,51 @@
...
 
-    static void spoiler(int length, long count) {
-        for (long i = 0; i < count; i++) {
-            sink = indexOfTest(length);
+    static Unsafe getUnsafe() {
+        try {
+            Field f = Unsafe.class.getDeclaredField("theUnsafe");
+            f.setAccessible(true);
+            return (Unsafe) f.get(null);
+        } catch(Exception exception) {
```
apangin, line 26: Formatting

test/test/jfr/JfrTests.java
fandreuz: Yes, failure here: https://github.com/async-profiler/async-profiler/actions/runs/15116002786/job/42486783966#step:10:113

I'll fix the failure in this PR

test/test/jfr/Ttsp.java
fandreuz: 363e56d42d715a1937ef462c40f2adb3bc9da442

test/test/jfr/JfrTests.java
fandreuz: d3784df4b887503ef4baae1e5d0ae2d1cd4a336d

src/profiler.cpp
```diff
@@ -1214,6 +1214,10 @@ Error Profiler::start(Arguments& args, bool reset) {
     }
 
     if (_event_mask & EM_ALLOC) {
+        if (args._all && VM::hotspot_version() < 11) {
+            Log::debug("'live' option was quietly removed, only supported on OpenJDK 11+");
```
apangin, line 1218: This is not accurate: `_live` is also supported with OpenJ9.
Why not checking `_all` in `AllocTracer::check`?

test/test/jfr/JfrTests.java
```diff
@@ -195,4 +197,9 @@ private boolean containsSamplesOutsideWindow(TestProcess p) throws Exception {
             return entryEnd.isBefore(event.getStartTime());
         });
     }
+
+    private static boolean checkJdkVersionEarlierThan11() {
+        String javaVersion = System.getProperty("java.version");
+        return javaVersion.startsWith("1.") || Integer.parseInt(javaVersion.split("\\.")[0]) < 11;
+    }
```
apangin, line 204: This returns the version of the host JDK (test runner), which may differ from the JDK version used by the test itself.
You may find the version from the JFR recording.

src/allocTracer.cpp
```diff
@@ -71,6 +71,9 @@ void AllocTracer::recordAllocation(void* ucontext, EventType event_type, uintptr
 
 Error AllocTracer::check(Arguments& args) {
     if (args._live) {
+        // This engine is only going to be selected in Profiler::selectAllocEngine
+        // when can_generate_sampled_object_alloc_events is not available, i.e.
+        // JDK<11.
```
apangin, line 76: Just came across this yesterday: https://lkml.org/lkml/2020/5/29/1038
I'm fully with Linus on this issue.

fandreuz: 045908b9001ccfd87ecc3e55ce82136f0cb5eabb

apangin: From one extreme to another :)
There are no strict requirements how long a line should be. Just keep it in a shape that looks fine and reads fine.
```
        // This engine is only going to be selected in Profiler::selectAllocEngine
        // when can_generate_sampled_object_alloc_events is not available, i.e. JDK<11.
```

test/test/jfr/JfrTests.java
fandreuz: 790b08607ecee5f13fdc4a5bc63a2b3516208f5b

src/allocTracer.cpp
fandreuz: e42181b113f763ce3a1c6104e0abde808f93059a

Issue #1193, opened 2025-03-25
Incomplete Symbol parsing in MacOs


Baraa-Hasheesh:
### Describe the bug

While working on the #1174 issue it was observed that the Async profiler doesn't detect all JVM symbols while parsing the JVM/Java library on the MacOs (symbols_macos.cpp).

For example the `JVM_StartThread` symbol was never detected in the calls made to the `CodeCache::addImport` function call on MacOs while it was detected normally on Linux.

### Expected vs. actual behavior

Expected: Symbol parsing on MacOs to detect all symbols similar to Linux.
Actual: Some symbols aren't being detected during the loading of Libs on MacOs

### Reproduction Steps

To reproduce the issue you can enable the `asyncProfilerFirstNoneNativeAppTest` test on MacOs & observe the parsed symbols via a simple logging statement on the `CodeCache::addImport` function

You can find the test here #1192

### Additional Information/Context

You may use the changes here #1192 to observe the issue more easily 

### Async-profiler version

latest

Baraa-Hasheesh:
Fixed here #1279

PR #1278, opened 2025-05-02
Improve symbol parsing on MacOs & Basic Native memory profiling support

Baraa-Hasheesh:
### Description
Improving Macos symbol parsing to better detect dynamic symbols & patch them more effectively 
As well as basic native memory profiling support on MacOs

### Related issues
#1193

### Motivation and context
MacOs Symbol parsing was missing many symbols which was causing problems when trying to make improvements on Linux side.

By making the Symbol parsing to be more inline with Linux it makes it easier to make improvements that effect both platforms 

### How has this been tested?
Nativemem tests have been enabled on MacOs
`make test`

---

### DIFF
--- a/src/symbols_macos.cpp
+++ b/src/symbols_macos.cpp
@@ -14,11 +14,15 @@
 #include "symbols.h"
 #include "log.h"
 
+#ifndef SEG_DATA_CONST
+#define SEG_DATA_CONST  "__DATA_CONST"
+#endif
 
 class MachOParser {
   private:
     CodeCache* _cc;
     const mach_header* _image_base;
+    intptr_t _slide;
 
     static const char* add(const void* base, uint64_t offset) {
         return (const char*)base + offset;
@@ -27,7 +31,7 @@ class MachOParser {
     const section_64* findSymbolPtrSection(const segment_command_64* sc) {
         const section_64* section = (const section_64*)add(sc, sizeof(segment_command_64));
         for (uint32_t i = 0; i < sc->nsects; i++) {
-            if (strcmp(section->sectname, "__la_symbol_ptr") == 0) {
+            if ((section->flags & SECTION_TYPE) == S_LAZY_SYMBOL_POINTERS || (section->flags & SECTION_TYPE) == S_NON_LAZY_SYMBOL_POINTERS) {
                 return section;
             }
             section++;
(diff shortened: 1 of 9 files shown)


src/symbols_macos.cpp
```diff
@@ -13,26 +13,30 @@
...
     CodeCache* _cc;
     const mach_header* _image_base;
+    intptr_t _slide;
 
     static const char* add(const void* base, uint64_t offset) {
         return (const char*)base + offset;
     }
 
-    const section_64* findSymbolPtrSection(const segment_command_64* sc) {
+    void findSymbolPtrSection(const segment_command_64* sc, std::vector<const section_64*>* symbol_sections) {
```
fandreuz, line 32: Is there a reason to prefer a ptr to `std::vector` over a reference?

src/symbols_macos.cpp
```diff
@@ -54,24 +58,27 @@ class MachOParser {
         _cc->setDebugSymbols(debug_symbols);
     }
 
-    void loadImports(const symtab_command* symtab, const dysymtab_command* dysymtab,
-                     const section_64* la_symbol_ptr, const char* link_base) {
-        const nlist_64* sym = (const nlist_64*)add(link_base, symtab->symoff);
-        const char* str_table = add(link_base, symtab->stroff);
+    void loadImports(const section_64* section, intptr_t slide, nlist_64 *symtab, char *strtab, uint32_t *indirect_symtab) {
```
fandreuz, line 61: `strtab` and `indirect_symtab` could be const

fandreuz: and `symtab` as well

src/symbols_macos.cpp
fandreuz: `std::vector<const section_64*>& symbol_sections`

src/symbols_macos.cpp
```diff
@@ -54,24 +58,27 @@ class MachOParser {
...
-        uint32_t isym_count = la_symbol_ptr->size / sizeof(void*);
-        void** slot = (void**)add(_image_base, la_symbol_ptr->addr);
+        for (uint64_t i = 0; i < section->size / sizeof(void *); i++) {
+            uint32_t symtab_index = indirect_symbol_indices[i];
+            if (symtab_index == INDIRECT_SYMBOL_ABS || symtab_index == INDIRECT_SYMBOL_LOCAL || symtab_index == (INDIRECT_SYMBOL_LOCAL | INDIRECT_SYMBOL_ABS)){
+                continue;
+            }
+            uint32_t strtab_offset = symtab[symtab_index].n_un.n_strx;
+            char *symbol_name = strtab + strtab_offset;
+            bool symbol_name_longer_than_1 = symbol_name[0] && symbol_name[1];
```
fandreuz, line 72: Can we have something like `symbol_name[1] == '\0'` instead?

src/symbols_macos.cpp
```diff
@@ -85,40 +92,54 @@ class MachOParser {
...
             lc = (const load_command*)add(lc, lc->cmdsize);
         }
 
+        // Parse Debug symbols
+        if (text_base != UNDEFINED && link_base != UNDEFINED) {
+            loadSymbols(symtab_cmd, text_base, link_base);
+        }
+
+        // Can't load imports from library
+        if (!symtab_cmd || !dysymtab_cmd || !linkedit_segment || !dysymtab_cmd->nindirectsyms || symbol_sections.size() == 0) {
```
fandreuz, line 128:
```suggestion
        if (!symtab_cmd || !dysymtab_cmd || !linkedit_segment || !dysymtab_cmd->nindirectsyms || symbol_sections.empty()) {
```

PR #1286, opened 2025-05-08
Add cpu sampling

JugadK:
### Description
Adds a "cpu" virtual frame at the bottom of each stack when one passes in --record-cpu as a flag

### How has this been tested?

---

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -110,6 +110,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     reverse          - generate stack-reversed FlameGraph / Call tree (defaults to icicle graph)
 //     inverted         - toggles the layout for reversed stacktraces from icicle to flamegraph
 //                        and for default stacktraces from flamegraph to icicle
+//     record-cpu       - record which cpu a sample was taken on (default: false)
 //
 // It is possible to specify multiple dump options at the same time
 
@@ -432,6 +433,9 @@ Error Arguments::parse(const char* args) {
 
             CASE("inverted")
                 _inverted = true;
+
+            CASE("record-cpu")
+                _record_cpu = true;
 
             DEFAULT()
                 if (_unknown_arg == NULL) _unknown_arg = arg;
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -301,6 +301,12 @@ const char* FrameName::name(ASGCT_CallFrame& frame, bool for_matching) {
         case BCI_ERROR:
             return _str.assign("[").append((const char*)frame.method_id).append("]").c_str();
 
+        case BCI_CPU: {
+            int cpu = (int)(uintptr_t)frame.method_id;
+            char buf[sizeof(int)];
+            snprintf(buf, sizeof(buf), "%d", cpu);
+            return _str.assign("[cpu").append(buf).append("]").c_str();
+        }
         default: {
             const char* type_suffix = typeSuffix(FrameType::decode(frame.bci));
 
@@ -354,6 +360,7 @@ FrameTypeId FrameName::type(ASGCT_CallFrame& frame) {
         case BCI_THREAD_ID:
         case BCI_ADDRESS:
         case BCI_ERROR:
+        case BCI_CPU:
             return FRAME_NATIVE;
 
         default:
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -531,6 +531,7 @@ PerfEvent* PerfEvents::_events = NULL;
 PerfEventType* PerfEvents::_event_type = NULL;
 bool PerfEvents::_alluser;
 bool PerfEvents::_kernel_stack;
+bool PerfEvents::_record_cpu;
 int PerfEvents::_target_cpu;
 
 int PerfEvents::createForThread(int tid) {
@@ -582,13 +583,17 @@ int PerfEvents::createForThread(int tid) {
 
 #ifdef PERF_ATTR_SIZE_VER5
     if (_cstack == CSTACK_LBR) {
-        attr.sample_type |= PERF_SAMPLE_BRANCH_STACK | PERF_SAMPLE_REGS_USER;
+        attr.sample_type |= PERF_SAMPLE_BRANCH_STACK | PERF_SAMPLE_REGS_USER | PERF_SAMPLE_CPU;
         attr.branch_sample_type = PERF_SAMPLE_BRANCH_USER | PERF_SAMPLE_BRANCH_CALL_STACK;
         attr.sample_regs_user = 1ULL << PERF_REG_PC;
     }
 #else
 #warning "Compiling without LBR support. Kernel headers 4.1+ required"
 #endif
+
+    if (_record_cpu) {
+        attr.sample_type |= PERF_SAMPLE_CPU;
+    }
 
     int fd;
     if (FdTransferClient::hasPeer()) {
@@ -797,6 +802,7 @@ Error PerfEvents::start(Arguments& args) {
     }
 
     _target_cpu = args._target_cpu;
+    _record_cpu = args._record_cpu;
 
     if (args._interval < 0) {
         return Error("interval must be positive");
@@ -880,7 +886,14 @@ int PerfEvents::walk(int tid, void* ucontext, const void** callchain, int max_de
 
         while (tail < head) {
             struct perf_event_header* hdr = ring.seek(tail);
+
             if (hdr->type == PERF_RECORD_SAMPLE) {
+
+                if (_record_cpu) {
+                  u64 cpu = ring.next();
+                  java_ctx->cpu = cpu+1;
+                }
+
                 u64 nr = ring.next();
                 while (nr-- > 0) {
                     u64 ip = ring.next();
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -698,6 +698,11 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     if (_add_sched_frame) {
         num_frames += makeFrame(frames + num_frames, BCI_ERROR, OS::schedPolicy(0));
     }
+    if (_add_cpu_frame) {
+      // +1 is to prevent the method_id from being read as a null pointer
+      // when cpu = 0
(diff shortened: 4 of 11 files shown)


src/frameName.cpp
```diff
@@ -301,6 +301,12 @@ const char* FrameName::name(ASGCT_CallFrame& frame, bool for_matching) {
         case BCI_ERROR:
             return _str.assign("[").append((const char*)frame.method_id).append("]").c_str();
 
+        case BCI_CPU: {
+            int cpu = (int)(uintptr_t)frame.method_id;
+            char buf[32];
+            snprintf(buf, sizeof(buf), "%d", cpu-1);
```
JugadK, line 307: an alternative idea I had was to make the line in the beginning of this function

if (frame.method_id == NULL) {

be

if(frrame.bci != BCI_CPU && frame.method_id == NULL)

so it will skip this if method_id is zero, not familiar with the codebase though so not sure if that would have a bad side effect?

src/arguments.cpp
```diff
@@ -433,6 +434,9 @@ Error Arguments::parse(const char* args) {
             CASE("inverted")
                 _inverted = true;
 
+            CASE("record-cpu")
```
apangin, line 437: This is a section for FlameGraph options as the comment above states.
I'd put new option next to `sched`, which is kind of similar behavior-wise.

src/frameName.cpp
apangin: I'd set some bit instead - to make sure `method_id` is never NULL and to distinguish it from BCI_THREAD, which uses the same encoding.

src/frameName.cpp
```diff
@@ -301,6 +301,12 @@ const char* FrameName::name(ASGCT_CallFrame& frame, bool for_matching) {
...
+            return _str.assign("[cpu").append(buf).append("]").c_str();
```
apangin, line 308: Why building string part-by-part, if you can format the whole thing in the above `snprintf`?

src/perfEvents_linux.cpp
```diff
@@ -582,14 +583,18 @@ int PerfEvents::createForThread(int tid) {
 
 #ifdef PERF_ATTR_SIZE_VER5
     if (_cstack == CSTACK_LBR) {
-        attr.sample_type |= PERF_SAMPLE_BRANCH_STACK | PERF_SAMPLE_REGS_USER;
+        attr.sample_type |= PERF_SAMPLE_BRANCH_STACK | PERF_SAMPLE_REGS_USER | PERF_SAMPLE_CPU;
```
apangin, line 586: Have you checked this actually works? It seems that the field is always added, but the reader only parses it when `_record_cpu` is set.

src/profiler.cpp
```diff
@@ -698,6 +698,11 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     if (_add_sched_frame) {
         num_frames += makeFrame(frames + num_frames, BCI_ERROR, OS::schedPolicy(0));
     }
+    if (_add_cpu_frame) {
+      // +1 is to prevent the method_id from being read as a null pointer
+      // when cpu = 0
+      num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu + 1);
```
apangin, line 704: Please fix indentation.

src/stackWalker.h
```diff
@@ -16,11 +16,20 @@ struct StackContext {
     const void* pc;
     uintptr_t sp;
     uintptr_t fp;
+    u64 cpu;
```
apangin, line 19: Why did you change it to 64-bit?

src/perfEvents_linux.cpp
JugadK: ah this is from when i was hacking it together not sure why I left it in

src/stackWalker.h
JugadK: ring.next() returns u64 so i thought it was nicer to just keep it as the type that was originally returned. Though I don't think it makes much of a difference

apangin: ok, thanks for the explanation

src/frameName.cpp
JugadK: Like just set the first bit then unset it when we are reading it? that works

src/frameName.cpp
JugadK: originally I was using std::to_string() and appending, will change it to snprintf

src/frameName.cpp
JugadK: hm what if I shift left by one and set, then shift right when reading? don't have to make assumptions on the size of a pointer on the system

src/stackWalker.h
```diff
@@ -16,11 +16,20 @@ struct StackContext {
...
 
     void set(const void* pc, uintptr_t sp, uintptr_t fp) {
         this->pc = pc;
         this->sp = sp;
         this->fp = fp;
+        this->cpu = 0;
+    }
+
+    void set(const void* pc, uintptr_t sp, uintptr_t fp, u64 cpu) {
```
apangin, line 28: Where is this function used?

src/perfEvents_linux.cpp
```diff
@@ -880,7 +886,14 @@ int PerfEvents::walk(int tid, void* ucontext, const void** callchain, int max_de
 
         while (tail < head) {
             struct perf_event_header* hdr = ring.seek(tail);
+
             if (hdr->type == PERF_RECORD_SAMPLE) {
+
+                if (_record_cpu) {
+                  u64 cpu = ring.next();
+                  java_ctx->cpu = cpu;
```
apangin, line 894: Use 4 spaces indent.
IMO, local variable is redundant here, just set `java_ctx->cpu = ring.next();`

src/profiler.cpp
```diff
@@ -698,6 +698,9 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     if (_add_sched_frame) {
         num_frames += makeFrame(frames + num_frames, BCI_ERROR, OS::schedPolicy(0));
     }
+    if (_add_cpu_frame) {
+        num_frames += makeFrame(frames + num_frames, BCI_CPU, (java_ctx.cpu << 1) | 0x1);
```
apangin, line 702: I thought of setting a higher bit instead. In this case, there will be no need for shift, and frame values for BCI_CPU will not clash with values for BCI_THREAD.

src/profiler.cpp
```diff
@@ -1125,6 +1128,8 @@ Error Profiler::start(Arguments& args, bool reset) {
         _add_event_frame = args._output != OUTPUT_JFR;
         _add_thread_frame = args._threads && args._output != OUTPUT_JFR;
         _add_sched_frame = args._sched;
+        _add_cpu_frame = args._record_cpu;
```
apangin, line 1131: Since `record_cpu` is only supported with PerfEvents, do you want to return an error when `record_cpu` is requested for any other engine?

JugadK: yeah that sounds like a good idea

src/stackWalker.h
JugadK: It is not, since it seems like we have a function to set all fields in that class I thought it would make sense to add another function with the option of passing CPU for completion sake/future usage. Can remove if you feel its not needed

src/profiler.cpp
JugadK: I liked this approach as it prevents as errors from occurring from mistakes in casting, changes in intermediary datatypes and different pointer/int sizes on various platforms (i.e one of them being smaller than whatever nth bit we set) in practice just setting the 16th bit would probably have the same effect so I can do that

Im curious what you mean by "clashing". they are differentiated by the bci enum, but is there some problem with them being the same value in the method_id pointer? maybe I'm missing something.

src/stackWalker.h
```diff
@@ -16,11 +16,13 @@ struct StackContext {
     const void* pc;
     uintptr_t sp;
     uintptr_t fp;
+    u64 cpu;
 
     void set(const void* pc, uintptr_t sp, uintptr_t fp) {
         this->pc = pc;
         this->sp = sp;
         this->fp = fp;
+        this->cpu = 0;
```
apangin, line 25: This seems to overwrite `cpu` value set earlier. Did you test different stack walking modes (fp, dwarf, vm)?

src/profiler.cpp
apangin: The same value for method_id can be a problem as long as `MethodMap` uses `jmethodID` as a key.

src/arguments.cpp
```diff
@@ -87,6 +87,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     filter=FILTER    - thread filter
 //     threads          - profile different threads separately
 //     sched            - group threads by scheduling policy
+//     record-cpu       - record which cpu a sample was taken on (default: false)
```
apangin, line 90: Do you mind updating `ProfilingOptions.md` doc as well?

src/stackWalker.h
JugadK: hm yeah default,vm,vmx work but fp, dwarf do not, think i'll just remove setting it as 0

src/arguments.cpp
JugadK: sure

test/test/cpu/CpuTests.java
```diff
@@ -58,6 +58,13 @@ public void perfEventsTargetCpuEventsCount(TestProcess p) throws Exception {
         Output outRightCpu = p.profile("-d 2 -e cpu-clock -i 100ms --total -o collapsed --target-cpu 0");
         assertCloseTo(outRightCpu.total(), 2_000_000_000, "perf_events total should match profiling duration");
     }
+    
+    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
+    public void perfEventsRecordCpuEventsCount(TestProcess p) throws Exception {
+
+        Output outRightCpu = p.profile("-d 2 -e cpu-clock -i 100ms --total -o collapsed --record-cpu");
+        assertCloseTo(outRightCpu.total(), 2_000_000_000, "perf_events total should match profiling duration");
```
apangin, line 66: This does not actually test the feature.
I suggest you pin the process to CPU 1 and check that output contains `[CPU-1]`, but not `[CPU-0]`.

JugadK: ah thought you mean't by sanity test just something that sees if it does not crash, will do

apangin:
All good now. Thank you for the contribution!

PR #1290, opened 2025-05-08
Move out `Buffer` from `flightRecorder.cpp`

fandreuz:
### Description
In this PR I move `Buffer` out of `flightRecorder.cpp`.

### Related issues
#1188 

### Motivation and context
I plan to use `Buffer` as the base for a Protobuf writer.

### How has this been tested?
GHA

---

### DIFF
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -16,9 +16,12 @@ constexpr int MAX_STRING_LENGTH = 8191;
 class Buffer {
   private:
     std::size_t _offset;
-    char* _data;
+    char _data[0];
 
   protected:
+    Buffer() : _offset(0) {
+    }
+
     void set(char v, std::size_t idx) {
         _data[idx] = v;
     }
@@ -28,9 +31,6 @@ class Buffer {
     }
 
   public:
-    Buffer(char* data) : _offset(0), _data(data) {
-    }
-
     const char* data() const {
         return _data;
     }
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -11,7 +11,10 @@
 #include <cstring>
 #include <cstdint>
 
-constexpr int MAX_STRING_LENGTH = 8191;
+const std::size_t MAX_STRING_LENGTH = 8191;
+// https://github.com/openjdk/jmc/blob/master/core/org.openjdk.jmc.flightrecorder/src/main/java/org/openjdk/jmc/flightrecorder/internal/parser/v1/SeekableInputStream.java
+const char STRING_ENCODING_UTF8_BYTE_ARRAY = 3;
+const char STRING_ENCODING_LATIN1_BYTE_ARRAY = 5;
 
 class Buffer {
   private:
@@ -47,11 +50,6 @@ class Buffer {
 
     void reset() {
         _offset = 0;
-    }
-
-    void put(const char* v, std::size_t len) {
-        std::memcpy(_data + _offset, v, len);
-        _offset += (int)len;
     }
 
     void put8(char v) {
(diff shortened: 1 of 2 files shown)


krk (approved):

src/buffer.h
```diff
@@ -0,0 +1,129 @@
...
+        }
+        while (v > 0x7f) {
+            _data[_offset++] = (char)v | 0x80;
+            v >>= 7;
+        }
+        _data[_offset++] = (char)v;
+    }
+
+    void put(const char* v, std::size_t len) {
+        std::memcpy(_data + _offset, v, len);
```
apangin, line 124: What's the benefit of prepending "std::" everywhere?
C-style equivalents are shorter. We are not going to add more libstdc++ dependencies.

src/buffer.h
```diff
@@ -0,0 +1,129 @@
...
+    void putVar64(u64 v) {
+        int iter = 0;
+        while (v > 0x1fffff) {
+            _data[_offset++] = (char)v | 0x80;
+            v >>= 7;
+            _data[_offset++] = (char)v | 0x80;
+            v >>= 7;
+            _data[_offset++] = (char)v | 0x80;
+            v >>= 7;
+            if (++iter == 3) return;
```
apangin, line 114: IIRC, protobuf and JFR encoding for full-length 64-bit varints is different. Can you confirm?

fandreuz: Actually I wasn't aware of this, and I was just trying to "fix" a unit test I wrote assuming JFR and Protobuf varints are the same :)

fandreuz: Moved the varint stuff to `FlightRecorderBuffer`

src/buffer.h
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _BUFFER_H
+#define _BUFFER_H
+
+#include "os.h"
+#include <cstdint>
```
apangin, line 10: ?

src/buffer.h
```diff
@@ -0,0 +1,78 @@
...
+        _data[_offset++] = v;
+    }
+
+    void put8(size_t offset, char v) {
+        _data[offset] = v;
+    }
+
+    virtual void put16(u16 v) = 0;
+    virtual void put32(u32 v) = 0;
+    virtual void put64(u64 v) = 0;
```
apangin, line 50: Virtual calls will kill performance. Fortunately, these functions are not used too often right now, but it's still not a good desgin for functions that are supposed to be a single memory wrtie operation to go through vtable.
We can probably have `put64le` and `put64be`, although I'm not a big fan of such a naming.
In fact, there's not much code left to share. float->int and double->u64 conversion can be moved to arch.h to make it even less. So maybe let's try to implement a separate protobuf writer first and then see what refactoring is needed if any?

src/flightRecorder.cpp
```diff
@@ -291,87 +294,55 @@ class Lookup {
...
-
-        u.f = v;
-        put32(u.i);
-    }
-
     void putVar32(u32 v) {
-        while (v > 0x7f) {
-            _data[_offset++] = (char)v | 0x80;
+        while (v > 0b01111111) {
+            _data[_offset++] = (char)v | 0b10000000;
```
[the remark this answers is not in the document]
apangin: Yes, please. A number of 21 ones is no way more readable (can you even count digits without a pointer?)

src/buffer.h
fandreuz: ec524fbbc3bdb557d8fa82e67de73b530652f4f4

src/buffer.h
Baraa-Hasheesh: @fandreuz @apangin 
How about using function pointers? `void (*put64)(u64) = NULL;` & assign the pointers to the correct functions in the constructors (I.E. do polymorphism using C-style)

This will keep the function name the same but save the performance as we won't hit the vtable

fandreuz:
Superseded by #1292

PR #1292, opened 2025-05-12
Protobuf writer

fandreuz:
### Description
In this PR I implement a simple Protobuf writer.

### Related issues
#1188

### Motivation and context
I plan to use this as the base for the implementation of `dumpOtlp`.

### How has this been tested?
Unit tests in `protobufBufferTest.cpp`

---


test/native/protobufBufferTest.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+  buf.field(3, (u32)0);
+
+  CHECK_EQ(buf.offset(), 2);
+  CHECK_EQ((unsigned char)buf.data()[0], (3 << 3) | VARINT);
+  CHECK_EQ((unsigned char)buf.data()[1], 0);
+}
+
+TEST_CASE(Buffer_test_var32_150) {
+  char *data = (char *)alloca(100);
```
Baraa-Hasheesh, line 23: The data is never freed is this intentional?

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+#include <type_traits>
+
+typedef const u32 protobuf_t;
+static protobuf_t VARINT = 0;
+static protobuf_t LEN = 2;
+
+typedef u32 protobuf_index_t;
+
+// We assume the length of a nested field can be represented with 3 varint
+// bytes.
```
apangin, line 20: Can you please re-format the code according to the project's style? (indents, * pointers, line length, etc.)

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+  size_t offset() const { return _offset; }
+
+  // VARINT
+  void field(protobuf_index_t index, bool b);
+  void field(protobuf_index_t index, u32 n);
+  void field(protobuf_index_t index, u64 n);
+  // LEN
+  void field(protobuf_index_t index, const char *s);
+  void field(protobuf_index_t index, const char *s, size_t len);
+  void field(protobuf_index_t index, const ProtobufBuffer buffer, size_t len);
```
apangin, line 64: Maybe reference or pointer instead of passing `ProtobufBuffer` by value?

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+const size_t nested_field_byte_count = 3;
+
+class ProtobufBuffer {
+private:
+  ProtobufBuffer *_parent_message;
+  char *_data;
+  size_t _offset;
+
+  ProtobufBuffer(ProtobufBuffer *parent) :
+    _data(parent->_data + parent->offset()),
```
apangin, line 30: Why `_data` but `offset()`?

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+    _offset(0),
+    _parent_message(parent) {}
+
+  template <typename T>
+  typename std::enable_if<std::is_unsigned<T>::value, void>::type
```
apangin, line 35: Is this template really useful? I think `u64` variant covers all cases. Writing `putVarInt<>` with angle brackets does not look nice.

test/native/protobufBufferTest.cpp
fandreuz: https://man7.org/linux/man-pages/man3/alloca.3.html

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+typename std::enable_if<std::is_unsigned<T>::value, void>::type
+ProtobufBuffer::putVarInt(T n) {
+  _offset = putVarInt(_offset, n);
+}
+
+template <typename T>
+typename std::enable_if<std::is_unsigned<T>::value, size_t>::type
+ProtobufBuffer::putVarInt(size_t offset, T n) {
+  while ((n >> 7) != 0) {
+    _data[offset++] = (char)(0b10000000 | (n & 0b01111111));
```
apangin, line 79: No binary literals, please. My eyes are bleeding when counting ones and zeros.
Every system developer should know `0x80` and `0x7f`.

src/protobuf.h
```diff
@@ -0,0 +1,86 @@
...
+typedef u32 protobuf_index_t;
+
+// We assume the length of a nested field can be represented with 3 varint
+// bytes.
+const size_t nested_field_byte_count = 3;
+
+class ProtobufBuffer {
+private:
+  ProtobufBuffer *_parent_message;
+  char *_data;
```
apangin, line 26: Maybe `unsigned char` to make less casts?
I know that FlightRecorder buffer uses `char`, so I don't mind either option.

src/protobuf.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+void ProtobufBuffer::field(protobuf_index_t index, const char *s) {
+  field(index, s, strlen(s));
+}
+
+void ProtobufBuffer::field(protobuf_index_t index, const char *s, size_t len) {
+  tag(index, LEN);
+  putVarInt<>(len);
+  memcpy(_data + _offset, s, len);
```
apangin, line 41: Do you check that message fits the buffer or do you automatically grow the buffer?

fandreuz: That's something I wanted to discuss, I'm not sure if this should be a smarter buffer like `Proto.java`, or a dumb one like what we have in JFR. For now, it's dumb and it gets a `char*` from the user, but I'm open to feedback

fandreuz: I'm trying to think of a scenario in which it makes sense not to have it smart enough to resize itself, but I can't find any. I'll make it self-resizable

src/protobuf.h
fandreuz: d5644f8deed2e5c350ae62f74c32c07fafacaf05

src/protobuf.h
fandreuz: `u64` variant covers all cases, but can we accept the additional overhead of having a new `u64` stack variable when we have a `u32` already? In the meantime I removed the templating: ad85b8800cac40ccef1e0079834d6ec54948c003

src/protobuf.h
fandreuz: a11b31bcc0ae503aa3bbb688470bd863ffa84547

d7258c4ab618db1a731109c120bb3787d978ba92

Let me know if there's anything else I should change

src/protobuf.h
fandreuz: I'm going to remove this whole logic since it does not play well in a self-resizing buffer

fandreuz:
@apangin all review comments addressed

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+void ProtobufBuffer::ensureCapacity(size_t new_data_size) {
+  size_t expected_capacity = _offset + new_data_size;
+  while (expected_capacity > _capacity) {
+    _capacity *= 2;
+  }
+  _data = (unsigned char*) realloc(_data, _capacity * sizeof(unsigned char));
+}
+
+void ProtobufBuffer::putVarInt(u64 n) {
+  _offset = putVarInt(_offset, n);
```
apangin, line 31: Please use 4 spaces indent.

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+  if (value <= 0x7F) return 1;
+  if (value <= 0x3FFF) return 2;
+  if (value <= 0x1FFFFF) return 3;
+  if (value <= 0xFFFFFFF) return 4;
+  if (value <= 0x7FFFFFFFF) return 5;
+  if (value <= 0x3FFFFFFFFFF) return 6;
+  if (value <= 0x1FFFFFFFFFFFF) return 7;
+  if (value <= 0xFFFFFFFFFFFFFF) return 8;
+  if (value <= 0x7FFFFFFFFFFFFFFF) return 9;
+  return 10;
```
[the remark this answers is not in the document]
fandreuz: How does this work?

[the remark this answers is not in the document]
apangin: The function counts number of bytes required to encode a value as varint: 7-bit value is encoded with 1 byte, a value up to 14 bits is encoded with 2 bytes, up to 21 bits with 3 bytes, and so on. In other words, `size = size_in_bits(value) / 7` rounded up. To get size in bits of an integer, we count number of leading zeros in its 64-bit representation and subtract this from 64. In many architectures, there is a dedicated CPU instruction to count leading zeros, e.g., `BSR` on x86 and `CLZ` on ARM. `__builtin_clzll` is GCC/clang intrinsic for this operation. Since this intrinsic is not defined for zero, we ensure the value is always non-zero by setting its lowest bit: `value | 1`.

So, `size = ceil[(64 - __builtin_clzll(value | 1)) / 7]`
Considering that `ceil[N / 7] = floor[(N + 6) / 7]`, we get to
`size = (70 - __builtin_clzll(value | 1)) / 7`

This is already a good branchless formula. I took it one step further by applying manual [strength reduction](https://en.wikipedia.org/wiki/Strength_reduction) to get rid of division. Even though compilers can replace integer division with multiplication automatically, I can optimize it better, knowing that the dividend is always in the range [0-63] here.
`N / 7 ≈ (N + 1) * 9 / 64`, where division by 64 is a simple bitwise shift. Note that this formula is accurate for all integers between 0 and 63.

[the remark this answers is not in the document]
Baraa-Hasheesh:
```
size_t size = 1;
while ((value = value >> 7)) {
   size++;
}
return size
```
should achieve the same & more readable IMO compared to using __builtin_clzll as using it requires multiple steps to understand how the evaluation work

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void ProtobufBuffer::ensureCapacity(size_t new_data_size) {
+  size_t expected_capacity = _offset + new_data_size;
+  while (expected_capacity > _capacity) {
```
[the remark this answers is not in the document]
Baraa-Hasheesh: Let's use `expected_capacity * N` to avoid constant resize operations  => I suggest `N = 1.5`
What do you think @apangin 

Also if _capacity isn't resized let's skip the `realloc`

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+  _data = (unsigned char*) realloc(_data, _capacity * sizeof(unsigned char));
+}
+
+void ProtobufBuffer::putVarInt(u64 n) {
+  _offset = putVarInt(_offset, n);
+}
+
+size_t ProtobufBuffer::putVarInt(size_t offset, u64 n) {
+  ensureCapacity(computeVarIntByteSize(n));
+  while ((n >> 7) != 0) {
```
Baraa-Hasheesh, line 36: `!= 0` isn't needed

src/protobuf.cpp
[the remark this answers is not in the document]
fandreuz: `while` loop will probably be less performant, and this method is used for every single operation on the buffer

src/protobuf.cpp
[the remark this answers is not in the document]
fandreuz: > Let's use `expected_capacity * N` to avoid constant resize operations => I suggest `N = 1.5` What do you think @apangin

Let's keep it simple, we don't need to introduce random guesses until we see it's not fast enough.

> Also if _capacity isn't resized let's skip the `realloc`

Good point

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void ProtobufBuffer::commitMessage(protobuf_mark_t mark) {
+  size_t message_length = _offset - mark;
+  for (size_t i = 0; i < nested_field_byte_count - 1; ++i) {
+    size_t idx = mark - nested_field_byte_count + i;
+    _data[idx] = (unsigned char) (0x80 | (message_length & 0x7f));
+    message_length >>= 7;
+  }
+  _data[mark - 1] = (unsigned char) message_length;
```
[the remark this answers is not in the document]
fandreuz: I'm thinking about changing this logic in the future. At the moment we allocate three bytes for the size (varint) of nested messages, since we don't known it in advance. If the message has size smaller than 128, we waste two bytes for each. For instance, for 100 nested messages of type `Sample`, that would be:
- 1 byte for the tag
- 3 bytes for the size
- 1 tag * 2 varints in the nested message

For small varints in the nested message, that's around 30% wasted space.

Instead, I would allow the caller to optimistically specify an expected message size, and we relocate the content if needed after the message is committed and we know the final size. I wouldn't do that in the initial PR since it's an optimization and we can leave it for after the prototyping phase.

What do you think @apangin ?

src/protobuf.h
```diff
@@ -0,0 +1,65 @@
...
+
+  void tag(protobuf_index_t index, protobuf_t type);
+
+  void ensureCapacity(size_t new_data_size);
+
+public:
+  ProtobufBuffer(size_t initial_capacity) : _capacity(initial_capacity), _offset(0) {
+    if (_capacity == 0) {
+      _capacity = 1;
+    }
```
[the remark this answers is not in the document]
fandreuz: c569c51fb465931f7a03cdd4b753a682a5a4ce4b

src/protobuf.cpp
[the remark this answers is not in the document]
fandreuz: d6722199c0e03089ce42c53e69ef7e7fdf02fcdd

src/protobuf.cpp
```diff
@@ -0,0 +1,90 @@
...
+  if (value <= 0x7FFFFFFFFFFFFFFF) return 9;
+  return 10;
+}
+
+void ProtobufBuffer::ensureCapacity(size_t new_data_size) {
+  size_t expected_capacity = _offset + new_data_size;
+  while (expected_capacity > _capacity) {
+    _capacity *= 2;
+  }
+  _data = (unsigned char*) realloc(_data, _capacity * sizeof(unsigned char));
```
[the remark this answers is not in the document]
fandreuz: 6d24b0a9f4a3dc53f350a42dc345472a54dc044c

src/protobuf.cpp
fandreuz: 5fecdff813299f251e03b1ce1ed236bcd7df4fae

src/protobuf.cpp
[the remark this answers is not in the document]
Baraa-Hasheesh: `while` & `__builtin_clzll` should have somewhat similar performance as both will be around `O(log(N))`
Now for exact performance between them you will need to actually check implementation of `__builtin_clzll` & count operations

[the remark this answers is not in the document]
fandreuz: 1c4b6ee33215329e50b1b113dc04ce80b5ce1630

fandreuz:
Thank you guys for the feedback, all comments addressed

src/protobuf.cpp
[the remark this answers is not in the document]
fandreuz: @Baraa-Hasheesh there you go:
```
$ g++ varIntByteSizeBench.cpp -Isrc
$ ./a.out 
Francesco: 02268160
Andrei   : 01645742
Bara     : 05314510
$ g++ -O3 varIntByteSizeBench.cpp -Isrc
$ ./a.out 
Francesco: 00255122
Andrei   : 00403302
Bara     : 01693702
```
This uses `rtdsc()` from `tsc.h`. Code [here](https://github.com/fandreuz/async-profiler/blob/varint-byte-size-bench/varIntByteSizeBench.cpp).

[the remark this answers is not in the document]
fandreuz: > In your benchmark, input data is consecutive, which makes it perfect for branch prediction

Yeah I figured there was some branch prediction going on in the benchmark, I'll keep Andrei's formula for this PR

src/protobuf.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "protobuf.h"
+#include <string.h>
```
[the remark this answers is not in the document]
fandreuz: 8b8c1dc0d02e2818069811613cc9d619d9ec9b1e

src/protobuf.h
```diff
@@ -0,0 +1,67 @@
...
+typedef u32 protobuf_mark_t;
+
+// We assume the length of a nested field can be represented with 3 varint bytes.
+const size_t nested_field_byte_count = 3;
+const size_t minimum_initial_size = 16;
+
+size_t computeVarIntByteSize(u64 value);
+
+class ProtobufBuffer {
+private:
```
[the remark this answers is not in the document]
fandreuz: d43a3b605e34553edd34351a49f283cd8eb3e3b1

src/protobuf.h
```diff
@@ -0,0 +1,67 @@
...
+typedef const u32 protobuf_t;
+static protobuf_t VARINT = 0;
+static protobuf_t LEN = 2;
+
+typedef u32 protobuf_index_t;
+typedef u32 protobuf_mark_t;
+
+// We assume the length of a nested field can be represented with 3 varint bytes.
+const size_t nested_field_byte_count = 3;
+const size_t minimum_initial_size = 16;
```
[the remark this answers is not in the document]
fandreuz: d677f7320afe8bd5da8c6b85b79475fa603960f7

test/native/protobufBufferTest.cpp
```diff
@@ -0,0 +1,184 @@
...
+    const protobuf_mark_t max_tag = 536870911;
+    buf.field(max_tag, (u64) 3);
+
+    CHECK_EQ(buf.offset(), 6);
+    // Check the value of the first 5 bytes as a varint
+    u32 sum = buf.data()[5] & 0x7f;
+    for (int idx = 4; idx >= 0; --idx) {
+            sum <<= 7;
+            sum += ( buf.data()[idx] & 0x7f);
+    }
```
apangin, line 144: Extraneous spaces

src/protobuf.h
```diff
@@ -0,0 +1,69 @@
...
+protobuf_t LEN = 2;
+
+typedef u32 protobuf_index_t;
+typedef u32 protobuf_mark_t;
+
+// We assume the length of a nested field can be represented with 3 varint bytes.
+const size_t NESTED_FIELD_BYTE_COUNT = 3;
+const size_t MINIMUM_INITIAL_SIZE = 16;
+
+size_t varIntSize(u64 value);
```
apangin, line 26: ?

src/protobuf.h
```diff
@@ -0,0 +1,67 @@
...
+typedef u32 protobuf_index_t;
+typedef u32 protobuf_mark_t;
+
+// We assume the length of a nested field can be represented with 3 varint bytes.
+const size_t nested_field_byte_count = 3;
+const size_t minimum_initial_size = 16;
+
+size_t computeVarIntByteSize(u64 value);
+
+class ProtobufBuffer {
```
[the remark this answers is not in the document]
apangin: That is true. Yet, "bufBuffer" looks weird. I think it's fine to shorten the class name to `ProtoBuffer` or `Protobuf` if you wish. Even in Protobuf documentation, you may find many occurrences of "Proto [something]": "Proto definition", "Proto Best Practices", "ProtoJSON", "Proto Limits", etc.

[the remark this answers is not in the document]
fandreuz: Sounds good: b4be7e418112e35a09dfae7622f60408548f9ca0

src/protobuf.h
fandreuz: 82700227fa7c6684430027d42869932078d50730

src/protobuf.h
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _PROTOBUF_H
+#define _PROTOBUF_H
+
+#include <sys/param.h>
+#include <string.h>
+#include <type_traits>
```
apangin, line 11: Please cleanup unused imports

test/native/protobufBufferTest.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // https://protobuf.dev/programming-guides/proto3/#assigning-field-numbers
+    const protobuf_mark_t max_tag = 536870911;
+    buf.field(max_tag, (u64) 3);
+
+    CHECK_EQ(buf.offset(), 6);
+    // Check the value of the first 5 bytes as a varint
+    u32 sum = buf.data()[5] & 0x7f;
+    for (int idx = 4; idx >= 0; --idx) {
+            sum <<= 7;
+            sum += (buf.data()[idx] & 0x7f);
```
apangin, line 143: Indentation is still wrong :(

src/protobuf.h
fandreuz: 2edff1037b195a1eb765e1a70437175269f3f5a2

test/native/protobufBufferTest.cpp
fandreuz: a834b878181dc5223ba687879e393f49fd18318d

src/protobuf.h
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _PROTOBUF_H
+#define _PROTOBUF_H
+
+#include <sys/param.h>
+#include <string.h>
+#include <stdlib.h>
```
apangin, line 11: It's generally preferable to keep number of #includes minimal in header files (this may improve compilation speed when sources are not merged).
In this case, all these standard headers can be relocated to .cpp. Implementation of ProtoBuffer constructor/destructor will need to move to .cpp as well, but that's fine.

test/native/protobufBufferTest.cpp
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "protobuf.h"
+#include "testRunner.hpp"
+#include <cstdint>
```
apangin, line 8: What for?

fandreuz: Do you have something which tells you that some headers are not needed?

3792066c4a29b3f9ed70655bad323d3426b3786c

src/protobuf.h
fandreuz: eeca655089c0818afbf61695210789c79c1918d6

test/native/protobufBufferTest.cpp
apangin: Intuition :) Also, a rule of thumb: if it is a C++ header, it is likely not needed :)

fandreuz:
@apangin can this be merged? I'll raise another PR for review with the next stuff for OTEL support

PR #1297, opened 2025-05-16
Poc unify hooks

### LINKED ISSUE #1245: async-profiler doesn't patch previously patched library in case of library reload
### Describe the bug

If async-profiler detects a `dlopen` call, it will patch the opened library to hook certain method calls such as `malloc` & nested `dlopen` calls

The problem happens if a lib has the sequence of `dlopen -> dlclose -> dlopen` operations done one it, 
Due to how the linker & memory allocation is handled it is possible that the second dlopen will return the library allocated to the same address as the original `dlopen`, this means that even though the lib was unloaded & reloaded (Shared objects were deleted & recreated) it will appear identical to the original opened library.

This will cause the async-profiler to incorrectly mark this new library as already patched and skip it causing the async-profiler to not correctly monitor certain calls made inside this library such as nested dlopen (Propagation of issue) & memory allocations (possible hidden memory leaks)

async-profiler shouldn't skip a previously patched library if the shared objects were recreated (unloaded & loaded back again)

### Reproduction Steps

For simplicity I would recommend using the new test found here #1243 

replace the **`if (!dlopenFirst)`** body to look like the following

```
lib = dlopen("libcallsmalloc.so", RTLD_NOW);
dlclose(lib);
lib = dlopen("libcallsmalloc.so", RTLD_NOW);
...
```

_No response_

### Async-profiler version

latest + changes at #1243

### Environment details

_No response_

### LINKED ISSUE #1273: recursive dlopen calls when Profiler is loaded via LD_PRELOAD
### Describe the bug

```
(gdb) where
(gdb) c
Continuing.

(gdb) where
(gdb)
```

_No response_

### Reproduction Steps

_No response_

### Async-profiler version

latest

### Environment details

_No response_

### LINKED ISSUE #1280: Missing profiling of None Java threads in Java applications
### Describe the bug

If a java application creates a none java thread via a JNI call that Java thread will not be profiled correctly in CPU/CTIMER modes 

This happens as java applications uses the `CpuEngine::setupThreadHook` function to find the pthread functions & hooks them to be monitored.

However native threads (c/c++) will typically depend on different hooks which are typically installed in the `Hooks::patchLibraries` function

However the `Hooks::patchLibraries` is never called for `dlopen` calls that are done via Java `System.load` 

This in turn creates the possibility of having threads that aren't monitored during the application run process

Profiler should monitor non java threads as well in java applications

### Reproduction Steps

You can use the following files found here https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/thread

Use the [runner_java.sh](https://github.com/Baraa-Hasheesh/asprof-reproducers/blob/main/thread/runner_java.sh) 
When inspecting the JFR output it will be observed that the `myNativeThread` thread is not reordered correctly  in the generated JFR

_No response_

### Async-profiler version

latest

### Environment details

_No response_

Baraa-Hasheesh:
### Description
Currently in Async profiler there are multiple locations were hooks are being installed 
This creates complexity in tracking what's installed & could create different behaviors depending on how async profiler is started 

This depends on the following PRs  
#1264 
#1279 

This will fix the following:
Partially fix #1245 if profiler is restarted (patches will be applied again)
Fix #1280 as patching is unified to use the same to all
Fix #1273 as preload hooks are removed

### Related issues
#1245 #1280 #1273

### Motivation and context
Simplify hook installation into a centralized method that is used by all

### How has this been tested?
`make test`

---

### DIFF
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -268,7 +268,8 @@ void CodeCache::makeImportsPatchable() {
     if (max_import != NULL) {
         uintptr_t patch_start = (uintptr_t)min_import & ~OS::page_mask;
         uintptr_t patch_end = (uintptr_t)max_import & ~OS::page_mask;
-        mprotect((void*)patch_start, patch_end - patch_start + OS::page_size, PROT_READ | PROT_WRITE);
+        uintptr_t patch_size = patch_end - patch_start + OS::page_size;
+        OS::protect(patch_start, patch_size, PROT_READ | PROT_WRITE);
     }
 }
 
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -29,12 +29,15 @@ size_t NativeFunc::usedMemory(const char* name) {
 
 
 CodeCache::CodeCache(const char* name, short lib_index, bool imports_patchable,
-                     const void* min_address, const void* max_address) {
+                     const void* min_address, const void* max_address,
+                     const char* image_base) {
     _name = NativeFunc::create(name, -1);
+
     _lib_index = lib_index;
     _min_address = min_address;
     _max_address = max_address;
     _text_base = NULL;
+    _image_base = image_base;
 
     _plt_offset = 0;
     _plt_size = 0;
@@ -234,7 +237,6 @@ void CodeCache::addImport(void** entry, const char* name) {
 void** CodeCache::findImport(ImportId id) {
     if (!_imports_patchable) {
         makeImportsPatchable();
-        _imports_patchable = true;
     }
     return _imports[id][PRIMARY];
 }
@@ -271,6 +273,8 @@ void CodeCache::makeImportsPatchable() {
         uintptr_t patch_size = patch_end - patch_start + OS::page_size;
         OS::protect(patch_start, patch_size, PROT_READ | PROT_WRITE);
     }
+
+    _imports_patchable = true;
 }
 
 void CodeCache::setDwarfTable(FrameDesc* table, int length) {
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -32,6 +32,10 @@ CodeCache::CodeCache(const char* name, short lib_index, bool imports_patchable,
                      const void* min_address, const void* max_address,
                      const char* image_base) {
     _name = NativeFunc::create(name, -1);
+    if (_name != NULL) {
+        _short_name = strrchr(_name, '/');
+        _short_name = _short_name == NULL ? _name : _short_name + 1;
+    }
 
     _lib_index = lib_index;
     _min_address = min_address;
@@ -43,6 +47,7 @@ CodeCache::CodeCache(const char* name, short lib_index, bool imports_patchable,
     _plt_size = 0;
 
     memset(_imports, 0, sizeof(_imports));
+    memset(_orig_import_references, 0, sizeof(_orig_import_references));
     _imports_patchable = imports_patchable;
     _debug_symbols = false;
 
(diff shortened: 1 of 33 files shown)


src/codeCache.cpp
```diff
@@ -32,6 +32,10 @@ CodeCache::CodeCache(const char* name, short lib_index, bool imports_patchable,
                      const void* min_address, const void* max_address,
                      const char* image_base) {
     _name = NativeFunc::create(name, -1);
+    if (_name != NULL) {
+        _short_name = strrchr(_name, '/');
```
fandreuz, line 36: Nit: can we call this "basename"?

src/hooks.cpp
```diff
@@ -182,11 +151,34 @@ void Hooks::patchLibraries() {
...
-            cc->patchImport(im_dlopen, (void*)dlopen_hook);
+            cc->patchImport(im_dlopen, (void*)dlopen_hook_impl);
+            cc->patchImport(im_pthread_setspecific, (void*)pthread_setspecific_hook);
         }
         cc->patchImport(im_pthread_create, (void*)pthread_create_hook);
         cc->patchImport(im_pthread_exit, (void*)pthread_exit_hook);
     }
 }
+
+void Hooks::unpatchLibraries() {
```
apangin, line 169: Uninstalling hooks is very dangerous, especially when there are two instances of the profiler loaded that can conflict with each other.
I would postpone this PR until we stabilize all symbol parsing changes.

src/hooks.cpp
```diff
@@ -87,83 +87,52 @@ static void pthread_exit_hook(void* retval) {
...
-
-// LD_PRELOAD hooks
-
-extern "C" WEAK DLLEXPORT
-int pthread_create(pthread_t* thread, const pthread_attr_t* attr, ThreadFunc start_routine, void* arg) {
-    if (_orig_pthread_create == NULL) {
-        _orig_pthread_create = ADDRESS_OF(pthread_create);
+// Intercept thread creation/termination by patching libjvm's GOT entry for pthread_setspecific().
+// HotSpot puts VMThread into TLS on thread start, and resets on thread end.
+static int pthread_setspecific_hook(pthread_key_t key, const void* value) {
```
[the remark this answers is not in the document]
Baraa-Hasheesh: @apangin I checked we do capture them but what if the thread is re-used?
Or is a  pthread_create created always when a thread starts (I'm referring to thread pools like a fork join pool)

src/hooks.cpp
Baraa-Hasheesh: I agree with that, as this PR inherits a lot of symbol parse fixes from 2 other PRs as mentioned in the description & should only be merged & considered after those fixes are integrated

Regarding the "two instances of the profiler loaded" can you give an example on this, 

Because if profilers share the shared objects this could be an existing issue that we should look into 
(Java hooks are currently uninstalled when profiler stops)

if profilers don't share the shared objects then it's also an existing issue as one profiler hooks will override the other.

krk: We can also check how [ddprof](https://github.com/DataDog/ddprof) is doing it, they have a different mechanism for hooking.

Baraa-Hasheesh: For two instances:

if profilers share the shared objects  the profiler loaded of profiler that shouldn't happen as profiler prevents 2 profiling sessions running in parallel

if profilers don't share the shared objects then it's something that currently exists & can't protect against it with the current hooks we have as one profiler will simply override the other

@krk for [ddprof](https://github.com/DataDog/ddprof) I will need to do a deep read on how they handle hooks

src/codeCache.cpp
```diff
@@ -179,8 +184,9 @@ const void* CodeCache::findSymbolByPrefix(const char* prefix, int prefix_len) {
 
 void CodeCache::saveImport(ImportId id, void** entry) {
     for (int ty = 0; ty < NUM_IMPORT_TYPES; ty++) {
-        if (_imports[id][ty] == nullptr) {
+        if (_imports[id][ty] == nullptr && entry != NULL) {
```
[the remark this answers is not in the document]
Baraa-Hasheesh: sanity check as I'm dereferencing the value, it shouldn't be NULL

PR #1299, opened 2025-05-18
Skip last 10% allocations for leak detection

apangin:
### Description

For native leak profiling, ignore allocations made in the last 10% of the profiling session.

### Motivation and context

#1064 introduced the feature to create a flamegraph of native memory allocations. With `--leak` option, only those allocations are displayed that do not have matching `free` calls - such allocations are considered memory leaks. However, this algorithm is biased towards allocations made closer to the end of the profiling session, because such allocations are more likely to be freed only after the session has ended.

To avoid such a bias, it is proposed to simply ignore younger allocations. For example, if a profiling session runs for 10 minutes, allocations made during the last minute won't be included in the leak flamegraph.

### How has this been tested?

Created an artificial memory leak by not closing `DirectoryStream` and ran
```
jfrconv --nativemem --leak --total --lines leak.jfr
```

**Before the change:**
leak-old

**After the change:**
leak-new

---

### DIFF
--- a/docs/ConverterUsage.md
+++ b/docs/ConverterUsage.md
@@ -49,6 +49,7 @@ JFR options:
     --live             Build allocation profile from live objects only during conversion
     --nativemem        Generate native memory allocation profile
     --leak             Only include memory leaks in nativemem
+    --tail RATIO       Ignore tail allocations for leak profiling (10% by default)
     --lock             Generate only Lock contention profile during conversion
  -t --threads          Split stack traces by threads
  -s --state LIST       Filter thread states: runnable, sleeping, default. State name is case insensitive
--- a/docs/ProfilingModes.md
+++ b/docs/ProfilingModes.md
@@ -117,9 +117,17 @@ jfrconv --total --nativemem --leak app.jfr app-leak.html
 jfrconv --total --nativemem app.jfr app-malloc.html
 ```
 
-When `--leak` option is used, the generated flame graph will show allocations without matching `free` calls. If `-nofree` is specified, every allocation will be reported as a leak:
+When `--leak` option is used, the generated flame graph will show allocations without matching `free` calls.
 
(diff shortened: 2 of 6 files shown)


Baraa-Hasheesh:
@apangin 
Using a % could lead to a very large ignore window for long profiling sessions 
For example if the profiling session is 1 hour => 60Minutes, we will ignore the last 6 minutes of allocations  for memory leakage 

At the same time using a static value like last minute could also lead to bad outcomes for short profiling sessions

I think we should have both a base static ignore window & % ignore window

We can define a base max ignore window of for example 10 seconds
And evaluate against the min value between the 10% & our base window

So for short profiling sessions we would use the % evaluation while for longer profiling sessions we would use the base max ignore window we define.

apangin:
> if the profiling session is 1 hour => 60Minutes, we will ignore the last 6 minutes of allocations

On the other hand, profiler will collect useful allocation data for 54 minutes - that's a lot!
If there is a continuous leak, then finding a meaningful signal in 54 minutes is just as efficient as in 59 minutes, but in the latter case, the noise will be likely higher due to young allocations that have not been freed. The problem with a constant size tail is that there always be such allocations, no matter how long profiler runs, whereas with the fixed ratio, results accuracy increases with the duration of profiling session.

The approach proposed in this PR is simple and efficient. There is no point in complicating it without obvious benefits.

fandreuz:
@apangin As a user, I'd want to have this behavior documented, and to have the tail size configurable

apangin:
@fandreuz I agree on the documentation point - will add a sentence.
As to making it configurable, I'm not sure. In my paradigm, software should just work well by default; the less things require manual configuration, the better. That said, we may add a knob in the future, if there is a demand for it.

apangin:
OK, made this configurable with `--tail` option. Updated docs.

fandreuz:
Looks good to me

PR #1300, opened 2025-05-19
Allow profiling kprobes/uprobes with --fdtransfer

apangin:
### Description

Extend fdtransfer protocol to pass probe name in perf_event requests.

### Motivation and context

Async-profiler can profile kprobes and uprobes, e.g. `asprof -e kprobe:fd_install`. However, this does not work with `--fdtransfer` option. The reason is, `perf_event_attr` structure contains a pointer to kprobe/uprobe name, but when `perf_event_attr` is passed to another process (fdtransfer server), this pointer becomes invalid, since the address belongs to a different process.

The proposed solution extends `perf_fd_request` structure with a new field that contains a name of kprobe/uprobe. For events other than kprobe/uprobe, this field contains an empty string. When fdtransfer server receives non-empty probe name, it adjusts pointer in `perf_event_attr`.

### How has this been tested?

Added `KernelTests.kprobe` test, which failed before the fix, but passes now.

---

### DIFF
--- a/src/fdtransfer.h
+++ b/src/fdtransfer.h
@@ -17,6 +17,8 @@
 
 #define RESTARTABLE(call)  ({ ssize_t ret; while ((ret = call) < 0 && errno == EINTR); ret; })
 
+#define MAX_PROBE_LEN  256
+
 
 // base header for all requests
 enum request_type {
@@ -34,6 +36,7 @@ struct perf_fd_request {
     int tid;
     int target_cpu;
     struct perf_event_attr attr;
+    char probe_name[MAX_PROBE_LEN];
 };
 
 struct fd_response {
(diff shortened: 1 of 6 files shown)


src/fdtransfer.h
```diff
@@ -17,6 +17,8 @@
 
 #define RESTARTABLE(call)  ({ ssize_t ret; while ((ret = call) < 0 && errno == EINTR); ret; })
 
+#define MAX_PROBE_LEN  256
```
fandreuz, line 20: Maybe this could be documented for users?

apangin: This will be superfluous, IMO. I can't imagine a reasonable function name with more than 256 characters. The only thing we must ensure is that async-profiler does not crash due to out-of-bounds access.

PR #1302, opened 2025-05-19
Add error frames for VM/VMX stack walking when it stops due to Frame/Stack size restrictions

Baraa-Hasheesh:
### Description
VM/VMX might stop midway in the unwinding process due to Stack depth restriction or due to Frame Size restrictions
This change simply adds an error frame at the bottom of such cases to indicate the incomplete unwind

### Related issues
#1230

### Motivation and context
Make incomplete stack traces more clear to user

### How has this been tested?
New tests added @ test/test/stackwalker/StackwalkerTests.java
`make test`

---


test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
...
+    while (i < 100000000) {
+        i++;
+        result += sqrt(i);
+        result += pow(i, sqrt(i));
+    }
+    return result;
+}
+
+__attribute__((unused)) JNIEXPORT double generateDeepStack(int count) {
+    char frame[0x20000];
```
fandreuz, line 22: Is there a specific reason for having this in hex and not in binary?

test/test/stackwalker/StackwalkerTests.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test(mainClass = Stackwalker.class, jvmArgs = "-Xss5m", args = "walkStackComplete",
+            agentArgs = "start,event=cpu,cstack=vmx,file=%f.jfr")
+    public void normalStackVMX(TestProcess p) throws Exception {
+        p.waitForExit();
+        assert p.exitCode() == 0;
+        Output output = Output.convertJfrToCollapsed(p.getFilePath("%f"));
+        assert output.contains("^([^\\[;]+;)?" + // Root can be different on different systems
+                "(start_thread;|thread_start;)?" + // Mac Vs Linux
+                "(_pthread_start;)?" + // Mac specific frame
```
fandreuz, line 57: All optional groups do not add any power to the test. I'd rather remove them, or (better) have dedicated tests or checks for MacOS and JDK>=24

test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
...
+    if (count == 0) {
+        return doCpuTask();
+    } else {
+        return generateDeepStack(count - 1);
+    }
+}
+
+__attribute__((unused)) JNIEXPORT void JNICALL Java_test_stackwalker_Stackwalker_walkStackLargeFrame(JNIEnv* env, jclass cls) {
+    char frame[0x50000];
+    (void)frame;
```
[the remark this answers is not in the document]
Baraa-Hasheesh: Avoid optimizing the `frame` variable by the compiler

test/test/stackwalker/StackwalkerTests.java
Baraa-Hasheesh: The optional groups will guarantee that the stack will have a specific structure 
If we remove them that guarantee will be lost

test/native/libs/jnistackwalker.c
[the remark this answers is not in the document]
fandreuz: Can we have a comment to explain that?

test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
...
+    (void)doCpuTask();
```
krk, line 35: Instead of casting to `void` to mark it as "used", we can return the result.

test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
...
+
+    while (i < 100000000) {
+        i++;
+        result += sqrt(i);
+        result += pow(i, sqrt(i));
+    }
+    return result;
+}
+
+__attribute__((unused)) JNIEXPORT double generateDeepStack(int count) {
```
krk, line 21: Why is `__attribute__((unused))` required?

src/stackWalker.cpp
```diff
@@ -402,7 +402,13 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
 
         // Check if the next frame is below on the current stack
-        if (sp < prev_sp || sp >= prev_sp + MAX_FRAME_SIZE || sp >= bottom) {
+        if (sp < prev_sp) {
+            break;
+        }
+
+        // Check unwind limits
+        if (sp >= prev_sp + MAX_FRAME_SIZE || sp >= bottom) {
+            fillFrame(frames[depth++], BCI_ERROR, "[possible-truncated]");
```
krk, line 411: Either `possible-truncation` or `possibly-truncated`.

test/test/stackwalker/StackwalkerTests.java
krk: Optionals are suspicious here, it may be better to split mac tests into `normalStackVMX_mac` or similar, if it will help simplify.

On the other hand, how many JVMs did you test this with? What is the intention of the test here, is it not testing if there is a `[possibly-truncated]` frame? What do we gain by asserting all these frames?

krk (changes requested):

test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
...
+    char frame[0x20000];
+    (void)frame;
+
+    if (count == 0) {
+        return doCpuTask();
+    } else {
+        return generateDeepStack(count - 1);
```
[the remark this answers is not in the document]
Baraa-Hasheesh: The test will fail if that happened 
I will add attribute to prevent optimization just in case

test/native/libs/jnistackwalker.c
Baraa-Hasheesh: Just to garantee the unused variable is never optimizated out
I'm also using the cast void logic to make that clear & more garanteed

test/test/stackwalker/StackwalkerTests.java
Baraa-Hasheesh: I will need to rephrase the comments 

However even in Linux distros the root frames could be different depending on the used Linux (debian is different from Alpine which is different from Amazon 2 & 2023)

Splitting into 2 tests won't simplify much

The check is to garantee the structure of the sampled stack frames

I've tested with JDK11->24

The intention of the test is to check that the stack frames are captured correctly

Baraa-Hasheesh: Just to note `(start_thread;|thread_start;)`
Could be missing on some Linux distros

test/native/libs/jnistackwalker.c
```diff
@@ -0,0 +1,46 @@
+/*
```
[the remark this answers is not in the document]
Baraa-Hasheesh: I will go with jninativestack

src/stackWalker.cpp
Baraa-Hasheesh: handled here [16f7505](https://github.com/async-profiler/async-profiler/pull/1302/commits/16f75056af64a3ae244fa48d0dd07512a983c307)

test/native/libs/jnistackwalker.c
Baraa-Hasheesh: handled here [16f7505](https://github.com/async-profiler/async-profiler/pull/1302/commits/16f75056af64a3ae244fa48d0dd07512a983c307)

test/native/libs/jnistackwalker.c
Baraa-Hasheesh: handled here [16f7505](https://github.com/async-profiler/async-profiler/pull/1302/commits/16f75056af64a3ae244fa48d0dd07512a983c307)

test/native/libs/jnistackwalker.c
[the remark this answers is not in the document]
Baraa-Hasheesh: handled here [16f7505](https://github.com/async-profiler/async-profiler/pull/1302/commits/16f75056af64a3ae244fa48d0dd07512a983c307)

test/test/stackwalker/StackwalkerTests.java
Baraa-Hasheesh: updated here [16f7505](https://github.com/async-profiler/async-profiler/pull/1302/commits/16f75056af64a3ae244fa48d0dd07512a983c307)

Baraa-Hasheesh:
After testing the changes here it was found to have a high error rate 

I couldn't find a clear way to know if the unwinding was complete or incomplete.
During research on the subject the option of checking if a "known root" was encountered however that is also not guaranteed.

Furthermore it's also possible to run on configs that don't save stack frames on certain system calls which make doing a complete stack unwind not possible (However the stackwalker could do a complete unwind of available frames).

I will remove the changes on the stackwalker & keep the new test to check some edge cases in VM stackwalker

Issue #1304, opened 2025-05-19
Multiple Async-Profilers conflict with one another


Baraa-Hasheesh:
### Describe the bug

**Senario Description:**
Assume you have a Java application that you want to continuously profile via the Java agent `-agentpath`

You want to use `asprof` to profile specific portions of the Process for specific events
This `asprof` will be started & stopped on demand as you profile your process 

--------
The problem is that each profiler instance will install it's hooks to better monitor the process.

The `asprof` installing it hooks will mean that the original Agent profiler `-agentpath` will lose the hooks it already installed.

This happens as the `asprof` will override the `GOT` entries for `dlopen` & `pthread_setspecific` that the `-agentpath` already modified.

This means that all future calls of `dlopen` & `pthread_setspecific` will be re-directed to the `asprof` profiling instance & the `-agentpath` will loose those capabilities

Furthermore once the `asprof` is stopped, it will uninstall the hooks over `dlopen` & `pthread_setspecific` & revert them back to original values.

This means after `asprof` is stopped the remaining `-agentpath` profiler will still be without hooks having limited profiling capabilities.

--------
In other words  this flow will split the process into 3 regions:

1. Region A: Before `asprof` is started 
Here only the `-agentpath` will be present & running & it will have full profiling capabilities 

2. Region B: after `asprof` is started 
Both the `-agentpath` &  `asprof` will be present & running.
Only the `asprof` will have full profiling capabilities 

3. Region C: After `asprof` is stopped
Both the `-agentpath` &  `asprof` will be present
Only -agentpath will be running
No agent will have full profiling capabilities

### Expected vs. actual behavior

At least one agent should have full profiling capabilities

### Reproduction Steps

Check then test here 
https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/javadoubleprofiler 

You can run with both `./runner.sh single` or `./runner.sh multi`

In **single** you will profile the malloc done by the call of `method`
in **multi** you will not profile the malloc done by the call of `method`

### Async-profiler version

latest

Baraa-Hasheesh:
Possible solutions include the following:
1. Save content of GOT entries & revert back to the found value when a profiler stops
This will be fragile & highly dependent on order of profilers,
Specially if we're using an API mode profiler & attach mode profiler which can be stopped at any moment from the process/cli.

2. Refactor profiler to have one main profiler.
When a profiler starts it will check if any other async-profiler is present
[...]

PR #1306, opened 2025-05-21
experimental fix for double profiling

Baraa-Hasheesh:
### Description
When using 2 async-profilers one profiler will override the other, making only one profiler fully functional.
Furthermore when a profiler exists it will leave the remaining profiler in-complete state making not profile correctly 

### Related issues
#1304

### Motivation and context
Make it possible to have multiple async-profilers living on same process

### How has this been tested?
Manual tests by using the test here => https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/javadoubleprofiler

## Notes
This PR isn't ready for full review as it's only applied on specific hooks as POC & only created as a proposal to fix the problem.
If we're happy with this approach will extend to part of all hooks installed in the system

Pros:
* This change will make profilers communicate with each other via callbacks
* Will make sure that profilers are alive with proper hook installs

Cons:
* If profilers are started/stopped in parallel some race conditions may happen making some profilers unable to install hooks
* When exiting it's possible to end up in a state were hooks are never uninstalled, as each profiler do extra checks as to not break callback chain across profilers & not cause infinite loops

---

### DIFF
--- a/src/cpuEngine.cpp
+++ b/src/cpuEngine.cpp
@@ -12,8 +12,10 @@
 #include "tsc.h"
 #include "vmStructs.h"
 
+typedef int (*pthread_setspecific_t)(pthread_key_t, const void*);
 
 void** CpuEngine::_pthread_entry = NULL;
+static pthread_setspecific_t _original_pthread_setspecific = NULL;
 CpuEngine* CpuEngine::_current = NULL;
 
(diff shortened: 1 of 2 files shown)


Baraa-Hasheesh:
Not needed => Fixed in #1323

PR #1312, opened 2025-05-23
Prevent parsing more libs than the MAX_NATIVE_LIBS limit

Baraa-Hasheesh:
### Description
It's possible for the profiler to try to parse more libs than it's specified limit which may result in an out of bounds exceptions 

This was discovered while writing dlopen heavy applications that load thousands of shared objects

### Related issues
N/A

### Motivation and context
Prevent the profiler from crashing & exceeding it's limits 

### How has this been tested?
the dlopen heavy application is no longer crashing 

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -675,7 +675,7 @@ Mutex Symbols::_parse_lock;
 bool Symbols::_have_kernel_symbols = false;
 bool Symbols::_libs_limit_reported = false;
 static std::unordered_set<u64> _parsed_inodes;
-static bool in_parse_libraries = false;
+static bool _in_parse_libraries = false;
 
 void Symbols::parseKernelSymbols(CodeCache* cc) {
     int fd;
@@ -779,10 +779,10 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
 void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     MutexLocker ml(_parse_lock);
 
-    if (in_parse_libraries || array->count() >= MAX_NATIVE_LIBS) {
+    if (_in_parse_libraries || array->count() >= MAX_NATIVE_LIBS) {
         return;
     }
-    in_parse_libraries = true;
+    _in_parse_libraries = true;
 
     if (kernel_symbols && !haveKernelSymbols()) {
         CodeCache* cc = new CodeCache("[kernel]");
@@ -831,12 +831,12 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
         array->add(cc);
     }
 
-    in_parse_libraries = false;
-
     if (array->count() >= MAX_NATIVE_LIBS && !_libs_limit_reported) {
         Log::warn("Number of parsed libraries reached the limit of %d", MAX_NATIVE_LIBS);
         _libs_limit_reported = true;
     }
+
+    _in_parse_libraries = false;
 }
 
 // Check that the base address of the shared object has not changed
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -757,13 +757,15 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
         }
 
         if (map.isExecutable()) {
+            if (libs.find(inode) == libs.end()) {
+                max_count--;
+            }
             SharedLibrary& lib = libs[inode];
             if (lib.file == nullptr) {
                 lib.file = strdup(map.file());
                 lib.map_start = map_start;
                 lib.map_end = map_end;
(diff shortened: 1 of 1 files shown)


src/symbols_linux.cpp
```diff
@@ -827,6 +827,11 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
         cc->sort();
         applyPatch(cc);
         array->add(cc);
+
+        // Don't continue parsing libs as limit was reached
```
fandreuz, line 831: Do we need this comment? I'd rather log something to either `debug` or `info`

src/symbols_linux.cpp
```diff
@@ -827,6 +827,11 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
...
+        if (array->count() >= MAX_NATIVE_LIBS) {
```
fandreuz, line 832: This should be merged with the `if` below:
```cpp
 if (array->count() >= MAX_NATIVE_LIBS) {
     if (!_libs_limit_reported) {
        Log::warn("Number of parsed libraries reached the limit of %d", MAX_NATIVE_LIBS);
        _libs_limit_reported = true;
     }
     break;
 }
```

Baraa-Hasheesh: I would rather keep as is & avoid the nesting

fandreuz: But also, I'm curious why we used to check for the array size, but did not do anything concrete about it

Baraa-Hasheesh: Handled here [975d270](https://github.com/async-profiler/async-profiler/pull/1312/commits/975d270b202cd1728844f8b5c21c21407b2f6149)

src/symbols_linux.cpp
```diff
@@ -824,15 +824,21 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
 
         free(lib.file);
 
+        // Due to async-profiler issue #1273, it's possible for a recursive dlopen to happen at UnloadProtection
```
krk, line 827: Parsing already happened, should we prevent the parse instead of ignoring its results?

src/symbols_linux.cpp
apangin: `collectSharedLibraries()` call above should not return more libraries that can fit into the array, so this condition should never happen. If it is not the case, the bug is elsewhere.

Baraa-Hasheesh: That's not fully true due to #1273 
* Profiler already has 2047 libs
* Profiler will detect a dlopen on LIB1
* Profiler will use `/proc/self/maps` to detect new libs & finds LIB1
* UnloadProtection causes a recursive dlopen call to happen
* Profiler will use `/proc/self/maps` to detect new libs & finds LIB2
* UnloadProtection causes a recursive dlopen call to happen
* Profiler will use `/proc/self/maps` to detect new libs & finds none
* profiler continues & adds LIB2 to list making number of libs 2048
* profiler continues & adds LIB1 to list making number of libs 2049

However I will also fix the root condition as well while keeping this check

src/symbols_linux.cpp
Baraa-Hasheesh: handled here [6b3c053](https://github.com/async-profiler/async-profiler/pull/1312/commits/6b3c05340ee5668fe5ea594d681e7d385537d047)

src/symbols_linux.cpp
apangin: Unfortunately, this is not a proper fix either.
If you say that the problem is due to recursive `Symbols::parseLibraries` calls, then fix the root cause instead. `Symbols::parseLibraries` should not be called recursively.

apangin: `if (_in_parse_libraries) return;`
will it be a simple way to get rid of recursion?

apangin: `if (_in_parse_libraries) return;`
will it be a simple way to terminate recursion?

Baraa-Hasheesh: Handled here [f0c8199](https://github.com/async-profiler/async-profiler/pull/1312/commits/f0c81996da4386bb6a0c52a275cf890443d10cc5)

src/symbols_linux.cpp
```diff
@@ -829,6 +831,8 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
         array->add(cc);
     }
 
+    in_parse_libraries = false;
```
apangin, line 834: Looks good.
I'd only put this at very end of the function (that will better reflect the meaning of the flag).

Also, global variables in async-profiler are usually prefixed with `_`.

Baraa-Hasheesh: Done here [57ffd73](https://github.com/async-profiler/async-profiler/pull/1312/commits/57ffd730940c35e593b69a7bc3b13b3b2976bc44)

PR #1315, opened 2025-05-25
Enable unit tests related to parsing for MacOs

Baraa-Hasheesh:
### Description
As mentioned in description, Enabling more tests to run on Mac that were previously disabled

### Related issues
N/A

### Motivation and context
Enable as much of the available tests to run on all platforms

### How has this been tested?
`make test`

---

### DIFF
--- a/test/native/libs/reladyn.c
+++ b/test/native/libs/reladyn.c
@@ -31,7 +31,7 @@ int reladyn() {
     indirect_pthread_setspecific = pthread_setspecific;
     indirect_pthread_setspecific(key, "Thread-specific value");
 
-    // static pointer don't get reported in lazy nor none lazy symbol sections when discovering the symbols for MacOs
+    // static pointer doesn't get reported in the lazy or the non-lazy sections when discovering the symbols for MacOs
 #ifdef __APPLE__
     static_pthread_exit = pthread_exit;
 #endif
--- a/test/native/libs/reladyn.c
+++ b/test/native/libs/reladyn.c
@@ -31,6 +31,10 @@ int reladyn() {
     indirect_pthread_setspecific = pthread_setspecific;
     indirect_pthread_setspecific(key, "Thread-specific value");
 
+    // static pointer don't get reported in lazy nor none lazy symbol sections when discovering the symbols for MacOs
+#ifdef __APPLE__
+    static_pthread_exit = pthread_exit;
+#endif
     // Use pthread_exit via the static pointer, forces into .rela.dyn as R_X86_64_64.
     static_pthread_exit(NULL);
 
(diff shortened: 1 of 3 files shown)


test/native/libs/reladyn.c
```diff
@@ -31,6 +31,10 @@ int reladyn() {
     indirect_pthread_setspecific = pthread_setspecific;
     indirect_pthread_setspecific(key, "Thread-specific value");
 
+    // static pointer don't get reported in lazy nor none lazy symbol sections when discovering the symbols for MacOs
```
apangin, line 34: I wonder why `pthread_exit` doesn't appear among indirect symbols in this case? How does compiled code refer to the address of `pthread_exit` then?

On a side note, consider running text through a grammar checker: there are some issues here, including your favorite "none lazy" -> "non-lazy".

apangin: Why I asked this is to clarify the reasons behind changing the test on macOS. The comment about a missing indirect symbol did not sound a clear justification to me. At a first glance, this seems to be a legitimate test failure rather than a test bug. If so, then we need to either fix the root cause or decide this case is not worth supporting on macOS and disable the corresponding test.

Baraa-Hasheesh: @apangin From what I read it seems that modern compilation will cause the `pthread_exit` to be resolved immediately in this case & not be loaded as a `S_*_LAZY_SYMBOL_POINTERS`

In order to detect such cases we will need to parse `LC_DYLD_CHAINED_FIXUPS` & check what symbols were resolved/optimized, I did a small POC on my local branch & did find the `pthread_exit` in the chained fix-up chain 

This could mean to potentially patch such cases, we will need to detect which normal symbols in `__data` & `__const` section of `__DATA` & `__DATA_CONST` refer to the dynamic symbols that we need such as `pthread_exit` that got into the `LC_DYLD_CHAINED_FIXUPS` chain & then patch those in particular (could be risky)

It's my opinion that the current symbol parsing should be good enough for MacOs as it will handle most normal cases,
=> I would suggest keeping the DT_REL unit test disabled on MacOs.

What do you think, should we dive deeper into reading data under LC_DYLD_CHAINED_FIXUPS, or keep the current solution as is & have this as a known limitation?

apangin: Thanks for the update. I checked `LC_DYLD_CHAINED_FIXUPS` section and it does not look too hard to parse, but I agree we can skip it for now for the sake of simplicity.

Baraa-Hasheesh: handled here [3ca4568](https://github.com/async-profiler/async-profiler/pull/1315/commits/3ca456830aa6912a7f0f134a4be76057453d009b)

PR #1322, opened 2025-05-29
Protect from looping on interrupted `clone()` syscall

### LINKED ISSUE #1114: Protect from looping on interrupted clone() syscall
Discussed in #1113

fandreuz:
### Description
In this PR I introduce protection against infinite looping following an interrupted `clone` system call. The idea is to hook `clone`, check whether it was interrupted, and disable the `CpuEngine` for the thread while we retry.

### Related issues
Fixes #1114 

### Motivation and context
Async-Profiler should not try to interrupt `clone` again.

### How has this been tested?
I could not reproduce the problem locally yet.

---

### DIFF
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -197,6 +197,9 @@ void CodeCache::addImport(void** entry, const char* name) {
             if (strcmp(name, "calloc") == 0) {
                 saveImport(im_calloc, entry);
             }
+            if (strcmp(name, "clone") == 0) {
+                saveImport(im_clone, entry);
+            }
             break;
         case 'd':
             if (strcmp(name, "dlopen") == 0) {
--- a/src/hooks.cpp
+++ b/src/hooks.cpp
@@ -8,6 +8,8 @@
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cstdarg>
+#include <errno.h>
 #include "hooks.h"
 #include "asprof.h"
 #include "cpuEngine.h"
@@ -27,6 +29,9 @@ struct ThreadEntry {
     ThreadFunc start_routine;
     void* arg;
 };
+
+typedef int (*clone_t)(int (*__fn) (void *__arg), void *__child_stack, int __flags, void *__arg, ...);
+static clone_t _orig_clone = NULL;
 
 typedef int (*pthread_create_t)(pthread_t*, const pthread_attr_t*, ThreadFunc, void*);
 static pthread_create_t _orig_pthread_create = NULL;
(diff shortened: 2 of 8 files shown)


src/codeCache.cpp
```diff
@@ -197,6 +197,9 @@ void CodeCache::addImport(void** entry, const char* name) {
             if (strcmp(name, "calloc") == 0) {
                 saveImport(im_calloc, entry);
             }
+            if (strcmp(name, "clone") == 0) {
```
fandreuz, line 200: I see there are several `clone`: https://man7.org/linux/man-pages/man2/clone.2.html

How can we address this? Should we hook them individually?

Baraa-Hasheesh: I don't believe this solution won't work for MUSL & MacOs
can you double check if we should have separate solution for those? 

If you inspect a MacOs call stack it will end with `thread_start`

```
thread_start;_pthread_start;ThreadJavaMain;JavaMain;jni_CallStaticVoidMethod;jni_invoke_static;JavaCalls::call_helper;call_stub;test/stackwalker/Stackwalker.main_[0];test/stackwalker/Stackwalker.walkStackComplete_[0];Java_test_stackwalker_Stackwalker_walkStackComplete;doCpuTask
```

On Musl it will end with `/lib/ld-musl-*`

```
/lib/ld-musl-x86_64.so.1;ThreadJavaMain;JavaMain;jni_CallStaticVoidMethod;jni_invoke_static;JavaCalls::call_helper;call_stub;test/stackwalker/Stackwalker.main_[0];test/stackwalker/Stackwalker.walkStackComplete_[0];Java_test_stackwalker_Stackwalker_walkStackComplete;doCpuTask 14
```

src/hooks.cpp
```diff
@@ -192,6 +230,7 @@ void Hooks::patchLibraries() {
             // Let libasyncProfiler always use original dlopen
             cc->patchImport(im_dlopen, (void*)dlopen_hook);
         }
+        cc->patchImport(im_clone, (void*)clone_hook);
```
Baraa-Hasheesh, line 233: This location will never be called in Java APPs as hooks are handled differently for Java

src/perfEvents.h
```diff
@@ -32,6 +32,9 @@ class PerfEvents : public CpuEngine {
     int createForThread(int tid);
     void destroyForThread(int tid);
 
+    void pause() { destroyForThread(OS::threadId()); }
```
Baraa-Hasheesh, line 35: should this fix also be done for WALL profiling mode?

src/hooks.cpp
fandreuz: Will fix later, once the solution I adopted is agreed on

src/codeCache.cpp
fandreuz: This is potentially a problem which potentially affects any system call which starts from scratch when it is interrupted

Issue #1310, opened 2025-05-22
Native memory profiling causing infinite loops when using multiple profilers


Baraa-Hasheesh:
### Describe the bug

Native memory profiling will cause a recursive call chain between multiple profilers 

Let's assume we have 2 profilers, profiler_1 & profiler_2

When profiler_1 starts it will find all shared objects & patch them to monitor native memory allocations, this will include the shared objects of profiler_2

profiler_2 will also do the same which will include the shared objects of profiler_1

This means any malloc call in profiler_1 will call malloc hook in profiler_2
At same time any profiler malloc in profiler_2 will malloc hook in profiler_1

this will create a callback chain 
malloc_hook_profiler_1 ==> malloc_hook_profiler_2 ==> malloc_hook_profiler_1  ==> malloc_hook_profiler_2 ==> etc...

### Expected vs. actual behavior

Having multiple profilers shouldn't cause infinite loops

### Reproduction Steps

You can find the test here https://github.com/Baraa-Hasheesh/asprof-reproducers/tree/main/javadoubleprofilerinfiniteloop

* `make`
* `./runner.sh multi`

### Async-profiler version

latest

Baraa-Hasheesh:
resolved in #1323

PR #1323, opened 2025-05-29
Enable native memory profiling on the async-profiler

Baraa-Hasheesh:
### Description
Enable Native memory profiling on the async-profiler shared objects

Previously this profiling was removed as to solve #1266 

This PR keeps that fix by reading the GOT entry before it is overwritten by the profiler & uses it to call the correct malloc via the dynamic linker

### Related issues
N/A

### Motivation and context
Allow async-profiler to detect memory leaks inside it self

### How has this been tested?
`make test`

---

### DIFF
--- a/src/mallocTracer.cpp
+++ b/src/mallocTracer.cpp
@@ -20,13 +20,12 @@
 #  define NO_OPTIMIZE __attribute__((optimize("O1")))
 #endif
 
-#define SAVE_IMPORT(TYPE) ({ \
-    _orig_##TYPE = (decltype(_orig_##TYPE))*lib->findImport(im_##TYPE); \
-})
+#define SAVE_IMPORT(FUNC) \
+    _orig_##FUNC = (decltype(_orig_##FUNC))*lib->findImport(im_##FUNC)
 
 static void* (*_orig_malloc)(size_t);
 static void (*_orig_free)(void*);
-static void* (*_orig_calloc)(size_t,size_t);
+static void* (*_orig_calloc)(size_t, size_t);
 static void* (*_orig_realloc)(void*, size_t);
 static int (*_orig_posix_memalign)(void**, size_t, size_t);
 static void* (*_orig_aligned_alloc)(size_t, size_t);
@@ -104,7 +103,8 @@ int MallocTracer::_patched_libs = 0;
 bool MallocTracer::_initialized = false;
 volatile bool MallocTracer::_running = false;
 
-// Function guarantees that memory allocation functions will exist in the generated Shared Objects as to be parsed by 'OS_symbols' files/methods
+// Call each intercepted function at least once to ensure
+// its GOT entry is updated with a correct target address
 static void resolveMallocSymbols() {
     static volatile intptr_t sink;
 


Baraa-Hasheesh:
I've been doing additional testing on this & found that it should also fix #1310.

This is due to how resolution of original memory allocation function is made, as it's done via the original GOT entry value which will contain the correct malloc to call for the shared object provided by the dynamic linker (For one profiler).

* When the first profiler starts that will always be the real function to call provided by the linker.
* When the second profiler is loaded that will be modified during the dlopen by the first profiler to call the first profiler hook
* When the second profiler is started it will find the first profiler hook as the real function to call in the GOT entry

This means any future memory allocation will start to have the following flow after all 3 steps are concluded 
`malloc` -> `second profiler hook` -> `first profiler hook` -> `original malloc`

The order of native memory profiling sessions start between profilers won't be important as the first profiler to start will always have the `original malloc` in it's GOT entries which will stop any potential infinite recursive chain from ever happening

src/mallocTracer.cpp
```diff
@@ -97,6 +111,34 @@ void MallocTracer::initialize() {
     CodeCache* lib = Profiler::instance()->findLibraryByAddress((void*)MallocTracer::initialize);
     assert(lib);
 
+    // Guarantee that memory allocation functions will exist in the generated Shared Objects as to be parsed by 'OS_symbols' files/methods
+    void* ptr = calloc(1, 1);
+    if (ptr) free(ptr);
+
+    ptr = malloc(1);
+    if (ptr) {
```
apangin, line 119: `free` perfectly works with NULL pointer. Code can be a lot simpler without `if` checks.

src/mallocTracer.cpp
```diff
@@ -97,6 +111,34 @@ void MallocTracer::initialize() {
...
+
+    ptr = NULL;
+    if (posix_memalign(&ptr, 1, 1) == 0) {
+        free(ptr);
+    }
+
+    ptr = aligned_alloc(1, 1);
+    if (ptr) free(ptr);
+
+    _orig_malloc = (malloc_t)*lib->findImport(im_malloc);
```
apangin, line 135: Why not standard `dlsym()` approach?

Baraa-Hasheesh: Using `dlsym` will cause an absolute symbol resolution rather than a lazy symbol 
The resolved absolute symbol via `dlsym` could be the wrong one (Not the one we want to call)

Using RTLD_NEXT previously was doing that, as it was resolving the wrong `malloc` function to call

I did try using RTLD_DEFAULT & a dlopen handle to evaluate the symbols but both resulted in test failures

As such I decided on using the lazy dynamic symbols as it was the safest choice as it will be equivalent to the current state of async profiler of using the dynamic linker.

The dlopen handle caused the following test to fail (evaluating wrong malloc)
NativememTests.preloadMalloc/LD_PRELOAD+profiler_first
NativememTests.preloadMalloc/LD_PRELOAD+profiler_second
NativememTests.preloadMalloc/api_test

The RTLD_DEFAULT caused the following tests to fail (Infinite callback loop)
ativememTests.malloc_plt_dyn

src/mallocTracer.cpp
Baraa-Hasheesh: I did oppressive some some sigfaults on MacOs that made me add all the existing `if` statements just for safety 

I will double check which if statements can be removed (I believe I can remove all other than the `realloc` if check)

Baraa-Hasheesh: I reduced the checks here [3dfc520](https://github.com/async-profiler/async-profiler/pull/1323/commits/3dfc520feb8c842c89e5def72567e3d2afd81864)

I've kept checks before calling realloc & after calling `posix_memalign`

I also moved the symbols to their own function as I noticed that MacOs was optimizing some of them out when removing the if checks,
This new function is Marked to not be optimized

src/mallocTracer.cpp
apangin: I checked that `malloc_plt_dyn` test indeed fails on Linux, depending on the compiler version and flags.
If the tests program has `BIND_NOW` flag, the test passes; if not, `dlsym` returns an address of PLT stub instead of the actual libc implementation.

Unfortunately, current code and the proposed `findImport` approach may also both fail depending on the compiler options. For example, if libasyncProfiler is compiled with `-fno-plt` flag, `findImport` will also return an address of PLT stub, resulting in subsequent recursion.

src/mallocTracer.cpp
```diff
@@ -20,16 +20,30 @@
...
+typedef void* (*realloc_t)(void*, size_t);
+typedef int (*posix_memalign_t)(void**, size_t, size_t);
+typedef void* (*aligned_alloc_t)(size_t, size_t);
+
+malloc_t _orig_malloc;
+free_t _orig_free;
+calloc_t _orig_calloc;
+realloc_t _orig_realloc;
+posix_memalign_t _orig_posix_memalign;
+aligned_alloc_t _orig_aligned_alloc;
```
apangin: All file-local variables and functions must be declared `static`.

src/mallocTracer.cpp
```diff
@@ -20,16 +20,30 @@
 #  define NO_OPTIMIZE __attribute__((optimize("O1")))
 #endif
 
+typedef void* (*malloc_t)(size_t);
+typedef void (*free_t)(void*);
+typedef void* (*calloc_t)(size_t,size_t);
+typedef void* (*realloc_t)(void*, size_t);
+typedef int (*posix_memalign_t)(void**, size_t, size_t);
+typedef void* (*aligned_alloc_t)(size_t, size_t);
```
apangin: Too many typedefs. It's possible to avoid all these definitions just by declaring `_orig_*` function with a proper type once and then using `decltype`.

src/mallocTracer.cpp
```diff
@@ -93,10 +107,34 @@ int MallocTracer::_patched_libs = 0;
...
     assert(lib);
 
+    guaranteeSymbols();
+
+    _orig_malloc = (malloc_t)*lib->findImport(im_malloc);
+    _orig_free = (free_t)*lib->findImport(im_free);
+    _orig_calloc = (calloc_t)*lib->findImport(im_calloc);
+    _orig_realloc = (realloc_t)*lib->findImport(im_realloc);
+    _orig_posix_memalign = (posix_memalign_t)*lib->findImport(im_posix_memalign);
+    _orig_aligned_alloc = (aligned_alloc_t)*lib->findImport(im_aligned_alloc);
```
apangin: Each function name is repeated 3 times, which makes it harder to read and to modify.
This is one the cases where macros are really good (althogh they are generally discouraged).
I'd write something like:
```
SAVE_ORIG(malloc);
SAVE_ORIG(free);
...
```

src/mallocTracer.cpp
Baraa-Hasheesh: @apangin thanks for the follow up analysis on `malloc_plt_dyn` & `dlsym`

Regarding the compile with `-fno-plt`, I tried compiling the `async-profiler` with `-fno-plt` however I didn't encounter any infinite loops
```
make clean; CFLAGS_EXTRA=-fno-plt CXXFLAGS_EXTRA=-fno-plt make test
```
can you provide me what gcc version you're currently using & how the compilation was made?

src/mallocTracer.cpp
Baraa-Hasheesh: handled here [a4a36cd](https://github.com/async-profiler/async-profiler/pull/1323/commits/a4a36cdf15ead816ef3f7f03913f7bf13033a55d)

src/mallocTracer.cpp
Baraa-Hasheesh: handled here [a4a36cd](https://github.com/async-profiler/async-profiler/pull/1323/commits/a4a36cdf15ead816ef3f7f03913f7bf13033a55d)

src/mallocTracer.cpp
Baraa-Hasheesh: handled here [a4a36cd](https://github.com/async-profiler/async-profiler/pull/1323/commits/a4a36cdf15ead816ef3f7f03913f7bf13033a55d)

src/mallocTracer.cpp
Baraa-Hasheesh: @apangin I was doing some experimentation & managed to create an infinite loop on async-profiler master branch 

This was done using the following command
```
CFLAGS_EXTRA="-fno-plt -Wl,-Bsymbolic" CXXFLAGS_EXTRA="-fno-plt -Wl,-Bsymbolic" make test TESTS=nativemem;  
```

**Same command doesn't fail on this branch**

From what I can tell, this happens as when the PLT stubs were removed, the async-profiler ended up using a reference for malloc that was later patched/changed by the profiler.

However this doesn't happen in this branch as the profiler saves the old Reference before it's changed by the profiler, hence avoiding any infinite loop possibility.

so effectively this change shouldn't have infinite loops as it doesn't depend on stubs existing or not,
It simply depends on the dynamic linker writing a correct value that would end up calling the correct malloc. 

As long as a correct value is written, an infinite loop should never happen as we save that value/reference before any patching is done to use it as our final callback (Breaks any potential chain).

apangin: I tried `make CXXFLAGS="-O3 -fno-plt" test`
`CXXFLAGS="-O3"` alone breaks some tests, but `-fno-plt` breaks even more.
(note these are `CXXFLAGS`, not `CXXFLAGS_EXTRA`)

src/mallocTracer.cpp
apangin: Unfortunately, newer GCC versions still optimize `calloc` and `aligned_alloc` in this code, leading to a crash.
I came up with [this implementation](https://github.com/async-profiler/async-profiler/compare/master...nativemem-test#diff-1a4d0bfbcdde34a1836c8bd36b8432b5f62afbe56df09dda8a35fcf9161cf0f5R106-R120) that retains all calls even with maximum optimization level.

src/mallocTracer.cpp
apangin: I mean, this was already a problem in existing code. Your change does not make it worse, so I'm happy to accept it once remaining issues are fixed.

src/mallocTracer.cpp
Baraa-Hasheesh: Thanks @apangin the implementation does look great, 
I will add it on my branch

Baraa-Hasheesh: @apangin applied here [376eeb0](https://github.com/async-profiler/async-profiler/pull/1323/commits/376eeb0a278b7e3896ddc8fa73d2508b09076df6)

src/mallocTracer.cpp
Baraa-Hasheesh: @apangin Thanks for the provided information

I've applied the latest feedback here [376eeb0](https://github.com/async-profiler/async-profiler/pull/1323/commits/376eeb0a278b7e3896ddc8fa73d2508b09076df6)

Issue #1319, opened 2025-05-27
SIGSEGV in PerfEvents::walk when profiling Renaissance akka-uct benchmark


WadeWalker:
### Describe the bug

Hi guys,

I love async-profiler, so thank you all sincerely for the work you do!

Anyway, to the bug report :) When I profile the Renaissance sub-benchmark akka-uct on Amazon EC2 m6a.metal instances, I pretty repeatably get a HotSpot SIGSEGV (file at [hs_err_pid4147.log](https://github.com/user-attachments/files/20462603/hs_err_pid4147.log)):

C  [libasyncProfiler.so+0x18109]  PerfEvents::walk(int, void*, void const**, int, StackContext*) [clone .constprop.1221]+0x29

It happens on other instance types too (m6i.metal, m7a.metal, m7g.metal, m7i.metal), so it affects both x86 and arm64. But it's most common on the m6a.metal instances, which are AMD EPYC 7R13 machines.

The error is somewhat intermittent. When I run it on most instance types, it might fail once or twice, but it'll eventually work. On m6a.metal, it's never run to completion for me.

I've tried to give a very complete description below, but please let me know if you need any more info.

Thanks!

### Expected vs. actual behavior

Expected behavior is for the profile to run to completion. Actual behavior is SIGSEGV after some amount of time. The unsuccessful output of the benchmark looks like this:

```
====== akka-uct (concurrency) [default], iteration 0 started ======
GC before operation: completed in 24.973 ms, heap usage 62.562 MB -> 11.223 MB.
final size= 199991
final height= 9
final size= 199991
final height= 9
final size= 199991
final height= 9
final size= 199991
final height= 9
final size= 199991
final height= 9
final size= 199991
final height= 9
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000079116ac9d109, pid=20286, tid=22946
#
# JRE version: OpenJDK Runtime Environment Temurin-24.0.1+9 (24.0.1+9) (build 24.0.1+9)
# Java VM: OpenJDK 64-Bit Server VM Temurin-24.0.1+9 (24.0.1+9, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  [libasyncProfiler.so+0x18109]  PerfEvents::walk(int, void*, void const**, int, StackContext*) [clone .constprop.1221]+0x29
#
[...]

PR #1324, opened 2025-05-30
#1319: Accessing osThreadId of a terminating thread may fail

apangin:
### Description

The problem happens when a sample lands in HotSpot's `Thread::~Thread()` in a short window between

```
  if (osthread() != nullptr) os::free_thread(osthread());
```
and
```
  if (this == Thread::current_or_null()) {
    Thread::clear_thread_current();
  }
```
and freed `osthread` memory is concurrently reused by some other thread.

### Related issues

#1319

### How has this been tested?

Fix confirmed by the original bug reporter.
New stress test is underway.

---

### DIFF
--- a/src/safeAccess.h
+++ b/src/safeAccess.h
@@ -52,11 +52,14 @@ class SafeAccess {
     }
 
     static uintptr_t skipLoadArg(uintptr_t pc) {
-#if defined(__aarch64__)
         if ((pc - (uintptr_t)load32) < 16 || (pc - (uintptr_t)loadPtr) < 16) {
-            return 4;
-        }
+#if defined(__x86_64__) || defined(__i386__)
+            if (*(u8*)pc == 0x8b) return 2;      // mov eax, [reg]
+            if (*(u16*)pc == 0x8b48) return 3;   // mov rax, [reg]
+#else
+            return sizeof(instruction_t);
 #endif
+        }
         return 0;
     }
 };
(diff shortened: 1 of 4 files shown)


src/safeAccess.h
```diff
@@ -52,11 +52,14 @@ class SafeAccess {
...
     static uintptr_t skipLoadArg(uintptr_t pc) {
-#if defined(__aarch64__)
         if ((pc - (uintptr_t)load32) < 16 || (pc - (uintptr_t)loadPtr) < 16) {
-            return 4;
-        }
+#if defined(__x86_64__) || defined(__i386__)
+            if (*(u8*)pc == 0x8b) return 2;      // mov eax, [reg]
+            if (*(u16*)pc == 0x8b48) return 3;   // mov rax, [reg]
+#else
+            return sizeof(instruction_t);
```
jbachorik, line 60: Just out of curiosity - why is this needed in this PR?

apangin: `VMThread::osThreadId()` now uses `SafeAccess::load32` to protect from unlikely case when a terminating thread is sampled after its `osthread` structure has been already freed, but the thread has not been yet excluded from profiling.

jbachorik: I see. But the updated skip logic is generic, right?

apangin: Previously, usages of `SafeAccess::load32` were on aarch64 only. Now it's called on all platforms and therefore `skipLoadArg` is adjusted to support other architectures.

jbachorik: This explains it completely. Thanks!

Issue #1327, opened 2025-06-05
Fix memory leaks reported by Infer


tteokbokki-master:
### Describe the bug

Hello :),

The `read_response()` function in `jattach_openj9.c` contains multiple code paths where dynamically allocated memory is not freed, which can lead to memory leaks.

1. When `read()` returns `0` (EOF) or a value less than `0` (indicating an error), the function exits early without freeing the buffer allocated by `malloc` (lines 126 and 129)
2. If `realloc()` fails and returns `NULL`, the original memory is lost because the result is assigned directly to `buf`, resulting in a classic `realloc` memory leak (line 140)

These issues were identified through static analysis using `Infer` and occur in code paths that return without calling `free()`

### Expected vs. actual behavior

The expected behavior is that all memory allocated via `malloc()` or `realloc()` should be freed in every return path of the function.

The actual behavior is that memory is not freed in the following code paths:

- If `read()` returns `0` (EOF), `buf` is not freed (line 126)
- If `read()` returns `< 0` (error), `buf` is not freed (line 129)
- If `realloc()` fails, `buf` is overwritten with `NULL`, making the originally allocated memory unreachable (line 140)

### Reproduction Steps

This issue is visible via static analysis using Infer.

1. Run `Infer` on the project:
    
    ```bash
    infer run -- make
    ```
    

2.  You’ll see memory leak warnings reported in `read_response()`

### Additional Information/Context

- Minor but meaningful memory cleanup
- Discovered using static analysis tool (Infer)
- Not a crash bug, but improves memory safety

### Async-profiler version

latest

### Environment details

- Source file: `src/jattach/jattach_openj9.c`
- Function: `read_response`
- Analyzed with: Infer static analyzer
- Platform: macOS (Apple Silicon), Clang 17

tteokbokki-master:
I've opened [PR #1328](https://github.com/async-profiler/async-profiler/pull/1328) to address this issue (commit: [`d2ebbda`](https://github.com/async-profiler/async-profiler/pull/1328/commits/d2ebbda1189c1f14d84e39be0219d63465dd0c15))

Thanks in advance for reviewing!

apangin:
Integrated

PR #1328, opened 2025-06-05
Fix memory leaks reported by Infer

tteokbokki-master:
### Description
This pull request fixes memory leaks in the `read_response()` function (`jattach_openj9.c`) identified via static analysis using Infer.

- Ensures `buf` is freed when `read()` returns `0` or `< 0`
- Adds `NULL` check after `realloc()` to avoid losing the original buffer
- All exit paths now properly release allocated memory

These are minor but important fixes to improve memory safety

### Related issues
#1327

### Motivation and context
Improve memory safety by ensuring all dynamic memory is properly freed in early return paths and in case of allocation failure

This change was guided by results from the Infer static analyzer

### How has this been tested?
Verified using Infer that the memory leak warnings no longer appear after the patch

---

### DIFF
--- a/src/jattach/jattach_openj9.c
+++ b/src/jattach/jattach_openj9.c
@@ -125,9 +125,11 @@ static int read_response(int fd, const char* cmd, int print_output) {
         ssize_t bytes = read(fd, buf + off, size - off);
         if (bytes == 0) {
             fprintf(stderr, "Unexpected EOF reading response\n");
+            free(buf);
             return 1;
         } else if (bytes < 0) {
             perror("Error reading response");
+            free(buf);
             return 1;
         }
 
@@ -137,7 +139,13 @@ static int read_response(int fd, const char* cmd, int print_output) {
         }
 
         if (off >= size) {
-            buf = realloc(buf, size *= 2);
+            char* temp_buf = realloc(buf, size *= 2);
+            if (temp_buf == NULL) {
+                free(buf);
+                fprintf(stderr, "Failed to reallocate memory for response\n");
+                return 1;
+            }
+            buf = temp_buf;
         }
     }
 


src/jattach/jattach_openj9.c
```diff
@@ -137,7 +139,13 @@ static int read_response(int fd, const char* cmd, int print_output) {
         }
 
         if (off >= size) {
-            buf = realloc(buf, size *= 2);
+            char* temp_buf = realloc(buf, size *= 2);
+            if (temp_buf == NULL) {
+                free(buf);
```
fandreuz, line 144: I think `realloc` is responsible for freeing the old buffer if needed, am I missing something?

apangin:
Thank you for your interest in async-profiler. Generally, we do not accept PRs whose only purpose is to satisfy static analyzers. Proposed changes must address real problems.

While I agree with freeing a buffer after read error (although it is not a real leak, since this code path belongs to a short-lived process which exits immediately after attach), the second part of the change (realloc) does not make sense to me. When `realloc` fails, meaning out of memory condition, process has already run into unrecoverable state and will likely crash in the next moment. Replacing one `realloc` with 7 lines of futile code only adds noise. This fix looks even more questionable to me because it covers only one instance of `realloc`, whereas async-profiler has a lot more.

Please note that `jattach/*` source files are copied verbatim from the sister project [`jattach`](https://github.com/jattach/jattach). To propose a change in these sources, open an issue/PR in jattach project instead. Thanks.

apangin:
I'm closing this PR for the reasons mentioned above.
However, I'll be happy to merge the fix in `jattach` project with the exception of `realloc` part.

tteokbokki-master:
Thank you for your feedback :)

Based on your suggestion, I’ve opened a [PR](https://github.com/jattach/jattach/pull/84) in the jattach project that includes only the buffer free fix, leaving out the realloc part

I’d appreciate it if you could take a look when convenient.

Issue #1317, opened 2025-05-27
A bug report about VM Structs stack walker on Alpine


yanglong1010:
### Describe the bug

Hi,

Using my reproducer code, I found the following problems, and I'm investigating...

- On Alpine 3.10,  the stacks are not correct if use VM Structs stack walker, while the stacks of `dwarf` and `agct` walker are correct.
- On Alpine 3.21, the stacks are not correct if use VM Structs stack walker, the stacks of `dwarf` and `agct` walker are not correct too (many `unknown` frames between kernel frames and java frames), but still readable. If use VM Structs stack walker, I have to add `--jstackdepth 512`, otherwise the html file can not be displayed by Chrome browser, the some stacks are too deep (default 2048), beyond the capability of Chrome.

#### Alpine 3.10
- JDK 8 (cstack vm) [alpine3.10-jdk8-vm.html.zip](https://github.com/user-attachments/files/20447730/alpine3.10-jdk8-vm.html.zip) (the picture is bellow)

![Image](https://github.com/user-attachments/assets/78225c18-c05c-4054-874b-f7851d16503d)

- JDK 8 (cstack dwarf)  [alpine3.10-jdk8-dwarf.html.zip](https://github.com/user-attachments/files/20447777/alpine3.10-jdk8-dwarf.html.zip)

![Image](https://github.com/user-attachments/assets/3a7c6623-23dd-43b1-84b4-6112756c8d17)

- JDK 8 (cstack agct) [alpine3.10-jdk8-agct.html.zip](https://github.com/user-attachments/files/20447838/alpine3.10-jdk8-agct.html.zip)

![Image](https://github.com/user-attachments/assets/be8fd678-1f68-457d-aadd-bcd9da9eaf8c)

- JDK 11 (cstack vm)[alpine3.10-jdk11-vm.html.zip](https://github.com/user-attachments/files/20447749/alpine3.10-jdk11-vm.html.zip)

![Image](https://github.com/user-attachments/assets/b7605809-b433-4ea0-bfd6-a5a58009590e)

- JDK 11 (cstack dwarf) [alpine3.10-jdk11-dwarf.html.zip](https://github.com/user-attachments/files/20447793/alpine3.10-jdk11-dwarf.html.zip)

![Image](https://github.com/user-attachments/assets/c5132f45-e0ba-4820-a74c-308d2fba7e3c)

[...]

PR #1329, opened 2025-06-05
A bug report about VM Structs stack walker on Alpine (#1317)

yanglong1010:
### Description
Hi, could I have a review of this patch ?
If `FrameDesc` can not be found for the `pc`, then it means we don't know how to unwind the current native frame.
Instead, we add a `skipped frames` dummy frame to indicate some frames are skipped, and use `Java Frame Anchor` to unwind the Java part.

The current version of `CodeCache::findFrameDesc` may return an incorrect `FrameDesc`, as described in this comment:
https://github.com/async-profiler/async-profiler/issues/1317#issuecomment-2911569497
I added a `fde_loc_end` to `FrameDesc`, and `CodeCache::findFrameDesc` checks the `pc` is in the range of the returned `FrameDesc`'s FDE.

I deleted the redundant `FrameDesc` to fix the possible bug described in this comment:
https://github.com/async-profiler/async-profiler/issues/1317#issuecomment-2918545682

This fix also works on JDK 11 aarch64, as I described in the comment:
https://github.com/async-profiler/async-profiler/issues/1317#issuecomment-2933584880

### Related issues
#1317

### How has this been tested?
[Reproducer.java.zip](https://github.com/user-attachments/files/20607620/Reproducer.java.zip)
```
$JAVA_HOME/bin/java Reproducer
/root/async-profiler/build/bin/asprof -e cpu -f /root/profile.html -d 10 --cstack vm jps
```
Without the patch, most native stacks are incomplete, missing the Java stack.
image

With the patch, native stacks and Java stacks are linked together
image

---

### DIFF
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -302,6 +302,11 @@ FrameDesc* CodeCache::findFrameDesc(const void* pc) {
     }
 
     if (low > 0) {
+      	if (target_loc < _dwarf_table[low - 1].fde_loc_end) {
+			return &_dwarf_table[low - 1];
+      	} else {
+			return NULL;
+      	}
         return &_dwarf_table[low - 1];
     } else if (target_loc - _plt_offset < _plt_size) {
         return &FrameDesc::empty_frame;
--- a/src/dwarf.cpp
+++ b/src/dwarf.cpp
@@ -55,8 +55,8 @@ enum {
 };
 
 
-FrameDesc FrameDesc::empty_frame = {0, DW_REG_SP | EMPTY_FRAME_SIZE << 8, DW_SAME_FP, -EMPTY_FRAME_SIZE};
-FrameDesc FrameDesc::default_frame = {0, DW_REG_FP | LINKED_FRAME_SIZE << 8, -LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE + DW_STACK_SLOT};
+FrameDesc FrameDesc::empty_frame = {0, DW_REG_SP | EMPTY_FRAME_SIZE << 8, DW_SAME_FP, -EMPTY_FRAME_SIZE, 0};
+FrameDesc FrameDesc::default_frame = {0, DW_REG_FP | LINKED_FRAME_SIZE << 8, -LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE + DW_STACK_SLOT, 0};
 
 
 DwarfParser::DwarfParser(const char* name, const char* image_base, const char* eh_frame_hdr) {
@@ -124,12 +124,12 @@ void DwarfParser::parseFde() {
 
     u32 range_start = getPtr() - _image_base;
     u32 range_len = get32();
+    u32 fde_loc_end = range_start + range_len;
     _ptr += getLeb();
-    parseInstructions(range_start, fde_start + fde_len);
-    addRecord(range_start + range_len, DW_REG_FP, LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE + DW_STACK_SLOT);
+    parseInstructions(range_start, fde_start + fde_len, fde_loc_end);
 }
 
-void DwarfParser::parseInstructions(u32 loc, const char* end) {
+void DwarfParser::parseInstructions(u32 loc, const char* end, u32 fde_loc_end) {
     const u32 code_align = _code_align;
     const int data_align = _data_align;
 
@@ -153,15 +153,15 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
                         _ptr = end;
                         break;
                     case DW_CFA_advance_loc1:
-                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off);
+                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off, fde_loc_end);
                         loc += get8() * code_align;
                         break;
                     case DW_CFA_advance_loc2:
-                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off);
+                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off, fde_loc_end);
                         loc += get16() * code_align;
                         break;
                     case DW_CFA_advance_loc4:
-                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off);
+                        addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off, fde_loc_end);
                         loc += get32() * code_align;
                         break;
                     case DW_CFA_offset_extended:
@@ -257,7 +257,7 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
                 }
                 break;
             case DW_CFA_advance_loc:
-                addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off);
+                addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off, fde_loc_end);
                 loc += (op & 0x3f) * code_align;
                 break;
             case DW_CFA_offset:
@@ -275,7 +275,7 @@ void DwarfParser::parseInstructions(u32 loc, const char* end) {
         }
     }
 
-    addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off);
+    addRecord(loc, cfa_reg, cfa_off, fp_off, pc_off, fde_loc_end);
 }
 
 // Parse a limited subset of DWARF expressions, which is used in DW_CFA_val_expression
@@ -332,15 +332,15 @@ int DwarfParser::parseExpression() {
     return pc_off;
 }
 
-void DwarfParser::addRecord(u32 loc, u32 cfa_reg, int cfa_off, int fp_off, int pc_off) {
+void DwarfParser::addRecord(u32 loc, u32 cfa_reg, int cfa_off, int fp_off, int pc_off, u32 fde_loc_end) {
     int cfa = cfa_reg | cfa_off << 8;
     if (_prev == NULL || (_prev->loc == loc && --_count >= 0) ||
             _prev->cfa != cfa || _prev->fp_off != fp_off || _prev->pc_off != pc_off) {
-        _prev = addRecordRaw(loc, cfa, fp_off, pc_off);
+        _prev = addRecordRaw(loc, cfa, fp_off, pc_off, fde_loc_end);
     }
 }
 
-FrameDesc* DwarfParser::addRecordRaw(u32 loc, int cfa, int fp_off, int pc_off) {
+FrameDesc* DwarfParser::addRecordRaw(u32 loc, int cfa, int fp_off, int pc_off, u32 fde_loc_end) {
     if (_count >= _capacity) {
         _capacity *= 2;
         _table = (FrameDesc*)realloc(_table, _capacity * sizeof(FrameDesc));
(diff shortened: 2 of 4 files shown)


fandreuz:
Hi @yanglong1010, thanks for this PR. It seems that your changes would break some of our [tests](https://github.com/async-profiler/async-profiler/actions/runs/15465720643/job/43537266307?pr=1329) on Alpine. Could you please have a look?

yanglong1010:
> Hi @yanglong1010, thanks for this PR. It seems that your changes would break some of our [tests](https://github.com/async-profiler/async-profiler/actions/runs/15465720643/job/43537266307?pr=1329) on Alpine. Could you please have a look?

I added `sp` check, and the test can pass now.

src/codeCache.cpp
```diff
@@ -302,6 +302,11 @@ FrameDesc* CodeCache::findFrameDesc(const void* pc) {
     }
 
     if (low > 0) {
+      	if (target_loc < _dwarf_table[low - 1].fde_loc_end) {
+            return &_dwarf_table[low - 1];
+      	} else {
+            return NULL;
+      	}
         return &_dwarf_table[low - 1];
```
Baraa-Hasheesh: unreachable return statement.

src/dwarf.h
```diff
@@ -66,6 +66,7 @@ struct FrameDesc {
     int cfa;
     int fp_off;
     int pc_off;
+    u32 fde_loc_end;
```
apangin, line 69: This will increase unwind tables, which already take several megabytes, by 25%.
At the same time, in a normal executable with a valid .eh_frame, FDEs typically follow one after another without gaps.

src/stackWalker.cpp
```diff
@@ -388,6 +397,30 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
         uintptr_t prev_sp = sp;
         CodeCache* cc = profiler->findLibraryByAddress(pc);
         FrameDesc* f = cc != NULL ? cc->findFrameDesc(pc) : &FrameDesc::default_frame;
+        if (f == NULL) {
+            if (anchor && anchor->lastJavaSP() != 0) {
+                if (detail == VM_EXPERT && (unwound_from_anchor || has_java_frame)) break;
```
apangin: This condition looks strange.
`anchor` cannot be used more than once regardless of unwinding mode.

src/stackWalker.cpp
```diff
@@ -255,11 +262,13 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
             }
             return depth;
         }
+        anchor = vm_thread->anchor();
     }
 
     // Walk until the bottom of the stack or until the first Java frame
     while (depth < max_depth) {
         if (CodeHeap::contains(pc)) {
+            has_java_frame = true;
```
apangin: This is not always true. Besides Java methods, `CodeHeap` contains stubs that can be called from non-Java threads.

src/dwarf.cpp
```diff
@@ -124,12 +124,12 @@ void DwarfParser::parseFde() {
 
     u32 range_start = getPtr() - _image_base;
     u32 range_len = get32();
+    u32 fde_loc_end = range_start + range_len;
     _ptr += getLeb();
-    parseInstructions(range_start, fde_start + fde_len);
-    addRecord(range_start + range_len, DW_REG_FP, LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE, -LINKED_FRAME_SIZE + DW_STACK_SLOT);
+    parseInstructions(range_start, fde_start + fde_len, fde_loc_end);
```
apangin, line 129: I remember there were cases (in libc?) when a few functions in otherwise normal library were not covered by FDEs. This extra record helped to unwind such frames using frame pointer. Now, this will result in a broken stack trace.
This code is very fragile, and I'm very reluctant to modifying it without comprehensive study of consequences on different OSes and JDKs.

src/stackWalker.cpp
yanglong1010: This condition is added before [`SP` check](https://github.com/yanglong1010/async-profiler/blob/2a0f607964f114b933f6d0bfecb84efc3c1b961a/src/stackWalker.cpp#L414-L417) does not exist yet.

Without `SP` check, the reason to add VM_EXPERT to condition is:

- If use VM_BASIC mode, the stack walker recovers JavaFrameAnchors from stack until the end of Java part, and then skips the methods from call_stub to platform-dependent root frame, and ends the walking.

- But if use VM_EXPERT mode, after unwinding the Java part, the stack walker will unwind the native part (from call_stub to platform-dependent root frame), however, on alpine, the functions in musl can not be unwinded (because the missing FDEs), so if now we use Java Frame Anchor from Java Thread to continue unwinding, then we can get impossible stack like bellow, even infinite loops until reaches the stack depth limit. So VM_EXPERT mode needs `unwound_from_anchor` to prevent infinite loops, nees `has_java_frame` to prevent invalid duplicate.
image

But, after the `SP` check is added, the line `if (detail == VM_EXPERT && (unwound_from_anchor || has_java_frame)) break;` is not needed any more. Because `sp < prev_sp` check achieves the same effect.

src/stackWalker.cpp
yanglong1010: The later added [`SP` check](https://github.com/yanglong1010/async-profiler/blob/2a0f607964f114b933f6d0bfecb84efc3c1b961a/src/stackWalker.cpp#L414-L417) achieves the same effect, this variable has been deleted.

src/codeCache.cpp
yanglong1010: Thanks, pushed ~

src/dwarf.cpp
yanglong1010: I didn't notice the code here avoids duplicate FrameDescs. 
https://github.com/async-profiler/async-profiler/blob/981619680ea6c93768df6253e3a6d87d4c33cd2b/src/dwarf.cpp#L337

The change has been reverted.

src/dwarf.h
yanglong1010: How about adding an FDE table, the size of FDE table is much less than the size of FrameDesc table, then save more memory.

```C
struct FDE {
    u32 loc_start; // inclusive
    u32 loc_end; // exclusive
    u32 frame_desc_idx_start; // inclusive
    u32 frame_desc_idx_end; // inclusive
}
```

Another benefit is, we can boost the speed of FrameDesc searching. Instead of search the FDEs, we first search the    FDE table, and then use the FDE's min and max FrameDesc index as the range to search the FrameDesc.


```
                          FDE 0                  FDE 1                    FDE 2
       FDEs         |-----------------|     |-----------------|     |------------------|
              (loc) 0                 10    10                30    30                 100

    (FrameDesc idx) 0                 2     3                 4     5                  7
    
                    |                 |     |                 |     |                  |
                    V                 V     V                 V     V                  V

       FrameDescs   |---| |------| |--|     |--------| |------|     |----| |---| |-----| 
              (loc) 0   3 4      8 9  10    11      15 16     30    31  35 36 40 40    100

```

apangin: If I understand correctly, you propose to create a 2-level table instead of a flat table.
How will this save memory, if total number of FrameDescs remains the same? In the current design, at most 1 trailing FrameDesc is added for each FDE (but often none, when there are no gaps between FDEs). In the proposed solution, there will be always a FDE of the same size.
And how will this increase speed? If lookups on both levels use binary search, the algorithm complexity will be the same as with one flat table, since log(ab) = log(a) + log(b). Or do you propose something else?

yanglong1010: - Assume that each FDE has 10 FrameDescs, and If we have 1000 FDEs, `add 8 bytes per FrameDesc` vs `add 16 bytes per FDE`, then the result is `1000 * 8` vs `16 * (1000 / 10)`, that is 8000 vs 1600. So, using 2-level table saves more memory than adding a field to each FrameDesc. Yes, more often, there is no gaps between FDEs, so the extra FrameDesc added in the last FDE is overwritten by the first FrameDescs in the next FDE. Compares to your original design, both of my proposals require more memory. But in order to make sure the returned `FrameDesc` really belongs to the containing `FDE` of the `pc`, otherwise we fallback to `Java Frame Anchor`, I think the overhead to save the address range for each `FDE` is necessary, or add this check on Alpine only ?

-  It's my mistake, the speed is the same.

src/dwarf.cpp
yanglong1010: > I remember there were cases (in libc?) when a few functions in otherwise normal library were not covered by FDEs. This extra record helped to unwind such frames using frame pointer. Now, this will result in a broken stack trace.

In this case, using frame pointer may also fail to unwind. Is fall back to `Java Frame Anchor` better ?

src/dwarf.h
apangin: > add 8 bytes per FrameDesc

That's exactly what I propose against. Instead, you can add a special FrameDesc at the end of each FDE. In a good case, it will be overwritten by a FrameDesc from the next FDE. In the worst case, it will be just 16 extra bytes per FDE. That's what async-profiler currently does, except that it inserts a default FrameDesc instead of a special marker.
As I mentioned, always replacing a default FrameDesc with a marker is likely a wrong thing to do, since it will break otherwise legitimate use cases. So, the main question here is how to differentiate these two situations when unwinding info is completely unreliable vs. when a few FDEs are missing.

yanglong1010: If we can get the function symbol count and the FDE count, then it is easy to judge. Unfortunately, the symbols in the `.symtab` are optional, in a stripped library, there is no symbol in `.symtab`.

I propose a method:
- Read the size of the `.text` sections, this is the size of instructions.
- Sum the pc ranges of all FDEs from `.eh_frame`, then we get the instruction size covered by FDEs.
- If over XX% of instructions are covered by FDEs, we think a few FDEs are missing, otherwise, we think unwinding info is completely missing.

In my environment, the size of `.text` if 285417 bytes, while the total pc range of all FDEs is 2425, that's only 0.8%.
```
~ # readelf --sections /lib/ld-musl-x86_64.so.1
There are 20 section headers, starting at offset 0x8d530:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 8] .text             PROGBITS         0000000000015060  00015060
       0000000000045ae9  0000000000000000  AX       0     0     1
```

```
Contents of the .eh_frame section:


00000000 0000000000000014 00000000 CIE "zR" cf=1 df=-8 ra=16
   LOC           CFA      ra
0000000000000000 rsp+8    c-8

00000018 0000000000000024 0000001c FDE cie=00000000 pc=0000000000015000..0000000000015060
   LOC           CFA      ra
0000000000015000 rsp+16   c-8
0000000000015006 rsp+24   c-8
0000000000015010 exp      c-8

00000040 0000000000000014 00000044 FDE cie=00000000 pc=000000000005a230..000000000005a4b0

00000058 0000000000000014 0000005c FDE cie=00000000 pc=000000000005a4b0..000000000005a759

00000070 0000000000000010 00000074 FDE cie=00000000 pc=000000000005a759..000000000005ab49
```
> So, the main question here is how to differentiate these two situations when unwinding info is completely unreliable vs. when a few FDEs are missing.

- If unwinding info is missing, then we fallback to use Java Frame Anchor
- Otherwise, try to find the FrameDesc (Including the default FrameDesc)

We can add an option to adjust this threshold.

What do you think ?

src/stackWalker.cpp
```diff
@@ -255,6 +260,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
             }
             return depth;
         }
+        anchor = vm_thread->anchor();
```
apangin, line 263: When I came across this issue again I realized that it's illegal to access `JavaFrameAnchor` for an arbitrary `VMThread`. This is only safe for `JavaThread`.

apangin:
Fallback to JavaFrameAnchor was implemented in #1517.
I'm closing this PR as outdated.

PR #1331, opened 2025-06-06
Test OTLP output format

fandreuz:
### Description
- Added Protobuf generated Java classes to the repository
- Added `Makefile` machinery to update Protobuf generated classes
- Added high-level OTLP tests in `OtlpTests` to validate the format with a real-world Protobuf reader

Remarks:
- New tests in this PR are failing on purpose, this PR needs #1188 (which is yet to be merged). I kept them separated to simplify the diff.
- I'm using `otel/build-protobuf` from [open-telemetry/build-tools](https://github.com/open-telemetry/build-tools) to compile Protobuf classes. Unfortunately it's using an old version of `libprotobuf` (open-telemetry/build-tools#410), and this forces us to use an old runtime version of Protobuf in Java (protocolbuffers/protobuf#19540). Hopefully this gets resolved soon, otherwise we can either:
    + Use `protoc` directly
    + Build our own Docker image

### Related issues
Needs #1188 

### Motivation and context
We must properly test OTLP output with high-level tests (e.g. the profile contains frame `X`) and that the output we write is well-formed.

### How has this been tested?
Tests are failing on purpose now, see remarks above.

---


fandreuz:
@apangin Protobuf generated files are failing the license test, they have an autogenerated header which looks like this:
```
// Generated by the protocol buffer compiler.  DO NOT EDIT!
// source: opentelemetry/proto/profiles/v1development/profiles.proto

package io.opentelemetry.proto.profiles.v1development;
...
```

We can fix this easily, but do we need to?

Makefile
```diff
@@ -285,3 +297,29 @@ clean-coverage:
 
 clean:
 	$(RM) -r build
+
+.ONESHELL:
+update-otlp:
```
fandreuz: `update-otlp-java-classes`

Makefile
```diff
@@ -285,3 +300,34 @@ clean-coverage:
...
 clean:
 	$(RM) -r build
+
+$(TEST_DEPS_DIR)/$(PROTOBUF_JAVA_JAR): Makefile
+	mkdir -p $(TEST_DEPS_DIR)
+	curl -o $(TEST_DEPS_DIR)/$(PROTOBUF_JAVA_JAR) $(PROTOBUF_JAVA_JAR_URL)
+	echo "$(PROTOBUF_JAVA_JAR_SHA256)  $(TEST_DEPS_DIR)/$(PROTOBUF_JAVA_JAR)" | $(SHA256SUM) -c
+
+.ONESHELL:
+update-otlp-classes-jar: $(TEST_DEPS_DIR)/$(PROTOBUF_JAVA_JAR)
```
apangin: A full page of git and docker command is not something that should appear in Makefile recipe.

fandreuz: Moved to a separate script in 090459ce681c457d768dc7821a9ab41a41d226af

test/test/otlp/OtlpTests.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import one.profiler.test.*;
+import io.opentelemetry.proto.profiles.v1development.*;
+
+public class OtlpTests {
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,otlp,file=%f.pb")
+    public void testOtlpReadable(TestProcess p) throws Exception {
+        ProfilesData profilesData = waitAndGetProfileData(p);
```
apangin: ProfilesData or ProfileData?

Makefile
```diff
@@ -285,3 +303,26 @@ clean-coverage:
 
 clean:
 	$(RM) -r build
+
+$(TEST_DEPS_DIR)/$(PB_JAVA_JAR): Makefile
+	mkdir -p $(TEST_DEPS_DIR)
```
apangin: Do you download it in a git-managed directory?

Makefile
```diff
@@ -285,3 +303,26 @@ clean-coverage:
...
+
+update-otlp-classes-jar: $(TEST_DEPS_DIR)/$(PB_JAVA_JAR)
+	$(RM) -rf $(OTEL_PROTO_PATH)
+	git clone --depth 1 git@github.com:open-telemetry/opentelemetry-proto.git $(OTEL_PROTO_PATH)
+	# Classes related to logs, metrics and traces are not needed
+	find $(OTEL_PROTO_PATH)/opentelemetry/proto/ \
+		-type f \( -name 'logs*.proto' -o -name 'metrics*.proto' -o -name 'trace*.proto' -o -name '*service.proto' \) \
+		-delete
+
+	$(RM) -rf $(PB_BIN_PATH)*
```
apangin: That's brave

Makefile
```diff
@@ -285,3 +303,26 @@ clean-coverage:
...
+	curl -L -o $(PB_BIN_PATH).zip $(PB_BIN_URL)
+	unzip $(PB_BIN_PATH).zip -d $(PB_BIN_PATH)
+
+	mkdir -p $(OTEL_PROTO_PATH)/gen/java
+	cd $(OTEL_PROTO_PATH) && $(PB_BIN_PATH)/bin/protoc --java_out=./gen/java $$(find . -name "*.proto")
+	find $(OTEL_PROTO_PATH)/gen/java -name "*.java" | \
+		xargs $(JAVAC) -source $(JAVA_TARGET) -target $(JAVA_TARGET) -cp $(TEST_DEPS_DIR)/$(PB_JAVA_JAR) -d $(OTEL_PROTO_PATH)/build
```
apangin: This calls `javac` on each file independently? Can be a single javac invocation.

test/test/otlp/OtlpTests.java
fandreuz: It's `ProfilesData`: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L134

Makefile
fandreuz: It's in `test/deps`, but I only commit `opentelemetry-gen-classes.jar`. Should I include this too in the VCS?

Makefile
fandreuz: 33753efe80fb00a1930e5100f044c02eb99e6f1f

Makefile
fandreuz: I made it a bit more sane here: a638566ed4f585d0be1dbe6b1e70d374336f54ad

test/test/otlp/OtlpTests.java
apangin: I know its `ProfilesData` in the protocol. The question was about the function called `*GetProfile_Data` that returns `Profile_s_Data`

fandreuz: Yeah I noticed, I fixed it here: 30aa65327137ce7de5f791cbf7a8d62b64a733cc

Makefile
apangin: No. You should not mix VCS managed files and dependencies in the same directory, and you can't just include third-party binaries in the source repository without following license conditions of the respective project.

Makefile
```diff
@@ -269,9 +283,13 @@ coverage: clean-coverage
 
 test: test-cpp test-java
 
-build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
+build/$(TEST_JAR): build/$(API_JAR) $(TEST_SOURCES) build/$(CONVERTER_JAR) $(TEST_DEPS_DIR)/$(PB_JAVA_JAR)
+	if [ ! -f "$(TEST_DEPS_DIR)/$(OTEL_PROTO_JAR)" ]; then \
+		echo "Missing $(TEST_DEPS_DIR)/$(OTEL_PROTO_JAR), build it with 'make update-otlp-classes-jar'"; \
```
apangin: How can it be missing, if it's in the repo?

fandreuz: Right, maybe this check is not needed

Makefile
fandreuz: Point taken, this is the tree now:
```
test/
├── deps
│   └── protobuf-java.jar
├── gen
│   └── opentelemetry-gen-classes.jar
```

fandreuz:
@apangin all comments addressed, I've also made the simplification we discussed [here](https://github.com/async-profiler/async-profiler/pull/1336#discussion_r2152274784)

Makefile
```diff
@@ -130,6 +133,18 @@ ifneq (,$(findstring $(ARCH_TAG),x86 x64 arm64))
   CXXFLAGS += -momit-leaf-frame-pointer
 endif
 
+# OTLP Protobuf
+OTEL_PROTO_PATH=$(TMP_DIR)/opentelemetry-proto
+OTEL_PROTO_JAR=opentelemetry-gen-classes.jar
+PB_JAVA_VERSION=4.31.1
+PB_JAVA_JAR_SHA256=d60dfe7c68a0d38a248cca96924f289dc7e1966a887ee7cae397701af08575ae
+PB_JAVA_JAR_URL=https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/$(PB_JAVA_VERSION)/protobuf-java-$(PB_JAVA_VERSION).jar
```
apangin: I already told that Makefile should not generally download anything, and after thinking more on the topic, I'm coming to conclusion this applies to dependencies too. Once we start downloading dependencies in the build system, we'll have to go through all burden of dependency management. Furthermore, there are build environments where it's not allowed to download 3p dependencies from untrusted sources.

Makefile can have a variable through which user provides path to dependencies. If dependency is not found, it fails with an error message clearly saying what is missing. It's then up to user to take care of downloads. This is similar to how OpenJDK build system works, for example.

fandreuz: That makes things a bit uglier since people will always have to download `protobuf-java.jar` manually, otherwise `make build-test-java` will always fail to compile.

fandreuz: 99b51f40513845e75156482c557579aa89bb88d3

Makefile
```diff
@@ -52,6 +53,8 @@ JAVAC_OPTIONS=--release $(JAVA_TARGET) -Xlint:-options
 
 TEST_LIB_DIR=build/test/lib
 TEST_BIN_DIR=build/test/bin
+TEST_DEPS_DIR=test/deps
```
apangin, line 56: Should it be added to `.gitignore`?

Makefile
```diff
@@ -65,6 +68,7 @@ JAVA_HELPER_CLASSES := $(wildcard src/helper/one/profiler/*.class)
 API_SOURCES := $(wildcard src/api/one/profiler/*.java)
 CONVERTER_SOURCES := $(shell find src/converter -name '*.java')
 TEST_SOURCES := $(shell find test -name '*.java')
+TEST_SOURCES_WITHOUT_OTEL := $(shell echo $(TEST_SOURCES) | tr ' ' '\n' | grep -v otlp)
```
apangin: Test classes do not depend on protobuf-java.jar, why can't they be compiled anyway?

Makefile
```diff
@@ -269,11 +272,35 @@ coverage: clean-coverage
...
+		-type f \
+		-name '*.proto' \
+		-not \( -name 'logs*.proto' -o -name 'metrics*.proto' -o -name 'trace*.proto' -o -name '*service.proto' \) \
+		| tr '\n' ' ')
+	$(JAVAC) -source $(JAVA_TARGET) \
+	    -target $(JAVA_TARGET) \
+	    -cp $(TEST_DEPS_DIR)/* \
+	    -d $(TMP_DIR)/build \
+		-Xlint:-options \
+	    $$(find $(TMP_DIR)/gen/java -name "*.java" | tr '\n' ' ')
```
apangin: Is `tr` really needed?

src/profiler.cpp
```diff
@@ -1661,43 +1661,36 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
-
         otlp_buffer.commitMessage(sample_mark);
 
-        frames_seen += si.num_frames;
+        for (int j = 0; j < trace->num_frames; j++) {
+            // To be written below in Profile.location_indices
+            location_indices.push_back(functions.indexOf(fn.name(trace->frames[j])));
+        }
+        frames_seen += trace->num_frames;
+    }
```
apangin: This change looks unrelated, I'd commit it separately.

test/test/otlp/OtlpTests.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,otlp,file=%f.pb")
+    public void testSamples(TestProcess p) throws Exception {
+        ProfilesData profilesData = waitAndGetProfilesData(p);
+
+        Profile profile = getFirstProfile(profilesData);
+        ProfilesDictionary dictionary = profilesData.getDictionary();
+
+        List<String> collapsed = toCollapsed(profile, dictionary);
+        assert collapsed.stream().anyMatch(s -> s.contains("test/otlp/CpuBurner.main;test/otlp/CpuBurner.burn"));
```
apangin: What if `toCollapsed` returned `Output` object, so we could use common methods like `Output.contains`?

src/profiler.cpp
fandreuz: Right, I wanted to get it merged but it doesn't really belong here indeed. Reverted in 2194ba810ad68eab24204bc9660c5eda4d07f2d4 and re-applied in #1361

Makefile
fandreuz: They do, for example this is the declaration of  `io.opentelemetry.proto.profiles.v1development.Profile`:
```java
public final class Profile extends com.google.protobuf.GeneratedMessageV3 implements ProfileOrBuilder
```

test/test/otlp/OtlpTests.java
fandreuz: Sure: 90ea6f588457c84dbbe29c99601bbccc85115f69

Makefile
fandreuz: Right: f0afb499d8ab0e3b8b8183ae75ce1cffec441f3b

Makefile
fandreuz: 37240345752ae13617934cd24466f42fd392e6ba

Makefile
```diff
@@ -269,11 +272,35 @@ coverage: clean-coverage
...
 	$(JAR) cf $@ -C build/test .
 
+update-otlp-classes-jar:
+	rm -rf $(TMP_DIR)/gen/java $(TMP_DIR)/build
+	mkdir -p $(TMP_DIR)/gen/java $(TMP_DIR)/build $(TEST_GEN_DIR)
+	cd $(OTEL_PROTO_PATH) && protoc --java_out=$(TMP_DIR)/gen/java $$(find . \
+		-type f \
+		-name '*.proto' \
+		-not \( -name 'logs*.proto' -o -name 'metrics*.proto' -o -name 'trace*.proto' -o -name '*service.proto' \) \
+		| tr '\n' ' ')
```
apangin: Redundant `tr`?

Makefile
apangin: Normally, class dependencies are not transitive. `OtlpTests.java` refers to classes in `opentelemetry-gen-classes.jar`, but not in `protobuf-java.jar`. But you are right: in this case, `javac` misses several protobuf classes. I found experimentally that only 5 protobuf classes are needed to compile OtlpTests. In fact, not even full classes - [stubs](https://gist.github.com/apangin/4da093f4ac9b3f2ff7cb13737c3d7499) are enough.

In that way, if we provide tiny `protobuf-stubs.jar` along with `opentelemetry-gen-classes.jar`, otlp tests will become compilable. They will still require full `protobuf-java.jar` at runtime, though.

Alternatively, it is also fine to fail build if `protobuf-java.jar` is missing, although less preferable to me.

Makefile
fandreuz: Right: e86aa22658833a7bdc63797d93e8e459b1171284

Makefile
fandreuz: Actually we also thought about something like this while discussing the matter with @krk, but we thought of stubbing stuff in `io.opentelemetry.proto.profiles.v1development`, not in `com.google.protobuf`, which after some more thinking I deemed unfeasible. 

I like your solution much more, I'll update the PR

fandreuz: @apangin This is the output in the latest revision if the real `protobuf-java.jar` is not provided:
```
Total test duration: 9.534 s
Results Summary:
PASS: 0 (NaN%)
FAIL: 0
SKIP (disabled): 0
SKIP (config mismatch): 0
SKIP (missing JAR): 3
TOTAL: 3
```

fandreuz: If it is provided, tests are going to run fine ignoring the stubs, which are the last jars in the classpath.

Do you think I should commit the `.java` version of the stub to the repository, and some Makefile recipe to regenerate the stub jar?

apangin: I noticed your stub jar contains `GeneratedMessageV3.class` - is it intentional? It was not needed for compilation when I tried.

Also, since these classes are never used at runtime, you can compile them with `-g:none` to reduce jar size.

fandreuz: Right, I added it to fix an error and then I forgot about it. Removed in e35f50faf62642453aa13f9e499599d260b0db63

> Also, since these classes are never used at runtime, you can compile them with -g:none to reduce jar size.

👍

Makefile
```diff
@@ -52,6 +53,9 @@ JAVAC_OPTIONS=--release $(JAVA_TARGET) -Xlint:-options
 
 TEST_LIB_DIR=build/test/lib
 TEST_BIN_DIR=build/test/bin
+TEST_DEPS_DIR=test/deps
+TEST_GEN_DIR=test/gen
+TEST_STUBS_DIR=test/stub-jars
```
apangin: Maybe put it in `test/gen`? Why another directory?
I'd also rename file to `protobuf-stubs.jar` to make it clear it has nothing to do with the original `protobuf-java`, i.e., it's not a derived work and is not subject to the protobuf license T&C.

Makefile
```diff
@@ -269,12 +273,31 @@ coverage: clean-coverage
 
 test: test-cpp test-java
 
-build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
+build/$(TEST_JAR): build/$(API_JAR) $(TEST_SOURCES) build/$(CONVERTER_JAR) $(TEST_DEPS_DIR) $(TEST_STUBS_DIR)
+	rm -rf build/test/classes
 	mkdir -p build/test/classes
-	$(JAVAC) -source $(JAVA_TARGET) -target $(JAVA_TARGET) -Xlint:-options -XDignore.symbol.file \
-	         -cp "build/jar/*:build/converter/*" -d build/test/classes $(TEST_SOURCES)
+	$(JAVAC) -source $(JAVA_TARGET) -target $(JAVA_TARGET) -Xlint:-options -cp "build/jar/*:$(TEST_DEPS_DIR)/*:$(TEST_GEN_DIR)/*:$(TEST_STUBS_DIR)/*" -d build/test/classes $(TEST_SOURCES)
```
apangin: Please resolve merge conflict, you've overwritten my previous change.

Makefile
fandreuz: Mixing stubs with other dirs makes it hard to reason about the jar order in the classpath. As it is now, we now for sure that stubs will always appear in the rightmost part of the classpath definition.

Makefile
apangin: I just thought of an alternative approach: what if we add these 5 protobuf stubs as Java sources in the repository under `/test` but do not include them in `TEST_SOURCES`? Then, if we add `-implicit:none` option, javac will not generate `.class` files for them, but will be able to compile OTLP tests normally.

Advantages:
- no another binary dependency, no extra generation step;
- stubs can be easily viewed and modified.

Drawbacks:
- 5 new files in `com.google.protobuf` package in async-profiler sources.

WDYT?

fandreuz: @apangin actually I got this error again:
```
test/test/otlp/OtlpTests.java:32: error: cannot access GeneratedMessageV3
        assert profile.getSampleTypeList().size() == 2;
                      ^
  class file for com.google.protobuf.GeneratedMessageV3 not found
1 error
```
so I guess `GeneratedMessageV3` is needed? This happens when I build `test.jar` with protobuf stub

apangin: Interesting. What javac version do you use?

fandreuz: It's `javac 21.0.7`

fandreuz: > I just thought of an alternative approach: [...]

I think I prefer this approach, I guess it's going to make it harder for us to lose track of where the stub `com.google.protobuf` is coming from. I'll give it a shot.

fandreuz: @apangin check the latest revision, I implemented your new idea

Makefile
fandreuz: 412ed552bc1798d0796c9cbfad716d6439da952a

test/stubs/com/google/protobuf/ByteString.java
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.google.protobuf;
+
+abstract class ByteString {}
```
apangin: Make them public? It was not possible when all classes were in the same .java file, but it's no longer the case.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -209,6 +209,12 @@ jobs:
           esac
           echo "jars_directory=jar_artifacts" >> $GITHUB_OUTPUT
           echo "release_directory=$(basename $(find . -type d -iname "async-profiler-*" ))" >> $GITHUB_OUTPUT
+      - name: Download Protobuf Java runtime
+        run: |
+          mkdir -p "test/deps"
+          curl -L -o "test/deps/protobuf-java.jar" "https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/$PB_JAVA_VERSION/protobuf-java-$PB_JAVA_VERSION.jar"
```
apangin: It should be fine to leave the original file name (with a version).

test/test/otlp/OtlpTests.java
apangin: Where is it?

test/test/otlp/CpuBurner.java
```diff
@@ -0,0 +1,29 @@
...
+
+package test.otlp;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Random;
+
+public class CpuBurner {
+    private static final Random RANDOM = new Random();
+    private static final Duration TEST_DURATION = Duration.ofSeconds(3);
```
apangin: I'd reduce test duration to save CI time; 1-2 seconds is probably enough.

Makefile
```diff
@@ -269,12 +272,35 @@ coverage: clean-coverage
 
 test: test-cpp test-java
 
-build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
+build/$(TEST_JAR): build/$(API_JAR) $(TEST_SOURCES) build/$(CONVERTER_JAR) $(TEST_DEPS_DIR)
+	rm -rf build/test/classes
```
apangin, line 276: Why is that? Generally, Makefile targets should prefer incremental builds.

Makefile
```diff
@@ -269,12 +272,35 @@ coverage: clean-coverage
...
+	mkdir -p $(TMP_DIR)/gen/java $(TMP_DIR)/build $(TEST_GEN_DIR)
+	cd $(OTEL_PROTO_PATH) && protoc --java_out=$(TMP_DIR)/gen/java $$(find . \
+		-type f \
+		-name '*.proto' \
+		-not \( -name 'logs*.proto' -o -name 'metrics*.proto' -o -name 'trace*.proto' -o -name '*service.proto' \))
+	$(JAVAC) -source $(JAVA_TARGET) \
+	    -target $(JAVA_TARGET) \
+	    -cp $(TEST_DEPS_DIR)/* \
+	    -d $(TMP_DIR)/build \
+		-Xlint:-options \
```
apangin: Something wrong with spacing.

Makefile
apangin: A change disappeared again - is that intentional?

fandreuz: Probably some git hiccup on my side, re-applied in 32966f45ddba55df93137664dc0a5c07ccab08ff

Makefile
fandreuz: Should be better now: a17c8ca

test/stubs/com/google/protobuf/ByteString.java
fandreuz: Sure, but does it change anything? fcbc9ad

.github/workflows/test-and-publish-nightly.yml
fandreuz: 80240a8

test/test/otlp/CpuBurner.java
fandreuz: 1a5776f

Makefile
fandreuz: Let's say I delete a directory under `tests`. Unless I'm missing something, the `.class` file will still be there until `build/test/classes` directory is cleaned manually.

test/test/otlp/OtlpTests.java
fandreuz: Mmm I must have got a merge conflict wrong at some point, I'm pushing both on my work laptop and personal laptop. Re-applied here: a850cbe

PR #1335, opened 2025-06-11
Prevent early start

Baraa-Hasheesh:
### Description
Currently async-profiler `ASPROF_COMMAND` env variable is used as an indication to start the profiler in PRELOAD mode. 
However this check isn't accurate & could be used to indirectly start the profiler when any `dlopen` happens for the profiler shared library, which introduces the following problems.
* Native(C/C++) profiler will start in an incorrect manner causing it to not install it's hooks correctly making it unable to monitor the process correctly.
* General problem of unintended control vector as it's possible for `ASPROF_COMMAND` env variable to automatically start the profiler with incorrect profiling option.

The introduced check is a best effort check, as it uses `dladdr` to resolve the file name which internally will return the relative path for the shared library & check that against the value of `LD_PRELOAD` via a string contain.

This check could fail due to the fact we're using  string contain operation on a relative path which could have potentially multiple matches

Note: `DYLD_INSERT_LIBRARIES` isn't checked as profiler doesn't currently PRELOAD over MacOs as that would require interposing & other changes to the profiler

### Related issues
N/A

### Motivation and context
Remove `ASPROF_COMMAND` control vector for `dlopen` of the profiler

### How has this been tested?
New tests have been added under test/test/control/ControlTests.java 
`make test`

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -107,7 +107,7 @@ jobs:
             name: ubuntu-latest
         java-version: [8, 11, 17, 21, 24]
         java-distribution: [corretto]
-        image: [public.ecr.aws/async-profiler/asprof-builder-x86:latest]
+        image: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
         include:
           - runson:
               display: macos-arm64
--- a/docs/OutputFormats.md
+++ b/docs/OutputFormats.md
@@ -58,3 +58,6 @@ async-profiler currently supports the following output formats:
   about the JVM as well as the Java application running on it. async-profiler can generate output in `jfr` format
   compatible with tools capable of viewing and analyzing `jfr` files. JDK Mission Control (JMC) and Intellij IDEA are
   some of many options to visualize `jfr` files. More details [here](JfrVisualization.md).
+
+- `otlp` - OpenTelemetry protocol format for [profiling data](https://opentelemetry.io/blog/2024/profiling).
+  Experimental feature: backward-incompatible changes may happen in future releases of async-profiler.
--- a/docs/ProfilerOptions.md
+++ b/docs/ProfilerOptions.md
@@ -114,5 +114,6 @@ By default, async-profiler merges stack traces starting from the outermost (e.g.
 - `flamegraph` - produce Flame Graph in HTML format.
 - `tree` - produce Call Tree in HTML format.
   - `--reverse` option will generate backtrace view.
+- `otlp` - dump events in OpenTelemetry format.
 
 It is possible to specify multiple dump options at the same time.
--- a/src/api/one/profiler/AsyncProfiler.java
+++ b/src/api/one/profiler/AsyncProfiler.java
@@ -232,6 +232,22 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
     }
 
     /**
+     * Dump collected data in OTLP format.
+     * <p>
+     * This API is UNSTABLE and might change or be removed in the next version of async-profiler.
+     *
+     * @return OTLP representation of the profile
+     */
+    @Override
+    public byte[] dumpOtlp() {
+        try {
+            return execute1("otlp");
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    /**
      * Add the given thread to the set of profiled threads.
      * 'filter' option must be enabled to use this method.
      *
@@ -270,6 +286,8 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
     private native void stop0() throws IllegalStateException;
 
     private native String execute0(String command) throws IllegalArgumentException, IllegalStateException, IOException;
+
+    private native byte[] execute1(String command) throws IllegalArgumentException, IllegalStateException, IOException;
 
     private native void filterThread0(Thread thread, boolean enable);
 }
--- a/src/api/one/profiler/AsyncProfilerMXBean.java
+++ b/src/api/one/profiler/AsyncProfilerMXBean.java
@@ -29,4 +29,5 @@ public interface AsyncProfilerMXBean {
     String dumpCollapsed(Counter counter);
     String dumpTraces(int maxTraces);
     String dumpFlat(int maxMethods);
+    byte[] dumpOtlp();
 }
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -69,6 +69,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     jfrsync[=CONFIG] - start Java Flight Recording with the given config along with the profiler
 //     traces[=N]       - dump top N call traces
 //     flat[=N]         - dump top N methods (aka flat profile)
+//     otlp             - dump in OpenTelemetry format
 //     samples          - count the number of samples (default)
 //     total            - count the total value (time, bytes, etc.) instead of samples
 //     chunksize=N      - approximate size of JFR chunk in bytes (default: 100 MB)
@@ -199,6 +200,9 @@ Error Arguments::parse(const char* args) {
             CASE("flat")
                 _output = OUTPUT_TEXT;
                 _dump_flat = value == NULL ? INT_MAX : atoi(value);
+
+            CASE("otlp")
+                _output = OUTPUT_OTLP;
 
             CASE("samples")
                 _counter = COUNTER_SAMPLES;
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -76,7 +76,8 @@ enum SHORT_ENUM Output {
     OUTPUT_COLLAPSED,
     OUTPUT_FLAMEGRAPH,
     OUTPUT_TREE,
-    OUTPUT_JFR
+    OUTPUT_JFR,
+    OUTPUT_OTLP
 };
 
 enum JfrOption {
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -778,11 +778,11 @@ class Recording {
         buf->putVar32(0);
         buf->putVar32(0x7fffffff);  // must not clash with JFR metadata ID, or 'jfr print' will break
 
-        std::vector<std::string>& strings = JfrMetadata::strings();
+        const Index& strings = JfrMetadata::strings();
         buf->putVar32(strings.size());
-        for (int i = 0; i < strings.size(); i++) {
-            buf->putUtf8(strings[i].c_str());
-        }
+        strings.forEachOrdered([&] (const std::string& s) {
+            buf->putUtf8(s.c_str());
+        });
 
         writeElement(buf, JfrMetadata::root());
 
--- a/src/index.h
+++ b/src/index.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _INDEX_H
+#define _INDEX_H
+
+#include <functional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+#include "arch.h"
+
+// Keeps track of values seen and their index of occurrence
+class Index {
+  private:
+    std::unordered_map<std::string, size_t> _idx_map;
+  
+  public:
+    Index() = default;
+
+    Index(const Index&) = delete;
+    Index(Index&&) = delete;
+    Index& operator=(const Index&) = delete;
+    Index& operator=(Index&&) = delete;
+
+    size_t indexOf(const std::string& value) {
+        return _idx_map.insert({value, _idx_map.size()}).first->second;
+    }
+
+    size_t size() const {
+        return _idx_map.size();
+    }
+
+    void forEachOrdered(std::function<void(const std::string&)> consumer) const {
+        std::vector<const std::string*> arr(_idx_map.size());
+        for (const auto& it : _idx_map) {
+            arr[it.second] = &it.first;
+        }
+        for (size_t idx = 0; idx < size(); ++idx) {
+            consumer(*arr[idx]);
+        }
+    }
+};
+
+#endif // _INDEX_H
--- a/src/javaApi.cpp
+++ b/src/javaApi.cpp
@@ -95,6 +95,37 @@ Java_one_profiler_AsyncProfiler_execute0(JNIEnv* env, jobject unused, jstring co
     return NULL;
 }
 
+extern "C" DLLEXPORT jbyteArray JNICALL
+Java_one_profiler_AsyncProfiler_execute1(JNIEnv* env, jobject unused, jstring command) {
+    Arguments args;
+    const char* command_str = env->GetStringUTFChars(command, NULL);
+    Error error = args.parse(command_str);
+    env->ReleaseStringUTFChars(command, command_str);
+
+    if (error) {
+        throwNew(env, "java/lang/IllegalArgumentException", error.message());
+        return NULL;
+    }
+    if (args.hasOutputFile()) {
+        throwNew(env, "java/lang/IllegalArgumentException", "execute1 calls should not specify an output file argument");
+        return NULL;
+    }
+
+    Log::open(args);
+
+    BufferWriter out;
+    // TODO: This is doing one more copy than necessary, from ProtoWriter to BufferWriter
+    error = Profiler::instance()->runInternal(args, out);
+    if (error) {
+        throwNew(env, "java/lang/IllegalStateException", error.message());
+        return NULL;
+    }
+
+    jbyteArray output = env->NewByteArray(out.size());
+    env->SetByteArrayRegion(output, 0, out.size(), (const jbyte*) out.buf());
+    return output;
+}
+
 extern "C" DLLEXPORT jlong JNICALL
 Java_one_profiler_AsyncProfiler_getSamples(JNIEnv* env, jobject unused) {
     return (jlong)Profiler::instance()->total_samples();
@@ -124,6 +155,7 @@ static const JNINativeMethod profiler_natives[] = {
     F(start0,        "(Ljava/lang/String;JZ)V"),
     F(stop0,         "()V"),
     F(execute0,      "(Ljava/lang/String;)Ljava/lang/String;"),
+    F(execute1,      "(Ljava/lang/String;)[B"),
     F(getSamples,    "()J"),
     F(filterThread0, "(Ljava/lang/Thread;Z)V"),
 };
--- a/src/jfrMetadata.cpp
+++ b/src/jfrMetadata.cpp
@@ -6,8 +6,7 @@
 #include "jfrMetadata.h"
 
 
-std::map<std::string, int> Element::_string_map;
-std::vector<std::string> Element::_strings;
+Index Element::_strings;
 
 JfrMetadata JfrMetadata::_root;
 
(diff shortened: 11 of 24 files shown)


test/test/control/ControlTests.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.control;
+
+import one.profiler.test.Output;
+import one.profiler.test.Test;
+import one.profiler.test.TestProcess;
+
+public class ControlTests {
```
fandreuz, line 15: `Control` is a bit vague

test/test/control/ControlTests.java
```diff
@@ -0,0 +1,22 @@
...
+    @Test(sh = "%testbin/start_control %f1.jfr", env = {"ASPROF_COMMAND=start,cpu,file=%f2.jfr"}, nameSuffix="fake-preload")
```
fandreuz: Can we use better qualifiers than `1` and `2`?

test/test/control/start_control.c
```diff
@@ -0,0 +1,79 @@
...
+void* getSymbol(void* lib, const char* name) {
+    void* ptr = dlsym(lib, name);
+    if (ptr == NULL) {
+        fprintf(stderr, "dlsym error: %s\n", dlerror());
+        exit(1);
+    }
+    return ptr;
+}
+
+
```
fandreuz: Double newline

test/test/control/ControlTests.java
```diff
@@ -0,0 +1,22 @@
...
+    @Test(sh = "%testbin/start_control %f1.jfr", nameSuffix="normal")
```
fandreuz: Two questions:
- why is this `@Test` needed?
- can/should we also check that `f2.jfr` is empty? Correct me if I'm wrong, AsyncProfiler is not `LD_PRELOADED`, `ASPROF_COMMAND` will be ignored and the file won't exist

src/zInit.cpp
```diff
@@ -32,7 +32,17 @@ class LateInitializer {
...
   private:
-    bool checkJvmLoaded() {
+    static bool checkPreload(Dl_info dl_info) {
+        const char* preload = getenv("LD_PRELOAD");
+        if (preload == NULL) {
+            return false;
+        }
+
+        // prevent profiler from auto starting on dlopen when ASPROF_COMMAND is defined
+        return strstr(preload, dl_info.dli_fname) != NULL;
```
apangin: Does this work for symlinks?
There should be a better way to check if the library was preloaded or not.

test/test/control/ControlTests.java
Baraa-Hasheesh: I'm open for suggestion here, I named it control to indicate it tests how the async profiler is controlled via the various options

src/zInit.cpp
```diff
@@ -16,14 +16,14 @@ class LateInitializer {
   public:
     LateInitializer() {
         Dl_info dl_info;
-        if (!OS::isMusl() && dladdr((const void*)Hooks::init, &dl_info) && dl_info.dli_fname != NULL) {
+        if (dladdr((const void*)Hooks::init, &dl_info) && dl_info.dli_fname != NULL) {
             // Make sure async-profiler DSO cannot be unloaded, since it contains JVM callbacks.
             // This is not relevant for musl, where dlclose() is no-op.
             // Can't use ELF NODELETE flag because of https://sourceware.org/bugzilla/show_bug.cgi?id=20839
-            dlopen(dl_info.dli_fname, RTLD_LAZY | RTLD_NODELETE);
+            if (!OS::isMusl()) dlopen(dl_info.dli_fname, RTLD_LAZY | RTLD_NODELETE);
```
roy-soumadipta: This can be moved to the existing check instead of creating a new nested condition

test/test/control/ControlTests.java
Baraa-Hasheesh: Both tests are needed as I'm testing for both `ASPROF_COMMAND` is defined or not

I will add the check for `f2.jfr`

src/zInit.cpp
Baraa-Hasheesh: I think I have found a way to do it. I will update the logic,

I will use dladdr to find the shared object related to `dlopen`

If I find that dlopen is part of the current shared object I will proceed if not I will stop

Basically I will do a similar check to the one inside => UnloadProtection::UnloadProtection

Baraa-Hasheesh: I think I found a way to do it.

```
CodeCache* dlopen_cc = Profiler::instance()->findLibraryByAddress((const void*)dlopen);
CodeCache* current_cc = Profiler::instance()->findLibraryByAddress((const void*)Hooks::init);

return (void*)dlopen_cc == (void*)current_cc;
```

src/zInit.cpp
Baraa-Hasheesh: Condition reverted back to original [cec75dc](https://github.com/async-profiler/async-profiler/pull/1335/commits/cec75dc222c2e7462fea91902798417c0c2b61cb)

test/test/control/ControlTests.java
Baraa-Hasheesh: handled here [cec75dc](https://github.com/async-profiler/async-profiler/pull/1335/commits/cec75dc222c2e7462fea91902798417c0c2b61cb)

test/test/control/start_control.c
Baraa-Hasheesh: handled here [cec75dc](https://github.com/async-profiler/async-profiler/pull/1335/commits/cec75dc222c2e7462fea91902798417c0c2b61cb)

test/test/control/ControlTests.java
Baraa-Hasheesh: handled here [cec75dc](https://github.com/async-profiler/async-profiler/pull/1335/commits/cec75dc222c2e7462fea91902798417c0c2b61cb)

src/zInit.cpp
Baraa-Hasheesh: handled here [cec75dc](https://github.com/async-profiler/async-profiler/pull/1335/commits/cec75dc222c2e7462fea91902798417c0c2b61cb)

But I had to explicitly turn off PRELOAD for MacOs due to how MacOs resolves symbols compared to linux [b08137d](https://github.com/async-profiler/async-profiler/pull/1335/commits/b08137d359c92c211ad11d3bf815745e3c381894)

Baraa-Hasheesh: I've updated the logic to use a standard `dladdr` approach using the same trick

src/zInit.cpp
```diff
@@ -32,7 +32,22 @@ class LateInitializer {
...
-    bool checkJvmLoaded() {
+    static bool checkPreload() {
+        Dl_info dlopen_info;
+        Dl_info current_info;
+
+        if (dladdr((const void*)Hooks::init, &current_info) == 0 || current_info.dli_fname == NULL) {
+            return false;
+        }
+
+        if (dladdr((const void*)dlopen, &dlopen_info) == 0 || dlopen_info.dli_fname == NULL) {
```
fandreuz, line 45: 2 things:
- I'd add a comment either here or on the function to explain the flow
- I'd move the declaration of `dlopen_info` on L42

test/test/control/ControlTests.java
fandreuz: I had a very deep thinking session about this and couldn't come up with anything better than `Control`, so it's fine for me. Thanks

src/zInit.cpp
Baraa-Hasheesh: updated

src/zInit.cpp
apangin: I believe `DYLD_INSERT_LIBRARIES` previously worked. What has happened to it?

Baraa-Hasheesh: While DYLD_INSERT_LIBRARIES did load the lib first, the PRELOAD hooks don't really work on MacOs as we aren't currently doing any Interposing (We're using Linux syntax to PRELOAD the hooks which doesn't work on MacOs)

So technically speaking previously the profiling did start but wasn't really doing a proper profiling on the process as the hooks didn't work (dlopen, pthread)

I simply removed the LD_PRELOAD hooks from compilation on MacOs, as they should be replaced with Interposing functions

The side effect is that the DYLD_INSERT_LIBRARIES will not work at all now rather than working with reduced capabilities

apangin: "Doesn't work" is not very informative, what is behavioral difference exactly?
Also, did you set `DYLD_FORCE_FLAT_NAMESPACE` with `DYLD_INSERT_LIBRARIES`?

Baraa-Hasheesh: Yes, I did set DYLD_FORCE_FLAT_NAMESPACE 

Please find the following patch, 

```
diff --git a/src/hooks.cpp b/src/hooks.cpp
index aa1a9df..ab6cd22 100644
--- a/src/hooks.cpp
+++ b/src/hooks.cpp
@@ -92,7 +92,7 @@ typedef void* (*dlopen_t)(const char*, int);
 static dlopen_t _orig_dlopen = NULL;
 
 static void* dlopen_hook_impl(const char* filename, int flags, bool patch) {
-    Log::debug("dlopen: %s", filename);
+    fprintf(stderr, "dlopen: %s\n", filename);
     void* result = _orig_dlopen(filename, flags);
     if (result != NULL && filename != NULL) {
         Profiler::instance()->updateSymbols(false);
diff --git a/test/test/nativemem/profile_with_dlopen.c b/test/test/nativemem/profile_with_dlopen.c
index 81b0d1d..5d7a5c7 100644
--- a/test/test/nativemem/profile_with_dlopen.c
+++ b/test/test/nativemem/profile_with_dlopen.c
@@ -40,51 +40,11 @@ int main(int argc, char** argv) {
     char* err = NULL;
     void* lib = NULL;
 
-    // first arg is the filename and required.
-    if (argc < 3) {
-        fprintf(stderr, "Usage: %s <dlopen_first | profile_first> <output.jfr>\n", argv[0]);
-        exit(1);
-    }
-
-    int dlopen_first = strcmp(argv[1], "dlopen_first") == 0 ? 1 : 0;
-    const char* filename = argv[2];
-
-    void* libprof = dlopen("build/lib/libasyncProfiler" LIB_EXT, RTLD_NOW);
-    ASSERT_NO_DLERROR(libprof);
-
-    asprof_init_t asprof_init = ((asprof_init_t)dlsym(libprof, "asprof_init"));
-    ASSERT_NO_DLERROR(asprof_init);
-    asprof_init();
-
-    asprof_execute_t asprof_execute = (asprof_execute_t)dlsym(libprof, "asprof_execute");
-    ASSERT_NO_DLERROR(asprof_execute);
-
-    asprof_error_str_t asprof_error_str = (asprof_error_str_t)dlsym(libprof, "asprof_error_str");
-    ASSERT_NO_DLERROR(asprof_error_str);
-
-    // Load libcallsmalloc.so before or after starting the profiler, based on args.
-    if (dlopen_first) {
-        lib = dlopen("build/test/lib/libcallsmalloc" LIB_EXT, RTLD_NOW);
-        ASSERT_NO_DLERROR(lib);
-    }
-
-    // Start profiler.
-    char start_cmd[2048] = {0};
-    snprintf(start_cmd, sizeof(start_cmd), "start,nativemem,cstack=dwarf,file=%s", filename);
-
-    asprof_error_t asprof_err = asprof_execute(start_cmd, outputCallback);
-    ASSERT_NO_ASPROF_ERR(asprof_err);
-
-    if (!dlopen_first) {
-        lib = dlopen("build/test/lib/libcallsmalloc" LIB_EXT, RTLD_NOW);
-        ASSERT_NO_DLERROR(lib);
-    }
+    lib = dlopen("build/test/lib/libcallsmalloc" LIB_EXT, RTLD_NOW);
+    ASSERT_NO_DLERROR(lib);
 
     call_malloc_t call_malloc = (call_malloc_t)dlsym(lib, "call_malloc");
     ASSERT_NO_DLERROR(call_malloc);
 
     free(call_malloc(1999993));
-
-    asprof_err = asprof_execute("stop", NULL);
-    ASSERT_NO_ASPROF_ERR(err);
 }
```

if you apply it & try to execute the test 
```
DYLD_INSERT_LIBRARIES=build/lib/libasyncProfiler.dylib ASPROF_COMMAND=start,nativemem,file=preload_file.jfr DYLD_FORCE_FLAT_NAMESPACE=1 build/test/bin/profile_with_dlopen
```

You will observe that the dlopen for `libcallsmalloc` was never redirected to the profiler, 
This means that the malloc made by calling the `libcallsmalloc` method was never recorded by the profiler 

From what I observed MacOs PRELOAD only really work on simple applications that don't dlopen any external dependency.

Furthermore the pthread calls are also not captured by the profiler making it enable to do proper cpu profiling for any new threads created in the main function.

The LD_PRELOAD capabilities are quite limited in Mac.

An alternative is to use the improved checking mechanism for Linux only while using the old best effort strstr for MacOs.

Baraa-Hasheesh: I've applied the logic here [e15fe59](https://github.com/async-profiler/async-profiler/pull/1335/commits/e15fe5956b92a1d7a2edc42621d820ab3356073a) 

To use best effort match for Mac while using the more accurate dladdr for Linux 
dladdr isn't used for Mac as it resolves to profiler when searching for dlopen (hooks.cpp)

Regarding preload not working well on Mac, fixing it should be relatively easy to do if that's desired (in a different PR later on)

src/zInit.cpp
```diff
@@ -32,7 +32,37 @@ class LateInitializer {
...
-    bool checkJvmLoaded() {
+    static bool checkPreload(){
+        Dl_info current_info;
+        if (dladdr((const void*)Hooks::init, &current_info) == 0 || current_info.dli_fname == NULL) {
+            return false;
+        }
+
+        // On Linux: Check if dlopen belong to the profiler shared objects
+        // If the dlopen is found inside the profiler shared objects that is a good indication that the profiler is preloaded
+        if (OS::isLinux()) {
```
fandreuz: Nit: Can this be a preprocessor directive?

src/zInit.cpp
```diff
@@ -32,7 +32,37 @@ class LateInitializer {
...
+            Dl_info dlopen_info;
+            if (dladdr((const void*)dlopen, &dlopen_info) == 0 || dlopen_info.dli_fname == NULL) {
+                return false;
+            }
+
+            return strcmp(dlopen_info.dli_fname, current_info.dli_fname) == 0;
+        }
+
+        // On MacOs: Check if the profiler is a part of the DYLD_INSERT_LIBRARIES environment variable
+        // dladdr for dlopen will always resolve to the current shared objects due to the declaration of dlopen in the hooks.cpp file
```
fandreuz: Nit: I wouldn't explicitly mention a file by its name. We can just say where we do it at an high level

src/zInit.cpp
Baraa-Hasheesh: @apangin I did some investigation here. For preload mode to work normally on MacOs we will need to correctly interpose the symbols via APPLE logic/methods 

I'm attaching a small example on  how to do that,
Please note for this to work the function can't be defined as `WEAK`

My recommendation is to have 2 different sets of preload functions, one is for APPLE & one is for LINUX
Those function shall call the hooks defined in `hooks.cpp`

```
preload_linux.cpp
preload_mac.cpp
```

Please note, for Mac we shouldn't use `dlsym` to resolve the original symbols but rather use them as is,
that's because for `dlopen` the `dlsym` call will end up causing a recursive loop:
* `dlsym` calls `dlopen` which is redirected  to our function 
* our `dlopen` calls `dlsym`
* loop

IMO it's better to simply remove the preload hooks & replace them with patching of the GOT entries as we do in API mode, I would however keep a small `dlopen` preload hook that does nothing.

This `dlopen` preload function, will be used to find if the lib is started in preload mode or not. 

```
#include <stdio.h>
#include <dlfcn.h>

int count = 0;
void* (*orig_dlopen)(const char*, int) = dlopen;

extern "C" {
// Replacement function
void* my_dlopen(const char* path, int mode) {
    fprintf(stderr, "[interpose] dlopen called with: %s - %d\n", path, count++);
    return orig_dlopen(path, mode);
}

// Interpose struct
__attribute__((used)) static struct {
    const void* replacement;
    const void* original;
} interposers[] __attribute__((section("__DATA,__interpose"))) = {
    { (const void*)my_dlopen, (const void*)dlopen }
};

}
```

src/zInit.cpp
```diff
@@ -32,7 +32,37 @@ class LateInitializer {
     }
 
   private:
-    bool checkJvmLoaded() {
+    static bool checkPreload(){
+        Dl_info current_info;
+        if (dladdr((const void*)Hooks::init, &current_info) == 0 || current_info.dli_fname == NULL) {
```
krk, line 37: What are we really checking here?

src/zInit.cpp
```diff
@@ -32,7 +32,37 @@ class LateInitializer {
     }
 
   private:
-    bool checkJvmLoaded() {
+    static bool checkPreload(){
```
krk: nit: space

also, name: `checkPreloaded`?

src/zInit.cpp
```diff
@@ -32,7 +32,37 @@ class LateInitializer {
...
+            if (dladdr((const void*)dlopen, &dlopen_info) == 0 || dlopen_info.dli_fname == NULL) {
+                return false;
+            }
+
+            return strcmp(dlopen_info.dli_fname, current_info.dli_fname) == 0;
+        }
+
+        // On MacOs: Check if the profiler is a part of the DYLD_INSERT_LIBRARIES environment variable
+        // dladdr for dlopen will always resolve to the current shared objects due to the declaration of dlopen in the hooks.cpp file
+        const char* preload = getenv("DYLD_INSERT_LIBRARIES");
```
krk, line 53: Why do we need to check this in linux builds?

test/test/control/ControlTests.java
krk: `EarlyStartTests`?

test/test/control/start_control.c
```diff
@@ -0,0 +1,78 @@
+/*
```
krk, line 1: What does the name `start_control` refer to?

src/zInit.cpp
Baraa-Hasheesh: we're checking that we're able to get the shared objects of the current async-profiler

src/zInit.cpp
Baraa-Hasheesh: It will never be checked as ` if (OS::isLinux()) {` check above it guarantees that this will only be checked on Mac

test/test/control/start_control.c
Baraa-Hasheesh: control on starting the profiler

test/test/control/ControlTests.java
Baraa-Hasheesh: The reason I picked control is that it can be used to check other aspects of controlling the profiler, if the need arises

src/zInit.cpp
Baraa-Hasheesh: handled here [df4fefa](https://github.com/async-profiler/async-profiler/pull/1335/commits/df4fefadc00ef48fc3f3ff69d2f183d26ecdd9ef)

src/zInit.cpp
Baraa-Hasheesh: handled here [df4fefa](https://github.com/async-profiler/async-profiler/pull/1335/commits/df4fefadc00ef48fc3f3ff69d2f183d26ecdd9ef)

src/zInit.cpp
Baraa-Hasheesh: handled here [df4fefa](https://github.com/async-profiler/async-profiler/pull/1335/commits/df4fefadc00ef48fc3f3ff69d2f183d26ecdd9ef)

Baraa-Hasheesh:
due to various observed problems when trying to enable preload hooks on macOS, I will close this PR & create a different one later on to refactor the logic in async-profiler to use GOT patching in LD_PRELOAD mode

PR #1336, opened 2025-06-11
JFR to OTLP converter

fandreuz:
### Description

In this PR I add a JFR to OTLP converter.

### Motivation and context

This piece of work was scheduled as part of the ongoing effort to integrate Async-Profiler with OTLP. Having an OTLP converter will allow our users to work with historical data.

### How has this been tested?

Local testing: [out_otlp.txt](https://github.com/user-attachments/files/20712505/out_otlp.txt)

---


src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,173 @@
...
+        int sampleTypeTotalMark = otlpProto.startField(Otlp.Profile.SAMPLE_TYPE);
+        otlpProto.field(Otlp.ValueType.TYPE_STRINDEX, stringPool.index(args.getValueType()));
+        otlpProto.field(Otlp.ValueType.UNIT_STRINDEX, stringPool.index(args.getTotalUnits()));
+        otlpProto.field(Otlp.ValueType.AGGREGATION_TEMPORALITY, Otlp.AggregationTemporality.CUMULATIVE);
+        otlpProto.commitField(sampleTypeTotalMark);
+
+        locationIndicesMark = otlpProto.startField(Otlp.Profile.LOCATION_INDICES);
+    }
+
+    // TODO: Do we need to aggregate across different chunks? Is it handled by the JfrReader?
```
fandreuz, line 55: ?

src/converter/one/convert/Arguments.java
```diff
@@ -134,4 +134,20 @@ private static long parseTimestamp(String time) {
...
+    }
+
+    public String getSampleUnits() {
+        return "count";
+    }
+
+    public String getTotalUnits() {
+        if (nativemem || alloc || live) return "bytes";
+        return "nanoseconds";
+    }
```
apangin: `Arguments` is a parser and holder for converter options, it is not supposed to contain business logic for interpreting options. `JfrConverter` is a better place for utility functions shared between different converters.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,184 @@
...
+            int[] locations = stackTrace.locations;
+
+            for (int i = methods.length; --i >= 0; ) {
+                String methodName = getMethodName(methods[i], types[i]);
+                int location;
+                if (args.lines && (location = locations[i] >>> 16) != 0) {
+                    methodName += ":" + location;
+                } else if (args.bci && (location = locations[i] & 0xffff) != 0) {
+                    methodName += "@" + location;
+                }
```
apangin: It's unnatural to have line numbers included in method names when the format provides a dedicated field for line numbers.

src/converter/one/jfr/JfrReader.java
```diff
@@ -106,6 +108,13 @@ public void close() throws IOException {
         }
     }
 
+    public ByteBuffer currentChunk() {
+        ByteBuffer chunk = buf.duplicate().asReadOnlyBuffer();
+        chunk.position((int) currentChunkStart);
+        chunk.limit((int) currentChunkSize);
+        return chunk;
+    }
```
apangin: That is not correct. `buf` is a limited window that contains a part of the recording. Chunk does not necessarily fit into the buffer.

src/converter/one/convert/Otlp.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.convert;
+
+final class Otlp {
+
+    static final class ProfilesDictionary {
```
apangin: Classes are not namespaces, do not multiply classes just to hold stuff.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,184 @@
...
+        scopeProfilesMark = otlpProto.startField(Otlp.ResourceProfiles.SCOPE_PROFILES);
+    }
+
+    @Override
+    protected void convertChunk() {
+        int profileMark = otlpProto.startField(Otlp.ScopeProfiles.PROFILES);
+
+        writeSampleTypes();
+        writeTimingInformation();
+        otlpProto.field(Otlp.Profile.ORIGINAL_PAYLOAD, jfr.currentChunk());
```
apangin: That kills one of the benefits of the format - compact representation.
I can't imagine a case where we may need JFR embedded in protobuf.

fandreuz: I think the original idea for having this in the OTLP format is in the case when the conversion form source to OTLP is lossy. When we go through a chunk, we filter away all the events which weren't selected as part of the converter CLI arguments.

Maybe this could be optional?

src/converter/one/convert/Arguments.java
fandreuz: 97cbf40667345788c5b9f8e58395b80007a924eb

src/converter/one/convert/Otlp.java
fandreuz: Better now? Replaced with enums: abcbfbecca556352c54cdd5a948930ebd00068fa

apangin: Worse.
Long 4-level expressions like `Otlp.Sample.LOCATIONS_START_INDEX.index` are neither nice nor efficient.
A dozen of extra classes loaded at startup (as well as a dozen of .class files in an archive) for things that should be compile-time constants is also not good.

What about a single `OtlpConstants` file with all the constants grouped into sections? E.g.
```
public static final int
        ValueType_type_strindex = 1,
        ValueType_unit_strindex = 2,
        ValueType_aggregation_temporality = 3;

public static final int
        Sample_locations_start_index = 1,
        Sample_locations_length = 2,
        Sample_value = 3;
```

Then converter can `import static OtlpConstants.*` and use constants without additional boilerplate:
```
proto.field(Sample_locations_start_index, framesSeen);
```
Naming style is discussible. I'm fine both with `Sample_locations_start_index` and `SAMPLE_LOCATIONS_START_INDEX` or maybe `SAMPLE_locations_start_index`.

fandreuz: Looks good, thanks: da5e16fa8f05c16212948491b19cccf1dd155bf7

src/converter/one/convert/JfrToOtlp.java
fandreuz: Fixed

src/converter/one/jfr/JfrReader.java
fandreuz: TODO, let's first discuss if we want to keep `ORIGINAL_PAYLOAD`. I'll fix this if we do

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/** Converts .jfr output to OpenTelemetry protocol. */
+public class JfrToOtlp extends JfrConverter {
```
fandreuz, line 22: I see information about loaded native libraries is available via `jdk.NativeLibrary` events. Do we need to encode that information? It's not particularly useful unless we have the address for instructions in each sample frame.

src/converter/one/convert/JfrConverter.java
```diff
@@ -260,16 +260,41 @@ protected boolean isNativeFrame(byte methodType) {
...
+    protected abstract class AggregatedEventVisitor extends NormalizedEventVisitor {
+        @Override
+        protected final void visitImpl(Event event, long samples, long value) {
+            visit(event, args.total ? value : samples);
+        }
+
+        protected abstract void visit(Event event, long value);
+    }
+
+    protected abstract class NormalizedEventVisitor implements EventCollector.Visitor {
```
apangin: Redundant abstraction layer. For reusing a single line of code, you add a dozen of new lines, complicate class hierarchy and add an extra indirection.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+    private final Index<String> stringPool = new Index<>(String.class, "");
+    private final Index<String> functionPool = new Index<>(String.class, "");
+    private final Index<Line> linePool = new Index<>(Line.class, Line.EMPTY);
+    private final Index<KeyValue> attributesPool = new Index<>(KeyValue.class, KeyValue.EMPTY);
+
+    private final Proto otlpProto = new Proto(1024);
+
+    private final int resourceProfilesMark;
+    private final int scopeProfilesMark;
```
apangin: These are not converter properties; they can be local variables.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+
+    public JfrToOtlp(JfrReader jfr, Arguments args) {
+        super(jfr, args);
+
+        resourceProfilesMark = otlpProto.startField(PROFILES_DATA_resource_profiles);
+        scopeProfilesMark = otlpProto.startField(RESOURCE_PROFILES_scope_profiles);
```
apangin: As mentioned above, they should not be fields.
Override `convert` method (or create a new method), where you will start these messages, call original convert, and then finalize messages.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    protected void convertChunk() {
+        int profileMark = otlpProto.startField(SCOPE_PROFILES_profiles);
+
+        writeSampleTypes();
+        writeTimingInformation();
+        otlpProto.field(PROFILE_original_payload, jfr.currentChunk());
```
apangin: IMO, original payload is not needed, it will just bloat output in vain.
If format `B` can be always reconstructed from format `A`, there is no point in storing `AB` with an intention to get format `A` back.
Also, this may be non-trivial to implement, given that original JFR can be gigabytes in size.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+
+        int locationIndicesMark = otlpProto.startField(PROFILE_location_indices);
+        List<SampleInfo> sampleInfos = new ArrayList<>();
+        collector.forEach(new OtlpEventVisitor(sampleInfos));
```
apangin: Are you going to record each individual sample, or a profile aggregated by stack traces? In either case, the code is not doing it right.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+    private void writeSample(SampleInfo si, long framesSeen) {
+        int sampleMark = otlpProto.startField(PROFILE_sample);
+        otlpProto.field(SAMPLE_locations_start_index, framesSeen);
+        otlpProto.field(SAMPLE_locations_length, si.numFrames);
+        otlpProto.field(SAMPLE_timestamps_unix_nano, si.timeNanos);
+
+        KeyValue threadNameAttribute = new KeyValue("thread.name", si.threadName);
+        otlpProto.field(SAMPLE_attribute_indices, attributesPool.index(threadNameAttribute));
+
+        int sampleValueMark = otlpProto.startField(SAMPLE_value);
```
apangin: We'll likely need `startField` with expected size, similar to C++ counterpart.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+            long[] methods = stackTrace.methods;
+            int[] locations = stackTrace.locations;
+
+            for (int i = 0; i < methods.length; ++i) {
+                String methodName = getMethodName(methods[i], stackTrace.types[i]);
+                int lineNumber = locations[i] >>> 16;
+                int functionIdx = functionPool.index(methodName);
+
+                Line line = new Line(functionIdx, lineNumber);
+                otlpProto.writeLong(linePool.index(line));
```
apangin: You are writing full stack trace every time for each event. That's against the ideas behind otlp design.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,267 @@
...
+        public boolean equals(Object o) {
+            if (!(o instanceof Line)) return false;
+
+            Line other = (Line) o;
+            return functionIdx == other.functionIdx && lineNumber == other.lineNumber;
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(functionIdx, lineNumber);
```
apangin: Never use `Objects.hash`. It does array allocation and boxing on every call.

src/converter/one/convert/JfrConverter.java
fandreuz: So you'd prefer to have the factors logic duplicated in two places?

src/converter/one/convert/JfrToOtlp.java
fandreuz: Reverted in 11116b0b5a005992310b2cf1dd1054ca08d78bab and 31823a8cdf911302687c73aea1ac506d579a9e07

src/converter/one/jfr/JfrReader.java
fandreuz: Not needed anymore

src/converter/one/convert/JfrToOtlp.java
fandreuz: Not needed as discussed below

src/converter/one/convert/JfrToOtlp.java
fandreuz: afd3d830fca08cd94902815666b967da92d4eed5

src/converter/one/convert/JfrToOtlp.java
fandreuz: afd3d830fca08cd94902815666b967da92d4eed5

src/converter/one/convert/JfrToOtlp.java
fandreuz: Didn't consider that, thanks: e6378d52f8b3cf7c4fbee5096f7fc79b3fa48796

src/converter/one/convert/JfrToOtlp.java
fandreuz: Can you elaborate? This is the code handling writing a single sample
```java
private void writeSample(SampleInfo si) {
        int sampleMark = otlpProto.startField(PROFILE_sample);
        otlpProto.field(SAMPLE_locations_start_index, si.locationsRange.start);
        otlpProto.field(SAMPLE_locations_length, si.locationsRange.length);
        otlpProto.field(SAMPLE_timestamps_unix_nano, si.timeNanos);

        KeyValue threadNameAttribute = new KeyValue("thread.name", si.threadName);
        otlpProto.field(SAMPLE_attribute_indices, attributesPool.index(threadNameAttribute));

        int sampleValueMark = otlpProto.startField(SAMPLE_value);
        otlpProto.writeLong(si.samples);
        otlpProto.writeLong(si.value);
        otlpProto.commitField(sampleValueMark);

        otlpProto.commitField(sampleMark);
 }
```

The presence of `timestamps_unix_nano` and `thread.name` pushes me not to aggregate samples, am I missing something?

src/converter/one/convert/JfrToOtlp.java
fandreuz: I've been thinking about this. Here's my first attempt: 37d19ef060fd76e01bb45bf29946fc63867ea952. 

I use an hash map to remember previous occurrences of a full array of location indices. If I find a previous occurrence, I'm going to reuse the same range from `Profile.location_indices`, otherwise I write a new range.

Deduplication could be made better by using a Trie instead of an hash map, but it's also probably going to be a bit worse on cpu. Would you accept a trie-based solution?

src/converter/one/convert/JfrConverter.java
apangin: It's a matter of a single expression: `double factor = args.lock ? 1e9 / jfr.ticksPerSec : 1.0`
I wouldn't even call it a duplication. But if you wish, you may create a method in `JfrConverter`.

src/converter/one/convert/JfrToOtlp.java
apangin: Library name is already encoded as class name in JFR Method structure (`MethodRef.cls`).

src/converter/one/convert/JfrToOtlp.java
apangin: C++ OTLP writer produces an aggregated profile, but here you are writing each single sample. Is this asymmetry intentional?

When writing individual samples, why collecting them in a list then? You may write them out as you go right in the `visit` method.

src/converter/one/convert/JfrToOtlp.java
apangin: Stacktraces are already deduplicated for you in JFR constant pool.

src/converter/one/convert/JfrToOtlp.java
fandreuz: f1e1437642e9eeade259c3ec9e7cf8a1fd571150

src/converter/one/convert/JfrConverter.java
fandreuz: 48853b15fd041095e50bd21036c34fbe4bd8d11e

src/converter/one/convert/JfrToOtlp.java
fandreuz: > C++ OTLP writer produces an aggregated profile, but here you are writing each single sample. Is this asymmetry intentional?

In the absence of information to discriminate between a sample and another (i.e. `timestamps_unix_nano` and `thread.name`) I don't see why we should not aggregate them if they share a common stacktrace. Consumers will just aggregate them again when producing e.g. a flamegraph.

But if we have additional information, it makes sense not to aggregate samples. I have some commits hanging (I'll push them once I have a bit less open PRs) which use the artificial frame injected by the `--threads` flag to add `thread.name` for the C++ OTLP writer too. In that case, we're not going to aggregate different samples taken on different threads. Similarly, we could have an artificial frame for the nanotime of a sample, and that could be used to fill `timestamps_unix_nano`.

Let me know if you see anything wrong in my reasoning.

fandreuz: > When writing individual samples, why collecting them in a list then? You may write them out as you go right in the visit method.

If we're writing a `repeated` primitive in Protobuf (`Profile.location_indices`), we cannot interleave primitives from this list and `Profile.sample` objects. Thus, we either need to collect samples in a list, or location indices. The latter is probably easier on resources, I could do that.

src/converter/one/convert/JfrToOtlp.java
fandreuz: Right, missed that. Fixed in 885b0e9646f0f35cd0f0b73cc1875feaf48af52c

src/converter/one/convert/JfrConverter.java
```diff
@@ -260,13 +260,32 @@ protected boolean isNativeFrame(byte methodType) {
...
+    }
+
     // Select sum(samples) or sum(value) depending on the --total option.
     // For lock events, convert lock duration from ticks to nanoseconds.
     protected abstract class AggregatedEventVisitor implements EventCollector.Visitor {
-        final double factor = !args.total ? 0.0 : args.lock ? 1e9 / jfr.ticksPerSec : 1.0;
-
         @Override
         public final void visit(Event event, long samples, long value) {
+            double factor = counterFactor();
```
apangin: The value does not change throughout chunk. Please leave it as a final field.

src/converter/one/convert/JfrToOtlp.java
apangin: Sounds good.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,313 @@
...
+            return start == range.start && length == range.length;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = 17;
+            result = (int) (31 * result + start);
+            return (int) (31 * result + length);
+        }
+    }
```
apangin: If a class is not used as a key in collections, these methods are not needed, they just add noise.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,313 @@
...
+            this.value = value;
+            this.locationsRange = locationsRange;
+            this.threadName = threadName;
+            this.timeNanos = timeNanos;
+        }
+    }
+
+    private static final class Range {
+        final long start;
+        final long length;
```
apangin: Isn't `int` enough?

src/converter/one/proto/Proto.java
```diff
@@ -63,27 +66,45 @@ public Proto field(int index, byte[] bytes) {
         return this;
     }
 
+    public Proto field(int index, ByteBuffer bytes) {
```
apangin: Is this method used?

src/converter/one/proto/Proto.java
```diff
@@ -140,4 +166,14 @@ private void ensureCapacity(int length) {
             buf = Arrays.copyOf(buf, newLength < 0 ? 0x7ffffff0 : Math.max(newLength, pos + length));
         }
     }
+
+    public static final class Mark {
```
apangin: As in C++, mark can be just `long`. In Java, this makes even more sense, since JVM capabilities of allocation elimination are very limited.

src/converter/one/proto/Proto.java
```diff
@@ -63,27 +66,45 @@ public Proto field(int index, byte[] bytes) {
...
+        if (actualLength >= (long) 1 << (7 * mark.maxLenByteCount)) {
             throw new IllegalArgumentException("Field too large");
         }
 
-        buf[mark - 3] = (byte) (0x80 | (length & 0x7f));
-        buf[mark - 2] = (byte) (0x80 | ((length >>> 7) & 0x7f));
-        buf[mark - 1] = (byte) (length >>> 14);
+        int lenBytesStart = mark.messageStart - mark.maxLenByteCount;
+        for (int i = 0; i < mark.maxLenByteCount - 1; ++i) {
+            buf[lenBytesStart + i] = (byte) (0x80 | (actualLength & 0x7f));
```
apangin: Redundant `& 0x7f`.
I know this was before, but since you touch this code, it's a good opportunity to fix this.

src/converter/one/proto/Proto.java
```diff
@@ -63,27 +66,45 @@ public Proto field(int index, byte[] bytes) {
...
+    public void commitField(int start) {
+        commitField(new Mark(start, DEFAULT_MAX_LEN_BYTE_COUNT));
     }
 
-    public void commitField(int mark) {
-        int length = pos - mark;
-        if (length >= 1 << (7 * 3)) {
+    public void commitField(Mark mark) {
+        int actualLength = pos - mark.messageStart;
+        if (actualLength >= (long) 1 << (7 * mark.maxLenByteCount)) {
```
apangin: `1L`

src/converter/one/proto/Proto.java
```diff
@@ -6,13 +6,16 @@
...
 import java.nio.charset.StandardCharsets;
+import java.nio.ByteBuffer;
 import java.util.Arrays;
 
 /**
  * Simplified implementation of Protobuf writer, capable of encoding
  * varints, doubles, ASCII strings and embedded messages
  */
 public class Proto {
+    private static final int DEFAULT_MAX_LEN_BYTE_COUNT = 3;
```
apangin: It needs a better name.
5 words with `default` and `max`, `len` and `count`, all within the same name, sounds gibberish.

src/converter/one/proto/Proto.java
```diff
@@ -63,27 +66,45 @@ public Proto field(int index, byte[] bytes) {
...
+    }
+
     public Proto field(int index, Proto proto) {
         tag(index, 2);
         writeBytes(proto.buf, 0, proto.pos);
         return this;
     }
 
     public int startField(int index) {
+        return startField(index, DEFAULT_MAX_LEN_BYTE_COUNT).messageStart;
```
apangin: There is just one usage of this method - let's change it instead.

src/converter/one/proto/Proto.java
fandreuz: Leftover from a previous iteration, fixed in 2359e43708a736b01714e3842adfd6a91ad6e203

src/converter/one/convert/JfrToOtlp.java
fandreuz: Right, this is a leftover from a previous iteration. Fixed in 2359e43708a736b01714e3842adfd6a91ad6e203

src/converter/one/convert/JfrConverter.java
fandreuz: 2359e43708a736b01714e3842adfd6a91ad6e203

src/converter/one/proto/Proto.java
fandreuz: Removed in 2359e43708a736b01714e3842adfd6a91ad6e203

src/converter/one/proto/Proto.java
fandreuz: Done: 2359e43708a736b01714e3842adfd6a91ad6e203

src/converter/one/proto/Proto.java
fandreuz: 1daeb9af3f5e77b126d0f2328f4d7ae6bab238ae

src/converter/one/proto/Proto.java
fandreuz: 1daeb9af3f5e77b126d0f2328f4d7ae6bab238ae

src/converter/one/convert/JfrToOtlp.java
fandreuz: 5f4f10d7698fc4485125ff8e986d28f7876654b5

src/converter/one/convert/JfrToOtlp.java
fandreuz: 32ecbb4312f596d0db9a80f416abc682b5f56523

fandreuz:
Latest revision validated with Pyroscope, looks good:

image

fandreuz:
Before this is merged, I'd like to tackle the problem which @Baraa-Hasheesh mentioned during the last demo. Moving to draft again.

fandreuz:
> Before this is merged, I'd like to tackle the problem which @Baraa-Hasheesh mentioned during the last demo. Moving to draft again.

I thought about this, I think we could have a flag `allEvents` which forces the JFR collector to collect all events. Then, each event from a given type will be part of a separate profile. This could be implemented in another PR. The current implementation handles only one type per call.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,275 @@
...
+        long locationIndicesMark =
+                proto.startField(PROFILE_location_indices, BIG_MESSAGE_BYTE_COUNT);
+        locationIndices.forEach(proto::writeInt);
+        proto.commitField(locationIndicesMark);
+
+        proto.commitField(pMark);
+    }
+
+    private void writeSampleTypes() {
+        long stsMark = proto.startField(PROFILE_sample_type, 1);
```
Baraa-Hasheesh: is there a reason why we don't a for loop here & mention fields being handled via a comment?

```
// stsMark & sttMark fields
for (int i = 0; i < 2; i++) {
    long field = proto.startField(PROFILE_sample_type, 1);
    proto.field(VALUE_TYPE_type_strindex, stringPool.index(getValueType()));
    proto.field(VALUE_TYPE_unit_strindex, stringPool.index(getTotalUnits()));
    proto.field(VALUE_TYPE_aggregation_temporality, AGGREGATION_TEMPORARALITY_cumulative);
    proto.commitField(field);
}
```

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,275 @@
...
+                proto.startField(PROFILES_DATA_dictionary, BIG_MESSAGE_BYTE_COUNT);
+
+        // Mapping[0] must be a default mapping according to the spec
+        long mappingMark = proto.startField(PROFILES_DICTIONARY_mapping_table, 1);
+        proto.commitField(mappingMark);
+
+        // Write function table
+        for (String name : functionPool.keys()) {
+            long fMark = proto.startField(PROFILES_DICTIONARY_function_table, 1);
+            int nameStrindex = stringPool.index(name);
```
Baraa-Hasheesh: do you really need `nameStrindex` variable?
```suggestion
            proto.field(FUNCTION_name_strindex, stringPool.index(name));
```

src/converter/one/convert/JfrToOtlp.java
fandreuz: You would need to have an ugly `i == 0 ? getSampleUnits() : getTotalUnits()`, that's the only reason why I'm not doing it. If the body was more than 5 lines of code I'd do it, but all things being equal I don't want the additional conditional and indentation

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,275 @@
...
+            proto.field(FUNCTION_name_strindex, nameStrindex);
```
fandreuz:
```suggestion
            proto.field(FUNCTION_name_strindex, stringPool.index(name));
```

src/converter/one/convert/JfrToOtlp.java
fandreuz: Not really, I applied your change.

FYI, I couldn't apply the suggestion you posted because you didn't select both lines (`int nameStringex = ...` and `proto.field(....` when you created the suggestion

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+    private final Index<Line> linePool = new Index<>(Line.class, Line.EMPTY);
+    private final Index<KeyValue> attributesPool = new Index<>(KeyValue.class, KeyValue.EMPTY);
+
+    private final Proto proto = new Proto(1024);
+
+    public JfrToOtlp(JfrReader jfr, Arguments args) {
+        super(jfr, args);
+
+        // string_table[0] must always be the empty string
+        stringPool.index("");
```
apangin: `Index` constructor already initializes empty element.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+import one.proto.Proto;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.*;
+
+/** Converts .jfr output to OpenTelemetry protocol. */
+public class JfrToOtlp extends JfrConverter {
+    private static final int BIG_MESSAGE_BYTE_COUNT = 5;
```
apangin: Name can be shorter. Also, you use a named constant for large messages and `1` for small ones.
IMO, having two constants (`MSG_LARGE` / `MSG_SMALL`) will be a good alternative. It's not necessary to repeat `BYTE_COUNT` every time, it does not add much clarity. Instead, you can add a comment here explaining what the number means.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+        long pMark = proto.startField(SCOPE_PROFILES_profiles, BIG_MESSAGE_BYTE_COUNT);
+
+        writeSampleTypes();
+        writeTimingInformation();
+
+        List<Integer> locationIndices = new ArrayList<>();
+        collector.forEach(new OtlpEventToSampleVisitor(locationIndices));
+
+        long locationIndicesMark =
+                proto.startField(PROFILE_location_indices, BIG_MESSAGE_BYTE_COUNT);
```
apangin: Make names shorter and keep it on a single line. In other places, too.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+
+        // Write string table
+        for (String s : stringPool.keys()) {
+            proto.field(PROFILES_DICTIONARY_string_table, s);
+        }
+
+        // Write attributes table
+        for (KeyValue kv : attributesPool.keys()) {
+            long aMark =
+                    proto.startField(PROFILES_DICTIONARY_attribute_table, BIG_MESSAGE_BYTE_COUNT);
```
apangin: Here is a good example why enforcing line length is not a good idea.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+        // Next index to be used for a location into Profile.location_indices
+        private int nextLocationIdx = 0;
+
+        public OtlpEventToSampleVisitor(List<Integer> locationIndices) {
+            this.locationIndices = locationIndices;
+        }
+
+        @Override
+        public void visit(Event event, long samples, long value) {
+            long msFromStart = (event.time - jfr.chunkStartTicks) * 1_000 / jfr.ticksPerSec;
```
apangin: Why do you artificially downscale resolution to milliseconds?

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,274 @@
...
+            Range range = idToRange.computeIfAbsent(event.stackTraceId, this::computeLocationRange);
+
+            long sMark = proto.startField(PROFILE_sample, 1);
+            proto.field(SAMPLE_locations_start_index, range.start);
+            proto.field(SAMPLE_locations_length, range.length);
+            proto.field(SAMPLE_timestamps_unix_nano, timeNanos);
+
+            proto.field(
+                    SAMPLE_attribute_indices,
+                    attributesPool.index(new KeyValue("thread.name", getThreadName(event.tid))));
```
apangin: My gut feeling tells me this can be a bottleneck due to lots of garbage created on the hot path.
I'm not asking you to rewrite it, though: it will be an interesting exercise to profile this later on.

src/converter/one/convert/JfrToOtlp.java
fandreuz: I see, I added a bit of documentation to `Index`: a5ae70aa0bd7a68b55b927f6674ef5b212cb5afd

src/converter/one/convert/JfrToOtlp.java
fandreuz: Sure will do, do you know any low overhead sampling profiler for Java that does not suffer from the Safepoint bias problem?

src/converter/one/convert/JfrToOtlp.java
fandreuz: Fixed in fd012768671c1162b4297121c1ce5cf172f263b2

src/converter/one/convert/JfrToOtlp.java
fandreuz: Sure: fd012768671c1162b4297121c1ce5cf172f263b2

src/converter/one/convert/JfrToOtlp.java
fandreuz: Should be better now: fac1d97255bf6691dabb3e02fab31c017bfb6b82

src/converter/one/convert/JfrToOtlp.java
fandreuz: Should be better now: 5b46e441862d79143267ec996e44d4ed8094d108

By the way, I got this snippet from [here](https://github.com/async-profiler/async-profiler/blob/master/src/converter/one/convert/JfrToHeatmap.java#L45), do you think it should be fixed there too?

fandreuz: Ok I see the additional precision is not needed there

apangin: Yes, heatmap intentionally downscales timestamps to trade precision for size.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,269 @@
...
+import one.proto.Proto;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.*;
+
+/** Converts .jfr output to OpenTelemetry protocol. */
+public class JfrToOtlp extends JfrConverter {
+    // Size in bits to be allocated in the buffer to hold the varint containing the length of the message
```
apangin: in bytes

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,269 @@
...
+        // Next index to be used for a location into Profile.location_indices
+        private int nextLocationIdx = 0;
+
+        public OtlpEventToSampleVisitor(List<Integer> locationIndices) {
+            this.locationIndices = locationIndices;
+        }
+
+        @Override
+        public void visit(Event event, long samples, long value) {
+            long nanosFromStart = (event.time - jfr.chunkStartTicks) * 1_000_000_000 / jfr.ticksPerSec;
```
apangin: Potential overflow when multiplying by such a large value.
You may compute `1_000_000_000.0 / jfr.ticksPerSec` as double and cache it in a private final field.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,269 @@
...
+            this.key = key;
+            this.value = value;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (!(o instanceof KeyValue)) return false;
+
+            KeyValue other = (KeyValue) o;
+            return Objects.equals(key, other.key) && Objects.equals(value, other.value);
```
apangin: Since neither key nor value can be `null`, you may call `key.equals` and `value.equals` directly.

src/converter/one/jfr/JfrReader.java
```diff
@@ -118,6 +118,14 @@ public long durationNanos() {
         return endNanos - startNanos;
     }
 
+    public long getChunkStartNanos() {
```
apangin: Accessors here do not start with `get`.
In fact, accessor is not needed since the field is public (which is against common design pattern, I know), but it's OK to leave the accessor if you wish.

src/converter/one/convert/JfrToOtlp.java
fandreuz: Right, fixed in 1ce32137e4d3d3804f54d06e165508f71d491ee6

src/converter/one/jfr/JfrReader.java
fandreuz: Got it, I removed the accessor: c1d2fb7ded8219728a97047f67c516d9b7fa8edf

src/converter/one/convert/JfrToOtlp.java
fandreuz: a5a48ce9566bd69b1106f9b72af365a780596f25

src/converter/one/convert/JfrToOtlp.java
fandreuz: Right, 13ebaaec32bfcbf05a7617dfd31b1653ce3349e7

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,271 @@
...
+    private final Index<String> functionPool = new Index<>(String.class, "");
+    private final Index<Line> linePool = new Index<>(Line.class, Line.EMPTY);
+    private final Index<KeyValue> attributesPool = new Index<>(KeyValue.class, KeyValue.EMPTY);
+
+    private final Proto proto = new Proto(1024);
+    private final double ticksPerNanosecond;
+
+    public JfrToOtlp(JfrReader jfr, Arguments args) {
+        super(jfr, args);
+        ticksPerNanosecond = jfr.ticksPerSec / 1_000_000_000.0;
```
apangin: No, move this to OtlpEventToSampleVisitor. This should not be cached across chunks, since chunks may have different `ticksPerSec`

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,271 @@
...
+        proto.field(VALUE_TYPE_aggregation_temporality, AGGREGATION_TEMPORARALITY_cumulative);
+        proto.commitField(sttMark);
+    }
+
+    private void writeTimingInformation() {
+        proto.field(PROFILE_time_nanos, jfr.chunkStartNanos);
+        proto.field(PROFILE_duration_nanos, jfr.chunkDurationNanos());
+    }
+
+    public void dump(OutputStream out) throws IOException {
```
apangin: The method is lost somewhere in the middle between private methods.
Move it next to other public methods, either after constructor or before `convert`.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -0,0 +1,271 @@
...
+
+        // Write function table
+        for (String name : functionPool.keys()) {
+            long fMark = proto.startField(PROFILES_DICTIONARY_function_table, MSG_SMALL);
+            proto.field(FUNCTION_name_strindex, stringPool.index(name));
+            proto.commitField(fMark);
+        }
+
+        // Write location table
+        for (Line l : linePool.keys()) {
```
apangin: You took my suggestion about shortening variable names too literally :)
`line` would be perfect here. `l` is not good as identifier - it resembles `1` or `I`.

src/converter/one/convert/JfrToOtlp.java
fandreuz: Ouch, I thought this was unique per JFR

fandreuz: 93679fc742741caa2671920c25d9fb7112d823a6

src/converter/one/convert/JfrToOtlp.java
fandreuz: 93679fc742741caa2671920c25d9fb7112d823a6

src/converter/one/convert/JfrToOtlp.java
fandreuz: 93679fc742741caa2671920c25d9fb7112d823a6

apangin:
Good job, thanks!
Merged

PR #1337, opened 2025-06-12
Use `Index` in `jfrMetadata`

fandreuz:
### Description
We can clean up `jfrMetadata.cpp` a bit by using `index.h`.

### How has this been tested?
GHA

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -778,7 +778,7 @@ class Recording {
         buf->putVar32(0);
         buf->putVar32(0x7fffffff);  // must not clash with JFR metadata ID, or 'jfr print' will break
 
-        Index strings = JfrMetadata::strings();
+        const Index& strings = JfrMetadata::strings();
         buf->putVar32(strings.size());
         strings.forEachOrdered([&] (const std::string& s) {
             buf->putUtf8(s.c_str());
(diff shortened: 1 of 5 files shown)


src/jfrMetadata.h
```diff
@@ -236,8 +229,12 @@ class JfrMetadata : Element {
...
     }
 
-    static std::vector<std::string>& strings() {
-        return _strings;
+    static void forEachStringOrdered(std::function<void(const std::string&)> consumer) {
+        _strings.forEachOrdered(consumer);
+    }
+
+    static size_t stringsSize() {
+        return _strings.size();
```
apangin: Maybe just expose `_strings`? These `forEachStringOrdered` and `stringsSize` look too artificial.

fandreuz: b9785616cb1dc262d64635206629cce317c2f162

apangin: I meant both `forEachStringOrdered` and `stringsSize` could be replaced with `strings()`

fandreuz: Right, d536fdd03014c33029b9ebaf0d86e1258e6b037b

src/flightRecorder.cpp
```diff
@@ -778,11 +778,11 @@ class Recording {
         buf->putVar32(0);
         buf->putVar32(0x7fffffff);  // must not clash with JFR metadata ID, or 'jfr print' will break
 
-        std::vector<std::string>& strings = JfrMetadata::strings();
+        Index strings = JfrMetadata::strings();
```
apangin: `Index&` ?

fandreuz: Actually this made me realize `Index` should probably not have any copy/move constructor/assignment. Thanks for catching that: 392114a62b9ded865f3ce7bd75f42082cd87cc80

PR #1338, opened 2025-06-12
C++ linting via clang-tidy

fandreuz:
### Description
In this PR I add some lint checks based on [clang-tidy](https://clang.llvm.org/extra/clang-tidy). They are mostly aimed at reducing time spent on trivial problems in code review, which can easily be automated.

### Checks

The checks I enabled required minimal changes in the code and are not too invasive.

Some examples:
- naming of local variables `like_this`
- naming of private variables `_like_this`
- https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unhandled-self-assignment.html
- some performance issues:
```cpp
/workspaces/async-profiler/src/index.h:21:29: error: the parameter 'value' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    u32 indexOf(std::string value) {
                            ^
                const      &
/workspaces/async-profiler/src/index.h:29:65: error: the parameter 'consumer' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    void forEachOrdered(std::function<void(const std::string&)> consumer) const {
                                                                ^
                        const                                  &
```
- some common mistakes:
```cpp
/workspaces/async-profiler/src/threadLocalData.cpp:33:60: error: the comparison always evaluates to false because pthread_setspecific always returns non-negative values [bugprone-posix-return,-warnings-as-errors]
    if (pthread_setspecific(profiler_data_key, (void*)val) < 0) {
                                                           ^
                                                           >
```

Full list of available linter checks: https://clang.llvm.org/extra/clang-tidy/checks/list.html

---


.clang-tidy
```diff
@@ -0,0 +1,26 @@
...
+  -bugprone-narrowing-conversions,
+  -bugprone-branch-clone,
+  -bugprone-not-null-terminated-result,
+  -bugprone-signed-char-misuse,
+  -bugprone-sizeof-expression
+CheckOptions:
+  - key: readability-identifier-naming.LocalVariableCase
+    value: lower_case
+  - key: readability-identifier-naming.LocalVariableIgnoredRegexp
+    value: '(KB|Thread|setDaemon|klassOop|nVMs|loadLibrary|getTicksFrequency|counterTime|System|M|R)'
```
apangin: This one concerns me. I don't want to explicitly list all java identifiers used in C++ code.

src/frameName.cpp
```diff
@@ -44,6 +44,8 @@ Matcher::Matcher(const Matcher& m) {
 }
 
 Matcher& Matcher::operator=(const Matcher& m) {
+    if(this == &m) return *this;
```
apangin: Why didn't linter complain about no space after `if`? :)

src/rustDemangle.cpp
```diff
@@ -45,10 +45,10 @@ struct demangle_legacy {
 
 // private version of memrchr to avoid _GNU_SOURCE
 static void *demangle_memrchr(const void *s, int c, size_t n) {
-    const uint8_t *s_ = (uint8_t *)s;
+    const uint8_t *bytes = (uint8_t *)s;
```
apangin: This file came from another project. I don't want it to diverge from the original source.

src/frameName.cpp
fandreuz: I guess that's more of a formatting problem. clang-tidy doesn't even have a way to check the desired pointer alignment.

.clang-tidy
fandreuz: I can do it :)

Jokes aside, I think it's a good compromise for what we get in exchange.

src/rustDemangle.cpp
fandreuz: Fixed in 5e3d572737fdb2cac2498349010e3949fd1c1962

src/frameName.cpp
fandreuz: `clang-format` would do that, but unlike `clang-tidy` there's no option to just check for one thing. It's designed to execute all checks

fandreuz: 259c11ef2f3626e78e89563b516a6854bf8d01c5

src/ctimer_linux.cpp
```diff
@@ -38,7 +38,7 @@ int CTimer::createForThread(int tid) {
     sev.sigev_value.sival_ptr = NULL;
     sev.sigev_signo = _signal;
     sev.sigev_notify = SIGEV_THREAD_ID;
-    ((int*)&sev.sigev_notify)[1] = tid;
+    ((&sev.sigev_notify))[1] = tid;
```
apangin: Double parentheses?

Makefile
```diff
@@ -285,3 +285,24 @@ clean-coverage:
 
 clean:
 	$(RM) -r build
+
+LINT_USE_DOCKER?=false
+LINT_DIFF_ONLY=true
+cpp-lint:
```
apangin: Do not place any targets after `clean`. Separate targets with a newline.

Makefile
```diff
@@ -285,3 +285,24 @@ clean-coverage:
 
 clean:
 	$(RM) -r build
+
+LINT_USE_DOCKER?=false
+LINT_DIFF_ONLY=true
```
apangin: Why `?=` in one case and `=` in another?

Makefile
```diff
@@ -285,3 +285,24 @@ clean-coverage:
...
+		LINT_SOURCES=$$(GLOBIGNORE=src/rustDemangle.cpp; ls src/*.cpp); \
+	fi; \
+	if [ -z "$$LINT_SOURCES" ]; then \
+		echo "Nothing to check"; exit 0; \
+	fi; \
+	if command -v clang-tidy >/dev/null 2>&1 && [ "$(LINT_USE_DOCKER)" == "false" ]; then \
+		clang-tidy $$LINT_SOURCES -- $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS); \
+	else \
+		# TODO: Replace with identifier to Docker image when it's available \
+		set -x; docker run -v $$(pwd):/async-profiler -it --rm \
```
apangin: As mentioned earlier, Makefile targets should not generally run docker etc. Environment should be provided by a caller. It's especially confusing that the same target runs docker in one situations and does not in other, depending on multiple conditions.

fandreuz: It's not that uncommon to have tools like formatters and linters in a Docker image. The main benefit is that the whole team can use the same version of the tool, which is convenient because newer versions typically introduce new checks.

This same image is also what we are going to use in GitHub Actions, and having locally the same tools we have in the CI means we are always going to receive the same set of warnings in both cases.

> It's especially confusing that the same target runs docker in one situations and does not in other, depending on multiple conditions.

Would it be ok if we always used Docker? That would be even better for me.

src/ctimer_linux.cpp
fandreuz: Fixed, that wasn't intended

Makefile
fandreuz: Right, no reason. Fixed in b636cf90343877a300a0a3c00f9d1e397e578c83

Makefile
```diff
@@ -285,3 +285,25 @@ clean-coverage:
 
 clean:
 	$(RM) -r build
+
+LINT_USE_DOCKER=false
+LINT_DIFF_ONLY=true
+cpp-lint:
+	@if [ "$(LINT_DIFF_ONLY)" == "true" ]; then \
+		LINT_SOURCES=$$(git diff --name-only | grep -e ".cpp" -e ".h"); \
```
Baraa-Hasheesh: Won't this lint the entire changed file rather than lint the changed lines only?
Is this on purpose?

Makefile
```diff
@@ -285,3 +285,25 @@ clean-coverage:
...
+	else \
+		LINT_SOURCES=$$(ls src/*.cpp); \
+	fi; \
+	LINT_SOURCES=$$(echo $$LINT_SOURCES | grep -v src/rustDemangle.cpp); \
```
Baraa-Hasheesh: why are we excluding `rustDemangle.cpp`

Makefile
```diff
@@ -285,3 +285,25 @@ clean-coverage:
...
+	if [ -z "$$LINT_SOURCES" ]; then \
+		echo "Nothing to check"; exit 0; \
+	fi; \
+	if command -v clang-tidy >/dev/null 2>&1 && [ "$(LINT_USE_DOCKER)" == "false" ]; then \
```
Baraa-Hasheesh: how about we print a message to the user requesting that he installs `clang-tidy`?

Makefile
```diff
@@ -285,3 +285,25 @@ clean-coverage:
...
+		clang-tidy $$LINT_SOURCES -- -x c++ $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS); \
+	else \
+		# TODO: Replace with identifier to Docker image when it's available \
+		docker run -v $$(pwd):/async-profiler -it --rm \
```
Baraa-Hasheesh: I would suggest checking if docker is installed before calling it,

src/ctimer_linux.cpp
```diff
@@ -38,7 +38,7 @@ int CTimer::createForThread(int tid) {
     sev.sigev_value.sival_ptr = NULL;
     sev.sigev_signo = _signal;
     sev.sigev_notify = SIGEV_THREAD_ID;
-    ((int*)&sev.sigev_notify)[1] = tid;
+    (&sev.sigev_notify)[1] = tid;
```
Baraa-Hasheesh, line 41: This change looks weird & if I'm not wrong it's not equivalent for all use cases,

Makefile
Baraa-Hasheesh: @apangin should we include `.c` & `.hpp` files here as well ?

If I'm not wrong the linter here won't check those files currently

Makefile
```diff
@@ -285,3 +285,25 @@ clean-coverage:
...
 clean:
 	$(RM) -r build
+
+LINT_USE_DOCKER=false
+LINT_DIFF_ONLY=true
+cpp-lint:
+	@if [ "$(LINT_DIFF_ONLY)" == "true" ]; then \
+		LINT_SOURCES=$$(git diff --name-only | grep -e ".cpp" -e ".h"); \
+	else \
+		LINT_SOURCES=$$(ls src/*.cpp); \
```
Baraa-Hasheesh: let's also include `.h` files here

fandreuz: It's going to check non-system header files automatically when they're included by the cpp files

src/ctimer_linux.cpp
fandreuz: This is documented as int in the [standard library](https://man7.org/linux/man-pages/man3/sigevent.3type.html), so I don't think your concern applies here. If you think that this specific line can be implementation dependent, let's discuss about it. If you talk general, I propose to discuss the topic when it shows up in a PR

Makefile
fandreuz: Part of a different project, as Andrei said during the meeting last friday

Makefile
fandreuz: I check clang-tidy because I want to have an automatic fallback to Docker if clang-tidy is not installed. If we check for docker, we are going to have something like this:
```bash
[...]
        if command -v clang-tidy >/dev/null 2>&1 && [ "$(LINT_USE_DOCKER)" == "false" ]; then \
		clang-tidy $$LINT_SOURCES -- -x c++ $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS); \
	elif command -v docker >/dev/null; then \
		# TODO: Replace with identifier to Docker image when it's available \
		docker run -v $$(pwd):/async-profiler -it --rm \
			--workdir /async-profiler \
			tidy -- \
			$$LINT_SOURCES -- -x c++ $(CXXFLAGS) $(DEFS) $(LIBS); \
	else
		if [ "$(LINT_USE_DOCKER)" == "false" ]; then \
			echo "clang-tidy and Docker not found"; \
		else \
			echo "Docker not found" \
		fi; \
	fi
```
just to print an error message that makes sense. I honestly don't see the added value of checking for Docker, we'll have a natural failure from the shell when we try to use Docker anyway

Baraa-Hasheesh: I would recommend in that just not checking either & always use `LINT_USE_DOCKER` variable 

if you don't want to always do `LINT_USE_DOCKER=true make cpp-lint`
You can define LINT_USE_DOCKER in your `bashrc` as `true` 

Having multiple control knobs on a logic that already have a clear control knob doesn't look logical to me

fandreuz: Why would I want to have an environment variable for something as simple as this? Even worse if it has to be part of my bashrc, since I have two sync it between three different hosts. 

Bash gives us a way to check if a command exist or not, I don't see why we should invent our own scaffolding just for this.

Makefile
fandreuz: > Won't this lint the entire changed file rather than lint the changed lines only? Is this on purpose?

Yes, but since we'll get no warnings with the current state, it's not a big deal.

There's a solution based on a Python script: https://clang.llvm.org/extra/doxygen/clang-tidy-diff_8py_source.html

It ingests the diff and it outputs problems on that diff only. But this approach assumes that this file is available locally (must be curl-ed) and Python is installed. Thus I would use it only if we decide to only support the Docker based approach in the Makefile.

fandreuz: > should we include .c

If you refer to `jattach` files, Andrei said clearly that code coming from a different project shouldn't be stlye-checked.

> .hpp files here as well

will be checked automatically since they are included by the cpp files.

Makefile
Baraa-Hasheesh: I don't understand this comment at all
* you're already using env variable for this
* The make file already uses multiple env variables.

==> `echo "export LINT_USE_DOCKER=true" >> ~/.bashrc`

I don't think executing this command on 3 different hosts is complex,

fandreuz: Makefile variable is not the same as env variable.

Baraa-Hasheesh: https://www.math.utah.edu/docs/info/make_6.html#SEC66

Every environment variable that make sees when it starts up is transformed into a make variable with the same name and value

`LINT_USE_DOCKER ?= false`

we already do the same for `CC, CXX`

fandreuz: > https://www.math.utah.edu/docs/info/make_6.html#SEC66

That sounds like "every environment variable can be read as a makefile variable", I still don't see where I'm defining new environment variables though. But I get your point. My point is that I don't feel like what you suggest is a simplification in terms of usability.

I'd personally prefer to have a Docker-only approach. If that's not accepted, I think we should fallback to Docker automatically when clang-tidy is not found. No additional variables should be needed in the happy path (clang-tidy is found), or the almost-happy-path (clang-tidy not found and Docker is found).

Makefile
fandreuz: If we could always use Docker, that would simplify dependency management. We could use `clang-tidy-diff`, which is an official wrapper written in Python to run clang-tidy only on the diff for a PR. That would be the cleanest solution in my opinion.

Makefile
fandreuz: Fixed: f71100946951be1510659ae86fd0cb5729fcf961

Makefile
fandreuz: I don't think that's really needed, we don't do that for e.g. java or g++

Makefile
Baraa-Hasheesh: > > should we include .c
> 
> If you refer to `jattach` files, Andrei said clearly that projects coming from a different project shouldn't be stlye-checked.

I'm referring to the case where a user defines a new c file as we do in test cases
on same point we have many `.c` files under `test` that I don't think we are currently checking right?

fandreuz: We don't need to enable all checks on the whole codebase as soon as we merge this PR. We can proceed incrementally, especially because we're probably going to see lots of warnings in test code. I'd argue that would cause more harm than good overall.

Baraa-Hasheesh: I agree with that's it's better to enable test checks later, but this doesn't answer my original question if we should enable checks on C files under normal src

fandreuz: https://github.com/async-profiler/async-profiler/pull/1338#discussion_r2161033409

Makefile
apangin: > The main benefit is that the whole team can use the same version of the tool

The plan is to run `clang-tidy` on GHA, right? Developers are free to use whatever version they want.
It's `make` that runs in a docker container, not vice versa.
I'll be worried if `make` starts pulling something from internet without asking.

apangin: Anyway, no downloads of third-party stuff from makefile, please. No matter if it is `wget`, `curl`, `docker run` or `git clone`.

fandreuz: Point taken, fixed here: 84262927bfc0129770f0edc0c2ea808f039051b2

I added some notes about the usage here: c9448ac1c9ac1cc87f23d035e0a8984fa15e16ca

Makefile
fandreuz: Not relevant anymore after the latest changes

CONTRIBUTING.md
```diff
@@ -44,6 +44,26 @@ GitHub provides additional document on [forking a repository](https://help.githu
 [creating a pull request](https://help.github.com/articles/creating-a-pull-request/).
 
 
+### Static code analysis
+We use tools like [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) to inspect new code and find problems quickly. 
+Before opening a PR, you should run the analyzers on the diff you are going to propose and fix all the reported warnings:
```
apangin: No, static analyzers should _help_ developers contribute instead of imposing extra routine.
Why would we ask people to do something manually what can be automated in CI?

CONTRIBUTING.md is a generic contribution guideline taken from AWS template, I prefer not to overload it with custom things that are not very important.

Makefile
```diff
@@ -274,6 +274,15 @@ build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
...
 	$(JAR) cf $@ -C build/test .
 
+.ONESHELL:
+cpp-lint:
+	@LINT_SOURCES=$$(ls -1 src/*.cpp src/*/*.cpp | grep -v rustDemangle.cpp)
+	clang-tidy $$LINT_SOURCES -- -x c++ $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS)
+
+cpp-lint-diff:
+	git diff -U0 -- '**/*.cpp' '**/*.h' ':!**/rustDemangle.cpp' | \
+		clang-tidy-diff $$LINT_SOURCES -- -x c++ $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS)
```
apangin: `LINT_SOURCES` is not defined in this target?

Makefile
```diff
@@ -274,6 +274,15 @@ build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
 	$(JAVAC) -source $(JAVA_TARGET) -target $(JAVA_TARGET) -Xlint:-options -cp "build/jar/*:build/converter/*" -d build/test $(TEST_SOURCES)
 	$(JAR) cf $@ -C build/test .
 
+.ONESHELL:
```
apangin: Make this explicit by combining below commands into one (`&&`)

Makefile
```diff
@@ -274,6 +274,15 @@ build/$(TEST_JAR): $(TEST_SOURCES) build/$(CONVERTER_JAR)
...
+cpp-lint:
+	@LINT_SOURCES=$$(ls -1 src/*.cpp src/*/*.cpp | grep -v rustDemangle.cpp)
```
apangin: Can we use "``" so that we don't need to escape `$`?

docker/code-check.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Image for all tasks related to static code analysis in Async-Profiler
+FROM public.ecr.aws/docker/library/amazoncorretto:11-alpine-jdk
+
+RUN apk add --no-cache clang-extra-tools linux-headers make wget python3 git && \
+    wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py -O /usr/bin/clang-tidy-diff && \
```
apangin: You can use `ADD` command to add a file to an image directly without need to install wget.

If the script is called `clang-tidy-diff.py`, why not leaving the original name?

Finally, use specific commit when referring to third-party tools on github.

CONTRIBUTING.md
fandreuz: reverted: d42d35682334666fb26d522812a8d5fa2a9865b0

Makefile
fandreuz: Should not be here indeed: 4796581bd4b252563b5c038596b9ba40d3920823

Makefile
fandreuz: b4fa8c55677bec5245a312cfe917d28d481d3fd7

Makefile
fandreuz: Sure: 6bcd07823c128eb3d91c2d53b1cb31aee924c729

docker/code-check.Dockerfile
fandreuz: Thanks, didn't think about it: 75f6beb835c3fbfe2bfac29ba5e06b27f8e3fc44

docker/code-check.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Image for all tasks related to static code analysis in Async-Profiler
+FROM public.ecr.aws/docker/library/amazoncorretto:11-alpine-jdk
+
+ADD https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-20.1.7/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py /usr/bin/
+RUN apk add --no-cache clang-extra-tools linux-headers make python3 git && \
+    chmod +x /usr/bin/clang-tidy-diff.py
```
apangin: I see `ADD` has [`--chmod`](https://docs.docker.com/reference/dockerfile/#add---chown---chmod) option, does it work here?

docker/code-check.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Image for all tasks related to static code analysis in Async-Profiler
+FROM public.ecr.aws/docker/library/amazoncorretto:11-alpine-jdk
+
+ADD https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-20.1.7/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py /usr/bin/
```
apangin: That is better. However, tags are still mutable. Specific commit id would be ideal.

docker/code-check.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Image for all tasks related to static code analysis in Async-Profiler
```
apangin: async-profiler

docker/code-check.Dockerfile
fandreuz: Had no idea it existed, thanks: a6cb3044ee21773c134b64b251e28b41ce373e36

docker/code-check.Dockerfile
fandreuz: Sure: a6cb3044ee21773c134b64b251e28b41ce373e36

But on a side not, maybe mutability is not a bad thing for tags? If a tag was overwritten, maybe there was a bug in the release. And a commit wouldn't protect for that.

docker/code-check.Dockerfile
fandreuz: a6cb3044ee21773c134b64b251e28b41ce373e36

docker/code-check.Dockerfile
apangin: So you no longer need to run `chmod +x`, right?

fandreuz: Right: 32d9bb3b12ce8af463d50789bec9fba3d6467cbb

docker/code-check.Dockerfile
apangin: Mutability is not bad for tags, but my intention is to minimize dependencies outside our control, i.e., to mitigate supply chain threats when a third party can unilaterally change code that is run on our servers.

PR #1341, opened 2025-06-13
Change stackwalker test checks to be more restrictive

Baraa-Hasheesh:
### Description
Change the assertions on Stackwalker tests to be more restrictive to avoid false negatives
The check now explicitly checks for main with native stack frame from the JNI call which shouldn't exist due to unwinding checks 

### Related issues
N/A

### Motivation and context
reduce random failures on GHA

### How has this been tested?
`make test`

---

### DIFF
--- a/test/test/stackwalker/StackwalkerTests.java
+++ b/test/test/stackwalker/StackwalkerTests.java
@@ -25,8 +25,8 @@ public class StackwalkerTests {
         assert output.contains("^Java_test_stackwalker_StackGenerator_largeFrame;" +
                 "doCpuTask");
 
-        // There will be no stack frame that contains both main & largeFrame method
-        assert !output.contains(".*main.*largeFrame");
+        // There will be no stack frame that contains both main & largeFrame native method
+        assert !output.contains(".*main.*Java_test_stackwalker_StackGenerator_largeFrame");
     }
 
     @Test(mainClass = StackGenerator.class, jvmArgs = "-Xss5m", args = "deepFrame",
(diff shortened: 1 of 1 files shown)


test/test/stackwalker/StackwalkerTests.java
```diff
@@ -25,8 +25,8 @@ public void largeFrame(TestProcess p) throws Exception {
         assert output.contains("^Java_test_stackwalker_StackGenerator_largeFrame;" +
                 "doCpuTask");
 
-        // There will be no stack frame that contains both main & largeFrame method
-        assert !output.contains(".*main.*largeFrame");
+        // There will be no stack frame that contains both main & largeFrame native method
+        assert !output.contains(".*main.*Java_test_stackwalker_StackGenerator_largeFrame");
```
apangin: This condition looks wrong. `main` does call `Java_test_stackwalker_StackGenerator_largeFrame`, and there is a high chance there will be such a stack trace.
What if a sample occurs immediately after calling `largeFrame` but before a stack frame is constructed?

Baraa-Hasheesh: would that be even possible? 

https://github.com/async-profiler/async-profiler/blob/master/test/native/libs/jninativestacks.c#L34

The `char frame[0x50000]` guarantees that the stack unwinding won't continue past `Java_test_stackwalker_StackGenerator_largeFrame`

For Java_test_stackwalker_StackGenerator_deepFrame that could be possible (a stack is captured during one of the recursive calls of `generateDeepStack`)

How about we make the condition 

`.*main.*Java_test_stackwalker_StackGenerator_deepFrame.*doCpuTask`  as that should be impossible in both cases?

apangin: > would that be even possible?

Yes, those 0x50000 bytes are allocated on stack after entering the function, so there is a period of time when async-profiler can walk from `Java_test_stackwalker_StackGenerator_largeFrame` down to `main`.

> `.*main.*Java_test_stackwalker_StackGenerator_deepFrame.*doCpuTask`

This will likely work.

Baraa-Hasheesh: Updated here [cb02008](https://github.com/async-profiler/async-profiler/pull/1341/commits/cb02008d4964bb4d0451cc0193de369a815ce1fe)

PR #1342, opened 2025-06-16
Add mapping info to OTLP profiling data

fandreuz:
### Description
In this PR I add [`Mapping`](https://github.com/open-telemetry/opentelemetry-proto/blob/fe18c4b149ed308a7af4b47075c32742e30ebd59/opentelemetry/proto/profiles/v1development/profiles.proto#L417) information to OTLP profiles. I fill the following fields:
- `memory_start`
- `memory_limit`
- `file_name_strindex`

Other changes:
- made `Index` generic, such that it can be used for other key types than `std::string`
- changed constant naming in `otlp.h` (`mapping_table` -> `MAPPING_TABLE`)
- converted `Location`, `Line` and `Function` (`otlp.h`) from namespaces to hashable classes, such that they can be used with `Index<T>` for deduplication.

### Motivation and context
Having mapping data is useful for consumers of OTLP data, since it allows showing more information about the profile. It's optional but nice to have.

---

### DIFF
--- a/src/frameName.h
+++ b/src/frameName.h
@@ -49,6 +49,49 @@ class Matcher {
     bool matches(const char* s);
 };
 
+class FrameInfo {
+  private:
+    const char* _name;
+    // E.g. non demangled name
+    const char* _system_name;
+    const char* _file_name;
+    const CodeCache* _cc;
+    const u64 _address;
+    const u64 _line;
+
+  public:
+    FrameInfo(const char* name, const char* system_name, const char* file_name, const CodeCache* cc, u64 address, u64 line) :
+      _name(name), _system_name(system_name), _file_name(file_name), _cc(cc), _address(address), _line(line) {}
+
+    FrameInfo(const char* name, const char* system_name, const CodeCache* cc) :
+      FrameInfo(name, system_name, nullptr, cc, 0, 0) {}
+
+    FrameInfo(const char* name) : FrameInfo(name, nullptr, nullptr, nullptr, 0, 0) {}
+
+    const char* get_name() const {
+      return _name;
+    }
+
+    const char* get_system_name() const {
+      return _system_name;
+    }
+
+    const char* get_file_name() const {
+      return _file_name;
+    }
+
+    const CodeCache* get_lib() const {
+      return _cc;
+    }
+
+    u64 get_address() const {
+      return _address;
+    }
+
+    u64 get_line() const {
+      return _line;
+    }
+};
 
 class FrameName {
   private:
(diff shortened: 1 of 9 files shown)


fandreuz:
I'd like to add some high-level tests, but I'm waiting for #1331 to be merged.

src/frameName.h
```diff
@@ -49,6 +50,49 @@ class Matcher {
...
 
+class FrameInfo {
+  private:
+    const char* _name;
+    // E.g. non demangled name
+    const char* _system_name;
+    const char* _file_name;
+    const CodeCache* _cc;
+    const u64 _address;
+    const u64 _line;
```
apangin, line 61: There should be a simpler way to do what you want without extensive changes in `FrameName`.
But if you aim at refactoring of symbol resolution, it needs to be done differently and in a separate PR. This will likely require unification with `Lookup` in `flightRecorder.cpp`. Not sure if you want to take this path.

fandreuz: The changes I have now in `frameName` are quite minor, I just take information which is already there and provide it in addition to the frame name.

> But if you aim at refactoring of symbol resolution, it needs to be done differently and in a separate PR.

The goal is to fill the following information for all frames:
```
FrameInfo {
    const char* _name;
    const char* _system_name;
    const char* _file_name;
    const CodeCache* _cc;
    const u64 _address;
    const u64 _line;
}
```

- `name` is already there
- `system_name` is already there
- `file_name` is probably not easy to get for native symbols unless we have debug information, but for JVM we could maybe use [this](https://docs.oracle.com/javase/8/docs/platform/jvmti/jvmti.html#GetSourceFileName);
- `cc` is already there
- `address` is discarded when we store the frame as an `ASGCT_CallFrame`, but we could keep it somewhere.
- `line`: same as `file_name`, for Java frames we could maybe use [this](https://docs.oracle.com/javase/8/docs/platform/jvmti/jvmti.html#GetLineNumberTable)

fandreuz: > But if you aim at refactoring of symbol resolution, it needs to be done differently and in a separate PR. This will likely require unification with Lookup in flightRecorder.cpp.

@apangin could you elaborate on this?

apangin: `FrameName`, as the class suggests, is supposed to resolve a frame to its symbolic name, not something else. And returning a composite value from a function is generally discouraged.

There is `Lookup` class in FlightRecorder that deals with extra frame info, such as line numbers, bytecode locations, frame types, etc. If you plan to emit all this information in OTLP, it makes sense to refactor `Lookup` to make it reusable across different formats. However, since this PR only adds library names and mangled symbols, it can do this right in `dumpOtlp` without touching `FrameName` at all.

Issue #1320, opened 2025-05-28
Parse DWARF unwinding info in .debug_frame section


apangin:
### Describe the feature

Currently, `dwarf` and `vm` stack walking modes use unwinding info from `.eh_frame` section on Linux, or, more precisely, async-profiler parses sorted FDEs referenced in `.eh_frame_hdr`.

However, in some cases, `.eh_frame` section is empty, and the unwinding info is available in `.debug_frame` instead. Async-profiler should support both `.eh_frame` and `.debug_frame` sections.

### Use Case

1. Stack walking does not work properly for musl binary on certain versions of Alpine Linux, see #1317.
2. GraalVM native images do not have `.eh_frame`, but they may be compiled with `.debug_frame` section in a separate debuginfo file. By supporting `.debug_frame`, it will be possible to profile GraalVM native images with async-profiler.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

visheshruparelia:
https://github.com/oracle/graal/issues/11622

PR #1347, opened 2025-06-25
Parse .debug_frame section if exists

visheshruparelia:
### Description
While parsing the symbols for libraries, prefer `.debug_frame` section over `.eh_frame_hdr` section for better unwinding. 
Also, add a new terminate condition while stack walking via `cstack=vm` or `cstack=dwarf`: stop the walk if `prev_pc == next_pc and prev_sp = next_sp`. This fixes the issue where in the flamegraph we see multiple "unknown" frames whereas in reality there is only one (see "before" section in the manual testing section for incorrect flamegraph).

### Related issues
https://github.com/async-profiler/async-profiler/issues/1320

### How has this been tested?
We create a `libdwarfdebugframe.so` from `dwarfdebugframe.c`. This `so` is stripped of `.eh_frame` and `.eh_frame_hdr`. 
#### Unit Testing
Added unit test for `linux` which initialize the `DwarfParser` class using the newly added constructor and attempts to parse the `libdwarfdebugframe.so` file. Also, we test scenarios where incorrect and invalid address are passed to the constructor arguments. 
#### Integration Testing
We have a new `StackGenForUnwindViaDebugFrame.java` which uses the `libdwarfdebugframe.so` file to generate a trace which we do cpu profiling for. In the test, we check whether the stack was unwound properly for `vm` and `vmx` stack walking modes.
#### Manual Testing
Setup:
Alpine Latest with `musl-dbg`, `openjdk8`, `openjdk8-dbg` and `openjdk11` (`dbg` pkg for java 11 and onwards does not exist).
Profiled application: There is a `repro.tar.gz` provided by [here](https://github.com/async-profiler/async-profiler/issues/1317#issue-3092439876) which contains a test application `Reproducer.java`. This application was profiled in the following scenarios:

###### OpenJDK 8 `-e cpu cstack=vm -d 10`
**NOTE:** There are "unknown" frames.
image
[jdk8_alpinelatest_before_vm.html.zip](https://github.com/user-attachments/files/20901791/jdk8_alpinelatest_before_vm.html.zip)

###### OpenJDK 11 `-e cpu cstack=vm -d 10`
image
[jdk11_alpinelatest_before_vm.html.zip](https://github.com/user-attachments/files/20901801/jdk11_alpinelatest_before_vm.html.zip)

###### OpenJDK 8 `-e cpu cstack=vm -d 10`
We can now see that there are no "unknown" frames. 
image
[jdk8_alpinelatest_after_vm.html.zip](https://github.com/user-attachments/files/20901667/jdk8_alpinelatest_after_vm.html.zip)
###### OpenJDK 11 `-e cpu cstack=vm -d 10`
**NOTE:** Since we don't have debug symbols for jdk11, method names are not seen. However, Native calls from `musl` like `_clone` can be seen now as opposed to "unknown" before this change.
image
[jdk11_alpinelatest_after_vm.html.zip](https://github.com/user-attachments/files/20901677/jdk11_alpinelatest_after_vm.html.zip)

---


src/dwarf.cpp
```diff
@@ -71,10 +72,25 @@ DwarfParser::DwarfParser(const char* name, const char* image_base, const char* e
     _code_align = sizeof(instruction_t);
     _data_align = -(int)sizeof(void*);
 
-    parse(eh_frame_hdr);
+    parseEhFrameHdr(eh_frame_hdr);
 }
 
-void DwarfParser::parse(const char* eh_frame_hdr) {
+DwarfParser::DwarfParser(const char* name, const char* image_base, const char* debug_frame_start, const char* debug_frame_end) {
```
visheshruparelia: @apangin alt approach I considered here was to create two classes `DwarfEhFrameParser` and `DwarfDebugFrameParser` which inherit a common class instead of having parser specific methods like `parseEhFrame()` and `parseDebugFrame()` in the same class. Let me know if you would prefer that.

Makefile
```diff
@@ -231,6 +231,7 @@ build-test-libs:
 	$(CC) -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjnimalloc.$(SOEXT) test/native/libs/jnimalloc.c
 	$(CC) -shared -fPIC -o $(TEST_LIB_DIR)/libmalloc.$(SOEXT) test/native/libs/malloc.c
 	$(CC) -fno-optimize-sibling-calls -shared -fPIC $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libjninativestacks.$(SOEXT) test/native/libs/jninativestacks.c
+	$(CC) -g -shared -fPIC -fno-dwarf2-cfi-asm -fomit-frame-pointer $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libdwarfdebugframe.$(SOEXT) test/native/libs/dwarfdebugframe.c
```
visheshruparelia: @apangin I tried using `objcopy --remove-section=.eh_frame --remove-section=.eh_frame_hdr` to ensure parsing happens via `.debug_frame` in the Integration tests but for some reason it was corrupting the file in `debian x64` and was giving `UnsatisfiedLinkError`. I also used `strip` but same result.
However, in my local machine using Amazon Linux, it worked just fine.

visheshruparelia: Tried some more stuff, no luck. The exact error is: `Exception in thread "main" java.lang.UnsatisfiedLinkError: /tmp/async-profiler/build/test/lib/libdwarfdebugframe.so: /tmp/async-profiler/build/test/lib/libdwarfdebugframe.so: ELF load command address/offset not properly aligned`

Output of `readelf -W -a`:

```
readelf -W -a build/test/lib/libdwarfdebugframe.so
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x1080
  Start of program headers:          64 (bytes into file)
  Start of section headers:          29416 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         8
  Size of section headers:           64 (bytes)
  Number of section headers:         31
  Section header string table index: 30

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .note.gnu.build-id NOTE            0000000000000238 000238 000024 00   A  0   0  4
  [ 2] .gnu.hash         GNU_HASH        0000000000000260 000260 000030 00   A  3   0  8
  [ 3] .dynsym           DYNSYM          0000000000000290 000290 0000f0 18   A  4   1  8
  [ 4] .dynstr           STRTAB          0000000000000380 000380 0000ca 00   A  0   0  1
  [ 5] .gnu.version      VERSYM          000000000000044a 00044a 000014 02   A  3   0  2
  [ 6] .gnu.version_r    VERNEED         0000000000000460 000460 000020 00   A  4   1  8
  [ 7] .rela.dyn         RELA            0000000000000480 000480 0000a8 18   A  3   0  8
  [ 8] .rela.plt         RELA            0000000000000528 000528 000060 18  AI  3  18  8
  [ 9] .init             PROGBITS        0000000000001000 001000 000017 00  AX  0   0  4
  [10] .plt              PROGBITS        0000000000001020 001020 000050 10  AX  0   0 16
  [11] .plt.got          PROGBITS        0000000000001070 001070 000008 08  AX  0   0  8
  [12] .text             PROGBITS        0000000000001080 001080 00019c 00  AX  0   0 16
  [13] .fini             PROGBITS        000000000000121c 00121c 000009 00  AX  0   0  4
  [14] .init_array       INIT_ARRAY      0000000000003e10 001e10 000008 08  WA  0   0  8
  [15] .fini_array       FINI_ARRAY      0000000000003e18 001e18 000008 08  WA  0   0  8
  [16] .dynamic          DYNAMIC         0000000000003e20 001e20 0001c0 10  WA  4   0  8
  [17] .got              PROGBITS        0000000000003fe0 001fe0 000020 08  WA  0   0  8
  [18] .got.plt          PROGBITS        0000000000004000 002000 000038 08  WA  0   0  8
  [19] .data             PROGBITS        0000000000004038 002038 000008 00  WA  0   0  8
  [20] .bss              NOBITS          0000000000004040 002040 000008 00  WA  0   0  1
  [21] .comment          PROGBITS        0000000000000000 002040 00001c 01  MS  0   0  1
  [22] .debug_aranges    PROGBITS        0000000000000000 00205c 000030 00      0   0  1
  [23] .debug_info       PROGBITS        0000000000000000 00208c 002f1b 00      0   0  1
  [24] .debug_abbrev     PROGBITS        0000000000000000 004fa7 000218 00      0   0  1
  [25] .debug_line       PROGBITS        0000000000000000 0051bf 0001e9 00      0   0  1
  [26] .debug_frame      PROGBITS        0000000000000000 0053a8 000090 00      0   0  8
  [27] .debug_str        PROGBITS        0000000000000000 005438 001695 01  MS  0   0  1
  [28] .symtab           SYMTAB          0000000000000000 006ad0 000528 18     29  46  8
  [29] .strtab           STRTAB          0000000000000000 006ff8 0001cb 00      0   0  1
  [30] .shstrtab         STRTAB          0000000000000000 0071c3 00011e 00      0   0  1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  l (large), p (processor specific)

There are no section groups in this file.

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x000588 0x000588 R   0x1000
  LOAD           0x001000 0x0000000000001000 0x0000000000001000 0x000225 0x000225 R E 0x1000
  LOAD           0x001225 0x0000000000000000 0x0000000000000000 0x000000 0x000000 R   0x1000
  LOAD           0x001e10 0x0000000000003e10 0x0000000000003e10 0x000230 0x000238 RW  0x1000
  DYNAMIC        0x001e20 0x0000000000003e20 0x0000000000003e20 0x0001c0 0x0001c0 RW  0x8
  NOTE           0x000238 0x0000000000000238 0x0000000000000238 0x000024 0x000024 R   0x4
  GNU_EH_FRAME   0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 R   0x8
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x8

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.build-id .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_r .rela.dyn .rela.plt
   01     .init .plt .plt.got .text .fini
   02
   03     .init_array .fini_array .dynamic .got .got.plt .data .bss
   04     .dynamic
   05     .note.gnu.build-id
   06
   07

Dynamic section at offset 0x1e20 contains 24 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000c (INIT)               0x1000
 0x000000000000000d (FINI)               0x121c
 0x0000000000000019 (INIT_ARRAY)         0x3e10
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x3e18
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x260
 0x0000000000000005 (STRTAB)             0x380
 0x0000000000000006 (SYMTAB)             0x290
 0x000000000000000a (STRSZ)              202 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x4000
 0x0000000000000002 (PLTRELSZ)           96 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x528
 0x0000000000000007 (RELA)               0x480
 0x0000000000000008 (RELASZ)             168 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0x460
 0x000000006fffffff (VERNEEDNUM)         1
 0x000000006ffffff0 (VERSYM)             0x44a
 0x000000006ffffff9 (RELACOUNT)          3
 0x0000000000000000 (NULL)               0x0

Relocation section '.rela.dyn' at offset 0x480 contains 7 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000003e10  0000000000000008 R_X86_64_RELATIVE                         1130
0000000000003e18  0000000000000008 R_X86_64_RELATIVE                         10f0
0000000000004038  0000000000000008 R_X86_64_RELATIVE                         4038
0000000000003fe0  0000000100000006 R_X86_64_GLOB_DAT      0000000000000000 _ITM_deregisterTMCloneTable + 0
0000000000003fe8  0000000200000006 R_X86_64_GLOB_DAT      0000000000000000 __gmon_start__ + 0
0000000000003ff0  0000000500000006 R_X86_64_GLOB_DAT      0000000000000000 _ITM_registerTMCloneTable + 0
0000000000003ff8  0000000600000006 R_X86_64_GLOB_DAT      0000000000000000 __cxa_finalize@GLIBC_2.2.5 + 0

Relocation section '.rela.plt' at offset 0x528 contains 4 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000004018  0000000700000007 R_X86_64_JUMP_SLOT     0000000000001135 useCpu + 0
0000000000004020  0000000300000007 R_X86_64_JUMP_SLOT     0000000000000000 pow + 0
0000000000004028  0000000400000007 R_X86_64_JUMP_SLOT     0000000000000000 sqrt + 0
0000000000004030  0000000900000007 R_X86_64_JUMP_SLOT     00000000000011ae doRecursiveWork + 0

The decoding of unwind sections for machine type Advanced Micro Devices X86-64 is not currently supported.

Symbol table '.dynsym' contains 10 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_deregisterTMCloneTable
     2: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
     3: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND pow
     4: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND sqrt
     5: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_registerTMCloneTable
     6: 0000000000000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize@GLIBC_2.2.5 (2)
     7: 0000000000001135   121 FUNC    GLOBAL DEFAULT   12 useCpu
     8: 00000000000011ef    45 FUNC    GLOBAL DEFAULT   12 Java_test_stackwalker_StackGenForUnwindViaDebugFrame_startWork
     9: 00000000000011ae    65 FUNC    GLOBAL DEFAULT   12 doRecursiveWork

Symbol table '.symtab' contains 55 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000000238     0 SECTION LOCAL  DEFAULT    1
     2: 0000000000000260     0 SECTION LOCAL  DEFAULT    2
     3: 0000000000000290     0 SECTION LOCAL  DEFAULT    3
     4: 0000000000000380     0 SECTION LOCAL  DEFAULT    4
     5: 000000000000044a     0 SECTION LOCAL  DEFAULT    5
     6: 0000000000000460     0 SECTION LOCAL  DEFAULT    6
     7: 0000000000000480     0 SECTION LOCAL  DEFAULT    7
     8: 0000000000000528     0 SECTION LOCAL  DEFAULT    8
     9: 0000000000001000     0 SECTION LOCAL  DEFAULT    9
    10: 0000000000001020     0 SECTION LOCAL  DEFAULT   10
    11: 0000000000001070     0 SECTION LOCAL  DEFAULT   11
    12: 0000000000001080     0 SECTION LOCAL  DEFAULT   12
    13: 000000000000121c     0 SECTION LOCAL  DEFAULT   13
    14: 0000000000003e10     0 SECTION LOCAL  DEFAULT   14
    15: 0000000000003e18     0 SECTION LOCAL  DEFAULT   15
    16: 0000000000003e20     0 SECTION LOCAL  DEFAULT   16
    17: 0000000000003fe0     0 SECTION LOCAL  DEFAULT   17
    18: 0000000000004000     0 SECTION LOCAL  DEFAULT   18
    19: 0000000000004038     0 SECTION LOCAL  DEFAULT   19
    20: 0000000000004040     0 SECTION LOCAL  DEFAULT   20
    21: 0000000000000000     0 SECTION LOCAL  DEFAULT   21
    22: 0000000000000000     0 SECTION LOCAL  DEFAULT   22
    23: 0000000000000000     0 SECTION LOCAL  DEFAULT   23
    24: 0000000000000000     0 SECTION LOCAL  DEFAULT   24
    25: 0000000000000000     0 SECTION LOCAL  DEFAULT   25
    26: 0000000000000000     0 SECTION LOCAL  DEFAULT   26
    27: 0000000000000000     0 SECTION LOCAL  DEFAULT   27
    28: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS crtstuff.c
    29: 0000000000001080     0 FUNC    LOCAL  DEFAULT   12 deregister_tm_clones
    30: 00000000000010b0     0 FUNC    LOCAL  DEFAULT   12 register_tm_clones
    31: 00000000000010f0     0 FUNC    LOCAL  DEFAULT   12 __do_global_dtors_aux
    32: 0000000000004040     1 OBJECT  LOCAL  DEFAULT   20 completed.7325
    33: 0000000000003e18     0 OBJECT  LOCAL  DEFAULT   15 __do_global_dtors_aux_fini_array_entry
    34: 0000000000001130     0 FUNC    LOCAL  DEFAULT   12 frame_dummy
    35: 0000000000003e10     0 OBJECT  LOCAL  DEFAULT   14 __frame_dummy_init_array_entry
    36: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS dwarfdebugframe.c
    37: 00000000000011ae    65 FUNC    LOCAL  DEFAULT   12 doRecursiveWork.localalias.0
    38: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS crtstuff.c
    39: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS
    40: 000000000000121c     0 FUNC    LOCAL  DEFAULT   13 _fini
    41: 0000000000004038     0 OBJECT  LOCAL  DEFAULT   19 __dso_handle
    42: 0000000000003e20     0 OBJECT  LOCAL  DEFAULT   16 _DYNAMIC
    43: 0000000000004040     0 OBJECT  LOCAL  DEFAULT   19 __TMC_END__
    44: 0000000000004000     0 OBJECT  LOCAL  DEFAULT   18 _GLOBAL_OFFSET_TABLE_
    45: 0000000000001000     0 FUNC    LOCAL  DEFAULT    9 _init
    46: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_deregisterTMCloneTable
    47: 00000000000011ef    45 FUNC    GLOBAL DEFAULT   12 Java_test_stackwalker_StackGenForUnwindViaDebugFrame_startWork
    48: 0000000000001135   121 FUNC    GLOBAL DEFAULT   12 useCpu
    49: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
    50: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND pow
    51: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND sqrt
    52: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_registerTMCloneTable
    53: 00000000000011ae    65 FUNC    GLOBAL DEFAULT   12 doRecursiveWork
    54: 0000000000000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize@@GLIBC_2.2.5

Histogram for `.gnu.hash' bucket list length (total of 3 buckets):
 Length  Number     % of total  Coverage
      0  1          ( 33.3%)
      1  1          ( 33.3%)     33.3%
      2  1          ( 33.3%)    100.0%

Version symbols section '.gnu.version' contains 10 entries:
 Addr: 000000000000044a  Offset: 0x00044a  Link: 3 (.dynsym)
  000:   0 (*local*)       0 (*local*)       0 (*local*)       0 (*local*)
  004:   0 (*local*)       0 (*local*)       2 (GLIBC_2.2.5)   1 (*global*)
  008:   1 (*global*)      1 (*global*)

Version needs section '.gnu.version_r' contains 1 entry:
 Addr: 0x0000000000000460  Offset: 0x000460  Link: 4 (.dynstr)
  000000: Version: 1  File: libc.so.6  Cnt: 1
  0x0010:   Name: GLIBC_2.2.5  Flags: none  Version: 2

Displaying notes found in: .note.gnu.build-id
  Owner                 Data size	Description
  GNU                  0x00000014	NT_GNU_BUILD_ID (unique build ID bitstring)	    Build ID: 88b03e4cf97cea56afa98dbb558dd3ca28400143
  ```

Comment (visheshruparelia) [reply Makefile]: @apangin I was able to make it work. We have to update `binutils` in our debian docker image (`2.31` is the one we get right now but that seems to have this bug which does not exist in `2.44`). I checked the versions in other images we have for Amazon Linux, Alpine and it is not `2.31` and that is probably why it works fine in those OS. Let me do that and update the PR.

Comment (apangin) [reply Makefile]: Instead of removing `.eh_frame_hdr` after compilation, can you just produce binary without this section in the first place? Try `-Wl,--no-eh-frame-hdr` compiler options.

Comment (visheshruparelia) [reply Makefile]: For some reason, I did not find this option on googling and ended up using `objcopy`/`strip`. This worked, thank you @apangin for the pointer!

Comment (visheshruparelia) [reply Makefile]: One more thing, earlier today I saw that just removing one section via `objcopy` or `strip` does not cause any problem, but I wanted to remove both `eh_Frame_hdr` and `eh_frame` and hence I started thinking about alternatives. With this approach (`-Wl,--no-eh-frame-hdr`) only `eh_frame_hdr` gets removed but the other one still exists. Are we okay with that? (The code only looks for `eh_frame_hdr` so it should be ok i guess).

Comment (apangin) [reply Makefile]: Yes, should be ok

File: test/native/dwarfLinuxTest.cpp [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+}
+
+TEST_CASE(Dwarf_NullPointers) {
+    // Test constructor with null pointers
+    DwarfParser parser("test", nullptr, nullptr, nullptr);
+    
+    ASSERT_EQ(parser.count(), 0);
+}
+
+#endif // __linux__
```
Baraa-Hasheesh: white space issues here & otherfiles

src/dwarf.cpp
Baraa-Hasheesh: The constructors are quite identical to one another,

I think if we keep a single class we should remove calls to `parseEhFrameHdr` & `parseDebugFrame` 

& explicitly call them after the constructor is defined (Basically the constructor just initialzes the class & stops doing the auto parse

src/symbols_linux.cpp
```diff
@@ -274,6 +274,7 @@ class ElfParser {
     void loadSymbolTable(const char* symbols, size_t total_size, size_t ent_size, const char* strings);
     void addRelocationSymbols(ElfSection* reltab, const char* plt);
     const char* getDebuginfodCache();
+    void loadSymbolsFromDebugFrame();
```
Baraa-Hasheesh: I think we should name it `parseDebugFrame()`

src/symbols_linux.cpp
```diff
@@ -468,6 +469,26 @@ void ElfParser::parseDwarfInfo() {
...
+    if (_cc->dwarfTableLength() > 0) {
+        return;
+    }
+    ElfSection* debug_frame_section = findSection(SHT_PROGBITS, ".debug_frame");
+    if (debug_frame_section != NULL) {
+        if (debug_frame_section->sh_size > 0) {
+            const char* section_start = at(debug_frame_section);
+            DwarfParser dwarf(_cc->name(), _base, section_start, section_start + debug_frame_section->sh_size);
+            _cc->setDwarfTable(dwarf.table(), dwarf.count());
+        } else if (strcmp(_cc->name(), "[vdso]") == 0) {
```
Baraa-Hasheesh: Is this condition needed? => `else if (strcmp(_cc->name(), "[vdso]") == 0) `

if this happens won't the eh_frame_hdr parsing portion handle it?

src/dwarf.cpp
visheshruparelia: If creating two different classes seems like an overkill, we can do that too.

apangin: Yes, I also prefer the same class with a single constructor but with two different functions.

src/codeCache.h
```diff
@@ -152,6 +152,10 @@ class CodeCache {
         return _image_base;
     }
 
+    const int dwarfTableLength() const {
```
apangin: You don't actually use length; should it be called `hasDwarfTable()` and return `_dwarf_table != NULL` instead?

src/dwarf.cpp
```diff
@@ -90,11 +106,70 @@ void DwarfParser::parse(const char* eh_frame_hdr) {
...
+            cie_id = get32();
+        }
+        if (cie_id == 0xffffffff || cie_id == (u64)-1) {
+            parseDebugFrameCie(entry_start, length);
+        } else {
+            parseDebugFrameFde(entry_start, length);
+        }
+        _ptr = entry_start + length;
+    }
+    // Sort the table by location since .debug_frame entries may not be ordered
```
apangin, line 124: Can you check if the table can be compacted after sorting? Two adjacent FrameDescs can be merged if they only differ by `loc`.

Also, do you handle gaps between functions? I.e. when some part of code is not covered by FDEs?

src/symbols_linux.cpp
apangin: `[vdso]` will never appear here, since this function is only called for libraries backed by a real file on the filesystem.

src/symbols_linux.cpp
```diff
@@ -507,6 +528,8 @@ void ElfParser::loadSymbols(bool use_debug) {
             }
         }
     }
+
+    loadSymbolsFromDebugFrame();
```
apangin: This will not work for objects that have `.symtab`, but where `.debug_frame` is in a separate file.
But I'm not sure how common such situation is, probably not. So does your code parse graalvm native images correctly?

src/symbols_linux.cpp
```diff
@@ -468,6 +469,26 @@ void ElfParser::parseDwarfInfo() {
...
+    ElfSection* debug_frame_section = findSection(SHT_PROGBITS, ".debug_frame");
+    if (debug_frame_section != NULL) {
+        if (debug_frame_section->sh_size > 0) {
+            const char* section_start = at(debug_frame_section);
+            DwarfParser dwarf(_cc->name(), _base, section_start, section_start + debug_frame_section->sh_size);
+            _cc->setDwarfTable(dwarf.table(), dwarf.count());
+        } else if (strcmp(_cc->name(), "[vdso]") == 0) {
+            FrameDesc* table = (FrameDesc*)malloc(sizeof(FrameDesc));
+            *table = FrameDesc::empty_frame;
+            _cc->setDwarfTable(table, 1);
```
fandreuz:
```suggestion
            _cc->setDwarfTable(&FrameDesc::empty_frame, 1);
```

src/dwarf.cpp
fandreuz: It seems to me that the class is only used once to extract `table()` and `count()`, do we need a class at all? We could have a simple function `parseDwardTable` returning a `std::vector<FrameDesc>`

test/native/dwarfLinuxTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+class ElfReader {
+private:
+    int _fd;
+    void* _mapped;
+    size_t _size;
+    const char* _base;
+
+public:
+    ElfReader(const char* filename) : _fd(-1), _mapped(nullptr), _size(0), _base(nullptr) {
```
fandreuz: This constructor has a bit too much logic and branching, how about you make it dumber, and you construct it from an helper function? The helper function could also check that `mmap` and `open` completed correctly, otherwise we don't even try to construct an `ElfReader`.

fandreuz: This would also simplify the destructor.

test/native/dwarfLinuxTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+    
+    bool isValid() const { 
+        const Elf64_Ehdr* ehdr = (const Elf64_Ehdr*)_base;
+        return _mapped != nullptr
+            && ehdr->e_ident[EI_MAG0] == ELFMAG0
+            && ehdr->e_ident[EI_MAG1] == ELFMAG1
+            && ehdr->e_ident[EI_MAG2] == ELFMAG2
+            && ehdr->e_ident[EI_MAG3] == ELFMAG3;
+    }
+    const char* base() const { return _base; }
```
fandreuz, line 59: Newline

test/native/dwarfLinuxTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+    
+    bool findDebugFrame(const char*& start, const char*& end) {
```
fandreuz: Nit: Reference to pointer is harder to read in the function signature than `const char**`

test/test/stackwalker/StackwalkerTests.java
```diff
@@ -84,4 +85,47 @@ public void normalStackVM(TestProcess p) throws Exception {
                 "Java_test_stackwalker_StackGenerator_leafFrame;" +
                 "doCpuTask");
     }
+
+    @Test(mainClass = StackGenForUnwindViaDebugFrame.class, jvmArgs = "-Xss5m",
+            agentArgs = "start,event=cpu,cstack=vm,file=%f.jfr", os = Os.LINUX)
+    public void walkStackUsingDebugFrameVM(TestProcess p) throws Exception {
+        p.waitForExit();
+        assert p.exitCode() == 0;
+        Output output = Output.convertJfrToCollapsed(p.getFilePath("%f"));
```
fandreuz, line 94: Is there a specific reason to use JFR in the first place? I think it adds one layer of complexity to the test

src/dwarf.cpp
Baraa-Hasheesh: The class handles many aspects of the parsing in the background
* Acts as a container for common variables such as `_image_base`
* auto resize the table => `DwarfParser::addRecord`
* storing dynamic info such as `_code_align`, `_data_align`, `_ptr`

While it's indeed technically possible to refactor this to use static functions, it would look worse & be more complex 

IMO keeping the class is better

fandreuz: If the user of `dwarf.h` uses the class as a simple sized array, exposing it as a class makes it harder to reason about things like ownership. Am I allowed to do this?

```cpp
DwarfParser dwarf(_cc->name(), _base, at(eh_frame_hdr));
_cc1->setDwarfTable(dwarf.table(), dwarf.count());
_cc2->setDwarfTable(dwarf.table(), dwarf.count());
```

I understand the benefits of implementing the parser as a class, my point is that we could consider exposing it as a static function backed by a class.

test/test/stackwalker/StackwalkerTests.java
```diff
@@ -84,4 +85,47 @@ public void normalStackVM(TestProcess p) throws Exception {
                 "Java_test_stackwalker_StackGenerator_leafFrame;" +
                 "doCpuTask");
     }
+
+    @Test(mainClass = StackGenForUnwindViaDebugFrame.class, jvmArgs = "-Xss5m",
```
fandreuz, line 89: Is `-Xss5m` actually needed? How big are the stacks?

src/dwarf.cpp
visheshruparelia: > Also, do you handle gaps between functions? I.e. when some part of code is not covered by FDEs?
@apangin regarding above^ i am curious on how that can be done? Do we do something similar for `.eh_frame`? 
Does something like the following make sense (in the `codeCache.cpp#findFrameDesc()`): 

```
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -302,7 +302,13 @@ FrameDesc* CodeCache::findFrameDesc(const void* pc) {
     }
 
     if (low > 0) {
-        return &_dwarf_table[low - 1];
+        FrameDesc* prev = &_dwarf_table[low - 1];
+        // Check for gaps: if PC is too far from the previous FDE, use default frame
+        u32 gap = target_loc - prev->loc;
+        if (gap > 0x10000) {  // If gap is greater than 64kb, we return the default_frame
+            return &FrameDesc::default_frame;
+        }
+        return prev;
     } else if (target_loc - _plt_offset < _plt_size) {
         return &FrameDesc::empty_frame;
     } else {
```

apangin: If there is a gap between functions, async-profiler inserts a FrameDesc of the default frame.
E.g., function `foo` spans from offset 10 to offset 36 and function `bar` starts at offset 48. Then async-profiler will place a synthetic FrameDesc at offset 36. When there is no gap (e.g. function `bar` start at offset 36), async-profiler will also insert default FrameDesc at offset 36, but it will be later overwritten by the first FrameDesc corresponding to `bar`.

src/dwarf.cpp
```diff
@@ -90,11 +106,70 @@ void DwarfParser::parse(const char* eh_frame_hdr) {
...
+    while (_ptr < debug_frame_end) {
+        u32 initial_length = get32();
+        u64 length;
+        u64 cie_id;
+        const char* entry_start;
+        if (initial_length == 0 || _ptr + initial_length > debug_frame_end) {
+            break;
+        }
+        if (initial_length == 0xffffffff) {
+            // 64-bit DWARF format
```
krk, line 107: Do we have tests for both formats?

src/dwarf.cpp
```diff
@@ -71,10 +72,25 @@ DwarfParser::DwarfParser(const char* name, const char* image_base, const char* e
...
+    _table = (FrameDesc*)malloc(_capacity * sizeof(FrameDesc));
+    _prev = NULL;
+
+    _code_align = sizeof(instruction_t);
+    _data_align = -(int)sizeof(void*);
+
+    parseDebugFrame(debug_frame_start, debug_frame_end);
+}
+
+void DwarfParser::parseEhFrameHdr(const char* eh_frame_hdr) {
```
krk, line 76: To ensure we have the most complete unwinding information, I suggest we always parse `.eh_frame` and then augment that data with the contents of `.debug_frame` if it's present.

test/native/dwarfLinuxTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include "dwarf.h"
+#include "testRunner.hpp"
+#include <elf.h>
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+class ElfReader {
```
krk, line 30: Can we reuse `ElfParser` from `symbols_linux.cpp` here instead?

test/native/dwarfLinuxTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+};
+
+TEST_CASE(Dwarf_SuccessfulParse, fileReadable("build/test/lib/libdwarfdebugframe.so")) {
+    ElfReader elf("build/test/lib/libdwarfdebugframe.so");
+    ASSERT_EQ(elf.isValid(), true);
+    
+    const char* debug_start;
+    const char* debug_end;
+    ASSERT_EQ(elf.findDebugFrame(debug_start, debug_end), true);
```
krk: Is this an existence check? `findDebugFrame` -> `hasDebugFrame`

test/test/stackwalker/StackwalkerTests.java
```diff
@@ -84,4 +85,47 @@ public void normalStackVM(TestProcess p) throws Exception {
...
+                "useCpu");
+    }
+
+    @Test(mainClass = StackGenForUnwindViaDebugFrame.class, jvmArgs = "-Xss5m",
+            agentArgs = "start,event=cpu,cstack=vmx,file=%f.jfr", os = Os.LINUX)
+    public void walkStackUsingDebugFrameVMX(TestProcess p) throws Exception {
+        p.waitForExit();
+        assert p.exitCode() == 0;
+        Output output = Output.convertJfrToCollapsed(p.getFilePath("%f"));
+        assert output.contains("^" +
```
krk, line 112: What does this stack look like without your changes?

krk (changes requested):

src/dwarf.cpp
visheshruparelia: Is that possible? Can we have scenarios where `.eh_frame` has more information than `.debug_frame`?

test/test/stackwalker/StackwalkerTests.java
visheshruparelia: For this test, no. I am doing it this way to maintain consistency because other tests in the class work like this. 
@Baraa-Hasheesh any reason why we kept it this way?

Baraa-Hasheesh: No particular reason,
If it's possible to provide collapsed as output in the java agent args it should be properly fine to switch here

test/native/dwarfLinuxTest.cpp
visheshruparelia: Yeah that is what I wanted to do as well. However, that one is coupled with `CodeCache` class. I can try passing "null" value for that and see if it works but it felt hacky so I created a new implementation which can be used in UT.

test/test/stackwalker/StackwalkerTests.java
visheshruparelia: This is how it looks like before the change:
```
__clone;start_thread;ThreadJavaMain;JavaMain;jni_CallStaticVoidMethod;jni_invoke_static;JavaCalls::call_helper;call_stub;test/stackwalker/StackGenForUnwindViaDebugFrame.main_[0];useCpu
```

src/dwarf.cpp
visheshruparelia: You mean the `addRecord()` we do [here](https://github.com/async-profiler/async-profiler/blob/master/src/dwarf.cpp#L129) for `.eh_frame` FDE? 
I am following a similar pattern for `.debug_frame` FDE [here](https://github.com/async-profiler/async-profiler/blob/8ee2701ce73b0c84ae8ba3ee8e13c4d4916c71a2/src/dwarf.cpp#L169).

test/native/dwarfLinuxTest.cpp
krk: Why do you have to pass null? You could create a real `CodeCache` if that eliminates this `ElfReader` implementation.

test/test/stackwalker/StackwalkerTests.java
krk: In that case, isn't testing for `doRecursiveWork` enough? Why are we trying to match the entire stack? Can we simplify this and remove the logic around `OPTIONAL_FRAME`?

src/dwarf.cpp
visheshruparelia: >I understand the benefits of implementing the parser as a class, my point is that we could consider exposing it as a static function backed by a class.

I am not sure if this would make sense. This `parse()` method if converted to static, is going to use class member variables - which means we have to reset the state every time this `parse()` is called. If we are saying that we create a different instance everytime so that we don't have to reset the state, then why keep the methods static in the first place.
I would say that we either make the whole parsing static or object based. For now I am inclined towards keeping this as is as it will involve refactoring code which is unrelated to the changes being proposed in the PR. If we decide to refactor it, a fresh PR would be the way to go.

test/test/stackwalker/StackwalkerTests.java
visheshruparelia: @krk just checking for `doRecursiveWork` should be enough for the test to pass. However, don't you think asserting on the complete stack trace makes it more complete? Since we are making changes to core component which is being used during stack walking, I would feel more confident if I am able to exactly assert the trace rather than just a part of it. Let me know what you think.

src/symbols_linux.cpp
visheshruparelia: f489ea9cc03a468119d3cd8415ef4d2c8362de2f --- you were right, this change was needed for graal VM parsing to work. 

I will add screenshots related graal VM parsing in the description.

src/dwarf.cpp
visheshruparelia: 71c6164b406e565234b9d123084db2864bf2f980

test/native/dwarfLinuxTest.cpp
visheshruparelia: 1dafa121b1724b9a4eea87102fcdb944cdf2d11c

src/symbols_linux.cpp
visheshruparelia: changed it to `parseDebugFrameSection()` 
70a8d5b0d7b9fcda67f5ed45cf890234a46acb55

src/symbols_linux.cpp
visheshruparelia: 9e9258857504e1054854ede4e52abbb89e14f9db

src/codeCache.h
visheshruparelia: e99d3c7

src/symbols_linux.cpp
visheshruparelia: Inclined to keep as is, we are invoking `free()` [here](https://github.com/async-profiler/async-profiler/blob/master/src/codeCache.cpp#L64) which can corrupt the static frame desc for any further use.

fandreuz: Right, missed that 👍

test/native/dwarfLinuxTest.cpp
visheshruparelia: - The constructor for `ElfParser` is private [here](https://github.com/async-profiler/async-profiler/blob/master/src/symbols_linux.cpp#L221). For the sake of testing, do we want to make it public?
- Alt would be to invoke static method `ElfParser::parseFile()` [here](https://github.com/async-profiler/async-profiler/blob/master/src/symbols_linux.cpp#L281). But then, I cannot explicitly invoke `DwarfParser::parseDebugFrame()` as the `ElfParser::parseFile()` will call it internally. It defies the purpose of unit test as I am trying to test DwarfParser and not ElfParser.
- If we do use "real" classes, wouldn't it make the "unit" tests not so "unit"? My understanding of writing the tests in this package is that we want to test the Class in question assuming everything else works fine. If I start using ElfParser, CodeCache's real instance, a bug there could cause tests for DwarfParser to fail. Is that desirable?

src/codeCache.h
visheshruparelia: How about `_dwarf_table != NULL && _dwarf_table_length > 0;` ?
5daf351

test/native/dwarfLinuxTest.cpp
visheshruparelia: No, `findDebugFrame()` populates the `debug_start` and `debug_end` with the right address values. It also returns false if it fails.

test/native/dwarfLinuxTest.cpp
visheshruparelia: 674d0ac

visheshruparelia: 9edb1bc7676d6e80244f680b4cf952074bd1a56c

test/native/dwarfLinuxTest.cpp
visheshruparelia: 9edb1bc7676d6e80244f680b4cf952074bd1a56c

test/native/dwarfLinuxTest.cpp
visheshruparelia: Using MACROs as discussed, helper function approach does not really solve the part where we skip object creation.
611f60e43f7c46121fe032849dd11002dbb26d13

visheshruparelia:
https://github.com/oracle/graal/issues/11622

apangin:
An updated version of this PR will follow shortly. I'm closing this one.

PR #1351, opened 2025-06-25
Fix invalid alignment in mallocTracer and zero-init buf in getTotalCpuTime

krk:
Initialize `buf` in `getTotalCpuTime`, to avoid any possibility of `sscanf` reading uninitialized data.

Also pass the minimum valid alignment, which is `sizeof(void*)` to the `posix_memalign` call in `mallocTracer` initialization. glibc is lenient in this case, we should still call it with a valid alignment.

### How has this been tested?
`make test`

---

### DIFF
--- a/src/os_linux.cpp
+++ b/src/os_linux.cpp
@@ -354,7 +354,7 @@ u64 OS::getTotalCpuTime(u64* utime, u64* stime) {
     char buf[128] = {0};
     if (read(fd, buf, sizeof(buf)) >= 12) {
         u64 user, nice, system, idle;
-        if (sscanf(buf + 4, "%llu %llu %llu  %llu", &user, &nice, &system, &idle) == 4) {
+        if (sscanf(buf + 4, "%llu %llu %llu %llu", &user, &nice, &system, &idle) == 4) {
             *utime = user + nice;
             *stime = system;
             real = user + nice + system + idle;


src/mallocTracer.cpp
```diff
@@ -114,7 +114,7 @@ static void resolveMallocSymbols() {
     void* p2 = calloc(1, 1);
     void* p3 = aligned_alloc(1, 1);
     void* p4 = NULL;
-    if (posix_memalign(&p4, 1, 1) == 0) free(p4);
+    if (posix_memalign(&p4, sizeof(void*), 1) == 0) free(p4);
```
apangin: Let allocation size match alignment?

src/os_linux.cpp
```diff
@@ -351,7 +351,7 @@ u64 OS::getTotalCpuTime(u64* utime, u64* stime) {
     }
 
     u64 real = (u64)-1;
-    char buf[512];
+    char buf[512] = {0};
```
apangin: Can you shrink buffer to the minimum required size? I think 128 will be good.
It did not make difference for uninitialized buffer, but should reduce amount of work for zeroing.

src/mallocTracer.cpp
krk: Will do.

PR #1354, opened 2025-06-30
Allow VMX to work on native applications

Baraa-Hasheesh:
### Description
Allow VMX stack walking to work on applications without checking if JVM is present or not.

Previously any continuous profiling agent on a native application would have to use other stack walking modes.
Those stack walking modes once a JVM is detected will start to use AsyncGetCallTrace method which is error prone.

The VMX makes sense for such native applications (applications that start a JVM instance) as they need both native stack & java stack unwinding

### Related issues
N/A

### Motivation and context
Allow continuous profiling on native application using VMX 

### How has this been tested?
test/test/nonjava/NonjavaTests.java was changed to use VMX 
`make test` 

---


src/profiler.cpp
```diff
@@ -1170,7 +1170,7 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("DWARF unwinding is not supported on this platform");
     } else if (_cstack == CSTACK_LBR && _engine != &perf_events) {
         return Error("Branch stack is supported only with PMU events");
-    } else if (_cstack >= CSTACK_VM && !VMStructs::hasStackStructs()) {
+    } else if (_cstack == CSTACK_VM && !VMStructs::hasStackStructs()) {
```
apangin: This check should not be simply disabled. We should not let users select `vmx` mode if the target JVM does not really support it (e.g. Zing, OpenJ9), or if it is new version of HotSpot where some VM structures have been changed. Otherwise this may lead to JVM crash.

Baraa-Hasheesh: updated check here [89f10a4](https://github.com/async-profiler/async-profiler/pull/1354/commits/89f10a4bd8234e5d6a454c710bef5dc96979ca53) 

The check specifically check that libjvm is loaded & attached to the profiler now

apangin: `VM::loaded()` is the canonical way to check JVM presence.

Baraa-Hasheesh: It seems I forgot about that function, thanks for the catch

updated here [d1a07f8](https://github.com/async-profiler/async-profiler/pull/1354/commits/d1a07f85457601e3539d46b0e14e26f545768eac)

src/profiler.cpp
```diff
@@ -1336,7 +1336,7 @@ Error Profiler::check(Arguments& args) {
             return Error("DWARF unwinding is not supported on this platform");
         } else if (args._cstack == CSTACK_LBR && _engine != &perf_events) {
             return Error("Branch stack is supported only with PMU events");
-        } else if (args._cstack >= CSTACK_VM && !VMStructs::hasStackStructs()) {
+        } else if (_cstack >= CSTACK_VM && VM::loaded() && !VMStructs::hasStackStructs()) {
```
apangin: args._cstack

test/test/nonjava/NonjavaTests.java
```diff
@@ -12,45 +12,47 @@
 public class NonjavaTests {
 
     // jvm is loaded before the profiling session is started
-    @Test(sh = "%testbin/non_java_app 1 %s.html", output = true)
+    @Test(sh = "%testbin/non_java_app 1 %s.jfr", output = true)
```
apangin: Why did you change it to use jfr format?

Baraa-Hasheesh: 2 main reasons,

* Most other tests use jfr so I changed it to be more inline 
* JFR offers more debug information & is easier to debug in CLI compared to flame-graph

src/profiler.cpp
Baraa-Hasheesh: Sorry my bad in copy & paste,

handled here [4de1eb4](https://github.com/async-profiler/async-profiler/pull/1354/commits/4de1eb408811e918727a2b40e5e006140024a6b9)

test/test/nonjava/NonjavaTests.java
apangin: This does not seem correct to me.
Originally, most of tests used `collapsed` format as it is easiest to parse and human readable at the same time.
Only those tests that require access to individual events used JFR, e.g. multi-event profiling and leak profiling.

Baraa-Hasheesh: Should I revert this to html or convert it to collapsed?

IMO for this context I think collapsed makes more sense than the original html as I'm checking if the stack contains certain function calls

apangin: Change it to collapsed

Baraa-Hasheesh: changed here [a03bba9](https://github.com/async-profiler/async-profiler/pull/1354/commits/a03bba95cdeed44de48875986a61d9b86881576f)

PR #1355, opened 2025-06-30
Refactor Preload mode to use patching & better detect preload of profiler

### LINKED ISSUE #1273: recursive dlopen calls when Profiler is loaded via LD_PRELOAD

### LINKED ISSUE #1304: Multiple Async-Profilers conflict with one another

Baraa-Hasheesh:
### Description
Refactor async-profiler preload mode to use patching rather than using LD_PRELOAD hooks,

This change was introduced for multiple reasons:
1. Simplify the flow for preload by re-using the C-API defined hooks
2. LD_PRELOAD currently isn't functional on macOS & while it's possible to fix them it will require maintaining two sets of preload functions (macOS vs LINUX) due to the differences in LD_PRELOAD hooks between macOS & Linux
3. LD_PRELOAD hooks can be dangerous when accounting for async-profiler behaviour of changing GOT entry values.

This comes into the picture when accounting for multiple profilers, one in PRELOAD mode (LD_PRELOAD hooks) while the other in API mode (changing GOT entries to install hooks).

The profiler needs to find a safe reference to the original method to call, this operation isn't straight forward on macOS as due to how interposing works in it.

Using dlsym on macOS won't find a safe reference due to the fact that the dynamic linker will change all references to point to the interposing function leaving only the preloaded lib unchanged. This in turn means that the preloaded profiler needs to use the value inside GOT to call the original functions.

Initially this is straight forward as you can call the function directly, however if different profiler starts in the process, it might patch that reference resulting in potential infinite loops.

For linux, similar issues might happen when using `dlsym`, as it might up resolving a stub which might not be safe to call after patching.

It's technically possible to account for these behaviors but I recommend against it as it will result in unneeded complexity.

### Related issues
will fix #1304 for Native applications (user can now use multiple async-profilers in a C/Rust application)
Will fix #1273 as dlopen is no longer preloaded
Will also fix the issue being addressed in the #1335 PR of wrong start of profiler on dlopen

### Motivation and context
Fix preload mode on macOS & remove dangerous parts of the code by re-using the C-API approach for LD_PRELOAD mode

### How has this been tested?
many tests were enabled on macOS
New tests added to validate different aspects of this change

`make test`

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,8 @@ endif
 
 CFLAGS_EXTRA ?=
 CXXFLAGS_EXTRA ?=
-CFLAGS=-O3 -Wl,-z,now -fno-exceptions $(CFLAGS_EXTRA)
-CXXFLAGS=-O3 -Wl,-z,now -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
+CFLAGS=-O3 -fno-exceptions $(CFLAGS_EXTRA)
+CXXFLAGS=-O3 -fno-exceptions -fno-omit-frame-pointer -fvisibility=hidden -std=c++11 $(CXXFLAGS_EXTRA)
 CPPFLAGS=
 DEFS=-DPROFILER_VERSION=\"$(PROFILER_VERSION)\"
 INCLUDES=-I$(JAVA_HOME)/include -Isrc/helper
@@ -76,7 +76,7 @@ endif
 
 OS:=$(shell uname -s)
 ifeq ($(OS),Darwin)
-  CXXFLAGS += -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE -Wl,-rpath,@executable_path/../lib -Wl,-rpath,@executable_path/../lib/server
+  CXXFLAGS += -Wl,-bind_at_load -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE -Wl,-rpath,@executable_path/../lib -Wl,-rpath,@executable_path/../lib/server
   INCLUDES += -I$(JAVA_HOME)/include/darwin
   SOEXT=dylib
   PACKAGE_EXT=zip
@@ -89,7 +89,7 @@ ifeq ($(OS),Darwin)
     MERGE=false
   endif
 else
-  CXXFLAGS += -U_FORTIFY_SOURCE -Wl,-z,defs -Wl,--exclude-libs,ALL -static-libstdc++ -static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections -ggdb
+  CXXFLAGS += -Wl,-z,now -U_FORTIFY_SOURCE -Wl,-z,defs -Wl,--exclude-libs,ALL -static-libstdc++ -static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections -ggdb
   ifeq ($(MERGE),true)
     CXXFLAGS += -fwhole-program
   endif


src/hooks.cpp
```diff
@@ -91,80 +89,58 @@ static void pthread_exit_hook(void* retval) {
...
 }
 
+// Call each intercepted function at least once to ensure
+// its GOT entry is updated with a correct target address
+static void resolveThreadSymbols() {
+    static volatile intptr_t sink;
 
-// LD_PRELOAD hooks
+    pthread_t thread;
+    pthread_create(&thread, NULL, callThreadSymbols, NULL);
```
apangin: Isn't there a better way to ensure linking is fully completed without actually calling a function?
Creating a new thread and waiting for it may not be the best idea.

Doesn't `-Wl,-z,now` achieve the desired effect?

src/preload.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+extern "C" DLLEXPORT
+int mac_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact) {
+    return sigaction(signum, act, oldact);
+}
+
+__attribute__((used)) static struct {
+    const void *replacement;
+    const void *original;
+} interposers[] __attribute__((section("__DATA,__interpose"))) = {
```
apangin: Well... it's a hack )
I'd rather prefer to leave current behavior than to maintain such tricks. Can we detect preloaded library somehow differently? Maybe checking order of loaded libraries? I know we discussed that, but I still hope for a cleaner solution. I'm inclined to accept the fact that checking LD_PRELOAD/DYLD_INSERT_LIBRARIES environment variable is a good enough approach.

src/hooks.cpp
Baraa-Hasheesh: `pthread_create` & `dlopen` are available in GOT but we're missing the `pthread_exit` I will change the logic in other locations in async profiler to safely add `pthread_exit` value.

Regarding `-Wl,-z,now`, I don't think it's needed, but I will need to read more/do some tests on this

src/preload.cpp
Baraa-Hasheesh: well, I see what you mean.
I will push the  order of loaded libraries way to detect if profiler is preloaded,
I think it will be hard to judge this without seeing the code.

src/hooks.cpp
Baraa-Hasheesh: function removed & replaced by guarantees in other locations

`-Wl,-z,now` is needed & achieves the same outcome as calling the function from what I observed

[eee3d59](https://github.com/async-profiler/async-profiler/pull/1355/commits/eee3d596157f122081dcff36eeea3f0252955a0a)

[0e08295](https://github.com/async-profiler/async-profiler/pull/1355/commits/0e08295d99b4eeb93b47963719b39c0e191dc790)

src/preload.cpp
Baraa-Hasheesh: Updated here [eee3d59](https://github.com/async-profiler/async-profiler/pull/1355/commits/eee3d596157f122081dcff36eeea3f0252955a0a)

Baraa-Hasheesh:
as requested, I will close this PR & separate this change into two different PRs
* new preload check 
* preload refactor & fixes

PR #1357, opened 2025-07-01
Correctly unwind the stack for malloc events in VM stack walking mode

Baraa-Hasheesh:
### Description
VM stack walker doesn't work on Malloc events which causes frames to not be unwound correctly 

```
profiler.Malloc {
  startTime = 14:45:34.088
  address = 0x7F5F73F6A810
  size = 27 bytes
  eventThread = "DestroyJavaVM" (javaThreadId = 24)
  stackTrace = [
    .no_Java_frame() line: 0
  ]
}
```

### Related issues
N/A

### Motivation and context
Fix VM stack walker on malloc to avoid bad samples

### How has this been tested?
added a native memory test with VM stack walking => was failing before change & now passes
`make test` 

---


src/profiler.cpp
```diff
@@ -669,7 +669,11 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
             num_frames += getJavaTraceAsync(ucontext, frames + num_frames, _max_stack_depth, &java_ctx);
         }
     } else if (event_type == MALLOC_SAMPLE) {
-        num_frames += getJavaTraceAsync(ucontext, frames + num_frames, _max_stack_depth, &java_ctx);
+        if (_cstack == CSTACK_VM) {
+            num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_NORMAL);
+        } else {
+            num_frames += getJavaTraceAsync(ucontext, frames + num_frames, _max_stack_depth, &java_ctx);
+        }
```
apangin: Unwinding logic here is similar to unwinding stacktraces of regular CPU samples, so can we merge it with the above branch (event_type <= WALL_CLOCK_SAMPLE)?

Baraa-Hasheesh: That's possible but should I keep the inner if check for CPU samples only?

java_frames > 0 && java_ctx.pc != NULL && VMStructs::hasMethodStructs()

Baraa-Hasheesh: as discussed offline updated here [48268c4](https://github.com/async-profiler/async-profiler/pull/1357/commits/48268c4aa2665db0be9d31cd270fab99e32effe5)

Issue #1358, opened 2025-07-01
Do not dereference jmethodIDs on JDK 26


apangin:
### Describe the bug

[JDK-8268406](https://bugs.openjdk.org/browse/JDK-8268406) "Deallocate jmethodID native memory" changed the way how jmethodIDs are represented, they are no longer pointers to `Method*`.

Async-profiler performs some tricks to workaround older JVM bugs where method-related JVM TI functions could lead to JVM crashes. These tricks are not needed with JDK 26 and therefore the related code should be removed.

### Expected vs. actual behavior

Expected: No regression with JDK 26.
Actual: JVM crash during resolution of jmethodIDs.

### Async-profiler version

4.0

apangin:
Fix has been merged.

PR #1362, opened 2025-07-03
#1358: Do not dereference jmethodIDs on JDK 26

apangin:
### Description

Fix crash when dumping profiling results on mainline JDK. The problem was caused by changes in jmethodIDs representation made in [JDK-8268406](https://bugs.openjdk.org/browse/JDK-8268406).

### Related issues

#1358 

### How has this been tested?

Class load-unload stress test

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -173,7 +173,7 @@ class Lookup {
     }
 
     bool fillJavaMethodInfo(MethodInfo* mi, jmethodID method, bool first_time) {
-        if (VMMethod::isStaleMethodID(method)) {
+        if (VMMethod::isStaleMethodId(method)) {
             return false;
         }
 
--- a/src/vmStructs.h
+++ b/src/vmStructs.h
@@ -388,12 +388,11 @@ class VMMethod : VMStructs {
   public:
     jmethodID id();
 
-    bool validate(jmethodID id) {
-        return !_can_dereference_jmethod_id || *(VMMethod**)id == this;
-    }
+    // Performs extra validation when VMMethod comes from incomplete frame
+    jmethodID validatedId();
 
     // Workaround for JDK-8313816
-    static bool isStaleMethodID(jmethodID id) {
+    static bool isStaleMethodId(jmethodID id) {
         if (!_can_dereference_jmethod_id) return false;
         VMMethod* vm_method = *(VMMethod**)id;
         return vm_method == NULL || vm_method->id() == NULL;
--- a/src/frameName.cpp
+++ b/src/frameName.cpp
@@ -155,7 +155,7 @@ const char* FrameName::typeSuffix(FrameTypeId type) {
 }
 
 void FrameName::javaMethodName(jmethodID method) {
-    if (VMMethod::isStaleMethodID(method)) {
+    if (VMMethod::isStaleMethodId(method)) {
         _str.assign("[stale_jmethodID]");
         return;
     }
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -53,10 +53,7 @@ static inline void fillFrame(ASGCT_CallFrame& frame, FrameTypeId type, int bci, 
 
 static jmethodID getMethodId(VMMethod* method) {
     if (!inDeadZone(method) && aligned((uintptr_t)method)) {
-        jmethodID method_id = method->id();
-        if (!inDeadZone(method_id) && aligned((uintptr_t)method_id) && method->validate(method_id)) {
-            return method_id;
-        }
+        return method->validatedId();
     }
     return NULL;
 }
--- a/src/vmStructs.cpp
+++ b/src/vmStructs.cpp
@@ -669,6 +669,16 @@ jmethodID VMMethod::id() {
     return NULL;
 }
 
+jmethodID VMMethod::validatedId() {
+    jmethodID method_id = id();
+    if (goodPtr(method_id)) {
+        if (!_can_dereference_jmethod_id || *(VMMethod**)method_id == this) {
+            return method_id;
+        }
+    }
+    return NULL;
+}
+
 NMethod* CodeHeap::findNMethod(char* heap, const void* pc) {
     unsigned char* heap_start = *(unsigned char**)(heap + _code_heap_memory_offset + _vs_low_offset);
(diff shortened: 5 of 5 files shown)


src/vmStructs.h
```diff
@@ -385,11 +386,18 @@ class VMThread : VMStructs {
 
 class VMMethod : VMStructs {
   public:
-    static VMMethod* fromMethodID(jmethodID id) {
-        return *(VMMethod**)id;
+    jmethodID id();
+
+    bool validate(jmethodID id) {
+        return !_can_dereference_jmethod_id || *(VMMethod**)id  == this;
```
fandreuz:
```suggestion
        return !_can_dereference_jmethod_id || *(VMMethod**)id == this;
```

src/vmStructs.h
```diff
@@ -385,11 +386,18 @@ class VMThread : VMStructs {
 
 class VMMethod : VMStructs {
   public:
-    static VMMethod* fromMethodID(jmethodID id) {
-        return *(VMMethod**)id;
+    jmethodID id();
+
+    bool validate(jmethodID id) {
```
fandreuz: As far as I understand, `method.validate(my_id) == false` does not imply that `my_id` is invalid, but simply that it does not point to this particular `method`. Is `validate` named properly?

Also, it seems that the only usage is something like this:
```cpp
jmethodID method_id = method->id();
if (... && method->validate(method_id)) {
```
so do we even need to take a parameter in `validate`? It could be something like this:
```cpp
bool validate() {
   return !_can_dereference_jmethod_id || *(VMMethod**) id()  == this;
}
```

Am I missing something?

src/flightRecorder.cpp
```diff
@@ -173,31 +173,24 @@ class Lookup {
...
+            (err = jvmti->GetMethodDeclaringClass(method, &method_class)) == 0 &&
+            (err = jvmti->GetClassSignature(method_class, &class_name, NULL)) == 0) {
             mi->_class = _classes->lookup(class_name + 1, strlen(class_name) - 2);
             mi->_name = _symbols.lookup(method_name);
             mi->_sig = _symbols.lookup(method_sig);
-        } else {
-            mi->_class = _classes->lookup("");
-            mi->_name = _symbols.lookup("jvmtiError");
-            mi->_sig = _symbols.lookup("()L;");
         }
```
fandreuz, line 194: Why isn't this needed anymore?

src/vmStructs.h
```diff
@@ -385,11 +386,18 @@ class VMThread : VMStructs {
...
+        return !_can_dereference_jmethod_id || *(VMMethod**)id  == this;
     }
 
-    jmethodID id();
+    // Workaround for JDK-8313816
+    static bool isStaleMethodID(jmethodID id) {
+        if (!_can_dereference_jmethod_id) return false;
```
Baraa-Hasheesh, line 396: As we can no longer deference jmethodID, this should increase the number of possible stale frames in the stack (Nothing we can do about that sadly)

Perhaps we should we consider aggregating the b2b stale frames into a single one to reduce the size of the flamegraph?

For me 3 b2b stale frames offer the same level of information as a single error frame

src/vmStructs.h
Baraa-Hasheesh: If I'm not wrong this is due to the `getMethodId` which might return NULL or the method_id

If i understand the logic correctly, the current logic is doing an implicit `!= NULL` check

`getMethodId` isn't available in the current scope

src/vmStructs.h
fandreuz: What is `b2b`?

Baraa-Hasheesh: b2b is a short hand for "back to back"

What I mean is if we have a stack like 

`valid-stale-stale-stale-valid-stale `

We shorten it to 

`valid-stale-valid-stale `

Note this isn't a blocker for this change as it's an issue that could happen now, I'm just raising it as a possible improvement as the frequency should increase

fandreuz: Ok, so you mean "adjacent stale frames"

src/vmStructs.h
apangin: `id()` is not a very cheap operation, I'd rather not call it twice.
But I took your point, refactored this logic a bit.

src/flightRecorder.cpp
apangin: This branch was never actually taken; problematic cases were previously caught by the above `isStale` check. This is no longer true after JDK 26 fix. Methods of unloaded classes started showing up as `jvmtiError`, which is not very nice. So I removed this branch entirely and now return `false`, so that the caller can handle this properly and display `stale_jmethodID` as before.

src/vmStructs.h
apangin: No, the amount of stale methods has not changed. This was a safeguard from the JVM bug [JDK-8313816](https://bugs.openjdk.org/browse/JDK-8313816), which is irrelevant for newer JDKs.

apangin: Anyhow, collapsing stale frames is probably not a good idea, because stack depth really matters.
Here is a self-descriptive example:
method-ids

src/vmStructs.h
```diff
@@ -385,12 +386,18 @@ class VMThread : VMStructs {
...
-    }
-
     jmethodID id();
 
+    // Performs extra validation when VMMethod comes from incomplete frame
+    jmethodID validatedId();
+
+    // Workaround for JDK-8313816
+    static bool isStaleMethodId(jmethodID id) {
+        if (!_can_dereference_jmethod_id) return false;
```
krk, line 396: Couldn't `id` still be stale even if it can not be dereferenced, if `Method::validate_jmethod_id(jmethodID mid)` returns false for the `id`.

apangin: Yes, it can, but that's fine. Subsequent JVMTI call (`GetMethodName` or other) will return `JVMTI_ERROR_INVALID_METHODID` which async-profiler catches and handles as the same `stale_jmethodID`. `isStaleMethodId` does not change logic of symbol resolution - it's only a [workaround](https://bugs.openjdk.org/browse/JDK-8313816) for older JVM versions to avoid calling JVMTI functions on knowingly bad jmethodIDs.

PR #1365, opened 2025-07-07
Ensure that only files under `src/` are checked by the linter in `cpp-lint-diff`

fandreuz:
### Description
I noticed that the current configuration does not filter out files out of `src/`. Example [here](https://github.com/fandreuz/async-profiler/actions/runs/16101824422/job/45431945162#step:6:109).

### Related issues
n/a

### Motivation and context
As discussed, static code analysis should not run on test code for now.

### How has this been tested?
Manual testing.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,7 @@ cpp-lint:
 
 DIFF_BASE=
 cpp-lint-diff:
-	git diff -U0 $(DIFF_BASE) -- '**/*.cpp' '**/*.h' ':!**/rustDemangle.cpp' | \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -282,7 +282,7 @@ cpp-lint:
 
 DIFF_BASE=
 cpp-lint-diff:
-	git diff -U0 $(DIFF_BASE) -- '**/*.cpp' '**/*.h' ':!**/rustDemangle.cpp' | \
+	git diff -U0 $(DIFF_BASE) -- 'src/*.cpp' 'src/**/*.cpp' 'src/*.h' 'src/**/*.h' ':!**/rustDemangle.cpp' | \
```
krk, line 285: There are c/cpp files under `test/native` that we should check.

PR #1366, opened 2025-07-07
Change jfrconv binary generation from cpp to shell.

### LINKED ISSUE #1344: Sign `jfrconv` executable on macOS
### Describe the feature

`jfrconv` is a self-executable JAR created by concatenating binary Java launcher with `jfr-converter.jar`. This makes `jfrconv` binary a valid executable and a valid JAR at the same time. The launcher looks for `java` executable at known locations and then invokes it passing itself (jfrconv) as a JAR to execute.

This works well, except that such a binary cannot be properly code-signed on macOS. Consequently, notarization of async-profiler package fails.

### Use Case

When async-profiler is downloaded with Safari, launching `jfrconv` will result in macOS showing a warning about unidentified developer.

### Proposed Solution

I'm thinking of two possible solutions:
1. Rewrite launcher (`src/launcher/launcher.cpp`) in shell script. Shell scripts don't require code signing, and they can also be concatenated with `.jar`. Need to double-check how different macOS versions handle such executables.
2. Embed `.jar` in data section of a program. It won't be possible to use the result binary as `.jar` directly, so a launcher will need to start the JVM by itself and provide a custom class loader for loading an embedded `.jar`.

### Acknowledgements

- [ ] I may be able to implement this feature request

roy-soumadipta:
### Description
The commit fixes #1344 and includes the following changes:

* Rewrite jfrconv launcher script in shell from cpp.
* Changes to Makefile to accomodate the shift to shell script.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1344

### How has this been tested?
jfrconv has been tested to show no Move to trash popup by using the binaries from the notarized zip.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,6 @@ build/$(JFRCONV): src/launcher/launcher.sh build/$(CONVERTER_JAR)
 	chmod +x $@
 	cat build/$(CONVERTER_JAR) >> $@
 
-
 build/$(LIB_PROFILER): $(SOURCES) $(HEADERS) $(RESOURCES) $(JAVA_HELPER_CLASSES)
 ifeq ($(MERGE),true)
 	for f in src/*.cpp; do echo '#include "'$$f'"'; done |\


src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
+# Copyright The async-profiler authors
+# SPDX-License-Identifier: Apache-2.0
+
+#!/bin/bash
+
+set -euo pipefail
+
+VERSION_STRING="JFR converter PROFILER_VERSION built on $(date '+%b %d %Y')"
```
apangin: It is today's date, not the build date, right?

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+if [[ "$1" == "-v" || "$1" == "--version" ]]; then
+    echo "$VERSION_STRING"
+    exit 0
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    SCRIPT_PATH="$(cd "$(dirname "$0")" && pwd)/$(basename "$0")"
+else
+    SCRIPT_PATH="$(readlink -f "$0")"
+fi
```
apangin: Is canonicalization necessary? `java -jar` accepts relative paths too.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+
+JAVA_CMD=("java" "-Xss2M" "-Dsun.misc.URLClassPath.disableJarChecking")
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -D*|-X*|-agent*)
+            if [[ ${#1} -gt 2 ]]; then
```
apangin: ~Maybe -gt 1?~

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+    exit 0
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    SCRIPT_PATH="$(cd "$(dirname "$0")" && pwd)/$(basename "$0")"
+else
+    SCRIPT_PATH="$(readlink -f "$0")"
+fi
+
+JAVA_CMD=("java" "-Xss2M" "-Dsun.misc.URLClassPath.disableJarChecking")
```
apangin: Why put `java` in the beginning, if it's always cut later?

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
+# Copyright The async-profiler authors
+# SPDX-License-Identifier: Apache-2.0
+
+#!/bin/bash
```
apangin: `bash` is not available on all systems. Script must use POSIX compatible `/bin/sh`.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+
+find_java() {
+    local java_path="$1"
+    [[ -x "$java_path" ]]
+}
+
+run_java() {
+    # 1. Try JAVA_HOME
+    if [[ -n "$JAVA_HOME" ]] && find_java "$JAVA_HOME/bin/java"; then
+        exec "$JAVA_HOME/bin/java" "${JAVA_CMD[@]:1}"
```
apangin: Script should continue trying next options if one of `exec` calls fails.
This is how the original C++ launcher worked.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+                fi
+            fi
+        done
+    fi
+    
+    echo "No JDK found. Set JAVA_HOME or ensure java executable is on the PATH" >&2
+    exit 1
+}
+
+run_java
```
apangin: Newline here is required not only because of the coding style, but also because it won't work otherwise after concatenation with .jar.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+        *)
+            break
+            ;;
+    esac
+    shift
+done
+
+JAVA_CMD+=("-jar" "$SCRIPT_PATH" "$@")
+
+find_java() {
```
apangin: Not very meaningful function - there is just one short expression that can be as well inlined.

src/launcher/launcher.sh
roy-soumadipta: this is such a great suggestion. Thanks can be simplified so much

src/launcher/launcher.sh
roy-soumadipta: yes changed here and other affected places

src/launcher/launcher.sh
roy-soumadipta: Yes replaced at build time like a macro. Will reflect in next commit

src/launcher/launcher.sh
roy-soumadipta: In the cpp launcher we had && (*argv)[2] check as well which check if something after -D or -X is not a null terminator. So -gt 2 mimics that right?

apangin: Sorry, I misread the expression. Yes, your translation is correct.

Makefile
```diff
@@ -295,4 +293,4 @@ clean-coverage:
 	$(RM) -rf build/test/cpptests build/test/coverage
 
 clean:
-	$(RM) -r build
+	$(RM) -r build
```
Baraa-Hasheesh: You have styling issue here & in other files 
The file should always end with a single empty new line

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
+# Copyright The async-profiler authors
+# SPDX-License-Identifier: Apache-2.0
+
+#!/bin/bash
+
+set -euo pipefail
```
Baraa-Hasheesh: Do we need this on user script?
This might end up causing weird failures on user side if we're not careful

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+
+VERSION_STRING="JFR converter PROFILER_VERSION built on $(date '+%b %d %Y')"
+
+if [[ "$1" == "-v" || "$1" == "--version" ]]; then
+    echo "$VERSION_STRING"
+    exit 0
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    SCRIPT_PATH="$(cd "$(dirname "$0")" && pwd)/$(basename "$0")"
```
Baraa-Hasheesh: `pwd)`?

wouldn't `realpath $0` solve for both?

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+}
+
+run_java() {
+    # 1. Try JAVA_HOME
+    if [[ -n "$JAVA_HOME" ]] && find_java "$JAVA_HOME/bin/java"; then
+        exec "$JAVA_HOME/bin/java" "${JAVA_CMD[@]:1}"
+    fi
+    
+    # 2. Try PATH
+    if command -v java >/dev/null 2>&1; then
```
Baraa-Hasheesh: Do we really need all this logic for finding `java`
can't we use PATH always?

if missing we can print a message telling the user to add it?
If the problem is with our jar being compatible with a specific minimum Java version, we can add a check for that as well 

This logic might be hiding bugs as if the user has 10 JDKs it could be failing for the first 9 while passing for the last one

@apangin what do you think?

src/launcher/launcher.sh
```diff
@@ -0,0 +1,87 @@
...
+set -euo pipefail
+
+VERSION_STRING="JFR converter PROFILER_VERSION built on $(date '+%b %d %Y')"
+
+if [[ "$1" == "-v" || "$1" == "--version" ]]; then
+    echo "$VERSION_STRING"
+    exit 0
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
Baraa-Hasheesh: why not use `uname -s` as we do in Makefile?

src/launcher/launcher.sh
roy-soumadipta: this is removed after Andrei's suggestion

src/launcher/launcher.sh
roy-soumadipta: We have an env variable ready to use. Why go through uname -s?

src/launcher/launcher.sh
apangin: I think the script is doing what it's intended to do - find any working Java Runtime, even if `java` is not on the `PATH`.

src/launcher/launcher.sh
roy-soumadipta: Isn't this better than silent failures?

Baraa-Hasheesh: Some failures that happen due to `set -euo pipefail` are due to reasons that would work fine without the flags 

Such as unbound variables 

I'm fine with keeping this but we should test all possible branches of the script 

If all possible branches were not tested then this isn't really safe

src/launcher/launcher.sh
Baraa-Hasheesh: For consistency

Makefile
roy-soumadipta: Addressed in latest commit

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: Which ones do you suggest removing?

src/launcher/launcher.sh
roy-soumadipta: Addressed in commit [444fb5d](https://github.com/async-profiler/async-profiler/pull/1366/commits/444fb5d0eb9f870bcb757f24a83606fba599d0ad)

src/launcher/launcher.sh
roy-soumadipta: I don't get how are we deviating from consistency by using a ready to use variable?

src/launcher/launcher.sh
Baraa-Hasheesh: All of them raise some level of risk 

The question shouldn't be "which to remove"

The proper question should be were all possible cases tested with these flags?

If we have enough testing then keeping this should be fine


For example the script should continue even if a previous exec fails right?

However if an exec fails wouldn't it cause an exit code of 1?

Wouldn't this break the script?

src/launcher/launcher.sh
Baraa-Hasheesh: Different approach than what we have in the Makefile

src/launcher/launcher.sh
```diff
@@ -0,0 +1,78 @@
...
+    esac
+    shift
+done
+
+JAVA_CMD+=("-jar" "$SCRIPT_PATH" "$@")
+
+run_java() {
+    # 1. Try JAVA_HOME
+    if [[ -n "$JAVA_HOME" ]] && [[ -x "$JAVA_HOME/bin/java" ]]; then
+        "$JAVA_HOME/bin/java" "${JAVA_CMD[@]}" && exit 0
```
krk: Why not `exec`?

roy-soumadipta: For fallback as per Andrei's comment on the previous commit. With prev commit when exec command itself failed it would proceed to try the next for example if java is not present in the path corresponding to the exec command. But now if the exec succeeds but then encounters an error, it won't try the subsequent alternatives. @apangin is this the change you wanted or the previous impl using exec suffices your expectation?

roy-soumadipta: @apangin I checked again and in the previous launcher cpp, there was no fallback if exec succeeds but then error encountered. So the previous impl in shell script sufficed the need as per the existing one. Should I revert again or do you want the changes now for a better fallback?

apangin: Script should proceed with the next attempt if `exec` itself fails, not the target command. There should be no fallback when `java` process is launched but fails for whatever reason. Basically, the script should retain the same behavior as in the original launcher.

roy-soumadipta: Ok I will change it accordingly. The previous one however wouldn't have proceeded due to set -e. There is a comment from @Baraa-Hasheesh regarding the `set -euo pipefail`. @apangin if you can share your thoughts so that I can make this change accordingly

src/launcher/launcher.sh
```diff
@@ -0,0 +1,78 @@
+# Copyright The async-profiler authors
+# SPDX-License-Identifier: Apache-2.0
+
+#!/bin/sh
```
apangin, line 1: Shebang must stay on the first line. License header shall come after.

src/launcher/launcher.sh
apangin: Let's not delve into philosophy. A simple answer is - leave options that are necessary and delete those that aren't.
To me, `-e` does not seem useful here: there are basically no commands that can fail except for `exec`, but I explicitly noted that the script should proceed if `exec` fails.
The same for `pipefail`: I don't see why it is here, given there are no pipes in the script.
While `-u` is generally useful, again, it has no behavior impact in this particular script.
Please correct me if I'm wrong, otherwise just remove the entire line. The less code, the better.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,78 @@
...
+
+set -euo pipefail
+
+VERSION_STRING="JFR converter PROFILER_VERSION built on BUILD_DATE"
```
apangin: There is no much value in introducing variables that are used only once in a local context.
Just inline the string where needed.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,78 @@
...
+
+if [[ "$1" == "-v" || "$1" == "--version" ]]; then
+    echo "$VERSION_STRING"
+    exit 0
+fi
+
+SCRIPT_PATH="$0"
```
apangin: Same as above.
Good code is like a poem: every word counts.

src/launcher/launcher.sh
```diff
@@ -0,0 +1,78 @@
...
+        *)
+            break
+            ;;
+    esac
+    shift
+done
+
+JAVA_CMD+=("-jar" "$SCRIPT_PATH" "$@")
+
+run_java() {
```
apangin: Is there a reason to put it in a function, if the following code is executed as is anyway?
E.g., the above argument parsing logic is not in a function, and that's fine.

roy-soumadipta: Since this is not a big script it can be removed

roy-soumadipta: Addressed in [385001b](https://github.com/async-profiler/async-profiler/pull/1366/commits/385001b1b33483353f6cf0e3907a904f468d2d72)

src/launcher/launcher.sh
roy-soumadipta: Addressed in [385001b](https://github.com/async-profiler/async-profiler/pull/1366/commits/385001b1b33483353f6cf0e3907a904f468d2d72)

src/launcher/launcher.sh
roy-soumadipta: Addressed in [385001b](https://github.com/async-profiler/async-profiler/pull/1366/commits/385001b1b33483353f6cf0e3907a904f468d2d72)

src/launcher/launcher.sh
roy-soumadipta: Addressed in [385001b](https://github.com/async-profiler/async-profiler/pull/1366/commits/385001b1b33483353f6cf0e3907a904f468d2d72)

src/launcher/launcher.sh
roy-soumadipta: Addressed in [385001b](https://github.com/async-profiler/async-profiler/pull/1366/commits/385001b1b33483353f6cf0e3907a904f468d2d72)

src/launcher/launcher.sh
roy-soumadipta: Addressed in [aeb211f](https://github.com/async-profiler/async-profiler/pull/1366/commits/aeb211ffd2db1e3e2b1746cb692cf7fc2df99438) to maintain uniformity and posix compliance

test.html
apangin, line 1: Added by mistake

src/launcher/launcher.sh
```diff
@@ -0,0 +1,68 @@
...
+            break
+            ;;
+    esac
+    shift
+done
+
+JAVA_CMD+=("-jar" "$0" "$@")
+
+# 1. Try JAVA_HOME
+if [[ -n "$JAVA_HOME" ]] && [[ -x "$JAVA_HOME/bin/java" ]]; then
```
apangin: AFAIK, `[[` is also a bash extension. Strictly POSIX-compliant shell only supports `[`.

roy-soumadipta: Had to make other changes as well for full compliance. Addressed in [52af9cc](https://github.com/async-profiler/async-profiler/pull/1366/commits/52af9cc2ac0d519159ef81cd07afa0bc0095f3d7)

src/launcher/launcher.sh
```diff
@@ -0,0 +1,72 @@
...
+
+while [ $# -gt 0 ]; do
+    case "$1" in
+        -D*|-X*)
+            if [ ${#1} -gt 2 ]; then
+                JAVA_OPTS="$JAVA_OPTS $1"
+            fi
+            ;;
+        -agent*)
+            if [ "$(echo "$1" | cut -c1-6)" = "-agent" ]; then
```
apangin: What does this expression verify? Don't we get in this `case` only if the argument starts with `-agent`?

roy-soumadipta: Yes this is redundant. Removed in [6f07677](https://github.com/async-profiler/async-profiler/pull/1366/commits/6f07677bbeb3bd678b5e9e3a71fa3c280a465d2c)

apangin:
Merged, thanks!

Issue #1369, opened 2025-07-08
Incomplete usage of VM & VMX & missing frame skipping for VM/VMX


Baraa-Hasheesh:
### Describe the bug

While doing some sanity checks on **VM/VMX** stack walking modes, I noticed that some features implemented for **DWARF** & **FP** are missing

These features are related to frame skipping found [here](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L336)

This results in some fixes/enhancements that were previously done in the profiler to missing such as this https://github.com/async-profiler/async-profiler/issues/993

**VMX not skipping the frames**
<img width="2552" height="253" alt="Image" src="https://github.com/user-attachments/assets/c5f205ba-29dd-45d9-928a-f7988e1beae9" />

**FP skipping the frames**
<img width="2552" height="253" alt="Image" src="https://github.com/user-attachments/assets/898600c3-7250-48b7-9255-09df63ca35a2" />

Furthermore various unintended bugs are being caused for certain types of event profiling, these bugs are related to unwinding of native frames due to the check [here](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L325) which skips native frame unwinding for VM & VMX

This result in incomplete stack frames for example the following flame graphs 

**VM stack walker for alloc samples**
<img width="1407" height="524" alt="Image" src="https://github.com/user-attachments/assets/d8076499-d8b6-497c-a8a5-071d4f32a0e2" />

**Dwarf stack walker for alloc samples**
<img width="936" height="415" alt="Image" src="https://github.com/user-attachments/assets/e0a1861e-4818-4d31-86c3-144560b3f1c3" />

Furthermore this, I did find the current [logic](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L648) for deciding which stack walker is to be used to unwind the stack is kinda convoluted 

While not strictly problematic, this logic in many cases ends up using the getJavaTraceJvmti which might not provide the same information level as VM/VMX which might hide some performance issues.

**VM Stack Walker showing a call stub**
<img width="2539" height="483" alt="Image" src="https://github.com/user-attachments/assets/834e49e4-469d-4eca-8db5-db3b53a954ff" />

**Dwarf not showing that call stub**
<img width="2538" height="129" alt="Image" src="https://github.com/user-attachments/assets/2eebf54c-3d29-4c1c-a759-c04b1bb0d96e" />

### Expected vs. actual behavior

[...]

PR #1370, opened 2025-07-08
Fully use VM stack walker when selected & general enhancements for VM/VMX modes

Baraa-Hasheesh:
### Description
In this PR, I move the stack unwinding to fully use VM stack walking mode when selected 
Previously it would have different handling depending on which event was being profiled

Furthermore this PR includes multiple enhancements on the VM/VMX stackwalking modes which include
* Skipping some async-profiler frames from the stack as those frames don't provide useful debugging information 
* Skipping some internal JVM methods that create noise without helping much in the profiling 
* augmenting the native stacks with `comptask` frames

**VM/VMX samples comparison (Before vs After)**
* LOCK
image
image

* ALLOC
image
image

* Malloc
image
image

* Comp Task (is Missing in current VM/VMX)
image
image

### Related issues
#1369

### Motivation and context
Fully expose the VM/VMX stack walkers when selected by the user

### How has this been tested?
multiple new tests have been added, these tests fails without the changes & pass with them
`make test`

---

### DIFF
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -265,7 +265,6 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
     unwind_loop:
     while (depth < max_depth) {
         if (CodeHeap::contains(pc)) {
-
             NMethod* nm = CodeHeap::findNMethod(pc);
             if (nm == NULL) {
                 if (anchor == NULL) {
--- a/test/test/comptask/ComptaskTests.java
+++ b/test/test/comptask/ComptaskTests.java
@@ -21,14 +21,18 @@ public class ComptaskTests {
         agentArgs = "start,features=comptask,collapsed,cstack=vm,interval=1ms,file=%profile",
         jvmArgs = "-Xcomp",
         jvm = Jvm.HOTSPOT,
-        nameSuffix = "VM"
+        nameSuffix = "VM",
+        // TODO: remove version filter after the fix for JDK-8367689 is released
+        jvmVer = {8, 24}
     )
     @Test(
         mainClass = Main.class,
         agentArgs = "start,features=comptask,collapsed,cstack=vmx,interval=1ms,file=%profile",
         jvmArgs = "-Xcomp",
         jvm = Jvm.HOTSPOT,
-        nameSuffix = "VMX"
+        nameSuffix = "VMX",
+        // TODO: remove version filter after the fix for JDK-8367689 is released
+        jvmVer = {8, 24}
     )
     public void testCompTask(TestProcess p) throws Exception {
         Output out = p.waitForExit("%profile");


src/lockTracer.cpp
```diff
@@ -137,6 +137,15 @@ Error LockTracer::initialize(jvmtiEnv* jvmti, JNIEnv* env) {
         return Error("setEntry method not found");
     }
 
+    CodeCache* lib = Profiler::instance()->findLibraryByAddress((void*)LockTracer::initialize);
+    lib->mark(
+        [](const char* s, const void* start, const void* end) -> bool {
+            return (start <= (const void*)LockTracer::UnsafeParkHook && (const void*)LockTracer::UnsafeParkHook <= end)
```
apangin: This does not make sense to me.
`<=` is likely incorrect, since `end` address may already belong to another function.
But why this range comparison is needed? How can a function address be in the middle of a code blob?
In any case, linear search is not suitable here; addresses in CodeCache are already sorted.

src/profiler.cpp
```diff
@@ -646,24 +646,22 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     }
 
     if (_cstack == CSTACK_VMX) {
-        num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_EXPERT);
+        num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_EXPERT, event_type);
+    } else if (_cstack == CSTACK_VM) {
+        num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_NORMAL, event_type);
```
apangin, line 628: How does this work with Java method samples? (INSTRUMENTED_METHOD)

src/profiler.h
```diff
@@ -254,6 +254,10 @@ class Profiler {
         instance()->onGarbageCollectionFinish();
     }
 
+    const StackWalkFeatures* features() const {
```
apangin: `StackWalkFeatures` is a register-width type, it can be passed by value, not need for a pointer.

src/codeCache.h
```diff
@@ -181,7 +181,7 @@ class CodeCache {
     inline void mark(NamePredicate predicate, char value) {
         for (int i = 0; i < _count; i++) {
             const char* blob_name = _blobs[i]._name;
-            if (blob_name != NULL && predicate(blob_name)) {
+            if (blob_name != NULL && predicate(blob_name, _blobs[i]._start, _blobs[i]._end)) {
```
roy-soumadipta: This change is only getting used in lockTracer but because of the predicate use across the code base there are changes across multiple files which has no use at the moment. Wouldn't it be better to either have a template with 3 params and have conditions to delegate to appropriate predicates? Or have 2 generic templates and 2 methods?

src/stackWalker.cpp
```diff
@@ -379,7 +382,32 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+            if (!java_pc_observed && mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
+                // Skip any frames above profiler hook methods
+                depth = 0;
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (!java_pc_observed && mark == MARK_ASYNC_PROFILER && (event_type == LOCK_SAMPLE || event_type == PARK_SAMPLE)){
+                // Skip internal frames for LOCK profiling
+                depth = 0;
+            }else if (!java_pc_observed && mark == MARK_COMPILER_ENTRY && Profiler::instance()->features()->comp_task) {
+                // Insert current compile task as a pseudo Java frame
+                jmethodID compile_task = Profiler::instance()->getCurrentCompileTask();
```
apangin: There is nothing in `Profiler` class that `getCurrentCompileTask` needs. To reuse this code, it's better to move it to vmStructs.cpp.

src/stackWalker.cpp
```diff
@@ -379,7 +382,32 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                 }
             }
         } else {
-            fillFrame(frames[depth++], BCI_NATIVE_FRAME, profiler->findNativeMethod(pc));
+            const char* method_name = profiler->findNativeMethod(pc);
+            char mark = method_name != NULL ? NativeFunc::mark(method_name) : 0;
+
+            if (!java_pc_observed && mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
```
apangin: `!java_pc_observed` is repeated in each `if` - it makes sense to hoist this check.

test/test/lock/SimpleLockTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static synchronized void syncLock() throws InterruptedException {
+        Thread.sleep(100);
+    }
+
+    private static void semLock() throws InterruptedException {
+        semaphore.acquire();
+        Thread.sleep(100);
+        semaphore.release();
+    }
```
apangin, line 34: This test is not reliable. It can easily happen that the first thread completes the operation before the second even enters the method., so there will be no contention.

src/codeCache.h
Baraa-Hasheesh: IMO that would create unneeded confusion for future development as the user will have to pick & choose one of the Templates/Methods 

I would rather keep a single template as using multiple generics/templates at once is convoluted 

Regarding creating a different method
In the future if we ever encounter a case where all 3 inputs are needed, should we create a third method?
=> in LockTracer we only use end & start

src/lockTracer.cpp
Baraa-Hasheesh: I wanted to avoid large refactoring here

LockTracer::UnsafeParkHook, LockTracer::recordContendedLock, LockTracer::MonitorContendedEntered names get mangled by the compiler

Checking the name here might be problematic, so we will need to refactor these methods as external "C" methods to avoid mangling & I wanted to avoid the need to do that, 

However from what I see we might have to sadly do that

src/profiler.cpp
Baraa-Hasheesh: I will need to double check on that

src/stackWalker.cpp
Baraa-Hasheesh: make sense

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: I will refactor the test to create more threads => 10
That should make the test reliable

src/codeCache.h
roy-soumadipta: What about changing the template to accept 3 params and then branch with few extra checks which was my first proposal?
```
template <typename NamePredicate, typename T1 = const void*, typename T2 = const void*>
inline void mark(NamePredicate predicate, char value) {
    for (int i = 0; i < _count; i++) {
        const char* blob_name = _blobs[i]._name;
        if (blob_name != NULL) {
            bool should_mark = false;
            if constexpr (std::is_invocable_v<NamePredicate, const char*>) {
                should_mark = predicate(blob_name);
            } else if constexpr (std::is_invocable_v<NamePredicate, const char*, T1, T2>) {
                should_mark = predicate(blob_name, _blobs[i]._start, _blobs[i]._end);
            }
            if (should_mark) {
                NativeFunc::mark(blob_name, value);
            }
        }
    }
    
    if (value == MARK_VM_RUNTIME && _name != NULL) {
        NativeFunc::mark(_name, value);
    }
}
```

This should keep existing usage unchanged add functionality for the new requirement in the PR.

src/lockTracer.cpp
Baraa-Hasheesh: For reference I'm basically searching for certain methods in the async-profiler & mark them to be skipped during the stalk walking

We already have similar logic in mallocTracer

apangin: I don't get what refactoring you are referring to. My idea is not to touch `CodeCache::mark` function at all, that is, not to do any refactoring. `mark` walks through all functions in a CodeCache, which is inefficient when the exact address is known.

apangin: And yes, `MallocTracer` could do better, too.

src/profiler.cpp
Baraa-Hasheesh: It looks fine, however I will need to add a frame skip on it same to melloc 

image

image

src/codeCache.h
Baraa-Hasheesh: This logic will be removed following other feedback
No need to discuss this for now

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: updated here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/stackWalker.cpp
Baraa-Hasheesh: updated here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/stackWalker.cpp
Baraa-Hasheesh: updated here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/profiler.h
Baraa-Hasheesh: updated here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/profiler.cpp
Baraa-Hasheesh: As we agreed I've moved Instrument & lock profiling to use jvmti  tracer here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

However I also fix the issues for VM as it was quite simple 
=> I did that so we're ready to move to VM if the need arises in the future

src/lockTracer.cpp
Baraa-Hasheesh: Logic removed as no longer needed [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/codeCache.h
Baraa-Hasheesh: as discussed logic reverted here [c5ea623](https://github.com/async-profiler/async-profiler/pull/1370/commits/c5ea623345d747696c032f40bb0da6465b0fff6a)

src/profiler.cpp
```diff
@@ -645,30 +622,30 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         }
     }
 
+    // Lock events and instrumentation events can safely call synchronous JVM TI stack walker.
+    bool use_jvmti_tracer = event_type == PARK_SAMPLE || event_type == LOCK_SAMPLE || event_type == INSTRUMENTED_METHOD;
```
fandreuz: Maybe something similar to `hasNativeStack` could also be done here? We are in `recordSample` so that may matter

src/stackWalker.cpp
```diff
@@ -379,7 +390,31 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+            const char* method_name = profiler->findNativeMethod(pc);
+            char mark = method_name != NULL ? NativeFunc::mark(method_name) : 0;
+
+            if (java_pc_observed) {
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
+                // Skip any frames above profiler hook methods
+                depth = 0;
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (mark == MARK_COMPILER_ENTRY && Profiler::instance()->features().comp_task) {
```
fandreuz: If `mark == MARK_COMPILER_ENTRY` and `Profiler::instance()->features().comp_task` is `false`, you are going to fall into the `else` statement and treat the frame as a normal native frame. Is that intended?

src/profiler.cpp
Baraa-Hasheesh: Not really, 
As we know that LOCK events & INSTRUMENTED are safe to use JVM TI stack walker

But other events such as User events might not be safe

src/stackWalker.cpp
Baraa-Hasheesh: yes that's intended

test/test/lock/SimpleLockTest.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package test.lock;
+
+import java.util.concurrent.Semaphore;
+import java.util.function.Supplier;
+
+public class SimpleLockTest {
+
+    private static final Semaphore semaphore = new Semaphore(1);
```
fandreuz: Why a semaphore with one permit and not a simple `ReentrantLock`?

test/test/lock/SimpleLockTest.java
```diff
@@ -0,0 +1,53 @@
...
+        semaphore.release();
+    }
+
+    public static void main(String[] args) throws InterruptedException {
+        if (args.length < 1) {
+            System.out.println("Please input lock type: sync or sem");
+            System.exit(1);
+        }
+
+        Supplier<Runnable> runnableSupplier = () -> () -> {
```
fandreuz: Why `Supplier<Runnable>>`? I think you can achieve the same with a simple `Runnable`. The task is stateless

test/test/lock/SimpleLockTest.java
fandreuz: I still think this test is still not deterministic, and it's relatively easy to fix.
- The first thread calls either `semLock` or `syncLock`
- It should be blocked there until all other threads are set up (you can do this with any sync primitive in Java, like `Barrier` or `Semaphore`)
- All other threads are set up and create contention on the lock
- You break the barrier from the main thread, so the first thread can exit the method

This will create deterministic contention. This test could definitely fail in resource-constrained environments like GitHub Actions runners as it is written now.

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: Both are simple & work fine for this case
Is there a reason to switch from one to the other?

fandreuz: I just like to reduce the number of states in the objects I use, and `ReentrantLock` is a semaphore with one permit

src/profiler.cpp
fandreuz: I meant something similar to its implementation, with a `constexpr` mask.

Baraa-Hasheesh: sure I can do that

apangin: Furthermore, I don't think a new method is even needed, because `hasNativeStack` serves exactly this purpose.
We can use JVM TI stack walker if and only if event does not have native stack.

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: updated here [72c9d03](https://github.com/async-profiler/async-profiler/pull/1370/commits/72c9d034ba42bd81aa4713bce918b11dbac1dfa9)

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: updated here [72c9d03](https://github.com/async-profiler/async-profiler/pull/1370/commits/72c9d034ba42bd81aa4713bce918b11dbac1dfa9)

src/profiler.cpp
Baraa-Hasheesh: updated here [72c9d03](https://github.com/async-profiler/async-profiler/pull/1370/commits/72c9d034ba42bd81aa4713bce918b11dbac1dfa9)

test/test/lock/SimpleLockTest.java
Baraa-Hasheesh: updated here [72c9d03](https://github.com/async-profiler/async-profiler/pull/1370/commits/72c9d034ba42bd81aa4713bce918b11dbac1dfa9)

src/profiler.h
```diff
@@ -214,6 +213,7 @@ class Profiler {
     void writeLog(LogLevel level, const char* message);
     void writeLog(LogLevel level, const char* message, size_t len);
 
+    jmethodID getCurrentCompileTask();
```
apangin: Should not be in `Profiler` class anymore

src/stackWalker.cpp
```diff
@@ -242,6 +242,9 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
 
     // Should be preserved across setjmp/longjmp
     volatile int depth = 0;
+    bool java_pc_observed = false;
+    bool skip_top_native_frames = !Event::hasNativeStack(event_type);
+    bool skip_instrument_frame = event_type == INSTRUMENTED_METHOD;
```
apangin: `walkVM` is no longer called for events without native stack, right?

src/stackWalker.cpp
```diff
@@ -379,7 +390,31 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+                jmethodID compile_task = VMStructs::getCurrentCompileTask();
+                if (compile_task != NULL) {
+                    fillFrame(frames[depth++], FRAME_INTERPRETED, 0, compile_task);
+                }
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (mark == MARK_VM_RUNTIME && event_type >= ALLOC_SAMPLE && event_type <= ALLOC_OUTSIDE_TLAB) {
+                // Skip internal frames for allocation profiling
+                depth = 0;
+            } else if (mark == MARK_INTERPRETER || skip_top_native_frames){
+                // Skip interpreter frames & internal frames for LOCK & INSTRUMENT profiling
```
apangin: This does not match the original logic. `Profiler::convertNativeTrace` does not simply skip an interpreter frame, but it rather stops walking native stack once it meets the first interpreter frame.

However, I'm not sure this code works at all. HotSpot C++ interpreter is quite special, I've never tested it with VMStructs stack walker.

src/vmStructs.cpp
```diff
@@ -627,6 +627,17 @@ void VMStructs::initThreadBridge() {
     }
 }
 
+jmethodID VMStructs::getCurrentCompileTask() {
```
apangin: Let's move `getCurrentCompileTask` into `VMThread` class, since a task belongs to a particular Compiler thread.

src/stackWalker.cpp
Baraa-Hasheesh: It's still called for VMX
I think this filtering logic would help in VMX specially for native apps that start a JVM instance

src/stackWalker.cpp
Baraa-Hasheesh: Yeah, I need to adjust the condition, I need to skip all frames between this one & the first java frame

src/stackWalker.cpp
apangin: The idea of `vmx` is to provide maximum level of details without hiding any internal frames.

Baraa-Hasheesh: I'm aware 
However for example for instrument profiling the extra frames in this case are the profiler injected hooks

Do we want to show those as part of the stack?

src/profiler.h
Baraa-Hasheesh: updated here [f000b75](https://github.com/async-profiler/async-profiler/pull/1370/commits/f000b75290c8317f24e89d963c774f11d44ea349)

src/stackWalker.cpp
Baraa-Hasheesh: updated here [f000b75](https://github.com/async-profiler/async-profiler/pull/1370/commits/f000b75290c8317f24e89d963c774f11d44ea349)

Now native frames are fully skipped until the first Java PC/frame is seen

src/vmStructs.cpp
Baraa-Hasheesh: handled here [f000b75](https://github.com/async-profiler/async-profiler/pull/1370/commits/f000b75290c8317f24e89d963c774f11d44ea349)

Baraa-Hasheesh:
while doing some testing on VM/VMX across multiple systems I've found that some crashes may happen for synchronous events such as ALLOC & MALLOC,

This should be fixed before proceeding with this change

I will create a separate issue & change the PR to draft in the meantime

src/stackWalker.cpp
apangin: Yes, `vmx` is not useful and is not recommended for Java method profiling, but if someone still wants to set it, let them see everything.

src/vmStructs.h
```diff
@@ -382,6 +382,8 @@ class VMThread : VMStructs {
         }
         return NULL;
     }
+
+    static jmethodID getCurrentCompileTask();;
```
apangin: ;;

src/stackWalker.cpp
apangin: Have you ever tested this?
My point was, VMStructs based stack walking was never supposed to work with C++ Bytecode Interpreter (aka Zero VM). What you've changed here is basically a dead code. Valid behavior is to return an error if user tries to use `vm/vmx` with Zero.

src/stackWalker.cpp
Baraa-Hasheesh: sure, I will remove the changes related for VMX & only keep what's needed for `alloc` & other profiling  modes that have Native frames for VM mode

src/stackWalker.cpp
Baraa-Hasheesh: I wasn't aware of this fact, 
I will generate local builds & test this out, 
If I do see such behaviors I will adjust accordingly

Baraa-Hasheesh: I've created some Zero VM JDK builds internally & as mentioned Zero builds will not work with VM/VMX stack walkers 

I will remove the logic here

src/vmStructs.h
Baraa-Hasheesh: handled here [9db28d8](https://github.com/async-profiler/async-profiler/pull/1370/commits/9db28d8160d48b88077999fe362410206e9e085d)

src/stackWalker.cpp
Baraa-Hasheesh: handled here [9db28d8](https://github.com/async-profiler/async-profiler/pull/1370/commits/9db28d8160d48b88077999fe362410206e9e085d) 

I had to re-order the conditions to include the `comptask` feature into VMX

src/stackWalker.cpp
```diff
@@ -382,7 +385,31 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+            const char* method_name = profiler->findNativeMethod(pc);
+            char mark = method_name != NULL ? NativeFunc::mark(method_name) : 0;
+
+            if (java_pc_observed) {
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (mark == MARK_COMPILER_ENTRY && Profiler::instance()->features().comp_task) {
+                // Insert current compile task as a pseudo Java frame
+                jmethodID compile_task = VMThread::getCurrentCompileTask();
+                if (compile_task != NULL) {
+                    fillFrame(frames[depth++], FRAME_INTERPRETED, 0, compile_task);
```
apangin: With other stack walking modes, such frames are displayed in bright green (JIT compiled).

src/stackWalker.cpp
```diff
@@ -382,7 +385,31 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+                }
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (detail == VM_EXPERT) { // don't skip anything for VMX
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
+            } else if (mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
+                // Skip any frames above profiler hook methods
+                depth = 0;
+                fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
```
apangin: This line is duplicated in almost all branches.
Can it be refactored to add a native frame in a single place?

src/stackWalker.cpp
```diff
@@ -260,6 +261,8 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
     // Walk until the bottom of the stack or until the first Java frame
     while (depth < max_depth) {
         if (CodeHeap::contains(pc)) {
+            java_pc_observed = true;
```
apangin: This is not always true. `CodeHeap` also contains generated stubs that can be called from native code.

src/stackWalker.cpp
Baraa-Hasheesh: I will see what can be done, but the ALLOC condition is what's holding us back here

src/stackWalker.cpp
Baraa-Hasheesh: I will update it

src/stackWalker.cpp
Baraa-Hasheesh: while it's true the use case for the enhancement logic we have this should good, I would change the name to `codeheap_pc_observed` to avoid calling it `java`

The use cases are: 
* Skip top frames for malloc samples (will be first couple of PCs)
* ALLOC frame skip which also happens on top before any other stub
* For MARK_COMPILER_ENTRY I'm not 100% confident I'm pretty sure it should be fine but I will double check

What do you think about this?

apangin: Compiler thread can certainly call out to runtime generated stubs (such as ARM64 atomic entries).

src/stackWalker.cpp
Baraa-Hasheesh: Updated here [314d07e](https://github.com/async-profiler/async-profiler/pull/1370/commits/314d07e27531ed5459716a04272e610959210e91)

src/stackWalker.cpp
Baraa-Hasheesh: Updated here [314d07e](https://github.com/async-profiler/async-profiler/pull/1370/commits/314d07e27531ed5459716a04272e610959210e91)

src/stackWalker.cpp
Baraa-Hasheesh: Updated here [314d07e](https://github.com/async-profiler/async-profiler/pull/1370/commits/314d07e27531ed5459716a04272e610959210e91)

src/stackWalker.cpp
Baraa-Hasheesh: handled in latest changes

krk:
> Furthermore this PR includes multiple enhancements on the VM/VMX stackwalking modes which include
>
>    Skipping some async-profiler frames from the stack as those frames don't provide useful debugging information
>    Skipping some internal JVM methods that create noise without helping much in the profiling
>    augmenting the native stacks with comptask frames


Could you split the enhancements into a new PR, to reduce the size of it?

Also, the images in the PR description produce HTTP 404 when clicked.

Baraa-Hasheesh:
> > Furthermore this PR includes multiple enhancements on the VM/VMX stackwalking modes which include
> > Skipping some async-profiler frames from the stack as those frames don't provide useful debugging information
> > Skipping some internal JVM methods that create noise without helping much in the profiling
> > augmenting the native stacks with comptask frames
> 
> Could you split the enhancements into a new PR, to reduce the size of it?
> 
> Also, the images in the PR description produce HTTP 404 when clicked.

The PR is relatively small 
Most changes are test additions if you think about it

Baraa-Hasheesh:
Closing this PR as https://github.com/async-profiler/async-profiler/pull/1537 handles most of what's needed

Need to create a new change to move alloc profiling to use VM

Also I do remember having a discussion about moving method trace samples to it as well  https://github.com/async-profiler/async-profiler/pull/1421#discussion_r2313683624

@apangin correct me if I'm wrong about the method trace

PR #1373, opened 2025-07-09
Save all generated logs for debug purposes

Baraa-Hasheesh:
### Description
keep all logs for debugging purposes 
This fix the following problems:
* If multiple profiling sessions are done, each session will overwrite the previous one logs 
* If a test depends on both profiler session logs & process logs only one of those logs will be kept
* Not all generated profiler logs were saved

### Related issues
N/A

### Motivation and context
Make debugging flakey tests easier

### How has this been tested?
`make test`

---

### DIFF
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -246,7 +246,7 @@ public class TestProcess implements Closeable {
 
             for (String key : tmpFiles.keySet()) {
                 if (!key.equals(STDERR) && !key.equals(STDOUT)) {
-                    moveLog(tmpFiles.get(key), key.replaceAll("%", ""), true);
+                    moveLog(tmpFiles.get(key), key.substring(1), true);
                 }
             }
         } catch (IOException e) {
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -34,9 +34,11 @@ public class TestProcess implements Closeable {
     public static final String TESTBIN = "%testbin";
     public static final String TESTLIB = "%testlib";
 
+    public int currentProfiler = 0;
+
     private static final String JAVA_HOME = System.getProperty("java.home");
 
-    private static final Pattern filePattern = Pattern.compile("(%[a-z]+)(\\.[a-z]+)?");
+    private static final Pattern filePattern = Pattern.compile("(%[a-z][a-z0-9_]*)(\\.[a-z]+)?");
 
     private static final MethodHandle pid = getPidHandle();
 
@@ -241,14 +243,16 @@ public class TestProcess implements Closeable {
         try {
             Files.createDirectories(Paths.get(logDir));
 
-            File stdout = tmpFiles.getOrDefault(PROFOUT, tmpFiles.get(STDOUT));
-            moveLog(stdout, "stdout", true);
+            moveLog(tmpFiles.get(STDOUT), "stdout", true);
+            moveLog(tmpFiles.get(STDERR), "stderr", false);
 
-            File stderr = tmpFiles.getOrDefault(PROFERR, tmpFiles.get(STDERR));
-            moveLog(stderr, "stderr", false);
+            for (String key : tmpFiles.keySet()) {
+                if (key.equals(STDERR) || key.equals(STDOUT)) {
+                    continue;
+                }
 
-            File profile = tmpFiles.get("%f");
-            moveLog(profile, "profile", true);
+                moveLog(tmpFiles.get(key), "profile" + key.replace('%', '-'), true);
+            }
         } catch (IOException e) {
             log.log(Level.WARNING, "Failed to move logs", e);
         }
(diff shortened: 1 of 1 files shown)


fandreuz:
> Make debugging flakey tests easier

How does this work?

Baraa-Hasheesh:
> > Make debugging flakey tests easier
> 
> How does this work?

Some logs are currently messing, I.E, we can't check them if a failure happens 
By saving those log files it would be possible to download them when a failure happens on GHA which makes it easier to debug

test/one/profiler/test/TestProcess.java
```diff
@@ -321,7 +327,7 @@ public Output profile(String args, boolean sudo) throws IOException, TimeoutExce
         log.log(Level.FINE, "Profiling " + cmd);
 
         Process p = new ProcessBuilder(cmd)
-                .redirectOutput(createTempFile(PROFOUT))
+                .redirectOutput(createTempFile(PROFOUT + profilerId))
```
apangin: This is not very helpful. Other outputs (stdout, stderr, custom `%f` files) will still overwrite each other.
More logs does not mean better, we'll get the last failed one anyway. I propose to leave this as it was.

test/one/profiler/test/TestProcess.java
```diff
@@ -241,14 +243,16 @@ private void moveLogs() {
...
-            File stderr = tmpFiles.getOrDefault(PROFERR, tmpFiles.get(STDERR));
-            moveLog(stderr, "stderr", false);
+            for (String key : tmpFiles.keySet()) {
+                if (key.equals(STDERR) || key.equals(STDOUT)) {
+                    continue;
+                }
 
-            File profile = tmpFiles.get("%f");
-            moveLog(profile, "profile", true);
+                moveLog(tmpFiles.get(key), "profile" + key.replace('%', '-'), true);
```
apangin: `profile-` prefix is redundant. E.g. `profile-perr` name does not make much sense to me.

test/one/profiler/test/TestProcess.java
```diff
@@ -241,14 +243,16 @@ private void moveLogs() {
...
-            File stdout = tmpFiles.getOrDefault(PROFOUT, tmpFiles.get(STDOUT));
-            moveLog(stdout, "stdout", true);
+            moveLog(tmpFiles.get(STDOUT), "stdout", true);
+            moveLog(tmpFiles.get(STDERR), "stderr", false);
 
-            File stderr = tmpFiles.getOrDefault(PROFERR, tmpFiles.get(STDERR));
-            moveLog(stderr, "stderr", false);
+            for (String key : tmpFiles.keySet()) {
+                if (key.equals(STDERR) || key.equals(STDOUT)) {
+                    continue;
```
apangin: nit: when the body of the loop is just one line, it's better to go ahead without `continue`:
```
if (!key.equals(STDOUT) && !key.equals(STDERR)) {
    moveLog(...);
}
```

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: what about cases where we replace other files

```
@Test(sh = "%testbin/non_java_app 3 %f.collapsed %s.collapsed", output = true)
public void jvmInBetween(TestProcess p) throws Exception {
```

ideally the test should give it a clear name so we have two options here 
* Add `profile-` & keep names as %f => `profile-f`
* Loop & give meaningful names for all file names

What do you think?

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: handled here [b447135](https://github.com/async-profiler/async-profiler/pull/1373/commits/b4471350bb4307a21f855f5fe007935de592c957)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: as discussed offline we will use the key as is 
For future we should try to give meaningful names 

Updated here [b447135](https://github.com/async-profiler/async-profiler/pull/1373/commits/b4471350bb4307a21f855f5fe007935de592c957)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: removed [b447135](https://github.com/async-profiler/async-profiler/pull/1373/commits/b4471350bb4307a21f855f5fe007935de592c957)

test/one/profiler/test/TestProcess.java
```diff
@@ -241,14 +241,14 @@ private void moveLogs() {
...
+            moveLog(tmpFiles.get(STDERR), "stderr", false);
 
-            File stderr = tmpFiles.getOrDefault(PROFERR, tmpFiles.get(STDERR));
-            moveLog(stderr, "stderr", false);
-
-            File profile = tmpFiles.get("%f");
-            moveLog(profile, "profile", true);
+            for (String key : tmpFiles.keySet()) {
+                if (!key.equals(STDERR) && !key.equals(STDOUT)) {
+                    moveLog(tmpFiles.get(key), key.replaceAll("%", ""), true);
```
apangin: Maybe `key.substring(1)` instead of `replaceAll`? This is more efficient and matches other usages in `TestProcess`.

Baraa-Hasheesh: updated here [347d09e](https://github.com/async-profiler/async-profiler/pull/1373/commits/347d09ed5e4ab42a2852f5f6fd4fefbbd4eacabf)

PR #1374, opened 2025-07-09
Correctly check if profiler is preloaded

Baraa-Hasheesh:
### Description
Add proper checking to validate that the profiler was preloaded as to avoid false starts

Currently if `ASPROF_COMMAND` env variable is defined any dlopen on the profiler will cause it to start which means the profiler might start with incorrect profiling options

Furthermore in this case async-profiler will assume that preload hooks exists which might not be true, which will to incomplete profiling by the profiler

### Related issues
N/A

### Motivation and context
Avoid false early starts for the async-profiler

### How has this been tested?
New test added
`make test`

---

### DIFF
--- a/src/os_linux.cpp
+++ b/src/os_linux.cpp
@@ -8,8 +8,10 @@
 #include <arpa/inet.h>
 #include <byteswap.h>
 #include <dirent.h>
+#include <dlfcn.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <link.h>
 #include <sched.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -387,5 +389,43 @@ void OS::freePageCache(int fd, off_t start_offset) {
 int OS::mprotect(void* addr, size_t size, int prot) {
     return ::mprotect(addr, size, prot);
 }
+
+static int checkPreloadedCallback(dl_phdr_info* info, size_t size, void* data) {
+    Dl_info* dl_info = (Dl_info*)data;
+
+    Dl_info current_info = dl_info[0];
+    Dl_info sigaction_info = dl_info[1];
+
+    if (info->dlpi_name == NULL) {
+        return 0;
+    } else if (strcmp(info->dlpi_name, current_info.dli_fname) == 0) {
+        return 1;
+    } else if (strcmp(info->dlpi_name, sigaction_info.dli_fname) == 0) {
+        return -1;
+    }
+
+    return 0;
+}
+
+bool OS::checkPreloaded() {
+    if (getenv("LD_PRELOAD") == NULL) {
+        return false;
+    }
+
+    Dl_info current_info;
+    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fname == NULL) {
+        return false;
+    }
+
+    Dl_info sigaction_info;
+    if (dladdr((const void*)sigaction, &sigaction_info) == 0 || sigaction_info.dli_fname == NULL) {
+        return false;
+    }
+
+    Dl_info info[2] = {current_info, sigaction_info};
+
+    return dl_iterate_phdr(checkPreloadedCallback, (void*)info) == 1;
+}
+
 
 #endif // __linux__
--- a/src/os_linux.cpp
+++ b/src/os_linux.cpp
@@ -394,14 +394,12 @@ static int checkPreloadedCallback(dl_phdr_info* info, size_t size, void* data) {
     Dl_info* dl_info = (Dl_info*)data;
 
     Dl_info current_info = dl_info[0];
-    Dl_info sigaction_info = dl_info[1];
+    Dl_info exit_info = dl_info[1];
 
-    if (info->dlpi_name == NULL) {
-        return 0;
-    } else if ((void*)info->dlpi_addr == current_info.dli_fbase) {
+    if ((void*)info->dlpi_addr == current_info.dli_fbase) {
         // async-profiler found first
         return 1;
-    } else if ((void*)info->dlpi_addr == sigaction_info.dli_fbase) {
+    } else if ((void*)info->dlpi_addr == exit_info.dli_fbase) {
         // libc found first
         return -1;
     }
@@ -418,17 +416,17 @@ bool OS::checkPreloaded() {
 
     // Find async-profiler shared objects
     Dl_info current_info;
-    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fname == NULL) {
+    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fbase == NULL) {
         return false;
     }
 
     // Find libc shared objects
-    Dl_info sigaction_info;
-    if (dladdr((const void*)sigaction, &sigaction_info) == 0 || sigaction_info.dli_fname == NULL) {
+    Dl_info exit_info;
+    if (dladdr((const void*)exit, &exit_info) == 0 || exit_info.dli_fbase == NULL) {
         return false;
     }
 
-    Dl_info info[2] = {current_info, sigaction_info};
+    Dl_info info[2] = {current_info, exit_info};
 
     return dl_iterate_phdr(checkPreloadedCallback, (void*)info) == 1;
 }
--- a/src/os_linux.cpp
+++ b/src/os_linux.cpp
@@ -398,25 +398,31 @@ static int checkPreloadedCallback(dl_phdr_info* info, size_t size, void* data) {
 
     if (info->dlpi_name == NULL) {
         return 0;
-    } else if (strcmp(info->dlpi_name, current_info.dli_fname) == 0) {
+    } else if ((void*)info->dlpi_addr == current_info.dli_fbase) {
+        // async-profiler found first
         return 1;
-    } else if (strcmp(info->dlpi_name, sigaction_info.dli_fname) == 0) {
+    } else if ((void*)info->dlpi_addr == sigaction_info.dli_fbase) {
+        // libc found first
         return -1;
     }
 
     return 0;
 }
 
+// Checks if the async is included in the preload list defined in the LD_PRELOAD env variable
+// This check is done by checking if the async-profiler is loaded before libc shared objects
 bool OS::checkPreloaded() {
     if (getenv("LD_PRELOAD") == NULL) {
         return false;
     }
 
+    // Find async-profiler shared objects
     Dl_info current_info;
     if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fname == NULL) {
         return false;
     }
 
+    // Find libc shared objects
     Dl_info sigaction_info;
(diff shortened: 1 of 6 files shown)


src/os_linux.cpp
```diff
@@ -388,4 +390,42 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
 
+static int checkPreloadedCallback(dl_phdr_info* info, size_t size, void* data) {
+    Dl_info* dl_info = (Dl_info*)data;
+
+    Dl_info current_info = dl_info[0];
+    Dl_info sigaction_info = dl_info[1];
+
+    if (info->dlpi_name == NULL) {
+        return 0;
+    } else if (strcmp(info->dlpi_name, current_info.dli_fname) == 0) {
```
apangin: Can we compare base address (dli_fbase) instead?

src/os_macos.cpp
```diff
@@ -367,4 +370,34 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+    }
+
+    uint32_t images = _dyld_image_count();
+    for (uint32_t i = 0; i < images; i++) {
+        const char* image_name = _dyld_get_image_name(i);
+        if (image_name == NULL) {
+            continue;
+        } else if (strcmp(image_name, current_info.dli_fname) == 0) {
+            return true;
+        }else if(strcmp(image_name, sigaction_info.dli_fname) == 0) {
```
apangin: Formatting

src/os_linux.cpp
```diff
@@ -388,4 +390,42 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+        return false;
+    }
+
+    Dl_info current_info;
+    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fname == NULL) {
+        return false;
+    }
+
+    Dl_info sigaction_info;
+    if (dladdr((const void*)sigaction, &sigaction_info) == 0 || sigaction_info.dli_fname == NULL) {
```
apangin: Just wondering why `sigaction`? Not the most obvious function in libc - may confuse someone who sees this for the first time. A short comment describing the idea of `checkPreloaded` would be useful anyway.

src/zInit.cpp
```diff
@@ -23,7 +23,7 @@ class LateInitializer {
             dlopen(dl_info.dli_fname, RTLD_LAZY | RTLD_NODELETE);
         }
 
-        if (!checkJvmLoaded()) {
+        if (!checkJvmLoaded() && OS::checkPreloaded()) {
```
apangin: A minor thing, but I'd propose to move this check _after_ `ASPROF_COMMAND` check, since using async-profiler through native API will be a common thing, while `LD_PRELOAD` is used rarely and mostly for ad-hoc profiling only.

src/os_linux.cpp
Baraa-Hasheesh: I wanted to avoid typical functions that some other developer might interpose/preload as that would effect the accuracy of the check 

For example `dlopen`on Linux `dladdr` will resolve to the async-profiler when it's preloaded

=> I wanted to avoid using functions that we patch/preload to keep the check consistent for all cases

I will add a comment for describing what the function does

src/os_linux.cpp
Baraa-Hasheesh: I will need to check on this

src/zInit.cpp
Baraa-Hasheesh: handled here [78edb2a](https://github.com/async-profiler/async-profiler/pull/1374/commits/78edb2a03c8d0babfd939ec7cf589712b61c63b3)

src/os_linux.cpp
Baraa-Hasheesh: doc added here [78edb2a](https://github.com/async-profiler/async-profiler/pull/1374/commits/78edb2a03c8d0babfd939ec7cf589712b61c63b3)

src/os_macos.cpp
Baraa-Hasheesh: handled here [78edb2a](https://github.com/async-profiler/async-profiler/pull/1374/commits/78edb2a03c8d0babfd939ec7cf589712b61c63b3)

src/os_linux.cpp
Baraa-Hasheesh: have been updated here [78edb2a](https://github.com/async-profiler/async-profiler/pull/1374/commits/78edb2a03c8d0babfd939ec7cf589712b61c63b3)

From the testing I did it looks like it does the trick

src/os_linux.cpp
```diff
@@ -388,4 +390,48 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
 }
 
+static int checkPreloadedCallback(dl_phdr_info* info, size_t size, void* data) {
+    Dl_info* dl_info = (Dl_info*)data;
+
+    Dl_info current_info = dl_info[0];
+    Dl_info sigaction_info = dl_info[1];
+
+    if (info->dlpi_name == NULL) {
+        return 0;
```
apangin: Redundant check?

src/os_linux.cpp
```diff
@@ -388,4 +390,48 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+// Checks if the async is included in the preload list defined in the LD_PRELOAD env variable
+// This check is done by checking if the async-profiler is loaded before libc shared objects
+bool OS::checkPreloaded() {
+    if (getenv("LD_PRELOAD") == NULL) {
+        return false;
+    }
+
+    // Find async-profiler shared objects
+    Dl_info current_info;
+    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fname == NULL) {
```
apangin: Is comparing `dli_fname` really necessary?

src/os_linux.cpp
apangin: Still, `sigaction` is a bad choice. Not only because it's some random unrelated function, but also because it does not meet the criteria you've mentioned:

> I wanted to avoid typical functions that some other developer might interpose/preload

It's not uncommon to intercept `sigaction` for signal chaining purposes. A prominent example is JDK's [`libjsig`](https://docs.oracle.com/en/java/javase/24/vm/signal-chaining.html) library.

I'd choose some basic function like `exit`.

src/os_macos.cpp
```diff
@@ -367,4 +370,40 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+
+    // Find libc shared objects
+    Dl_info sigaction_info;
+    if (dladdr((const void*)sigaction, &sigaction_info) == 0 || sigaction_info.dli_fname == NULL) {
+        return false;
+    }
+
+    uint32_t images = _dyld_image_count();
+    for (uint32_t i = 0; i < images; i++) {
+        const char* image_name = _dyld_get_image_name(i);
```
apangin: Is it possible to use base address on macOS, similarly to how you do this on Linux?

src/os_linux.cpp
Baraa-Hasheesh: I will update it to check `dli_fbase` instead

src/os_macos.cpp
Baraa-Hasheesh: I did some double checking, yes we can check against _dyld_get_image_header

Baraa-Hasheesh: Updated here [6e283d6](https://github.com/async-profiler/async-profiler/pull/1374/commits/6e283d672ff36465d6803886f03e3a42f918415a)

src/os_linux.cpp
Baraa-Hasheesh: updated here [6e283d6](https://github.com/async-profiler/async-profiler/pull/1374/commits/6e283d672ff36465d6803886f03e3a42f918415a)

src/os_linux.cpp
Baraa-Hasheesh: removed [6e283d6](https://github.com/async-profiler/async-profiler/pull/1374/commits/6e283d672ff36465d6803886f03e3a42f918415a)

src/os_linux.cpp
Baraa-Hasheesh: updated here [6e283d6](https://github.com/async-profiler/async-profiler/pull/1374/commits/6e283d672ff36465d6803886f03e3a42f918415a)

src/os_linux.cpp
```diff
@@ -388,4 +390,45 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+// Checks if the async is included in the preload list defined in the LD_PRELOAD env variable
+// This check is done by checking if the async-profiler is loaded before libc shared objects
+bool OS::checkPreloaded() {
+    if (getenv("LD_PRELOAD") == NULL) {
+        return false;
+    }
+
+    // Find async-profiler shared objects
+    Dl_info current_info;
+    if (dladdr((const void*)OS::checkPreloaded, &current_info) == 0 || current_info.dli_fbase == NULL) {
```
apangin: Why would `dli_fbase` be `NULL`? Is it even possible?

src/os_macos.cpp
```diff
@@ -367,4 +370,39 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
 }
 
+// Checks if the async is included in the preload list defined in the DYLD_INSERT_LIBRARIES env variable
+// This check is done by checking if the async-profiler is loaded before libc shared objects
+bool OS::checkPreloaded() {
+    if (getenv("DYLD_INSERT_LIBRARIES") == NULL) {
+        return false;
+    }
+
+    // Find async-profiler shared objects
```
apangin: object (singular)?

src/os_macos.cpp
```diff
@@ -367,4 +370,39 @@ int OS::mprotect(void* addr, size_t size, int prot) {
...
+    Dl_info current_info;
```
apangin: `current_info` and `exist_info` sound a bit vague.
What about `lib_async_profiler` (or maybe `libprofiler`) and `libc`?

Then conditions like `if (image_base == libc.dli_fbase)` will be self-descriptive.

src/os_linux.cpp
Baraa-Hasheesh: yeah it shouldn't be needed, will remove it

src/os_macos.cpp
Baraa-Hasheesh: I will use `libprofiler` & `libc`

Baraa-Hasheesh: Updated here [486dc9d](https://github.com/async-profiler/async-profiler/pull/1374/commits/486dc9ddfe5118477b20000dd09554f7ef95941e)

src/os_macos.cpp
Baraa-Hasheesh: [486dc9d](https://github.com/async-profiler/async-profiler/pull/1374/commits/486dc9ddfe5118477b20000dd09554f7ef95941e)

src/os_linux.cpp
Baraa-Hasheesh: [486dc9d](https://github.com/async-profiler/async-profiler/pull/1374/commits/486dc9ddfe5118477b20000dd09554f7ef95941e)

PR #1375, opened 2025-07-09
Create `test/deps` if it does not exist before running Makefile recipes

fandreuz:
### Description
`build/test.jar` has a dependency on `test/deps`, because the JAR should be rebuilt when a new dependency JAR is pulled in. In this PR I add an invocation to `mkdir` at the beginning of the Makefile to make sure that the directory exists.

### Related issues
@Baraa-Hasheesh experienced the problem in his local environment. We did not spot that during GHA runs because we pull in dependencies there (Protobuf Java runtime). 

### Motivation and context
The recipe should not fail if the directory does not exist.

### How has this been tested?
Tested locally.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,7 +273,7 @@ coverage: clean-coverage
 test: test-cpp test-java
 
 $(TEST_DEPS_DIR):
-	mkdir -p test/deps
+	mkdir -p $@
 
 build/$(TEST_JAR): build/$(API_JAR) $(TEST_SOURCES) build/$(CONVERTER_JAR) $(TEST_DEPS_DIR)
 	rm -rf build/test/classes
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,7 @@ ifneq (,$(findstring $(ARCH_TAG),x86 x64 arm64))
   CXXFLAGS += -momit-leaf-frame-pointer
 endif
 
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,6 +133,9 @@ ifneq (,$(findstring $(ARCH_TAG),x86 x64 arm64))
   CXXFLAGS += -momit-leaf-frame-pointer
 endif
 
+# TEST_DEPS_DIR is a dependency for build/$(TEST_JAR), thus it must exist before the
+# recipe dependencies are evaluated.
+$(shell mkdir -p $(TEST_DEPS_DIR))
```
apangin: Can `TEST_DEPS_DIR` be a target?
Doing modifications in the filesystem outside Makefile target does not seem a good idea.

fandreuz: Sure, I thought the recipes that depend on the directory would not re-run if a target exists with the same name, but I realized that's not the case. Fixed in c9acf1f686872babe84d096d15e4ea4d5eceff83

Makefile
```diff
@@ -272,6 +272,9 @@ coverage: clean-coverage
 
 test: test-cpp test-java
 
+$(TEST_DEPS_DIR):
+	mkdir -p test/deps
```
apangin:
```suggestion
	mkdir -p $@
```

PR #1378, opened 2025-07-10
Extract JFR symbol lookup

fandreuz:
### Description
In this PR I refactor JFR symbol lookup to let `dumpOtlp` use it to enhance the information we provide for each frame.

### Related issues
Follow-up of #1342

### Motivation and context
Ideally we would like to fill the following for all frames in `dumpOtlp`:
```cpp
- name
- system_name`
- file_name
- cc
- address
- line
- type
```
This PR allows `dumpOtlp` to easily consume the rich frame lookup from JFR.

### How has this been tested?
GHA

---


src/lookup.h
```diff
@@ -0,0 +1,267 @@
...
+
+        jvmtiEnv* jvmti = VM::jvmti();
+        jvmtiError err;
+
+        if ((err = jvmti->GetMethodName(method, &method_name, &method_sig, NULL)) == 0 &&
+            (err = jvmti->GetMethodDeclaringClass(method, &method_class)) == 0 &&
+            (err = jvmti->GetClassSignature(method_class, &class_name, NULL)) == 0 &&
+            (err = jvmti->GetSourceFileName(method_class, &source_name)) == 0) {
+            mi->_class = _classes->lookup(class_name + 1, strlen(class_name) - 2);
+            mi->_file = _classes->lookup(source_name);
```
apangin: This adds an entry in the classes dictionary. This name will then appear as a class in JMC or other JFR viewer.

src/lookup.h
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    void fillJavaClassInfo(MethodInfo* mi, u32 class_id) {
+        mi->_class = class_id;
+        mi->_name = _symbols->indexOf("");
+        mi->_sig = _symbols->indexOf("()L;");
+        mi->_modifiers = 0;
+        mi->_line_number_table_size = 0;
+        mi->_line_number_table = NULL;
+        mi->_type = FRAME_INLINED;
```
apangin: What about other fields? Is it OK to leave them uninitialized?

src/lookup.h
```diff
@@ -0,0 +1,267 @@
...
+        _jni(VM::jni()) {
+
+        if (_packages == nullptr) {
+            _owns_packages = true;
+            _packages = new Index();
+        }
+        if (_symbols == nullptr) {
+            _owns_symbols = true;
+            _symbols = new Index();
+        }
```
apangin: This "ownership" concept does not seem the best design solution to me.
OTLP format does not use packages, IIRC?

src/lookup.h
```diff
@@ -0,0 +1,267 @@
...
+            mi->_file = _classes->lookup(lib_name + 1, strlen(lib_name) - 2);
+        } else {
+            mi->_file = _classes->lookup(lib_name);
+        }
+
+        mi->_modifiers = 0x100;
+        mi->_line_number_table_size = 0;
+        mi->_line_number_table = NULL;
+
+        mi->_system_name = _symbols->indexOf(name);
```
apangin: This pollutes symbol table in JFR, even though `_system_name` is never used there.

src/lookup.h
fandreuz: Ok for now, but I'd very much like to fill them later. This PR is just a proposal/draft to probe the sentiment for these changes

src/lookup.h
fandreuz: Thanks, I wasn't aware of this problem. I'll fix it somehow

src/lookup.h
fandreuz: 520d2f1cf10720c5561880d479f990ba512b6baf

src/lookup.h
fandreuz: I agree it's a bit awkward, fixed in e1b90df72a7291d24a8103783d2f8d9308557339 and 4eef22b649c7b98b9d5b383f4c927bf9debd8390

src/lookup.h
fandreuz: Fixed in f59a96fc823498ac990920984e81019971fd258c

src/converter/one/jfr/Dictionary.java
```diff
@@ -37,10 +37,6 @@ public int size() {
     }
 
     public void put(long key, T value) {
-        if (key == 0) {
```
fandreuz, line 40: @apangin Is there a reason not to allows this? I've changed the JFR writer to use `Index`, so now `key=0` can happen

apangin: `0` is used as a marker for empty slots. `keys[]` array is initially filled with zeros.

fandreuz: 28d71a8cedac280518b22f98d87d309e78520fd5

fandreuz:
@apangin I removed the OTEL part from this PR, I think it will be easier to review if we only focus on the JFR changes. I'll open a follow-up PR with OTEL additions.

src/converter/one/jfr/JfrReader.java
```diff
@@ -503,7 +503,12 @@ private void readSymbols() {
             if (buf.get() != 3) {
                 throw new IllegalArgumentException("Invalid symbol encoding");
             }
-            symbols.put(id, getBytes());
+            byte[] bytes = getBytes();
+            if (id == 0 && bytes.length == 0) {
+                // Empty string at index 0
+                continue;
+            }
+            symbols.put(id, bytes);
```
apangin: Besides the fact this hack does not look very nice, it likely breaks JFR reader.
We never record an empty string (considering it has zero ID), and therefore will never get it from `symbols`. Trying to access a zero key will return `null` instead.

src/flightRecorder.cpp
```diff
@@ -1148,7 +929,7 @@ class Recording {
             if (mi._mark) {
                 mi._mark = false;
                 buf->putVar32(mi._key);
-                buf->putVar32(mi._class);
+                buf->putVar32(std::max(mi._class, mi._file));
```
apangin: This doesn't look right.
What if both a non-zero, but `mi._class > mi._file`?

src/lookup.h
```diff
@@ -0,0 +1,264 @@
...
+    u32 _key = 0;
+    u32 _class = 0;
+    u32 _file = 0;
+    u32 _name = 0;
+    u32 _system_name = 0;
+    u32 _sig = 0;
+    jint _modifiers;
+    jint _line_number_table_size;
+    jvmtiLineNumberEntry* _line_number_table;
+    FrameTypeId _type;
```
apangin: Shouldn't we initialize other fields here as well? So that we don't need to repeat
```
    mi->_line_number_table_size = 0;
    mi->_line_number_table = NULL;
```
in all other places and risking to leave them uninitialized?

src/flightRecorder.cpp
fandreuz: Right, it should rather be `mi._class != 0 ? mi._class : mi._file`

src/converter/one/jfr/JfrReader.java
fandreuz: Right, tests are running because we defend agains `null` values:
```java
String classStr = className == null || className.length == 0 || isNativeFrame(methodType) ? "" :
                toJavaClassName(className, 0, args.dot);
String methodStr = methodName == null || methodName.length == 0 ? "" :
                new String(methodName, StandardCharsets.UTF_8);
```
But to be fair, I don't think we should give up this refactoring because `Dictionary.java` does not accept `key=0`. I would expect it to be agnostic on the particular key we try to use. This feels more like a limitation of `Dictionary` than a problem in what I'm proposing.

I see from your comment [here](https://github.com/async-profiler/async-profiler/pull/1378#discussion_r2222367126) that the problem is that allowing `key=0` would conflict with the value used for empty keys. How about we do one of the following:
- We use `key=-1` for unset positions
- We maintain a separate `bool[]` to keep track of which positions are set.

src/flightRecorder.cpp
apangin: You mean, `_file` should have priority, not `_class`?

src/converter/one/jfr/JfrReader.java
apangin: Mmm, I wouldn't like to maintain a separate array for this. After all, `Dictionary` is not supposed to be universally applicable data structure (this is what `HashMap` is for).
I thought of using `-1` as an empty marker instead of `0`. But since `0` is naturally a default value, a smart trick would be to add `+1` when setting a key, and subtracting `1` on reading.

src/flightRecorder.cpp
fandreuz: Why `_file`? My goal would be to keep things as they were before this PR (for JFR). So we keep the class information by default, and fallback to file in other cases (native methods?)

src/lookup.h
fandreuz: Right: f9b14843e6c9db5e7a22091ea1a76bf60fa48106

src/converter/one/jfr/JfrReader.java
fandreuz: Sure, that's better: c645853af7d2a46154c7e5a4378861a5bd9214d7

src/flightRecorder.cpp
apangin: Ah, it's the _type_ field of the method, I misread the code, sorry.
But shouldn't then `_class` field always be set for methods when writing JFR, even for native functions? Because the field is of `T_CLASS` type.

src/converter/one/jfr/Dictionary.java
```diff
@@ -60,6 +67,8 @@ public void put(long key, T value) {
 
     @SuppressWarnings("unchecked")
     public T get(long key) {
+        key = remapKey(key);
```
apangin, line 70: What about `forEach`?

src/lookup.h
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        if (first_time && jvmti->GetMethodModifiers(method, &mi->_modifiers) != 0) {
+            mi->_modifiers = 0;
+        }
+
+        if (first_time && jvmti->GetLineNumberTable(method, &mi->_line_number_table_size, &mi->_line_number_table) != 0) {
+            mi->_line_number_table_size = 0;
+            mi->_line_number_table = NULL;
+        }
```
apangin: Since all members are initialized now, this can be simplified to
```
if (first_time) {
    jvmti->GetMethodModifiers(method, &mi->_modifiers);
    jvmti->GetLineNumberTable(method, &mi->_line_number_table_size, &mi->_line_number_table);
}
```

src/lookup.h
```diff
@@ -0,0 +1,259 @@
...
+
+    void fillJavaClassInfo(MethodInfo* mi, u32 class_id) {
+        mi->_class = class_id;
+        mi->_name = _symbols->indexOf("");
+        mi->_sig = _symbols->indexOf("()L;");
+        mi->_type = FRAME_INLINED;
+    }
+
+  public:
+    Lookup(MethodMap* method_map, Dictionary* classes, Index* packages, Index* symbols, ExportType export_type = ExportType::JFR) :
```
apangin: There is already `enum Output` with `OUTPUT_JFR` and `OUTPUT_OTLP` values. No need to introduce another entity with a similar meaning.
Also, make it the required argument.

src/vmEntry.h
```diff
@@ -25,6 +25,7 @@ enum FrameTypeId {
     FRAME_CPP          = 4,
     FRAME_KERNEL       = 5,
     FRAME_C1_COMPILED  = 6,
+    FRAME_UNSET        = 7,
```
apangin: No, don't add "undefined" constant with some random value.
`-1` would make sense, but I'd just leave it without a new constant.

src/lookup.h
fandreuz: I thought about this, but I couldn't determine from the docs whether `GetLineNumberTable` guarantees anything about `line_number_table_size` and `line_number_table` after an error. Is it possible they contain rubbish?

src/flightRecorder.cpp
fandreuz: Check what I'm doing [here](https://github.com/async-profiler/async-profiler/pull/1378/files#diff-73c664ab2d060d93784da8e82dba4aa7da5a38cd21c22d7b11f1671e741f9517R90) vs what was done [before](https://github.com/async-profiler/async-profiler/pull/1378/files#diff-1af8dc685ba7f64f7123d4fd2e8e16262cf7a59c0a43ab95bf83fb7d77475b04L144). 

### `_class` before
- class information for Java methods
- file information for native methods 

### `_class` after
- `_class` holds class information for Java methods
- unset for native methods (thus we need `_class != 0 ? _class : _file` to keep the previous JFR behavior)

And `_file` holds file information for both Java (`GetSourceFileName`) and native methods.

src/converter/one/jfr/Dictionary.java
fandreuz: Missed that, thanks: b21e396572ef19d5e70856d4b110aee10e589835

src/lookup.h
fandreuz: Missed the enum: e0f06ce1013f0d1e4848fb402d56789f6bcc6509

src/vmEntry.h
fandreuz: ef8735eee1868677a2cec16eabc82f56e0a8a510

fandreuz:
@apangin do you have any feedback on the open threads?

src/flightRecorder.cpp
apangin: `_file` became particularly overloaded, and this is what confuses me.
It can be source file name (Function.filename_strindex) for Java classes, or it can be object file name (Mapping.filename_strindex) for native functions. It can be also used as a class name in JFR output. BTW, where do you plan to store source file name for native functions? (if anywhere at all)

src/lookup.h
apangin: Yeah, spec is not clear about that. However, any reasonable implementation (HotSpot for sure) does not modify output arguments in case of error. OK, I'll leave it up to you, either way will do.

src/converter/one/jfr/Dictionary.java
```diff
@@ -36,10 +36,21 @@ public int size() {
...
+    // Thus, we map key=key+1, so the user can still use key=0.
+    private static long remapKey(long key) {
+        if (key < 0) {
+            throw new IllegalArgumentException("Negative keys not allowed");
         }
+        return key + 1;
+    }
+
+    private static long remapKeyBack(long key) {
+        return key - 1;
```
apangin: I understand why you did this, but it hurts me to see a whole method for doing `- 1`

fandreuz: Yeah I understand, I think we're doing something meaningful and I'd rather have an additional function than copy-pasting the same comment wherever this operation is used, to explain what is going on. It's only one place tough, let me know if you have a strong opinion about this.

src/flightRecorder.cpp
fandreuz: >  BTW, where do you plan to store source file name for native functions? (if anywhere at all)

Do we have this information for native function? I mean, isn't it the name of the shared object file, which would go into `Mapping.filename_strindex`?

If we do, there's [`Function.filename_strindex`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L501)

apangin: That's what I said above: sometimes `_file` denotes `Function.filename_strindex` but sometimes it denotes `Mapping.filename_strindex`, and I don't like this ambiguity. It is possible to get source file name for native code too, but I don't think it is worth doing so, even though for native code it is far more meaningful than for Java classes (meaning that I don't find `GetSourceFileName` useful for Java classes at all).

src/converter/one/jfr/Dictionary.java
apangin: Yes, it's just one place, and I prefer inlining `key - 1`.

Alternatively, if you are concerned about finding all places where `1` is added/subtracted, you can replace `1` with `BASE_ID`, but that's not necessary, IMO.

src/flightRecorder.cpp
fandreuz: Got it, fixed in bb092efe8655a2a4e69dfc9d0ac9c60fe8f4113e

I reverted `_class` to the previous semantic. I guess if we need any more fields I can address it in a later PR for OTLP stuff.

src/flightRecorder.cpp
```diff
@@ -1005,7 +784,9 @@ class Recording {
 
         buf->putVar32(11);
 
-        Lookup lookup(&_method_map, Profiler::instance()->classMap());
+        Index packages;
+        Index symbols;
+        Lookup lookup(&_method_map, Profiler::instance()->classMap(), &packages, &symbols, Output::OUTPUT_JFR);
```
apangin: Redundant qualifier - `Output::`

fandreuz: a448b9cbf40fe74cd24d825a51215f771781ac43

apangin: There are more places

fandreuz: 6a0b82de50ca4eafbb58594d98421a52aed4ec73

src/flightRecorder.cpp
```diff
@@ -1005,7 +784,9 @@ class Recording {
 
         buf->putVar32(11);
 
-        Lookup lookup(&_method_map, Profiler::instance()->classMap());
+        Index packages;
+        Index symbols;
```
apangin, line 788: After this change, async-profiler will produce JFRs with `0` mapped to the empty string.
Previously, symbol dictionary did not contain `0`, and `0` denoted a missing symbol.
There can be unobvious compatibility issues with that. Have you checked that FlameGraph and Heatmap converters work well with new recordings? Also, that JMC and IDEA can open and display new async-profiler recordings correctly?

fandreuz: - [x] FlameGraph
- [x] Heatmap
- [x] JMC
- [x] IDEA

fandreuz: I checked all four, I compared two JFRs from the same app and everything looks fine

src/lookup.h
```diff
@@ -0,0 +1,244 @@
...
+        size_t bytes = 0;
+        for (const_iterator it = begin(); it != end(); ++it) {
+            bytes += sizeof(jmethodID) + sizeof(MethodInfo);
+            bytes += it->second._line_number_table_size * sizeof(jvmtiLineNumberEntry);
+        }
+        return bytes;
+    }
+};
+
+class Lookup {
```
apangin, line 41: Header file is not the best place to hold non-trivial implementation, especially since you'll be using Lookup in more than one place. Please leave .h for interface only and move the implementation to .cpp.

fandreuz: 1cb7ab56c23550e038c28c6fa2b1511dd74759b9

src/lookup.h
```diff
@@ -0,0 +1,77 @@
...
+    u32 _sig = 0;
+    jint _modifiers = 0;
+    jint _line_number_table_size = 0;
+    jvmtiLineNumberEntry* _line_number_table = nullptr;
+    FrameTypeId _type;
+
+    jint getLineNumber(jint bci);
+};
+
+class MethodMap : public std::map<jmethodID, MethodInfo> {
```
apangin: Maybe unordered_map?

src/lookup.h
```diff
@@ -0,0 +1,78 @@
...
+#ifndef _LOOKUP_H
+#define _LOOKUP_H
+
+#include <assert.h>
+#include <jvmti.h>
+#include "arguments.h"
+#include "dictionary.h"
+#include "index.h"
+#include "vmEntry.h"
+#include "vmStructs.h"
```
apangin: I don't see usages of vmStructs here, are there?

src/lookup.h
```diff
@@ -0,0 +1,78 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _LOOKUP_H
+#define _LOOKUP_H
+
+#include <assert.h>
+#include <jvmti.h>
+#include "arguments.h"
+#include "dictionary.h"
```
apangin: nit: Dictionary class does not seem to be used here other than for `Dictionary*` pointer.
I'd just add forward declaration `class Dictionary;` instead of including full header file.

fandreuz: Sure: 0e6dc208d78ac9f930de7467974bd6130ed41759

src/lookup.h
fandreuz: Right, can be moved to cpp: c7abb85eb9292dd93c07f03a0f26bceec6157103

src/lookup.h
fandreuz: Doesn't seem like it's used as an ordered map indeed: c7abb85eb9292dd93c07f03a0f26bceec6157103

apangin:
Looks good now, thanks! I'll merge it once CI completes.

src/lookup.h
```diff
@@ -0,0 +1,78 @@
...
+    MethodInfo* resolveMethod(ASGCT_CallFrame& frame);
+    u32 getPackage(const char* class_name);
+
+    u32 getSymbol(const char* name) {
+        return _symbols->indexOf(name);
+    }
+
+    u32 getSymbol(const std::string& name) {
+        return _symbols->indexOf(name);
+    }
```
apangin: Sorry, one more thought: `_symbols` is already public, e.g. we access it like `lookup->_symbols->size()`. What if we remove `getSymbols` methods completely? We'll be able to call whatever overloads of `indexOf` we need directly. Consequently, `lookup.h` will not need to depend on `index.h`.

fandreuz: Sure: 8dc7741f6dc14332c932e262d1e0eca253dd6902

Issue #1343, opened 2025-06-16
`JfrTests.parseMultiModeRecording` fails sometimes on macOS


fandreuz:
### Describe the bug

https://github.com/async-profiler/async-profiler/actions/runs/15679466661/job/44167329739

```
java.lang.AssertionError: Expected 17.0 > 50.0
	>  test/test/jfr/JfrTests.java:78
	>  Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 50);
	at one.profiler.test.Assert.assertComparison(Assert.java:39)
	at one.profiler.test.Assert.isGreater(Assert.java:60)
	at test.jfr.JfrTests.parseMultiModeRecording(JfrTests.java:78)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at one.profiler.test.Runner.run(Runner.java:124)
	at one.profiler.test.Runner.main(Runner.java:184)
```

Baraa-Hasheesh:
Fixed here #1392

Baraa-Hasheesh:
Some failures are happening on the new check

Baraa-Hasheesh:
Should stop failing after #1428

fandreuz:
@Baraa-Hasheesh 

https://github.com/async-profiler/async-profiler/actions/runs/18401481701/job/52431659178?pr=1529#step:11:181
```
java.lang.AssertionError: Expected 0.78797942173913 > 0.8
	>  test/test/jfr/JfrTests.java:95
	>  Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.80);
	at one.profiler.test.Assert.assertComparison(Assert.java:39)
	at one.profiler.test.Assert.isGreater(Assert.java:60)
	at test.jfr.JfrTests.parseMultiModeRecording(JfrTests.java:95)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

PR #1383, opened 2025-07-10
Fix threading flaky tests

Baraa-Hasheesh:
### Description
Fixes flakiness of some tests in async-profiler 

[datagramSocketLock](https://github.com/async-profiler/async-profiler/actions/runs/16151504646/job/45583762864)
Tests checks that CPU samples are taken in locks, however if not enough contention is created that would fail 
The idea is to allow as many threads as possible to run to create more contention on the resources

[parseMultiModeRecording](https://github.com/async-profiler/async-profiler/actions/runs/16167564928/job/45633054150)
Tests checks that a certain amount of contention happens on a synchronous lock 
The idea is to allow as many threads as possible to run to create more contention on the resources

[regularPeek](https://github.com/async-profiler/async-profiler/actions/runs/16144510319/job/45559845528)
Tests that certain events/samples are captured within certain time periods 
From what I could tell the executor.scheduleAtFixedRate starts the execution of the thread without waiting the full 4 seconds 
The idea of the fix to replace that logic with a simple `sleep` & while loop

### Related issues
#1343

### Motivation and context
reduce random failures on GHA

### How has this been tested?
`make test`

---

### DIFF
--- a/test/test/jfr/JfrMultiModeProfiling.java
+++ b/test/test/jfr/JfrMultiModeProfiling.java
@@ -21,7 +21,7 @@ public class JfrMultiModeProfiling {
     private static int count = 0;
 
     public static void main(String[] args) throws InterruptedException {
-        ExecutorService executor = Executors.newFixedThreadPool(2);
+        ExecutorService executor = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
         for (int i = 0; i < 10; i++) {
             executor.submit(JfrMultiModeProfiling::cpuIntensiveIncrement);
         }
--- a/test/test/lock/DatagramTest.java
+++ b/test/test/lock/DatagramTest.java
@@ -50,7 +50,7 @@ public class DatagramTest {
         ch.bind(new InetSocketAddress(0));
         ch.configureBlocking(false);
 
-        Executor pool = Executors.newCachedThreadPool();
+        Executor pool = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
         for (int i = 0; i < 10; i++) {
             pool.execute(DatagramTest::sendLoop);
         }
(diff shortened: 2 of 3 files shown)


test/test/lock/DatagramTest.java
```diff
@@ -50,7 +50,7 @@ public static void main(String[] args) throws Exception {
         ch.bind(new InetSocketAddress(0));
         ch.configureBlocking(false);
 
-        Executor pool = Executors.newCachedThreadPool();
+        Executor pool = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
```
fandreuz, line 53: How does this fix the issue? Cached thread pool is not going to reuse any of the threads spawned to run `sendLoop`, the loop is too fast.

test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -21,7 +21,7 @@ public class JfrMultiModeProfiling {
     private static int count = 0;
 
     public static void main(String[] args) throws InterruptedException {
-        ExecutorService executor = Executors.newFixedThreadPool(2);
+        ExecutorService executor = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
```
fandreuz, line 24: Same as above

test/test/lock/DatagramTest.java
apangin: +1 This is not a fix.
Actual problem with this test is not about contention.

apangin:
> executor.scheduleAtFixedRate starts the execution of the thread without waiting the full 4 seconds

What do you mean by that? The test sets `initialDelay` to 4 seconds. If the code could run earlier, that would be a terrible JDK bug.

Baraa-Hasheesh:
> > executor.scheduleAtFixedRate starts the execution of the thread without waiting the full 4 seconds
> 
> What do you mean by that? The test sets `initialDelay` to 4 seconds. If the code could run earlier, that would be a terrible JDK bug.

I was reviewing the logs across multiple runs & realized I came to the wrong conclusion 
I will close this PR & investigate some anomalies I've observed that might be the reason for the common random failures on macOS

I'm noticing that sometimes `asprof` takes multiple seconds to attach it self to the java process which causes failures for time sensitive tests such as `ttsp` & `regularPeek`

PR #1384, opened 2025-07-10
Separate workflow for automated clang-tidy review

fandreuz:
### Description
In this PR, I introduce a separate job to create PR comments based on the output from clang-tidy. This new workflow is triggered via `workflow_run`, which runs with a `GITHUB_TOKEN` that can receive `write` permissions.

This is an adapted version of the example [here](https://github.com/platisd/clang-tidy-pr-comments?tab=readme-ov-file#using-this-action-to-safely-perform-analysis-of-pull-requests-from-forks).

### Related issues
#1382 

### Motivation and context
We need to re-enable automated clang-tidy review comment generation to profit from the common clang-tidy configuration.

### How has this been tested?
fandreuz/async-profiler#10

---


.github/workflows/clang-tidy-review.yml
```diff
@@ -0,0 +1,48 @@
...
+        name: code-check-artifacts
+        path: /tmp/code-check-artifacts/
+    - name: Read PR information
+      id: pr_info
+      run: |
+        cd /tmp/code-check-artifacts
+        echo "pr_id=$(cat pr-id.txt)" >> "$GITHUB_OUTPUT"
+        echo "pr_head_repo=$(cat pr-head-repo.txt)" >> "$GITHUB_OUTPUT"
+        echo "pr_head_sha=$(cat pr-head-sha.txt)" >> "$GITHUB_OUTPUT"
+    - uses: actions/checkout@v4
```
visheshruparelia, line 35: Worth checking this: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ 

>trigger workflow_run was introduced to enable scenarios that require building the untrusted code and also need write permissions to update the PR with e.g. code coverage results or other test results. To do this in a secure manner, the untrusted code must be handled via the pull_request trigger so that it is isolated in an unprivileged environment. The workflow processing the PR should then store any results like code coverage or failed/passed tests in artifacts and exit. The following workflow then starts on workflow_run where it is granted write permission to the target repository and access to repository secrets, so that it can download the artifacts and make any necessary modifications to the repository or interact with third party services that require repository secrets (e.g. API tokens).

fandreuz: Thank you for the reading, could you point if/how the approach I took in this PR differs from what is recommended in the article? 

What I'm doing:
- The user code is checked out in the workflow`code-check`
- The trigger is `pull_request`, thus `GITHUB_TOKEN` has no `write` permissions (enforced by AWS OSPO)
- clang-tidy runs on the diff
- clang-tidy output is stored as an artifact
- `clang-tidy-review` (triggered by `workflow_run`) obtains the artifact generated by `code-check`
- `clang-tidy-review` generates review comments

It seems to me that the recommendations of the article are being followed here, let me know if I'm missing something.

Also, workflows triggered by `workflow_run` always take the workflow definition from `master` ([ref](https://github.com/orgs/community/discussions/25288)). Thus, there's no way a malicious user could get to run code with `write` permissions via `GITHUB_TOKEN`.

visheshruparelia: I agree that we are not doing anything risky here after checking out the code since we are not building it. 
The article says that we should be "using" the artifacts (clang tidy output in this case) computed in the previous `pull_request` trigger and avoid `checkout` altogether in the `workflow_run`. I am wondering if it is possible to revamp this workflow in a way where we don't use `checkout`. 

Again, having said that, I don't necessarily see this as a blocker but something we should be conscious of while merging this PR.

fandreuz: >  I am wondering if it is possible to revamp this workflow in a way where we don't use checkout.

Unfortunately, that's not possible for now. The action we use to publish comments needs access to files in the PR to match clang-tidy warnings with what has already been published for housekeeping purposes, e.g. to skip comments which have been published already. 

That's a limitation of the action we use. We could open an issue there, but that shouldn't block this PR. There's no way to automatically execute code when a branch is checked out, and we vet whatever goes into the workflow triggered by `workflow_run`.

visheshruparelia: Yeah not a blocker IMO.

apangin (approved):

visheshruparelia (approved):

PR #1386, opened 2025-07-11
Refactor Preload mode to use patching

### LINKED ISSUE #1273: recursive dlopen calls when Profiler is loaded via LD_PRELOAD

### LINKED ISSUE #1304: Multiple Async-Profilers conflict with one another

Baraa-Hasheesh:
### Description
Refactor async-profiler preload mode to use patching rather than using LD_PRELOAD hooks,

This change was introduced for multiple reasons:

Simplify the flow for preload by re-using the C-API defined hooks
LD_PRELOAD currently isn't functional on macOS & while it's possible to fix them it will require maintaining two sets of preload functions (macOS vs LINUX) due to the differences in LD_PRELOAD hooks between macOS & Linux
LD_PRELOAD hooks can be dangerous when accounting for async-profiler behaviour of changing GOT entry values.
This comes into the picture when accounting for multiple profilers, one in PRELOAD mode (LD_PRELOAD hooks) while the other in API mode (changing GOT entries to install hooks).

The profiler needs to find a safe reference to the original method to call, this operation isn't straight forward on macOS as due to how interposing works in it.

Using dlsym on macOS won't find a safe reference due to the fact that the dynamic linker will change all references to point to the interposing function leaving only the preloaded lib unchanged. This in turn means that the preloaded profiler needs to use the value inside GOT to call the original functions.

Initially this is straight forward as you can call the function directly, however if different profiler starts in the process, it might patch that reference resulting in potential infinite loops.

For linux, similar issues might happen when using dlsym, as it might up resolving a stub which might not be safe to call after patching.

It's technically possible to account for these behaviors but I recommend against it as it will result in unneeded complexity & would require large refactors in the profiler.

### Related issues
will fix https://github.com/async-profiler/async-profiler/issues/1304 for Native applications (user can now use multiple async-profilers in a C/Rust application)
Will fix https://github.com/async-profiler/async-profiler/issues/1273 as dlopen is no longer preloaded

### Motivation and context
Fix preload mode on macOS & remove dangerous parts of the code by re-using the C-API approach for LD_PRELOAD mode

### How has this been tested?
many tests were enabled on macOS
New tests added

`make test`

---


Makefile
```diff
@@ -92,7 +92,7 @@ ifeq ($(OS),Darwin)
     MERGE=false
   endif
 else
-  CXXFLAGS += -U_FORTIFY_SOURCE -Wl,-z,defs -Wl,--exclude-libs,ALL -static-libstdc++ -static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections -ggdb
+  CXXFLAGS += -Wl,-z,now -U_FORTIFY_SOURCE -Wl,-z,defs -Wl,--exclude-libs,ALL -static-libstdc++ -static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections -ggdb
```
apangin: `resolveMallocSymbols` can be simplified after this change?

src/wallClock.h
```diff
@@ -32,7 +32,7 @@ class WallClock : public Engine {
 
     static void* threadEntry(void* wall_clock) {
         ((WallClock*)wall_clock)->timerLoop();
-        return NULL;
+        pthread_exit(NULL);
```
apangin: `pthread_exit` is declared in `pthread.h`, but I don't see that `wallClock.h` includes `pthread.h` even transitively.

src/profiler.h
```diff
@@ -143,7 +143,7 @@ class Profiler {
 
     static void* pthreadTimerEntry(void* arg) {
         instance()->timerLoop(arg);
-        return NULL;
+        pthread_exit(NULL);
```
apangin: It may be unobvious why `pthread_exit` is used when `return NULL` is sufficient.
Can we ensure the symbol is referenced without actually calling it?

test/native/libs/malloc.c
```diff
@@ -21,3 +23,24 @@ void* malloc(size_t size) {
...
+__attribute__((visibility("default")))
+void* mac_malloc(size_t size) {
+    void* ptr = _orig_malloc(size);
+    memset(ptr, 0xff, size);
+    return ptr;
+}
+
+__attribute__((used)) static struct {
+    const void *replacement;
+    const void *original;
```
apangin: `void*_` here and in all other places

test/one/profiler/test/TestProcess.java
```diff
@@ -123,6 +126,13 @@ public String testLibPath() {
         return "build/test/lib";
     }
 
+    public String preload() {
+        if (currentOs().equals(Os.MACOS)) {
+            return "DYLD_INSERT_LIBRARIES";
+        }
+        return "LD_PRELOAD";
+    }
```
apangin: `%preload=%lib` looks clumsy.
If we do substitution anyway, can we just leave `LD_PRELOAD` in the code as it is, but replace it with `DYLD_INSERT_LIBRARIES` in `TestProcess` constructor?
We can probably do the same with .so -> .dylib conversion, but I'm not sure if it is better than `%ext`.

Makefile
```diff
@@ -261,6 +262,8 @@ test-cpp: build-test-cpp
 
 test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
+	# generate copies of libasyncProfiler for testing multiple profilers
+	cp -f build/$(LIB_PROFILER) build/test/lib/libasyncProfiler.$(SOEXT)
```
apangin: Some other test already copies `libasyncProfiler` manually. This should be unified.

src/wallClock.h
Baraa-Hasheesh: it's included  here https://github.com/async-profiler/async-profiler/blob/master/src/wallClock.h#L11

src/profiler.h
Baraa-Hasheesh: I will check if I can inject it via compiler options

Makefile
Baraa-Hasheesh: I will check if I can inject references via compiler options

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: That's a nice suggestion, I will do it as you mentioned

Makefile
apangin: Maybe you can take an address of the function in the code to make sure it is referenced?

Baraa-Hasheesh: Sadly I would recommend keeping it as is, however I will refactor the code to relocate the function to a different location (I will reuse it to guarantee the pthread_exit symbol in a safe manner)

While I did find some simpler ways to guarantee the symbol in the generated so file, if we don't use the symbol explicitly somewhere in the code we might end up not generating some needed stubs 

This in turn might cause some undesired infinite loops (I've observed this in pthread_exit when I tried to remove the explicit call to it),
So I would recommend keeping explicit calls, at least once for symbols we want

src/profiler.h
Baraa-Hasheesh: I will remove it from here & guarantee it in a different way

Makefile
Baraa-Hasheesh: handled here [31de82f](https://github.com/async-profiler/async-profiler/pull/1386/commits/31de82fc5c7d7a6666889d70c715b412a0a2b369)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: handled here [31de82f](https://github.com/async-profiler/async-profiler/pull/1386/commits/31de82fc5c7d7a6666889d70c715b412a0a2b369)

test/native/libs/malloc.c
Baraa-Hasheesh: handled here [31de82f](https://github.com/async-profiler/async-profiler/pull/1386/commits/31de82fc5c7d7a6666889d70c715b412a0a2b369)

src/profiler.h
Baraa-Hasheesh: updated here [31de82f](https://github.com/async-profiler/async-profiler/pull/1386/commits/31de82fc5c7d7a6666889d70c715b412a0a2b369) 

Now pthread_exit is called in a more obvious location

Makefile
krk: Why do we need `resolveMallocSymbols`, if we use `-Wl,-z,now`?

Baraa-Hasheesh: it's needed for 2 reasons 
* I couldn't find a way to inject symbols into the generated shared object without explicitly referencing them somewhere in the code 
* Not calling the symbol explicitly somewhere in the code might cause the linker to provide a reference that points to a stub that might be patched later on (I noticed this happening in some of my testing on pthread_exit)
==> you can't do `void* ptr = malloc` & you need to do `void* ptr = malloc(1)` to guarantee it

Now as long the function isn't optimized out you can avoid actually executing it,
So theoretically you can have `resolveMallocSymbols` as a place holder to keep the explicit calls but never execute it

krk: > I noticed this happening in some of my testing on pthread_exit

To clarify, this happens with `-Wl,-z,now`?

Baraa-Hasheesh: Yes, but to clarify are 3 different unique issues here that we need to consider:
1. guaranteeing that symbol exists in GOT (shared object has it)
2. making sure GOT has a meaningful value when we read it (you can execute it safely)
3. making sure that the GOT value is actually a safe reference (it won't be patched by a different profiler)

`-Wl,-z,now` only addresses point 2.
without `-Wl,-z,now`, randomly reading GOT & calling the function using the value read from it might cause crashes

problem 1 is solved by referencing the symbol somewhere in the code.
For example you can do the following to make sure that both malloc & calloc exist in GOT for the shared object
```
static volatile u64 ptr = (u64)malloc | (u64)calloc;
return ptr;
```

problem 3 can be solved by actually calling the symbol somewhere in the code even if it was never executed (assuming `-Wl,-z,now` is used).
For example you can do the following to make sure that "pthread_exit" achieves that
```
static volatile void* ptr = malloc(1);
if (ptr == NULL) {
      pthread_exit(NULL);
}
```

The mentioned looping issue is related to problem 3 which isn't not linked to what `-Wl,-z,now` is trying to solve
=> Two different problems

------
A little bit of a history for context

currently async-profiler uses `dlsym` to find a "safe" reference to call the original method 

However `dlsym` can result with an incorrect reference depending on the order of library loading & the libraries being loaded as discussed here https://github.com/async-profiler/async-profiler/pull/1323#discussion_r2118327999 
=> we can't control how the user loads his libs or what symbols they export.

I couldn't find a "standard method/API" to find this safe reference as all methods/APIs I could think of hit the same limitations of load order.

Using the linker fully as done here #1269 would result in infinite recursive loops in case the profiler wants to profile it self or if multiple profilers are used on the same process

So a workaround was introduced to read this "safe" reference from GOT as provided from the linker.

krk: Do you have a clear minimal repro that crashes or has incorrect behavior, with `-Wl,-z,now`?

Baraa-Hasheesh: it's fixed in the current code version/state.

But if you need it, I can create a branch that would have that issue 
The new test `test/test/c/multiple_profilers.c` was what caught the issue for me, I think I can further simplify it if required

krk: Not sure I get it. When I remove `-Wl,-z,now`, `test/test/c/multiple_profilers.c` is still passing.

Baraa-Hasheesh: You should try running on ARM, You should see the crash on ARM, x86 usually works fine

krk: Thanks for mentioning. On arm, PR crashes while doing fixups. Intel has no problem so far.

```
#0  _dl_fixup (l=0xfffff7ff8000, reloc_arg=403623933) at dl-runtime.c:75
#1  0x0000fffff7fd1bf4 in _dl_runtime_resolve () at ../sysdeps/aarch64/dl-trampoline.S:98
#2  0x0000fffff7f2f2b4 in pthread_create_hook (thread=0xffffffffeea0, attr=0x0,
    start_routine=0xfffff7f685c0 <Profiler::pthreadTimerEntry(void*)>, arg=0x80000001) at src/hooks.cpp:74
#3  0x0000fffff7f63b98 in Profiler::startTimer (this=0x439340) at src/profiler.cpp:1840
#4  Profiler::start (this=this@entry=0x439340, args=..., reset=<optimized out>) at src/profiler.cpp:1250
#5  0x0000fffff7f66f9c in Profiler::runInternal (out=..., args=..., this=0x439340) at src/profiler.cpp:1907
#6  Profiler::run (this=0x439340, args=...) at src/profiler.cpp:1995
#7  0x0000fffff7f2dc48 in LateInitializer::startProfiler (this=<optimized out>, command=<optimized out>) at src/zInit.cpp:86
#8  LateInitializer::LateInitializer (this=<optimized out>) at src/zInit.cpp:56
#9  __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535) at src/zInit.cpp:92
#10 _GLOBAL__sub_I_ () at src/zInit.cpp:92
#11 0x0000fffff7fcc0b4 in call_init (l=<optimized out>, argc=argc@entry=2, argv=argv@entry=0xfffffffff0d8, env=env@entry=0xfffffffff0f0)
    at dl-init.c:72
#12 0x0000fffff7fcc1b0 in call_init (env=0xfffffffff0f0, argv=0xfffffffff0d8, argc=2, l=<optimized out>) at dl-init.c:119
#13 _dl_init (main_map=0xfffff7fff1a0, argc=2, argv=0xfffffffff0d8, env=0xfffffffff0f0) at dl-init.c:120
#14 0x0000fffff7fbf0c4 in _dl_start_user () from /lib/ld-linux-aarch64.so.1
```

src/zInit.cpp
```diff
@@ -5,10 +5,35 @@
...
+    free(p1);
+
+    sink = (intptr_t)p0 | (intptr_t)p1 | (intptr_t)p2 | (intptr_t)p3 | (intptr_t)p4;
+
+    // This condition will never be true, but the compiler won't optimize it out
+    // Here symbols can be called in a safe manner to make sure they're included in the GOT entries of the shared object
+    // Any newer symbols needed by the profiler can be called here
+    if (!sink) {
+        pthread_exit(NULL);
+    }
```
krk, line 35: Instead of the `sink`, if you declare the locals `volatile` optimizer would keep them.

src/hooks.cpp
```diff
@@ -91,80 +89,36 @@ static void pthread_exit_hook(void* retval) {
...
-        _orig_pthread_exit = ADDRESS_OF(pthread_exit);
-        _orig_dlopen = ADDRESS_OF(dlopen);
-        patchLibraries();
-    }
+    Profiler::instance()->updateSymbols(false);
+
+    CodeCache* lib = Profiler::instance()->findLibraryByAddress((void*)Hooks::init);
+    _orig_pthread_create = SAVE_IMPORT(pthread_create);
+    _orig_pthread_exit = SAVE_IMPORT(pthread_exit);
+    _orig_dlopen = SAVE_IMPORT(dlopen);
```
krk: `SAVE_IMPORT` already contains `_orig_##FUNC =`.

```suggestion
    SAVE_IMPORT(pthread_create);
    SAVE_IMPORT(pthread_exit);
    SAVE_IMPORT(dlopen);
```

src/zInit.cpp
```diff
@@ -5,10 +5,35 @@
 
 #include <dlfcn.h>
 #include <stdlib.h>
+#include <pthread.h>
 #include "hooks.h"
 #include "profiler.h"
 #include "vmStructs.h"
 
+// Reference intercepted functions at least once to ensure
+// GOT entries are present in the generated shared library
```
krk: Instead of just being referenced, they are invoked.

```suggestion
// Invoke functions we may intercept so they GOT entries are resolved
```

src/zInit.cpp
```diff
@@ -5,10 +5,35 @@
...
+static void resolveSymbols() {
```
krk:
```suggestion
static void resolveGotEntries {
```

test/native/libs/malloc.c
```diff
@@ -21,3 +23,24 @@ void* malloc(size_t size) {
...
+
+typedef void* (*malloc_t)(size_t);
+static malloc_t _orig_malloc = malloc;
+
+__attribute__((visibility("default")))
+void* mac_malloc(size_t size) {
+    void* ptr = _orig_malloc(size);
+    memset(ptr, 0xff, size);
+    return ptr;
+}
```
krk, line 37: This seems duplicated, let's merge with `malloc` above and ifdef only the `dlsym` call?

test/one/profiler/test/TestProcess.java
```diff
@@ -86,10 +86,11 @@ public TestProcess(Test test, Os currentOs, String logDir) throws Exception {
         for (String env : test.env()) {
             String[] keyValue = env.split("=", 2);
             if (keyValue.length == 2) {
-                pb.environment().put(keyValue[0], substituteFiles(keyValue[1]));
+                pb.environment().put(substituteFiles(substitutePlatform(keyValue[0])), substituteFiles(keyValue[1]));
             }
         }
         pb.environment().put("TEST_JAVA_HOME", JAVA_HOME);
+        pb.environment().put("DYLD_FORCE_FLAT_NAMESPACE", "1");
```
krk: Does macOS set this env var by default?

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,116 @@
...
+
+void executeAsyncProfilerCommand(char* cmd) {
+    asprof_error_t asprof_err = _asprof_execute(cmd, NULL);
+    if (asprof_err != NULL) {
+        fprintf(stderr, "%s\n", _asprof_error_str(asprof_err));
+        exit(1);
+    }
+}
+
+void initAsyncProfiler() {
```
krk: Filename is `multiple_profilers.c` but a single profiler is started here using the C API. Let's rename the file to match intent.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,116 @@
...
+asprof_error_str_t _asprof_error_str;
+asprof_execute_t _asprof_execute;
+asprof_init_t _asprof_init;
+
+typedef void (*pthread_exit_t)(void*);
+typedef void* (*pthread_entry_t)(void*);
+typedef int (*pthread_create_t)(pthread_t*, pthread_attr_t*, pthread_entry_t, void*);
+
+pthread_exit_t pthread_exit_ref = (pthread_exit_t)pthread_exit;
+pthread_create_t pthread_create_ref = (pthread_create_t)pthread_create;
```
krk, line 35: Why are these necessary? Let's add a short comment or remove them.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,116 @@
...
+    void* libprof = openLib("build/test/lib/libasyncProfiler" LIB_EXT);
+
+    _asprof_init = (asprof_init_t)getSymbol(libprof, "asprof_init");
+    _asprof_init();
+
+    _asprof_execute = (asprof_execute_t)getSymbol(libprof, "asprof_execute");
+    _asprof_error_str = (asprof_error_str_t)getSymbol(libprof, "asprof_error_str");
+}
+
+void sampleMalloc() {
```
krk: These are not doing sampling, let's rename to `doSleep`, `doMalloc` etc.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,116 @@
...
+    void* lib = openLib("build/test/lib/libcallsmalloc" LIB_EXT);
+
+    malloc_t call_malloc = (malloc_t)getSymbol(lib, "call_malloc");
+    free(call_malloc(1999993));
+}
+
+void sampleWall(unsigned long ms) {
+    struct timespec ts = {ms / 1000, (ms % 1000) * 1000000};
+    while (nanosleep(&ts, &ts) < 0 && errno == EINTR) ;
```
krk, line 97: Please add comment why this is needed: To sleep for the full duration even when interrupted.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void* threadEntry2(void* args) {
+    sampleMalloc();
+    sampleWall(100);
+    pthread_exit_ref(NULL);
+    return NULL;
+}
+
+int main(int argc, char** args) {
```
krk, line 114: This is too similar to the `native_api.c` test, can we reuse it instead?

Makefile
```diff
@@ -257,6 +258,8 @@ test-cpp: build-test-cpp
...
```
krk: Let's make it clear this is a copy and not to be loaded by anything other than the test that needs it.

```suggestion
	cp -f build/$(LIB_PROFILER) build/test/lib/libasyncProfiler-copy.$(SOEXT)
```

test/test/c/CTests.java
```diff
@@ -26,4 +26,16 @@ public void nativeApi(TestProcess p) throws Exception {
         File preloadFile = p.getFile("%preload_file");
         assert preloadFile == null || preloadFile.length() == 0;
     }
+
+    @Test(sh = "LD_PRELOAD=%lib %testbin/multiple_profilers %api_file.jfr", env = {"ASPROF_COMMAND=start,nativemem,file=%preload_file.jfr"})
+    public void twoProfilers(TestProcess p) throws Exception {
```
krk, line 42: Let's mention in a comment that one profiler is from LD_PRELOAD, other loaded by the test app.

krk, line 42: Let's add two more tests:

* both profilers are loaded via `dlopen`
* both via `LD_PRELOAD`

krk (changes requested):

src/zInit.cpp
Baraa-Hasheesh: Are you referring to individual pointers?

test/native/libs/malloc.c
Baraa-Hasheesh: Not really 
Check below I'm doing interposing to overwrite the malloc reference for the application being run

This interposing function can't be called malloc
Keeping them separate is cleaner & more readable

test/test/c/CTests.java
Baraa-Hasheesh: Let's not do 2"LD_PRELOAD", 
I can't think of a way to give each profiler a unique output file for that case 

& I don't think it's a use case anyone would ever use 

I will add the 2 dlopens

test/test/c/multiple_profilers.c
Baraa-Hasheesh: I would rather keep that test simple rather than complicate it with the details of having multiple profilers

test/test/c/multiple_profilers.c
Baraa-Hasheesh: I will keep it the same as I will add a 2 dlopen test as you suggested

test/test/c/CTests.java
Baraa-Hasheesh: handled here [ca9cff3](https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119) 

But there is a weird behavior on macOS for copied shared objects on native apps using dlopen,
It didn't look significant enough to chase down, so I didn't look in depth for an explanation 

I added a comment for the findings, if further investigation is required that can be done
=> Flow works perfectly fine on linux

Makefile
Baraa-Hasheesh: handled [ca9cff3](https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119)

test/test/c/multiple_profilers.c
Baraa-Hasheesh: added [ca9cff3](https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119)

src/zInit.cpp
Baraa-Hasheesh: handled [ca9cff3](https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: Variable is no longer needed & removed [ca9cff3](https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119)

test/test/c/multiple_profilers.c
Baraa-Hasheesh: added https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119

test/test/c/multiple_profilers.c
Baraa-Hasheesh: handled https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119

src/zInit.cpp
Baraa-Hasheesh: https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119

src/hooks.cpp
Baraa-Hasheesh: https://github.com/async-profiler/async-profiler/pull/1386/commits/ca9cff38de913ef0886826cccb629f8c9e9bc119

test/test/c/CTests.java
Baraa-Hasheesh: [7b9f98e](https://github.com/async-profiler/async-profiler/pull/1386/commits/7b9f98e82a3e844447e4e00d0e98a450e1064613)

test/one/profiler/test/TestProcess.java
krk: Why was it needed, why no longer needed?

src/hooks.cpp
```diff
@@ -91,80 +89,36 @@ static void pthread_exit_hook(void* retval) {
 typedef void* (*dlopen_t)(const char*, int);
 static dlopen_t _orig_dlopen = NULL;
 
-static void* dlopen_hook_impl(const char* filename, int flags, bool patch) {
+static void* dlopen_hook(const char* filename, int flags) {
     Log::debug("dlopen: %s", filename);
     void* result = _orig_dlopen(filename, flags);
-    if (result != NULL && filename != NULL) {
+    if (result != NULL && filename != NULL && (flags & RTLD_NOLOAD) == 0) {
```
krk, line 95: Do we have a test for `RTLD_NOLOAD`?

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: I think it was needed during early iterations of making preload hooks work that I was using interposing in which I gave up on due to having a large amount of complexity 

In this new approach I'm using patching which should works out of the box as long as the profiler is loaded into memory

src/profiler.cpp
```diff
@@ -1018,6 +1018,7 @@ Engine* Profiler::selectAllocEngine(long alloc_interval, bool live) {
     } else if (VM::isOpenJ9()) {
         return &j9_object_sampler;
     } else {
+        fprintf(stderr, "I'm ALIVE\n");
```
krk: debug remnant

test/one/profiler/test/TestProcess.java
```diff
@@ -179,10 +179,18 @@ private void addArgs(List<String> cmd, String args) {
         }
     }
 
+    private String substitutePlatform(String s) {
+        if (this.currentOs.equals(Os.MACOS)) {
+            s = s.replaceAll("LD_PRELOAD", "DYLD_INSERT_LIBRARIES");
```
krk, line 184: Safer to include the delimiter as well:

```suggestion
            s = s.replaceAll("LD_PRELOAD=", "DYLD_INSERT_LIBRARIES=");
```

test/one/profiler/test/TestProcess.java
```diff
@@ -86,7 +86,7 @@ public TestProcess(Test test, Os currentOs, String logDir) throws Exception {
         for (String env : test.env()) {
             String[] keyValue = env.split("=", 2);
             if (keyValue.length == 2) {
-                pb.environment().put(keyValue[0], substituteFiles(keyValue[1]));
+                pb.environment().put(substituteFiles(substitutePlatform(keyValue[0])), substituteFiles(keyValue[1]));
```
krk: `substitueFiles` calls `substituePlatform`, and we call both here. Can we simplify these substitutions?

test/test/c/CTests.java
```diff
@@ -26,4 +27,26 @@ public void nativeApi(TestProcess p) throws Exception {
...
+    // otool -D build/test/lib/libasyncProfiler-copy.dylib; otool -D build/lib/libasyncProfiler.dylib
+    // Preloading one of the shared objects bypasses this issue, AllocTests.objectSamplerWtihDifferentAsprofs isn't effected by this
+    @Test(sh = "LD_PRELOAD=%lib %testbin/multiple_profilers api %profiler_1.jfr %profiler_2.jfr", nameSuffix = "api", os = Os.MACOS)
+    @Test(sh = "%testbin/multiple_profilers api %profiler_1.jfr %profiler_2.jfr", nameSuffix = "api", os = Os.LINUX)
+    public void twoProfilers(TestProcess p) throws Exception {
+        p.waitForExit();
+        assert p.exitCode() == 0;
+
+        Output preloadProfiler = Output.convertJfrToCollapsed(p.getFilePath("%profiler_1"), "--nativemem");
+        Output apiProfiler = Output.convertJfrToCollapsed(p.getFilePath("%profiler_2"), "--wall");
```
krk: Let's test multiple combinations, e.g. both can be `nativemem` - these can be read from 

In a real scenario, a long running service may have async-profiler enabled from start with `nativemem`, and user may want to collect an ephemeral profile with perhaps different settings with `nativemem`, perhaps different interval bytes, for a short duration.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+
+asprof_error_str_t _asprof_error_str[2];
+asprof_execute_t _asprof_execute[2];
+asprof_init_t _asprof_init[2];
+
+typedef void (*pthread_exit_t)(void*);
+typedef void* (*pthread_entry_t)(void*);
+typedef int (*pthread_create_t)(pthread_t*, pthread_attr_t*, pthread_entry_t, void*);
+
+// This is to make sure invocations on references in .rela.dyn don't cause infinite loops
```
krk: Why does it cause infinite loops "in .rela.dyn"?

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+    if (asprof_err != NULL) {
+        fprintf(stderr, "%s\n", _asprof_error_str[profiler_id](asprof_err));
+        exit(1);
+    }
+}
+
+void initAsyncProfiler(int profiler_id) {
+    void* libprof;
+
+    if (profiler_id == 0) {
```
krk, line 66: Not that it matters for the test, but let's load the default name first.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+        libprof = openLib("build/test/lib/libasyncProfiler-copy" LIB_EXT);
+    } else {
+        libprof = openLib("build/lib/libasyncProfiler" LIB_EXT);
+    }
+
+    _asprof_init[profiler_id]= (asprof_init_t)getSymbol(libprof, "asprof_init");
```
krk: Spacing, you can run clang-format on new files with the profile from https://github.com/async-profiler/async-profiler/discussions/1227#discussioncomment-12840586

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+void* threadEntry1(void* args) {
+    doMalloc();
+    doSleep(100);
+    pthread_exit(NULL);
+}
+
+void* threadEntry2(void* args) {
+    doMalloc();
+    doSleep(1000);
+    pthread_exit_ref(NULL);
```
krk, line 110: Why is `ref` used here, why not `pthread_exit`?

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+    return NULL;
+}
+
+int main(int argc, char** args) {
+    if (argc < 3) {
+        fprintf(stderr, "Usage: %s api|preload <output_file_1> [output_file_2]\n", args[0]);
```
krk: Help message does not match expectations, e.g. `output_file_2` is not optional for `api`.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    initAsyncProfiler(0);
+
+    if (api_mode) {
+        initAsyncProfiler(1);
+    }
+
+    char start_cmd[2048] = {0};
+    snprintf(start_cmd, sizeof(start_cmd), "start,wall=10ms,file=%s", api_mode ? args[3] : args[2]);
```
krk: Check return value to see if the was truncated.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+        exit(1);
+    }
+
+    initAsyncProfiler(0);
+
+    if (api_mode) {
+        initAsyncProfiler(1);
+    }
+
+    char start_cmd[2048] = {0};
```
krk: Use `PATH_MAX` instead of `2048`.

test/test/c/multiple_profilers.c
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void* threadEntry2(void* args) {
+    doMalloc();
+    doSleep(1000);
+    pthread_exit_ref(NULL);
+    return NULL;
+}
+
+int main(int argc, char** args) {
```
krk, line 114: Can we simplify this by introducing `start_profiler(id)` method?

test/test/nativemem/NativememTests.java
```diff
@@ -172,31 +172,37 @@ public void ldpreload(TestProcess p) throws Exception {
...
 
-    @Test(sh = "LD_PRELOAD=%lib ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_plt_dyn", os = Os.LINUX)
+    @Test(sh = "LD_PRELOAD=%lib ASPROF_COMMAND=start,nativemem,file=%f.jfr %testbin/malloc_plt_dyn")
     public void malloc_plt_dyn(TestProcess p) throws Exception {
         Map<Long, Long> sizeCounts = assertNoLeaks(p);
 
         Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_SIZE, 0L), 1);
-        Assert.isEqual(sizeCounts.getOrDefault((long) MALLOC_DYN_SIZE, 0L), 1);
+
+        // macOS doesn't currently patch .rela.dyn
```
krk, line 181: Is this documented, will it be supported etc.?

krk (changes requested):

src/profiler.cpp
Baraa-Hasheesh: Yes :).

test/test/c/multiple_profilers.c
Baraa-Hasheesh: I don't see the value in checking that as it should not happen 

The only dynamic part is file which should be relatively short as it's a temp file

src/hooks.cpp
Baraa-Hasheesh: The profiler internally calls RTLD_NOLOAD

Double profilers will patch each other dlopen references 

This is basically an optimization to reduce the size of internal ping pong between the profilers when they parse/patch shared objects

test/test/c/multiple_profilers.c
Baraa-Hasheesh: I think a misunderstanding is happening here

.rela.dyn references are also patched in Linux 

This is to make sure that this patching is also happening safely 

I will update the comment to reflect that

test/test/c/multiple_profilers.c
Baraa-Hasheesh: This to make sure invocations happening through .rela.dyn references are also safe

test/test/nativemem/NativememTests.java
Baraa-Hasheesh: It was skipped for simplicity https://github.com/async-profiler/async-profiler/pull/1315#discussion_r2113564687

As if it's documented I don't believe it's documented clearly

test/test/c/CTests.java
Baraa-Hasheesh: I will try to find a way to do that without complicating the test

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: Invalid - won't change, `String[] keyValue = env.split("=", 2);`
It will not have always have the `=`

test/test/c/CTests.java
Baraa-Hasheesh: I will simply enable multiple options for each profiler as jfr supports that both profilers picked up different events 

Both profilers will have wall & native memory & I will check both for samples 

I will have different durations for each profiler

test/test/c/multiple_profilers.c
Baraa-Hasheesh: If a just for style I will keep it as the `libasyncProfiler` is loaded first by the LD_PRELOAD in one of the tests 

Changing the indexes would require having a more complex/more coding to handle always loading `libasyncProfiler` first which isn't really needed 

Let's keep it simple

test/test/c/multiple_profilers.c
Baraa-Hasheesh: will add documentation about this

test/test/c/multiple_profilers.c
Baraa-Hasheesh: Updated [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

test/test/c/multiple_profilers.c
Baraa-Hasheesh: The check & message was updated to better reflect the usage here [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

test/test/c/multiple_profilers.c
Baraa-Hasheesh: fixed in [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465) 

Regarding the style I will usually use Clion reformatter but I forgot to re-format while updating the test

test/test/c/multiple_profilers.c
Baraa-Hasheesh: Comment was update here [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

If still not clear I'm open for suggestions

test/test/c/CTests.java
Baraa-Hasheesh: Test adjusted to have both profilers collect wall & native memory with different intervals & results [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: was removed substitueFiles as it's not needed for a ENV key [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

src/profiler.cpp
Baraa-Hasheesh: removed [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

test/test/c/multiple_profilers.c
Baraa-Hasheesh: Suggestion was helpful & applied here [87f7dc3](https://github.com/async-profiler/async-profiler/pull/1386/commits/87f7dc393e5672f0fdd73a71fc9657969fd94465)

apangin:
As discussed offline, the risk-benefit tradeoff is not in favor of this change right now.
We may reconsider this later if we see demand for the preload mode.

Issue #1316, opened 2025-05-26
Per Thread Heatmap's flamegraph


franz1981:
### Describe the feature

Enables JFR heatmap converter to generate per-thread heatmap's flamegraph views i.e. the flamegraph view obtained by selecting heatmaps ranges shows frames in the flamegraph with a thread name as the root frame.

### Use Case

While working with normal flamegraphs (no heatmap) can be useful to have a per-thread view, to better understand how the different threads behave; this seems even more important as we need to observe changes over time for long/medium running profiling sessions e.g.  asymmetric workload distribution, spikes of created threads in elastic thread pools to sustain some increased bursty load, etc etc

I didn't thought yet how the diff option would behave with such feature in - but I suppose it will still work per-thread.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

franz1981:
> Do you mean just adding a synthetic thread frame on the generated flamegraphs while keeping the heatmap view as is

Yep

apangin:
@franz1981 Splitting heatmap's flamegraph by threads turned out to be not as trivial feature as it initially seemed to me. But thanks to @fandreuz, this has been now implemented. On a good side, while doing this task, we've refactored heatmap builder to make it simpler and easier to extend in the future if we need to add something similar.

The feature is activated by `-t`/`--threads` flag. You're welcome to give it a try and report back.

franz1981:
Thanks Andrei, will give it a shot soon ❤️ and report here 🙏 
At least I cannot feel too bad I wasn't able to achieve it in one weekend eheh

PR #1388, opened 2025-07-11
Simplify heatmap JFR converter

fandreuz:
### Description
In this PR I simplify the implementation of the JFR to heatmap converter.

Summary:
- splitted `Method.frequencyOrNewMethodId` in two fields, as it is its usage is confusing
- avoid sorting `EvaluationContext.orderedMethods` in-place in `renameMethodsByFrequency`
- removed `EvaluationContext.method`
- all methods in `StackStorage.index` now take two arrays `prefix` and `suffix`, which can be set to `null`

### Related issues
#1316 

### Motivation and context
I had some troubles adding the feature requested in #1316, some of the changes in this PR will help in this sense.

### How has this been tested?
Manual testing.

---

### DIFF
--- a/src/converter/one/heatmap/Heatmap.java
+++ b/src/converter/one/heatmap/Heatmap.java
@@ -221,16 +221,16 @@ public class Heatmap {
     }
 
     private void renameMethodsByFrequency(EvaluationContext context) {
-        Method[] descsByFrequency = context.orderedMethods.clone();
-        Arrays.sort(descsByFrequency, new Comparator<Method>() {
+        Method[] methodsByFrequency = context.orderedMethods.clone();
+        Arrays.sort(methodsByFrequency, new Comparator<Method>() {
             @Override
             public int compare(Method o1, Method o2) {
                 return Integer.compare(o2.frequency, o1.frequency);
             }
         });
 
-        for (int i = 0; i < descsByFrequency.length; i++) {
-            Method method = descsByFrequency[i];
+        for (int i = 0; i < methodsByFrequency.length; i++) {
+            Method method = methodsByFrequency[i];
             method.frequencyBasedId = i + 1; // zero is reserved for no method
         }
     }


src/converter/one/heatmap/MethodCache.java
```diff
@@ -31,84 +33,87 @@ public void clear() {
...
         }
 
-        Method last = null;
-        Method prototype = null;
-        while (method != null) {
-            if (method.originalMethodId == methodId) {
-                if (method.location == location && method.type == type && method.start == firstInStack) {
-                    return method.index;
+        FrameDesc prototype = null;
+        LinkedList<FrameDesc> list = getFarMethodList(methodId);
```
krk: `getFarMethodList` can be called once at the top.

src/converter/one/heatmap/MethodCache.java
```diff
@@ -6,20 +6,22 @@
...
     private final Index<String> symbolTable = new Index<>(String.class, "", 32768);
-    private final Index<Method> methodIndex = new Index<>(Method.class, new Method(symbolTable.index("all"), 0), 32768);
+    private final Index<FrameDesc> methodIndex = new Index<>(FrameDesc.class, new FrameDesc(symbolTable.index("all"), 0), 32768);
 
-    private final Method[] nearCache = new Method[256 * 256];
+    private final FrameDesc[] nearCache = new FrameDesc[256 * 256];
     // It should be better to create dictionary with linked methods instead of open addressed hash table
     // but in most cases all methods should fit nearCache, so less code is better
-    private final Dictionary<Method> farMethods = new Dictionary<>(1024);
+    private final Dictionary<LinkedList<FrameDesc>> farMethods = new Dictionary<>(1024);
```
krk: Dictionary of linked list, addressed by method id? What is the intention here? Why is this better than the hash table?

krk (changes requested):

apangin:
This does not look a simplification to me. At least, not all of it.
- While `Method` is probably not the very accurate name, renaming it to `FrameDesc` does not make it any better. To minimize changes, let's avoid renaming unless there is a clear benefit in doing so.
- `LinkedList` is almost never a good choice. ["I wrote it, and I never use it."](https://x.com/joshbloch/status/583813919019573248) - J.Bloch
- Adding _prefix_ and _suffix_ does not simplify `index`. What if we need multiple prefixes or suffixes?

fandreuz:
@apangin 

> While Method is probably not the very accurate name, renaming it to FrameDesc does not make it any better.

Isn't its content the description of a stack frame?

> LinkedList is almost never a good choice. ["I wrote it, and I never use it."](https://x.com/joshbloch/status/583813919019573248) - J.Bloch

Could you elaborate on this? A linked list was already there: `Method` contains a `next` field which is used to access the next node in the linked list. I think that's bad design, the object should not be aware of the container, do you agree? `LinkedList` is a good choice in this case because it gives us optimal complexity for what we're doing with the collection, but I'm fine with changing it to an `ArrayList`.

> Adding prefix and suffix does not simplify index. What if we need multiple prefixes or suffixes?

We can potentially call `index` multiple times nesting the results. That's not going to be efficient, but also I don't see why optimize something that we are not going to use now. Let me know if I'm missing something.

src/converter/one/heatmap/MethodCache.java
fandreuz: If you mean why `Dictionary<LinkedList>` instead of `HashMap<LinkedList>`, I'd be happy to replace all the occurrences of `Dictionary` with `HashMap`, I don't see other advantages than no boxing. Could you weigh in @apangin ?

If you mean why `LinkedList`, the linked list was already there, it was implemented with a field `FrameDesc.next`

krk: I meant the latter, thanks.

fandreuz:
@apangin I addressed all the comments you had during our last discussion. I finally decided to keep `Method.next` since I realized it's needed for other purposes. I tested the changes with multiple JFRs, the file produced is identical for `master` and `simplify-heatmap-converter`.

src/converter/one/heatmap/StackStorage.java
```diff
@@ -32,65 +32,31 @@ public int[] get(int id) {
...
-    public int index(int[] stack, int stackSize) {
-        int mask = meta.length - 1;
-        int hashCode = murmur(stack, stackSize);
-        int i = hashCode & mask;
-        while (true) {
-            long currentMeta = meta[i];
-            if (currentMeta == 0) {
-                break;
-            }
+    public int index(int[] prototype, int stackSize, int[] prefix, int[] suffix) {
```
apangin: Is it possible to get rid of prefix and suffix altogether and use a single array with all the frames?

fandreuz: We could do that, but that means we are going to make a full copy of the whole array (e.g. thread + stack trace) regardless of whether or not the combination is already known, just to pass it to the function. With the current implementation that's not going to happen, the full array is only constructed if the combination does not exist. 

Can we accept the additional overhead?

apangin: This is a good point. However, if we don't allocate an array for a stack trace every time but reuse the same pre-allocated array, we will only pay the price of a single copy. At the same time, we will eliminate allocations of suffix and prefix arrays, simplify murmur hash and equals, reduce number of dictionary lookups. Even if these changes don't fully compensate the overhead, I suppose performance penaltly will be small enough to justify nicer code.

I also have another idea that does not require copying. You can first use a streaming hash to compute fingerprint of the combined stack without creating prefix/suffix arrays. Then you instantiate a stack trace only when the fingerprint is seen for the first time. This is going to be a more extensive change though.

fandreuz: Check 0e1e5c7fe9fa7c30955beb6ae47aa60264368ad5, I took the first approach

> I also have another idea that does not require copying. You can first use a streaming hash to compute fingerprint of the combined stack without creating prefix/suffix arrays. Then you instantiate a stack trace only when the fingerprint is seen for the first time.

We would still need to pass in stack + prefix + suffix in some form, to allow `equals` to be called after an entry with matching hash is found. So we wouldn't be able to simplify that much with this approach.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -413,30 +409,28 @@ public void addEvent(int stackTraceId, int extra, byte type, long timeMs) {
...
-            }
-
-            int id = stackTracesCache.get((long) extra << 32 | stackTraceId, -1);
-            if (id != -1) {
-                sampleList.add(id, timeMs);
-                return;
-            }
 
             int prototypeId = stackTracesCache.get(stackTraceId);
             int[] prototype = stackTracesRemap.get(prototypeId);
```
apangin: Indirection via `prototypeId` was needed for the shortcut `long stackTraceId -> int compactStackId` when no extra frames were involved. You removed this fast path, so `prototypeId` is now nothing more than just a redundant indirection.

You can either return the shortcut back or get rid of `prototypeId`. In the latter case, you can still optimize away `arraycopy` when there are no extras.

fandreuz: Are you sure the indirection is redundant? `stackTracesRemap` contains both "natural" and artificial stack traces, so we cannot just use the original stacktrace ID as the key we pass to `sampleList.add`.

I made the flow simpler in b24a2285e86704d9974e3517c04fd16082e31fba and d8a176ef7dfdddc5053299ee11d6b624e2baa366

src/converter/one/heatmap/Heatmap.java
```diff
@@ -215,26 +215,24 @@ private void writeStartMethods(HtmlOut out, EvaluationContext context) {
...
             if (method.start) {
-                out.writeVar(method.frequencyOrNewMethodId);
+                out.writeVar(method.frequencyBasedId);
             }
         }
     }
 
     private void renameMethodsByFrequency(EvaluationContext context) {
-        Arrays.sort(context.orderedMethods, new Comparator<Method>() {
+        Method[] descsByFrequency = context.orderedMethods.clone();
```
apangin: `methodsByFrequency`?

fandreuz: 7382e92d52002876fb782791dc5d5923425b577d

Issue #1380, opened 2025-07-10
Test failure on Mac OS x86


parttimenerd:
### Describe the bug

Multiple tests related to leaf stack walking are failing

### Expected vs. actual behavior

```sh
FAIL [89/91] StackwalkerTests.normalStackVM (args: leafFrame) (agentArgs: start,event=cpu,cstack=vm,file=%f.jfr) took 9.470 s
java.lang.AssertionError
	at test.stackwalker.StackwalkerTests.normalStackVM(StackwalkerTests.java:85)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at one.profiler.test.Runner.run(Runner.java:124)
	at one.profiler.test.Runner.main(Runner.java:188)
```

The collected stack traces are:

```sh
thread_start;_pthread_start;thread_native_entry;Thread::call_run;JavaThread::thread_main_inner;MonitorDeflationThread::monitor_deflation_thread_entry;Monitor::wait_without_safepoint_check;PlatformMonitor::wait;__psynch_cvwait 1
thread_start;_pthread_start;thread_native_entry;Thread::call_run;JavaThread::thread_main_inner;CompileBroker::compiler_thread_loop;CompileBroker::invoke_compiler_on_method;Compiler::compile_method;Compilation::Compilation;Compilation::compile_method;Compilation::compile_java_method;Compilation::emit_code_body;LIR_Assembler::emit_code;LIR_Assembler::emit_lir_list;LIR_Assembler::emit_op2 1
thread_start;_pthread_start;ThreadJavaMain;JavaMain;JNI_CreateJavaVM;Threads::create_vm;JvmtiExport::post_vm_initialized;VM::VMInit;Profiler::run;Profiler::runInternal;Profiler::start;Profiler::startTimer;jvmti_RunAgentThread;JvmtiEnv::RunAgentThread;JvmtiAgentThread::JvmtiAgentThread;os::create_thread;_pthread_create 1
thread_start;_pthread_start;ThreadJavaMain;JavaMain;jni_DestroyJavaVM;attach_current_thread;os::create_attached_thread;PosixSignals::hotspot_sigmask;__pthread_sigmask 1
[...]

PR #1390, opened 2025-07-14
fix crashes when using instrument profiling

Baraa-Hasheesh:
### Description
From the looks of it for certain compilers an implicit casting was causing wrong logical & mathematical operations 
This in turn caused bad byte modification by the profiler for the Java class 

### Related issues
#1380

### Motivation and context
Fix bug in instrument profiling

### How has this been tested?
GHA added for macOs JDK 17 which was causing the issue to happen consistently 

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -123,18 +123,9 @@ jobs:
             # ARM MacOS should take fat binaries built on ARM
             asprof-binaries-job: macos
           - runson:
-              display: macos-x64
-              name: macos-13
-            java-version: 11
-            java-distribution: corretto
-            architecture: x64
-            image: ""
-            # x64 MacOS should take fat binaries built on ARM
-            asprof-binaries-job: macos
-          - runson:
               display: macos-arm64
               name: macos-15
-            java-version: 17
+            java-version: 21
             java-distribution: corretto
             # Not using container for mac-os as we have images only for linux
             image: ""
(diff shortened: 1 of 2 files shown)


src/instrument.cpp
```diff
@@ -356,7 +356,7 @@ void BytecodeRewriter::rewriteStackMapTable() {
         } else if (frame_type <= 254) {
             // append_frame
             put16(get16());
-            for (int j = 0; j < frame_type - 251; j++) {
+            for (u8 j = 0; j < frame_type - (u8)251; j++) {
```
apangin: Can you add a short comment in the code explaining this strange way of writing loop.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -131,6 +131,24 @@ jobs:
             image: ""
             # x64 MacOS should take fat binaries built on ARM
             asprof-binaries-job: macos
+          - runson:
+              display: macos-arm64
+              name: macos-15
+            java-version: 17
```
apangin: IMO, there are already too many test configurations but without much variety.
I'd leave 3 macOS configs testing 3 JDK versions: 11, 17 and 21, where 11 and 21 are on arm64, and jdk 17 is on x86.

src/instrument.cpp
Baraa-Hasheesh: handled here [9be8f8f](https://github.com/async-profiler/async-profiler/pull/1390/commits/9be8f8f4ec21f667a2ea6539512fa02345a934d4)

.github/workflows/test-and-publish-nightly.yml
Baraa-Hasheesh: done here [9be8f8f](https://github.com/async-profiler/async-profiler/pull/1390/commits/9be8f8f4ec21f667a2ea6539512fa02345a934d4)

PR #1392, opened 2025-07-15
Fix randomly failing tests

Baraa-Hasheesh:
### Description
Fixes 2 randomly failing tests

[datagramSocketLock](https://github.com/async-profiler/async-profiler/actions/runs/16151504646/job/45583762864)
Failure happens as no samples exists that matches `Unsafe_.ark`
Regex updated to `Unsafe_.{1,3}ark` to match both `Unsafe_Park` & `Unsafe_Unpark`

[parseMultiModeRecording](https://github.com/async-profiler/async-profiler/actions/runs/16167564928/job/45633054150)
Tests checks that a certain amount of Lock samples happen on a synchronous lock
The idea is to allow as many threads as possible to run as to create more samples for that lock

### Related issues
#1343

### Motivation and context
reduce amount of random failures on GHA

### How has this been tested?
`make test`

---

### DIFF
--- a/test/test/jfr/JfrMultiModeProfiling.java
+++ b/test/test/jfr/JfrMultiModeProfiling.java
@@ -21,7 +21,7 @@ public class JfrMultiModeProfiling {
     private static int count = 0;
 
     public static void main(String[] args) throws InterruptedException {
-        ExecutorService executor = Executors.newFixedThreadPool(2);
+        ExecutorService executor = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
         for (int i = 0; i < 10; i++) {
             executor.submit(JfrMultiModeProfiling::cpuIntensiveIncrement);
         }
--- a/test/test/lock/LockTests.java
+++ b/test/test/lock/LockTests.java
@@ -15,7 +15,7 @@ public class LockTests {
     public void datagramSocketLock(TestProcess p) throws Exception {
         Output out = p.profile("-e cpu -d 3 -o collapsed --cstack dwarf");
         assert out.ratio("(PlatformEvent::.ark|PlatformEvent::.npark)") > 0.1
-                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.ark"));
+                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.{1,3}ark"));
         out = p.profile("-e lock -d 3 -o collapsed");
         assert out.contains("sun/nio/ch/DatagramChannelImpl.send");
     }


test/test/lock/LockTests.java
```diff
@@ -15,7 +15,7 @@ public class LockTests {
     public void datagramSocketLock(TestProcess p) throws Exception {
         Output out = p.profile("-e cpu -d 3 -o collapsed --cstack dwarf");
         assert out.ratio("(PlatformEvent::.ark|PlatformEvent::.npark)") > 0.1
-                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.ark"));
+                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.{1,3}ark"));
```
fandreuz: Could you add the possible expected materializations of the new regex as a comment?

krk: >  Regex updated to Unsafe_.{1,3}ark to match both Unsafe_Park & Unsafe_Unpark

Can we match Unsafe_Park & Unsafe_Unpark instead?

Baraa-Hasheesh: > > Regex updated to Unsafe_.{1,3}ark to match both Unsafe_Park & Unsafe_Unpark
> 
> Can we match Unsafe_Park & Unsafe_Unpark instead?

I will need to check across multiple systems for this

Baraa-Hasheesh: I changed the regex here [682cd0a](https://github.com/async-profiler/async-profiler/pull/1392/commits/682cd0a568089330432bd3a53b4c427a15ddc480) 

Now it will match `Unsafe.park` & `Unsafe.unpark`  & `Unsafe_Park`  & `Unsafe_Unpark`

Baraa-Hasheesh: I've reverted the regex here [deaa7f6](https://github.com/async-profiler/async-profiler/pull/1392/commits/deaa7f64991b2e37944ed03189fa80e01d64957a)

Now it will only match `Unsafe_Park` & `Unsafe_Unpark`

test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -21,7 +21,7 @@ public class JfrMultiModeProfiling {
     private static int count = 0;
 
     public static void main(String[] args) throws InterruptedException {
-        ExecutorService executor = Executors.newFixedThreadPool(2);
+        ExecutorService executor = Executors.newFixedThreadPool(Math.max(2, Runtime.getRuntime().availableProcessors()));
```
apangin: This does not look a reliable fix to me.
If `availableProcessors()` is equal to `2`, nothing changes at all. For example, GitHub-hosted runners still provide 2 vCPUs for private repos. Also, people may run `make test` in their virtual machines with 2 vcores.

test/test/lock/LockTests.java
```diff
@@ -15,7 +15,7 @@ public class LockTests {
     public void datagramSocketLock(TestProcess p) throws Exception {
         Output out = p.profile("-e cpu -d 3 -o collapsed --cstack dwarf");
         assert out.ratio("(PlatformEvent::.ark|PlatformEvent::.npark)") > 0.1
-                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.ark"));
+                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_Park|Unsafe_Unpark"));
```
apangin: This changes the logic of the test.
`ReentrantLock.lock()` calls `Unsafe_Park` under the hood, so it is reasonable to assume that `Unsafe_Park` will be present in the profile.

Baraa-Hasheesh: That's true
But the test also checks for `ReentrantLock.unlock` which calls `Unsafe_Unpark`

`((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1`

Now regarding this problem, 
Basically as we're doing CPU sampling here, we're bounded by the sampling done via the engine
From the samples I saw for the failures I noticed that no sample contained `Unsafe_Park`, which is fine as for short profiling sessions you might not see specific method samples 
To address this I considered 3 changes & only opted for one of them as I didn't want to cause an over correction

* Increase the profiling period of the test 
* Increase the sampling rate for the test
* Change the regex to cover more samples

I opted for the `regex` change as it made sense to me as the test was already checking for both `lock` & `unlock` so it made sense to also check for `park` & `unpark`

Also for the cases I found were no sample was found for `Unsafe_Park` method, I did find samples for `Unsafe_Unpark`

What do you think?
If this is insufficient I would also change the sampling frequency here which should create more samples which in turn should make this situation much less likely

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: The problem it's kinda hard to guarantee a minimum number of samples for a `synchronized` lock 
The test checks for 50 samples,

I didn't want to create more Threads than the amount of vCPUs that we have as then we would be bounded by context switches & scheduler which also isn't reliable 

I would opt out to reducing the number of samples we expect in the test
From the failures I saw the lowest number of samples was around 17

What do you think about changing the number of expected samples to be greater than 10?

apangin: The idea of a test (any test, not just this one) is to verify that observed results match expectations. To me, it was expected that a test that acquires a lock 100K times concurrently in two threads should have many contended locks recorded. I can imagine why this is wrong, but I'd like to get confirmation that profiler does everything correctly.
What does happen in this case?

test/test/lock/LockTests.java
apangin: > no sample contained Unsafe_Park, which is fine as for short profiling sessions you might not see specific method samples

No, this is not fine. In the failed test, there are 50 samples (14%) of a Java method `jdk/internal/misc/Unsafe.park`, and every invocation of `jdk/internal/misc/Unsafe.park` ends with a call to `Unsafe_Park`.

Baraa-Hasheesh: > > no sample contained Unsafe_Park, which is fine as for short profiling sessions you might not see specific method samples
> 
> No, this is not fine. In the failed test, there are 50 samples (14%) of a Java method `jdk/internal/misc/Unsafe.park`, and every invocation of `jdk/internal/misc/Unsafe.park` ends with a call to `Unsafe_Park`.

Sorry for the confusion here, I should phrased it different as it's an Alpine issue,
Native frame unwinding is prone to fail the unwind process due to missing information in the `eh_frame_hdr` section

This can be observed by the fact that the `jdk/internal/misc/Unsafe.park` method was sampled, but native unwinding failed due to missing information (`unknown` frames)

Now `Unsafe_Unpark` is kinda safer for Alpine as it has less native frames to unwind from compared to  `Unsafe_Park`

However I would have to agree that this might be a bad idea as this is an alpine issue & this change might cause to hide bugs on other systems

I would opt for increasing the sampling frequency as that should increase the chance we pick a good sample for systems such as Alpine that have missing/incomplete unwind information , what do you think?

Baraa-Hasheesh: Please note the change here should fix the issue for us https://github.com/async-profiler/async-profiler/pull/1347 when it's integrated

apangin: I don't see how increasing sampling frequency will help in situations when unwinding does not work properly, because stack walking is broken no matter how often we execute this code. Unfrotunately, we cannot rely on #1347 either, since debuginfo is not usually available by default.

Baraa-Hasheesh: Increasing the frequency will increase the chance we pick up a sample that is able to unwind correctly without hitting the alpine issue 

The current compromise I have in mind is to check for `Unsafe_Unpark` only on Alpine where the issue happens

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: @apangin what would you think about using Java flight recorder as our base line 
Where the async-profiler should be within a certain % (2~3)  of the number of samples generated by the java flight recorder 

This will make our life easier as we have a better base line to compare against

**Note:** This would make the test Java 11+

apangin: These are uncomparable things.

test/test/lock/LockTests.java
apangin: So, the test uncovers a real stack walking bug. We can of course disable the test or make it pass otherwise, but the best solution would be to fix the stack walking issue (I am not yet sure how - this requires additional research).

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: @apangin I was checking & on the samples I have on hand did confirm that even if the number of samples is quite low the total lock duration is actually quite high, 
So I added a condition to validate either number of samples or total duration 

I found that in both cases normal & failure the total lock duration was actually high(around 400ms total), I've added the check as 200ms for safety margins [3e6eeb5](https://github.com/async-profiler/async-profiler/pull/1392/commits/3e6eeb546eeedf7ab07c95be6ee7bda9ab939978)

test/test/lock/LockTests.java
Baraa-Hasheesh: The Unsafe_Unpark is only checked for MUSL were the unwinding native traces issue exists [3e6eeb5](https://github.com/async-profiler/async-profiler/pull/1392/commits/3e6eeb546eeedf7ab07c95be6ee7bda9ab939978)

test/test/jfr/JfrMultiModeProfiling.java
apangin: Did you choose a value basing on the observation on your PC? What if we run tests on a 2x faster PC? To me, 200ms does not make much sense. If the test start failing at some point with 190ms, will it mean that async-profiler works wrong?

Back to the [expectations](https://github.com/async-profiler/async-profiler/pull/1392#discussion_r2213416223) - if two threads compete for the same resource all the time, I would expect that the total duration waiting on a lock will be  of the same order as the test run time, i.e. at every moment of time, one of the threads is running computation while the second one is waiting.

test/test/lock/LockTests.java
apangin: Checking the OS inside a test which is supposed to verify locks is against a spirit of the test, especially since this hides a real problem. Furthermore, the condition is just incorrect: whether stack unwinding works or not depends on how libc is compiled, not on the type of libc.

1. Link the test to an issue about stack unwinding problem on Alpine.
2. Modify the test to work on all systems while keeping the original purpose of the test - checking if CPU profile highlights locking and unlocking.

test/test/jfr/JfrTests.java
```diff
@@ -74,8 +76,14 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
             }
         }
 
+        double totalLockDurationMillis = 0;
+        try (JfrReader r = new JfrReader(p.getFile("%f").toURI().getPath())) {
```
apangin: The recording is already parsed in the above loop, no need to read the file twice.
Also, don't mix JFR API and async-profiler's JfrReader in the same test (unless you specifically intend to test JfrReader).

test/test/jfr/JfrTests.java
```diff
@@ -74,8 +76,14 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
...
+            List<ContendedLock> events = r.readAllEvents(ContendedLock.class);
+            totalLockDurationMillis = events.stream().map(e -> e.duration).reduce((Long::sum)).get() / 1_000_000.0;
```
apangin: `mapToLong().sum()`

test/test/jfr/JfrTests.java
```diff
@@ -74,8 +76,14 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
...
+        }
+
         Assert.isGreater(eventsCount.get("jdk.ExecutionSample"), 50);
-        Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 50);
+        assert eventsCount.get("jdk.JavaMonitorEnter") > 50 || totalLockDurationMillis >= 200;
```
apangin: Is it _expected_ that a recording may have only 1 lock event with 400ms duration?
Or if it has 1000 events with 10ms total duration?
To me, both would look suspiciously. IMO, it's better to check both event count and duration.

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: Regarding this 
I tested on both EC2, local & GHA

Regarding my thought process 
If you test on a powerful machine you will get more contention due to the threadpool change 
If you test on light machine you might get less lock contention in terms of number of samples but each sample is much more expensive in terms of duration 

When this fails I've seen samples of +70ms duration while locally the samples are in microseconds 

I don't believe we should test with both as it will make it harder to correctly apply conditions that would work across many different skews

test/test/lock/LockTests.java
Baraa-Hasheesh: I feel we keep going around in circles here due to the unwind issue when we're missing both unwind info & frame pointers

While it's true this might happen on any systems that does  compile libc with certain options I don't think we should account for them all 

At this point I would do as suggested
 keeping the test as is & just disable it on MUSL & track this unwind issue on a different issue

test/test/jfr/JfrTests.java
Baraa-Hasheesh: Yeah that makes sense to me 
That there should be a minimum amount of lock samples 
& Even if we have more samples there should be a minimum of time holding the locks

Baraa-Hasheesh:
@apangin In the latest revision I've disabled the datagram test for MUSL to avoid this unwind problem & added a TODO above it
We already have 2 different items for this I didn't see the need to create a third item 

For the test/test/jfr/JfrMultiModeProfiling.java test I've refactored the test to report how much time it took to complete all of the threads &  that time is checked against the total lock duration we have in the JFR 

Checking a flat number like 200ms isn't stable across multiple JVM & CPU skews 
However checking the percentage should be quite stable as if the thread isn't running it should be locked

For most of the time 1 of the 2 threads should be locked in our logic, this means in practice for most cpu cycles where something is being executed a different thread should be locked, so the total lock duration should be quite close to the total run duration. 

During my benchmarking I noticed numbers between 70 & 85% (total lock / total run) [depends on CPU/JVM].

The lowest percentage I saw during various runs was around ~70%, for safety I made the check to be higher than 50%

test/test/lock/LockTests.java
apangin: > just disable it on MUSL

Sorry if I was unclear, but I actually meant the opposite: *"Modify the test to work on all systems"*.
The intention of the test is to check that CPU profile discovers high lock contention inside the JDK standard class. This is not platform dependent: contention happens with glibc, musl and macOS.

Is it possible to discover `Unsafe_Park` or any other lock-related native frame on Alpine/x86? It seems like it is not, due to lack of symbols and unwinding info in ld-musl.so. I checked that gdb (which may be considered as the "reference" implementation for stack unwinding) cannot get a stack trace without musl-dbg package too.

This means, we cannot rely on native stacks for this test, unfortunately. So, to keep test working on all platforms, can we rely on Java frames instead? An obvious idea is to check for presense of all of `ReentrantLock.lock/unlock` and `Unsafe.park/unpark`. Sounds good, but the problem is, in JDK 8, `DatagramChannel` used `synchronize` instead of `ReentrantLock`. So, with JDK 8 on Alpine, output is completely unusable: it shows neither native frames nor Java stacks related to locking. This is, of course, not what one would expect from the profiler.

The problem is deeper than it seemed. And it's relevant not only for Alpine, as mentioned in https://github.com/async-profiler/async-profiler/issues/1317#issuecomment-2933584880.
Two issues need to be fixed:
1. Add support for external .debug_frame.
2. Unwind Java stacks using JavaFrameAnchor, when .eh_frame is missing or incomplete.

Before these are fixed, I suggest the following changes to suppress test failures:
1. Limit the test for JDK 11+ (as JDK 8 uses a different synchronization primitive in DatagramChannel).
2. Adjust assertions to verify Java frames only: `ReentrantLock.lock/unlock` and `Unsafe.park/unpark`.

Baraa-Hasheesh: @apangin thanks for the suggestions

I've applied as mentioned & simplified the condition as we no longer check for Java 8.
I also added a generic check to validate that `Unsafe.park/unpark` should exist somewhere on the stacks collected for `ReentrantLock.lock/unlock`

test/test/lock/LockTests.java
```diff
@@ -11,11 +11,11 @@
...
 public class LockTests {
 
-    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true)
+    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true, jvmVer = {11, Integer.MAX_VALUE})
     public void datagramSocketLock(TestProcess p) throws Exception {
         Output out = p.profile("-e cpu -d 3 -o collapsed --cstack dwarf");
-        assert out.ratio("(PlatformEvent::.ark|PlatformEvent::.npark)") > 0.1
-                || ((out.ratio("ReentrantLock.lock") + out.ratio("ReentrantLock.unlock")) > 0.1 && out.contains("Unsafe_.ark"));
+        assert out.ratio("(ReentrantLock.lock|ReentrantLock.unlock)") > 0.1;
+        assert out.contains("(ReentrantLock.lock|ReentrantLock.unlock).*(Unsafe.park|Unsafe.unpark)");
```
apangin: Can we check that _all_ of ReentrantLock.lock, unlock, Unsafe.park and unpark exist?

test/test/jfr/JfrTests.java
```diff
@@ -61,22 +60,29 @@ public void parseRecording(TestProcess p) throws Exception {
...
                 eventsCount.put(eventName, eventsCount.getOrDefault(eventName, 0) + 1);
             }
         }
 
-        Assert.isGreater(eventsCount.get("jdk.ExecutionSample"), 50);
-        Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 50);
-        Assert.isGreater(eventsCount.get("jdk.ObjectAllocationInNewTLAB"), 50);
+        assert eventsCount.get("jdk.ExecutionSample") > 50;
+        assert eventsCount.get("jdk.JavaMonitorEnter") > 10 && (totalLockDurationMillis / lockJobDurationMillis) >= 0.50;
+        assert eventsCount.get("jdk.ObjectAllocationInNewTLAB") > 50;
```
apangin: `Assert.isGreater` has an advantage that it prints actual values in case of failure. Why did you replace it with `assert`?

test/test/jfr/JfrTests.java
```diff
@@ -61,22 +60,29 @@ public void parseRecording(TestProcess p) throws Exception {
...
+        long lockJobDurationMillis = Long.parseLong(output.stream().findFirst().get());
+
+        double totalLockDurationMillis = 0;
         Map<String, Integer> eventsCount = new HashMap<>();
         try (RecordingFile recordingFile = new RecordingFile(p.getFile("%f").toPath())) {
             while (recordingFile.hasMoreEvents()) {
                 RecordedEvent event = recordingFile.readEvent();
                 String eventName = event.getEventType().getName();
+                if (eventName.equals("jdk.JavaMonitorEnter")) {
+                    totalLockDurationMillis += event.getDuration().toNanos() / 1000.0 / 1000.0;
```
apangin: `/ 1000.0 / 1000.0` looks weird to me. You could write `/ 1_000_000` instead.
Alternatively, you could just accumulate nanoseconds as `long`.

test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -25,9 +26,13 @@ public class JfrMultiModeProfiling {
 
     public static void main(String[] args) throws InterruptedException {
         ExecutorService executor = Executors.newFixedThreadPool(2);
+        List<CompletableFuture<Void>> completableFutures = new ArrayList<>();
+        long startTime = System.currentTimeMillis();
         for (int i = 0; i < 10; i++) {
-            executor.submit(JfrMultiModeProfiling::cpuIntensiveIncrement);
+            completableFutures.add(CompletableFuture.runAsync(JfrMultiModeProfiling::cpuIntensiveIncrement, executor));
         }
+        completableFutures.forEach(CompletableFuture::join);
```
apangin: This changes the original logic of the test which intended to run locking code and allocation code in parallel.

test/test/jfr/JfrTests.java
Baraa-Hasheesh: assert is easier to check multiple condition & used in most locations

I don't have a real preference for either I will change back to Assert.

test/test/lock/LockTests.java
Baraa-Hasheesh: I will need to check if that's doable without facing flakey failures

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: I read the test intension as generating multiple types of samples in the same JFR

I wasn't aware it wanted to achieve that in parallel manner as well 

I will add an extra thread for the Alloc logic to make it parallel again

test/test/jfr/JfrTests.java
apangin: Yes, you can replace compound `&&` with two `Assert.isGreater` statements.

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: @apangin Allowing the Alloc to run in parallel will cause some scheduling issues for the LOCK threads
This in turn we will need to make the new check more relaxed than one expected it to be 

The lowest I saw while running multiple runs was around 10%, 
This issue happens on GHA due to limited vCPUs available for the process 

To avoid going through the need to set a very relaxed check, I would just recommend the following:
* Keep this test running as with current revision (not in parallel) 
* Introduce a new test that would sample Alloc & Lock in parallel with simple sane checks (Lock happens & Alloc happens) 

What are your thoughts about this?

apangin: > some scheduling issues

What issues?

> around 10%

Do you mean that lock was held for 10% of the lock loop run time? What was the loop doing for the remaining 90% then?

Baraa-Hasheesh: @apangin Simply that the CPU is scheduled in such a manner for the running threads that the contention happening on the Lock is minimal 

As discussed offline I will update the logic to use ThreadMXBean which can track Lock time per thread

Baraa-Hasheesh:
@apangin Test changed to use ThreadMXBean 
Which should be quite reliable across all types of machines/CPU configs 

For datagramSocketLock, I did run the test +10 times on GHA & passed for all runs, so I kept the assertions for all methods

test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -23,13 +28,22 @@ public class JfrMultiModeProfiling {
     private static int count = 0;
     private static final List<byte[]> holder = new ArrayList<>();
 
+    private static ThreadMXBean tmx = ManagementFactory.getThreadMXBean();
+    private static Map<Long, Long> threadLockTimes = new ConcurrentHashMap<>();
```
fandreuz: Nit: how about a `ThreadLocal<Long>` here? You can print the value before the thread exits

Baraa-Hasheesh: I will check on this

Baraa-Hasheesh: Discussed offline, keeping as is

test/test/jfr/JfrTests.java
```diff
@@ -61,21 +61,29 @@ public void parseRecording(TestProcess p) throws Exception {
...
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock,jfr,file=%f")
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock,quiet,jfr,file=%f", output = true)
     public void parseMultiModeRecording(TestProcess p) throws Exception {
-        p.waitForExit();
+        Output output = p.waitForExit(TestProcess.STDOUT);
         assert p.exitCode() == 0;
+
+        long totalLockDurationMillis = output.stream().map(Long::parseLong).reduce((Long::sum)).orElse(0L);
```
apangin: `mapToLong().sum()`

test/test/jfr/JfrMultiModeProfiling.java
apangin: nit: `final` for both

apangin: Hooking thread termination in the pool will require more code. I'm fine with Bara's approach.

test/test/jfr/JfrTests.java
```diff
@@ -61,21 +61,29 @@ public void parseRecording(TestProcess p) throws Exception {
...
+                    jfrTotalLockDurationMillis += event.getDuration().toNanos() / 1_000_000.0;
+                }
                 eventsCount.put(eventName, eventsCount.getOrDefault(eventName, 0) + 1);
             }
         }
 
         Assert.isGreater(eventsCount.get("jdk.ExecutionSample"), 50);
-        Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 50);
+        Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 10);
+        Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.90);
```
apangin, line 86: That's a nice idea!

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: handled [c5a0cfa](https://github.com/async-profiler/async-profiler/pull/1392/commits/c5a0cfa3cb9d2d0ef1209fed7baa7fd9970a4374)

test/test/jfr/JfrTests.java
Baraa-Hasheesh: handled [c5a0cfa](https://github.com/async-profiler/async-profiler/pull/1392/commits/c5a0cfa3cb9d2d0ef1209fed7baa7fd9970a4374)

test/test/jfr/JfrTests.java
fandreuz: @Baraa-Hasheesh Maybe `0.9` is too high? https://github.com/async-profiler/async-profiler/actions/runs/16679310201/job/47213879714#step:11:158

Baraa-Hasheesh: I will need to investigate in depth both me & @apangin have doubt that we're missing something 
In mean time I will reopen the issue that was in place

Issue #1217, opened 2025-04-04
GHA cosmetic, do not show "publish" as skipped task in PRs


krk:
### Describe the feature

It is confusing to always have a skipped task in PRs, for publishing nightly builds:

![Image](https://github.com/user-attachments/assets/7e1cfddf-38fe-46d1-85df-b9c88ac8ec7a)

### Use Case

Create a PR and not see "skipped publish task".

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1398, opened 2025-07-17
Automated Maven artifact snapshot publication

fandreuz:
### Description
In this PR I introduce a new piece of GHA to create a snapshot release in Maven after a push to `master`.

### Related issues
n/a

### Motivation and context
We don't publish SNAPSHOT artifacts yet, but that would be nice for our users.

### How has this been tested?
https://github.com/fandreuz/maven-deploy-test/actions/runs/16350974154/job/46197387100

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -318,11 +318,9 @@ jobs:
           mvn versions:set -DnewVersion="${VERSION}-SNAPSHOT"
       - name: Create directory structure for Maven release
         run: |
-          mkdir -p native/linux-x64
+          mkdir -p native/{linux-x64,linux-arm64,macos}
           tar xfO async-profiler-*-linux-x64.tar.gz --wildcards "*/lib/*" > native/linux-x64/libasyncProfiler.so
-          mkdir -p native/linux-arm64
           tar xfO async-profiler-*-linux-arm64.tar.gz --wildcards "*/lib/*" > native/linux-arm64/libasyncProfiler.so
-          mkdir -p native/macos
           unzip -p async-profiler-*-macos.zip "*/lib/*" > native/macos/libasyncProfiler.so
       - uses: actions/setup-java@v4
         with:


.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
+name: Release on tag
```
fandreuz: This should be done for jfr-converter too

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Maven release
+        run: |
+          mvn versions:set -DnewVersion=${GITHUB_REF_NAME#v}
+          mvn deploy
+        env:
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+          GITHUB_USERNAME: ${{ secrets.GITHUB_ACTOR }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: GitHub release
+        uses: softprops/action-gh-release@v2
```
fandreuz: TODO add the other assets

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+          servers: '[{"id": "central", "username": "${{secrets.MAVEN_USER}}", "password": "${{secrets.MAVEN_TOKEN}}"}]'
+      - name: Maven release
+        run: |
+          mvn versions:set -DnewVersion=${GITHUB_REF_NAME#v}
+          mvn deploy
+        env:
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+          GITHUB_USERNAME: ${{ secrets.GITHUB_ACTOR }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: GitHub release
```
fandreuz: It would be nice to create the changelog automatically. I see it's not as simple as `git log ...` since commits are divided in "fixes", "features", ...

Still, we could just dump all the commits into the release message, and a human curator could edit it later. What do you think @apangin ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
+name: Release on tag
+
+on:
+  push:
+    tags:
```
fandreuz: I thought about the triggering strategy.

What we want:
- Run when a tag `v.*.*.*` is pushed
- Wait for `test-and-publish-nightly.yml` to complete, so we don't have to rebuild the artifacts, and we're sure we publish the binaries which have been tested earlier

We could have a `workflow_run` registered on the `CI` workflow, and then filter away all cases which are not related to tags. However, this means that we are going to see a spurious empty workflow run `Release on tag` for all commits not related to a tag, which is not pleasant.

A better solution could be to add new jobs in `test-and-publish-nightly.yml` (which we could rename to `test-and-publish.yml`) to handle Maven publication and GitHub release, similarly to how we now handle nightly publication there. This way we can also easily consume the artifacts. Problems with this approach:
- `test-and-publish.yml` will be relatively large, and will do lots of things
- We now have a spurious `publish(nightly)` job (#1217); after this change, there are going to be more spurious jobs

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -246,12 +246,14 @@ jobs:
           path: |
```
fandreuz, line 246: @apangin looking at `pom.xml`, it seems that the intention is to use a classifier (`native.platform`) to distinguish JARs targeted for macos, linux-x64 and arm64. But looking at what is published in Maven, I see this:
```
        0  04-11-2025 02:59   META-INF/
       81  04-11-2025 02:59   META-INF/MANIFEST.MF
        0  04-11-2025 02:59   linux-arm64/
        0  04-11-2025 02:59   linux-x64/
        0  04-11-2025 02:59   macos/
        0  04-11-2025 02:59   one/
        0  04-11-2025 02:59   one/profiler/
        0  04-11-2025 02:59   META-INF/maven/
        0  04-11-2025 02:59   META-INF/maven/tools.profiler/
        0  04-11-2025 02:59   META-INF/maven/tools.profiler/async-profiler/
-> 552200  04-11-2025 02:59   linux-arm64/libasyncProfiler.so
-> 530352  04-11-2025 02:59   linux-x64/libasyncProfiler.so
-> 855760  04-11-2025 02:59   macos/libasyncProfiler.so
     6240  04-11-2025 02:59   one/profiler/AsyncProfiler.class
      633  04-11-2025 02:59   one/profiler/AsyncProfilerMXBean.class
      996  04-11-2025 02:59   one/profiler/Counter.class
      512  04-11-2025 02:59   one/profiler/Events.class
     4896  04-09-2025 02:04   META-INF/maven/tools.profiler/async-profiler/pom.xml
       64  04-11-2025 02:59   META-INF/maven/tools.profiler/async-profiler/pom.properties
```
Which one should we adopt for nightly releases?

apangin: 4 artifacts should be included: `async-profiler.jar` with binaries for all platforms and 3 platform-specific jars: `async-profiler-{linux-x64,linux-arm64,macos}.jar`.

I changed `pom.xml` to produce all 4 jars at a time. Previously, they were deployed independently. Also, switched from OSSRH to Central Portal plugin. Please merge recent changes from master.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,42 @@ jobs:
...
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -DnewVersion="${VERSION}-SNAPSHOT"
+      - name: Create directory structure for Maven release
+        run: |
+          find . -type f -name "async-profiler-*.tar.gz" -not -name "*debug*" -exec tar -xf {} \;
+          find . -type f -name "async-profiler-*.zip" -not -name "*debug*" -exec unzip {} \;
+          for file in async-profiler-*/lib/*.{so,dylib}; do
+            arch=$(python3 -c 'import sys; print("-".join(sys.argv[1].split("/")[0].split("-")[4:]))' "$file")
+            ln -s $(dirname "$file") $arch
+          done
```
apangin: It's complicated but incorrect.
Can we be more straightforward?
```
tar xfO async-profiler-*-linux-x64.tar.gz --wildcards "*/lib/*" > native/linux-x64/libasyncProfiler.so
tar xfO async-profiler-*-linux-arm64.tar.gz  --wildcards "*/lib/*" > native/linux-arm64/libasyncProfiler.so
unzip -p async-profiler-*-macos.zip "*/lib/*" > native/macos/libasyncProfiler.so
```

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,42 @@ jobs:
...
+        shell: bash
+      - uses: actions/setup-java@v4
+        with:
+          distribution: corretto
+          java-version: 21
+          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+      - name: Maven MacOS release
+        run: mvn clean -Dnative.platform=macos deploy
```
apangin: No longer needed. Now `mvn deploy` needs to be called only 2 times: one for async-profiler and one for jfr-converter.

fandreuz: Good to know: 3ec2441eceb3ae2c4f4d122d5c27337b960ab9c9

.github/workflows/test-and-publish-nightly.yml
fandreuz: 3ec2441eceb3ae2c4f4d122d5c27337b960ab9c9

.github/workflows/test-and-publish-nightly.yml
fandreuz: The idea here was to support other possible architectures, e.g. in `AsyncProfiler.java` I see this:
```java
        if (os.contains("linux")) {
            if (arch.equals("amd64") || arch.equals("x86_64") || arch.contains("x64")) {
                return "linux-x64";
            } else if (arch.equals("aarch64") || arch.contains("arm64")) {
                return "linux-arm64";
            } else if (arch.equals("aarch32") || arch.contains("arm")) {
                return "linux-arm32";
            } else if (arch.contains("86")) {
                return "linux-x86";
            } else if (arch.contains("ppc64")) {
                return "linux-ppc64le";
            }
        }
```
But maybe it got a bit out of hand indeed. I'll take your solution

fandreuz: 461f9e48a1002eb44adcd42cb55f00042d969c6c

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,28 @@ jobs:
...
             });
+      - name: Set snapshot version
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -DnewVersion="${VERSION}-SNAPSHOT"
+      - name: Create directory structure for Maven release
+        run: |
+          tar xfO async-profiler-*-linux-x64.tar.gz --wildcards "*/lib/*" > native/linux-x64/libasyncProfiler.so
+          tar xfO async-profiler-*-linux-arm64.tar.gz --wildcards "*/lib/*" > native/linux-arm64/libasyncProfiler.so
+          unzip -p async-profiler-*-macos.zip "*/lib/*" > native/macos/libasyncProfiler.so
```
Baraa-Hasheesh, line 324: macos `.so`?

Shouldn't this be `dylib`

fandreuz: Spotted that too, it's also [in the artifact](https://github.com/async-profiler/async-profiler/pull/1398#discussion_r2219204675) we publish on Maven.

@apangin is that intended?

apangin: Yes, that's what async-profiler Java API historically [looks for](https://github.com/async-profiler/async-profiler/blob/cdedd3fb221385ca9dfdba9012c821eda99239b2/src/api/one/profiler/AsyncProfiler.java#L60).
I don't mind changing it or leaving as is - this does not seem to affect anything.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,28 @@ jobs:
...
+      - uses: actions/setup-java@v4
+        with:
+          distribution: corretto
+          java-version: 21
+          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+      - name: Maven release
+        run: mvn clean deploy
```
apangin, line 332: I noticed in https://github.com/fandreuz/maven-deploy-test/actions/runs/16350974154/job/46197387100 that Maven downloads half of the internet to do this simple job. Will this happen on every commit?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,28 @@ jobs:
...
+        run: |
+          tar xfO async-profiler-*-linux-x64.tar.gz --wildcards "*/lib/*" > native/linux-x64/libasyncProfiler.so
+          tar xfO async-profiler-*-linux-arm64.tar.gz --wildcards "*/lib/*" > native/linux-arm64/libasyncProfiler.so
+          unzip -p async-profiler-*-macos.zip "*/lib/*" > native/macos/libasyncProfiler.so
+      - uses: actions/setup-java@v4
+        with:
+          distribution: corretto
+          java-version: 21
+          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
```
apangin, line 330: Why is it _here_ in actions/setup-java?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,28 @@ jobs:
...
               draft: false,
             });
+      - name: Set snapshot version
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -DnewVersion="${VERSION}-SNAPSHOT"
+      - name: Create directory structure for Maven release
+        run: |
+          tar xfO async-profiler-*-linux-x64.tar.gz --wildcards "*/lib/*" > native/linux-x64/libasyncProfiler.so
+          tar xfO async-profiler-*-linux-arm64.tar.gz --wildcards "*/lib/*" > native/linux-arm64/libasyncProfiler.so
```
apangin, line 323: How does this work? I don't see where `native` directory is created.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,28 @@ jobs:
...
+        with:
+          distribution: corretto
+          java-version: 21
+          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+      - name: Maven release
+        run: mvn clean deploy
+        env:
+          GITHUB_USERNAME: ${{ secrets.GITHUB_ACTOR }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
apangin: For manual publishing, I setup token username and token password in `settings.xml`, and they are different from github credentials. How is it supposed to work?

fandreuz: Right, this is a residual from an example I had in another repository, but those vars are not needed here. Removed here: bf98d00dac4d507fd5261a03f8f7d994a1a770c8

.github/workflows/test-and-publish-nightly.yml
fandreuz: Right, when I tested the command locally the dir structure was already there. Fixed in d5d4249288b76571d353560582d6871ab282c131

.github/workflows/test-and-publish-nightly.yml
fandreuz: https://github.com/actions/setup-java?tab=readme-ov-file#maven-options

.github/workflows/test-and-publish-nightly.yml
fandreuz: Yes, every commit to `master`. There are some ways to "fix" this behavior:
- GHA [cache](https://github.com/actions/cache)
- Docker image

Should I take some time to investigate?

apangin: 3rd option: do not use Maven.

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -310,3 +312,25 @@ jobs:
               release_id: newReleaseId,
               draft: false,
             });
+      - name: Set snapshot version
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -DnewVersion="${VERSION}-SNAPSHOT"
+      - name: Create directory structure for Maven release
+        run: |
+          mkdir -p native/linux-x64
```
apangin: `mkdir -p native/{linux-x64,linux-arm64,macos}`

.github/workflows/test-and-publish-nightly.yml
fandreuz: 3rd seems a bit harsh. I understand where this is coming from, but the workflow is relatively common. Do you have security concerns?

.github/workflows/test-and-publish-nightly.yml
fandreuz: Nice, thanks: 6b4acdb9821c6e4e2e232c962f603962fe64b335

.github/workflows/test-and-publish-nightly.yml
apangin: Not really. When migrating from OSSRH to Central Portal, I realized that submitting a bundle for publishing is as simple as calling one REST API. Preparing a bundle is also not a big deal - this is just a plain .zip where each artifact is accompanied with a few checksum files.
Apparently, a script to prepare and upload a bundle will be smaller than one `pom.xml`, but at the same time, a lot lot faster, since it won't need to download hundreds of maven dependencies. Additionally, as you noted, a free bonus of eliminating supply chain risks.

fandreuz: Just so I understand, do you want to remove maven usages altogether, or keep using e.g. `mvn install` and handle the upload manually?

apangin: Try to build and publish without maven at all. But if happens to be more challenging than expected, please let me know.

Issue #1389, opened 2025-07-11
Crashes in VMX for synchronous events


Baraa-Hasheesh:
### Describe the bug

I was doing some testing for VM/VMX across different architectures & encountered some crashes for synchronous events such as MALLOC & ALLOC

```
#0  StackFrame::unwindStub (entry=0xf1fe03970880, name=name@entry=0xf1fe1c5408d0 "adapters", pc=@0xf1fe1b49d8e8: 266073284216960, sp=@0xf1fe1b49d8e0: 266073681812448, fp=@0xf1fe1b49d8d8: 266073703994512, this=0xf1fe1b49d8f8) at src/stackFrame_aarch64.cpp:40
40          return (uintptr_t)REG(regs[30], lr);
[Current thread is 1 (Thread 0xf1fe1b49f1a0 (LWP 14592))]
#0  StackFrame::unwindStub (entry=0xf1fe03970880, name=name@entry=0xf1fe1c5408d0 "adapters", pc=@0xf1fe1b49d8e8: 266073284216960, sp=@0xf1fe1b49d8e0: 266073681812448, fp=@0xf1fe1b49d8d8: 266073703994512, this=0xf1fe1b49d8f8) at src/stackFrame_aarch64.cpp:40
#1  0x0000f1fe1b1eac60 in StackWalker::walkVM (ucontext=ucontext@entry=0x0, frames=0xf1fe158b3a00, max_depth=2048, detail=detail@entry=VM_NORMAL, pc=<optimized out>, sp=<optimized out>, fp=<optimized out>) at src/stackWalker.cpp:370
#2  0x0000f1fe1b1f0b4c in StackWalker::walkVM (ucontext=0x0, detail=VM_NORMAL, max_depth=<optimized out>, frames=0xf1fe158b3a00) at src/stackWalker.cpp:203
#3  StackWalker::walkVM (detail=VM_NORMAL, max_depth=<optimized out>, frames=0xf1fe158b3a00, ucontext=0x0) at src/stackWalker.cpp:201
#4  Profiler::recordSample (this=0xf1fe1402cbb0, ucontext=ucontext@entry=0x0, counter=266073679010412, counter@entry=40, event_type=event_type@entry=MALLOC_SAMPLE, event=0xf1fe1b49db80, event@entry=0xf1fe1b49db60) at src/profiler.cpp:652
[...]

PR #1399, opened 2025-07-18
#1389: Incorrect top frame for synchronous events with cstack=vm on ARM64

apangin:
### Description

VMStructs-based stack walker could crash on ARM64 for allocation or nativemem events.

### Related issues

#1389 

### Motivation and context

The crash was caused by the combination of two issues:
1. gcc 13.3 changed frame prologue on AArch64. Previously, frame base `x29` was always pointing to `sp`, but now it is no longer the case. E.g.:
```
   2ba10:       d107c3ff        sub     sp, sp, #0x1f0
                                ...
   2ba20:       a91a7bfd        stp     x29, x30, [sp, #416]
   2ba24:       910683fd        add     x29, sp, #0x1a0
```
   Therefore, the value returned by `callerSP()` was incorrect. Since there is no simple way to get caller SP, we will start stack walking from the current frame rather than a caller frame on ARM64.

2. Because of the above error, stack unwinding code attempted to access signal context when it should not. `alloc` and `nativemem` events are not triggered by a signal and thus `ucontext` passed to the stack walking code is `NULL`. To avoid similar errors in the future, we will pass a dummy `empty_context` instead.

### How has this been tested?

`make test` previously failed on Ubuntu 24 with gcc 13.3. Now, all tests pass.
Manually verified that `nativemem` flamegraphs are good with cstack=vm.

---

### DIFF
--- a/src/arch.h
+++ b/src/arch.h
@@ -15,8 +15,6 @@
 #  define unlikely(x)  (__builtin_expect(!!(x), 0))
 #endif
 
-#define callerPC()     __builtin_return_address(0)
-
 #ifdef _LP64
 #  define LP64_ONLY(code) code
 #else // !_LP64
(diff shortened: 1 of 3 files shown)


Baraa-Hasheesh:
This looks fine for me, however I would suggest testing against VMX 
As for `alloc` profiling currently it will actually use the [getJavaTraceJvmti](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L674C23-L674C40) when using `VM`

There is a PR that fixes that [here](https://github.com/async-profiler/async-profiler/pull/1370) but that was put on hold until this bug was fixed as it was a blocker for that

apangin:
Thanks. I tested it both with `vm` and `vmx` modes.

Issue #1385, opened 2025-07-10
No stack frames for md5_implCompressMB on Arm


WadeWalker:
### Describe the bug

Profiles on Intel and AMD show full stack traces leading down to the md5_implCompressMB function; profiles on Arm show the md5_implCompressMB function directly under "all", with no other stack frames.

This is with `cstack=vm`, which I now use as the default after the resolution of https://github.com/async-profiler/async-profiler/issues/1325.

This happens on a variety of benchmarks; I'll pick a specific one for the example below.

### Expected vs. actual behavior

I expect a stack trace like this one I got on AMD:

<img width="873" height="712" alt="Image" src="https://github.com/user-attachments/assets/176e7018-e266-449a-a4e2-3ddf07eb585c" />

Instead I get one like this on Arm Graviton 4:

<img width="879" height="75" alt="Image" src="https://github.com/user-attachments/assets/0e1e14cc-6241-4232-9fb4-97d4fd5f8541" />

### Reproduction Steps

Run the DaCapo luindex benchmark (large size). See https://github.com/async-profiler/async-profiler/issues/1325 for details on how to download the benchmark and provision EC2 instances.

Profile the benchmark using this command:

`java -agentpath:async-profiler-4.0-f627b31-linux-arm64/lib/libasyncProfiler.so=start,cstack=vm,event=cpu,wall,file=profile.jfr -jar dacapo-23.11-MR2-chopin.jar luindex --size large -n 1`

Convert JFR to flame graph with this command:

`java -jar jfr-converter.jar --inverted --state default -o html profile.jfr profile.html`

Observe the missing stack frames :)

### Additional Information/Context

JFR files collected with async-profiler for AMD (m7a) and Arm (m8g):

[dacapo-luindex-large-m7a.metal-48xl-profile.jfr.gz](https://github.com/user-attachments/files/21170947/dacapo-luindex-large-m7a.metal-48xl-profile.jfr.gz)

[dacapo-luindex-large-m8g.metal-24xl-profile.jfr.gz](https://github.com/user-attachments/files/21170948/dacapo-luindex-large-m8g.metal-24xl-profile.jfr.gz)

### Async-profiler version

async-profiler-4.0-f627b31-linux-arm64

### Environment details

I always use nightly async-profiler, because the last stable release still crashes on Arm (see https://github.com/async-profiler/async-profiler/issues/1319). My scripts use the GitHub API to pull the most recent nightly at the time of the run.

[...]

PR #1400, opened 2025-07-20
Unwind checksum and digest intrinsics on ARM64

apangin:
### Description

Unwind certain ARM64 intrinsics.

### Related issues

#1385

### Motivation and context

HotSpot JVM has hand-coded assembly implementation for some CPU intensive checksum and digest computation algorithms suchs as CRC32, MD5, SHA256, etc. These compiler intrinsic functions are generated at runtime and they do not follow commonly used frame layout, especially on ARM64. That's why async-profiler (as well as JDK Flight Recorder and other profilers) cannot unwind Java stack trace during execution of these functions.

I divided stub routines for these intrinsics into 4 categories:
1. Stubs that do not modify stack pointer or frame pointer. Unwinding these stubs is trivial: return address stays in `x30` register.
2. Stubs that have fixed size frame layout and begin with `stp` instruction that decrement stack pointer by this fixed amount. To unwind these stubs, I find frame size directly from the instruction stream and increment `sp` back by this amount.
3. Stubs that setup regular frame pointer link. For them, I use `fp` (`x29`) register to find the previous frame.
4. All other stubs that are either non-leaf functions or have dynamic frame size (modify `sp` in the middle). These cases are rare, and I do not attempt to unwind such frames.

### How has this been tested?

Added new test `CodingIntrinsics.java` that executes the following intrinsics:
- updateBytesCRC32
- updateBytesAdler32
- md5_implCompress, md5_implCompressMB
- sha1_implCompress, sha1_implCompressMB
- sha256_implCompress, sha256_implCompressMB
- encodeBlock
- zero_blocks
- _large_arrays_hashcode_byte
- jbyte_disjoint_arraycopy

---

### DIFF
--- a/test/test/recovery/RecoveryTests.java
+++ b/test/test/recovery/RecoveryTests.java
@@ -56,11 +56,11 @@ public class RecoveryTests {
         Assert.isGreater(out.ratio("Suppliers.loop"), 0.5);
     }
 
-    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, inputs = "")
-    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, inputs = "--cstack vm", nameSuffix = "VM")
+    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.X64}, inputs = "")
+    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.X64}, inputs = "--cstack vm", nameSuffix = "VM")
     public void intrinsics(TestProcess p) throws Exception {
         Output out = p.profile("-d 3 -e cpu -i 1ms -o collapsed " + p.inputs()[0]);
         Assert.isLess(out.ratio("^\\[unknown"), 0.02, "No more than 2% of unknown frames");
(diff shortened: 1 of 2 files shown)


test/test/recovery/RecoveryTests.java
```diff
@@ -55,4 +55,12 @@ public void suppliers(TestProcess p) throws Exception {
         Assert.isGreater(out.ratio("itable stub"), 0.01);
         Assert.isGreater(out.ratio("Suppliers.loop"), 0.5);
     }
+
+    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.X64}, inputs = "")
+    @Test(mainClass = CodingIntrinsics.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.X64}, inputs = "--cstack vm", nameSuffix = "VM")
+    public void intrinsics(TestProcess p) throws Exception {
```
Baraa-Hasheesh, line 61: @apangin I've observed that this test could be quite fragile on Systems where we don't have dwarf support 

This can be observed on macOS in our GHA

I would recommend disabling this test for macOS until we have proper dwarf support in it, what do you think?

Example => https://github.com/async-profiler/async-profiler/actions/runs/16412814698/job/46371796767

apangin: Thanks for pointing this out. The problem has nothing to do with dwarf support, but it's likely that the test reveals an actual gap in the profiler's stack walking algorithm. I'll take a look.

PR #1404, opened 2025-07-23
Refactor RegularPeak test to use javaagent

Baraa-Hasheesh:
### Description
Regular Peak test is quite time sensitive & delay in `asprof` startup might cause instability in checks done in it
 https://github.com/async-profiler/async-profiler/actions/runs/16164199439/job/45622187714 

Here we write the test to use javaagent which is more stable compared to `asprof` as it will not have issues with startup time

### Related issues
N/A

### Motivation and context
Make RegularPeak test more stable

### How has this been tested?
`make test`

---


test/test/jfr/Cache.java
```diff
@@ -35,7 +33,7 @@ public String get(Long key) {
         return vt.value;
     }
 
-    private void calculateTop() {
+    protected void calculateTop() {
```
fandreuz: If the intent is to make it package-visible only, I'd rather use the default visibility:
```suggestion
    void calculateTop() {
```
`protected` kinda tells "I'm going to extend this class"

test/test/jfr/RegularPeak.java
```diff
@@ -11,37 +11,49 @@
...
+    public static void main(String[] args) throws Exception {
+        Thread thread = new Thread(RegularPeak::consistentCpuLoad);
+        thread.start();
+
+        Thread.sleep(4000);
+        cache.calculateTop();
+
+        Thread.sleep(2000);
+
+        System.exit(0);
```
fandreuz: Is this needed?

test/test/jfr/RegularPeak.java
```diff
@@ -11,37 +11,49 @@
...
-                .filter(Objects::nonNull)
-                .count();
+    public static void main(String[] args) throws Exception {
+        Thread thread = new Thread(RegularPeak::consistentCpuLoad);
+        thread.start();
+
+        Thread.sleep(4000);
+        cache.calculateTop();
+
+        Thread.sleep(2000);
```
fandreuz: Why is this needed? Can't we check the period after 4s to the JVM shut down?

test/test/jfr/RegularPeak.java
```diff
@@ -11,37 +11,49 @@
 
 /**
  * This example makes use of JFR compatible output.
- * Run with `-f profile.jfr`
+ * Run with `file=profile.jfr`
  * <p>
  * The program generates a continuous load with
- * the peaks of CPU usage every 5 seconds.
+ * the peaks of CPU usage every 4 seconds.
```
fandreuz: Is this still true? It looks like there is only one peak now, am I missing something?

test/test/jfr/RegularPeak.java
```diff
@@ -11,37 +11,49 @@
...
         }
     }
 
-    public RegularPeak() {
-        for (long id = 0; id < SIZE; id++) {
-            cache.put(id, Long.toString(ThreadLocalRandom.current().nextLong()));
+    private static void consistentCpuLoad() {
+        long count = 0;
+
+        while (true) {
```
fandreuz: I'd rather `while (!Thread.interrupted())` so the task is interruptible.

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: Yeah I forgot to update the comment here after some iterations
I initially made it still happen every 4 seconds

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: To keep in sync with original test it was taking 6s, I will update to remove the need for this sleep

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: yeah due to infinite thread, but I will change to no longer need it

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: replaced with a finite for loop of 300 iterations here [e375239](https://github.com/async-profiler/async-profiler/pull/1404/commits/e375239cb1252079c780f20af97f7177ead4f615)

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: handled here [e375239](https://github.com/async-profiler/async-profiler/pull/1404/commits/e375239cb1252079c780f20af97f7177ead4f615)

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: handled here [e375239](https://github.com/async-profiler/async-profiler/pull/1404/commits/e375239cb1252079c780f20af97f7177ead4f615)

test/test/jfr/Cache.java
Baraa-Hasheesh: applied here [e375239](https://github.com/async-profiler/async-profiler/pull/1404/commits/e375239cb1252079c780f20af97f7177ead4f615)

test/test/jfr/RegularPeak.java
Baraa-Hasheesh: [70be81b](https://github.com/async-profiler/async-profiler/pull/1404/commits/70be81bea8b8609e527b9623ac8f737f29f83da6)

test/test/jfr/RegularPeak.java
fandreuz: There seems to be a bit of skew between the [MacOS runtime](https://github.com/async-profiler/async-profiler/actions/runs/16466916386/job/46546864707?pr=1404#step:10:134) and [linux runtime](https://github.com/async-profiler/async-profiler/actions/runs/16466916386/job/46546864718?pr=1404#step:11:152). Maybe instead of 300 iterations we could iterate for a fixed amount of time?

Baraa-Hasheesh: updated here [43a19e4](https://github.com/async-profiler/async-profiler/pull/1404/commits/43a19e4a381ea72ad2b6aa8aff9835efdfd56675)

test/test/jfr/RegularPeak.java
```diff
@@ -11,37 +11,45 @@
...
     }
 
-    public RegularPeak() {
-        for (long id = 0; id < SIZE; id++) {
-            cache.put(id, Long.toString(ThreadLocalRandom.current().nextLong()));
+    private static void consistentCpuLoad() {
+        long count = 0;
+        long startTime = System.currentTimeMillis();
+
+        while (System.currentTimeMillis() - startTime < 6000) {
```
fandreuz: Do we really need 6 seconds to observe the behavior we want to observe?

Baraa-Hasheesh: The test checks for anything higher than 5 seconds
It's a good idea to keep some samples for that

Again I don't want to change the overall profiling duration the old profiler was doing, simply shifting it from `asprof` to `javaagent`

fandreuz: Fair enough

fandreuz (approved):

test/test/jfr/RegularPeak.java
apangin: Duration of the test was extended in a previous attempt to fix these macOS failures.
Now, once we switch to more reliable javaagent approach, it makes sense to shorten the test back to 4-5 seconds.

This test was originally made from a [presentation demo](https://github.com/apangin/java-profiling-presentation/blob/master/src/demo12/CacheTest.java). Now, since it is no longer a _regular_ peak, you can rename the file, too.

Make it simple. The idea of the test is to catch certain events in a specific timeframe and ensure there are no such events outside this timeframe. And, as a bonus, test `--from/--to` arguments of JfrConverter.

Baraa-Hasheesh: test recreated as test/test/jfr/CpuLoad.java here [5eb5602](https://github.com/async-profiler/async-profiler/pull/1404/commits/5eb56027bcb9abc1e13f4b5f4230f48f774ad793)

test/test/jfr/CpuLoad.java
```diff
@@ -0,0 +1,50 @@
...
+    private static void consistentCpuLoad() {
+        long startTime = System.currentTimeMillis();
+
+        while (System.currentTimeMillis() - startTime < 4000) {
+            cpuLoad();
+            sleep(20);
+        }
+    }
+
+    private static void inconsistentCpuLoad() {
```
apangin: I don't think "inconsistent" is the right term applicable to cpu load.
Maybe "normalCpuLoad" and "cpuSpike"?

test/test/jfr/CpuLoad.java
```diff
@@ -0,0 +1,50 @@
...
+        sleep(2000);
```
apangin: If this method is meant to cause a cpu spike, it should not have `sleep` in it.
Move sleep outside.

test/test/jfr/CpuLoad.java
```diff
@@ -0,0 +1,50 @@
...
+
+        for (int i = 0; i < 10; i++) {
```
apangin: Duration of the method may vary by an order of magnitude, depending on the system.
Why not use fixed duration expressed in milliseconds?

Baraa-Hasheesh: Sure, handled here [5bb4003](https://github.com/async-profiler/async-profiler/pull/1404/commits/5bb4003b1b4d03ab560a2294f6a3eab3d3995e21)

test/test/jfr/CpuLoad.java
Baraa-Hasheesh: That makes sense, updated here [5bb4003](https://github.com/async-profiler/async-profiler/pull/1404/commits/5bb4003b1b4d03ab560a2294f6a3eab3d3995e21)

test/test/jfr/CpuLoad.java
Baraa-Hasheesh: updated here [5bb4003](https://github.com/async-profiler/async-profiler/pull/1404/commits/5bb4003b1b4d03ab560a2294f6a3eab3d3995e21)

test/test/jfr/CpuLoad.java
```diff
@@ -0,0 +1,52 @@
...
+            sleep(20);
+        }
+    }
+
+    private static void cpuSpike() {
+        long startTime = System.currentTimeMillis();
+
+        while (System.currentTimeMillis() - startTime < 500) {
+            cpuLoad();
+            sleep(20);
```
apangin: This is exactly the same what `normalCpuLoad` does. For it to be a real spike, let it run without `sleep`

Baraa-Hasheesh: removed [438ac9e](https://github.com/async-profiler/async-profiler/pull/1404/commits/438ac9efa545aecc31a27d44580fb73145bdd89e)

PR #1409, opened 2025-07-24
Improve sample type information for OTEL output

fandreuz:
### Description
In this PR I improve the information provided in the fields `sample_type` and `period_type` for OTEL output.

### Related issues
n/a

### Motivation and context
There's no clear specification about this yet, but looking at the values used/expected by existing implementations made me realize we can improve what we do now.

### How has this been tested?
GHA

---

### DIFF
--- a/src/allocTracer.h
+++ b/src/allocTracer.h
@@ -30,6 +30,10 @@ class AllocTracer : public Engine {
         return "alloc_tracer";
     }
 
+    const char* otel_type() {
+        return "alloc";
+    }
+
     const char* title() {
         return "Allocation profile";
     }
--- a/src/instrument.cpp
+++ b/src/instrument.cpp
@@ -520,6 +520,10 @@ Error Instrument::check(Arguments& args) {
     return Error::OK;
 }
 
+const char* Instrument::otel_type() {
+    return _target_class;
+}
+
 Error Instrument::start(Arguments& args) {
     Error error = check(args);
     if (error) {
--- a/src/instrument.h
+++ b/src/instrument.h
@@ -23,6 +23,8 @@ class Instrument : public Engine {
         return "instrument";
     }
 
+    const char* otel_type();
+
     const char* title() {
         return "Java method profile";
     }
--- a/src/cpuEngine.h
+++ b/src/cpuEngine.h
@@ -45,6 +45,10 @@ class CpuEngine : public Engine {
         return "ns";
     }
 
+    virtual const char* otel_type() {
+        return "cpu";
+    }
+
     static void onThreadStart();
     static void onThreadEnd();
 };
--- a/src/engine.h
+++ b/src/engine.h
@@ -38,6 +38,10 @@ class Engine {
         return "noop";
     }
 
+    virtual const char* otel_type() {
+        return type();
+    }
+
     virtual const char* title() {
         return "Flame Graph";
     }
--- a/src/j9WallClock.h
+++ b/src/j9WallClock.h
@@ -30,6 +30,10 @@ class J9WallClock : public Engine {
         return "j9_wall";
     }
 
+    const char* otel_type() {
+        return "wall";
+    }
+
     const char* title() {
         return "Wall clock profile";
     }
--- a/src/lockTracer.h
+++ b/src/lockTracer.h
@@ -51,6 +51,10 @@ class LockTracer : public Engine {
         return "lock_tracer";
     }
 
+    const char* otel_type() {
+        return "lock";
+    }
+
     const char* title() {
         return "Lock profile";
     }
--- a/src/mallocTracer.h
+++ b/src/mallocTracer.h
@@ -31,6 +31,10 @@ class MallocTracer : public Engine {
         return "malloc_tracer";
     }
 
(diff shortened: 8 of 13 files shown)


src/allocTracer.h
```diff
@@ -30,6 +30,10 @@ class AllocTracer : public Engine {
         return "alloc_tracer";
     }
 
+    const char* otel_type() {
```
fandreuz: @apangin I see `type()` is only used in JFR to provide some more metadata information. Is it subject to any constraint/specification? Can we change it? I don't really like this `otel_type()` method

apangin: I don't like it either. But why should it be different from `type()`? Is there a spec that mandates certain values?

fandreuz: No spec yet, I'll ask around. These changes are based on what I see in existing implementations (mainly ebpf-profiler and Pyroscope). Should we wait to see if/when a spec is formalized?

fandreuz: I asked around, there's no spec on this. OTEL unofficially follows what [pprof does](https://github.com/google/pprof/blob/6e76a2b096b5fa52e4bb3f7f7a357bd6e6b3b7b1/proto/profile.proto#L49-L51), and pprof does not have a strong spec on this either.

apangin: Got it. It's OK to modify what `Engine::type()` returns as long as the value is different for CTimer vs. ITimer vs. PerfEvents, so that JFR contains information what engine exactly collected CPU samples.

fandreuz: 36b3d2cc53ccfe0f6e1e82c4190ff677bef7131f

src/instrument.h
```diff
@@ -7,6 +7,7 @@
 #define _INSTRUMENT_H
 
 #include <jvmti.h>
+#include "arch.h"
```
Baraa-Hasheesh, line 10: why is this needed now but not before?

src/instrument.cpp
```diff
@@ -520,6 +520,10 @@ Error Instrument::check(Arguments& args) {
     return Error::OK;
 }
 
+const char* Instrument::type() {
```
Baraa-Hasheesh: why not change it inside the header file rather than redefining here in the cpp file?

Nit: Also I don't really like the idea of an instance method returning a static class member

Baraa-Hasheesh:
@fandreuz both `event=cpu` & `event=wall` will have the same type on macOS is this intended?

Also both Live & alloc will have same type is this intended?

src/instrument.h
fandreuz: Actually that was in `instrument.cpp`, removed it from there in 156f731c9eb36f5b450658033057e6d7cfb3fe2d

I'd rather have it in the header because the header uses `u64`

fandreuz: The failure originated from a clang-tidy run, I guess because the diff only contains `instrument.h` and clang-tidy didn't know where to find `u64`

src/instrument.cpp
fandreuz: > why not change it inside the header file rather than redefining here in the cpp file?

e8a90774abb93314d05d5738968a0832f2aed15a

> Nit: Also I don't really like the idea of an instance method returning a static class member

Is there a reason why these are `static` and not normal instance fields?

fandreuz:
> both event=cpu & event=wall will have the same type on macOS is this intended?

I'd say that's not a problem, what matters is the effective engine being used.

On a second thought, @Baraa-Hasheesh pointed out that it's possible a sample is recorded with a different `EventType` depending not only on the engine type, but also on other arguments (e.g. `_mode == WALL_BATCH` for `WallClock`). 

This makes me think that we should inspect the `EventType` for a given event, like JFR does, rather than the active engine. What do you think @apangin ?

apangin:
OK, there's a sort of confusion between engine type and event.
Let's step back and define what problem you are solving exactly? What to write as `sample_type` in OTEL output, right?
A short answer is - doesn't matter as long as there is no semantics attached to this value. From a common sense, it should be something related to what is currently printed as a FlameGraph title, but in a more formal (identifier-style) way.
So, to get an idea, look at how `Engine::title()` is currently implemented.

fandreuz:
> OK, there's a sort of confusion between engine type and event. Let's step back and define what problem you are solving exactly? What to write as `sample_type` in OTEL output, right? A short answer is - doesn't matter as long as there is no semantics attached to this value. From a common sense, it should be something related to what is currently printed as a FlameGraph title, but in a more formal (identifier-style) way. So, to get an idea, look at how `Engine::title()` is currently implemented.

I think there are two different concepts we need to encode in the OTEL profile:
- [`sample_type`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L226) provides information about a specific sample value
- [`period_type`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L242) provides information about the sampler (what we call engine).

I think `sample_type` should be aware of how the specific sample is taken, so it should be a function of the `EventType`.

PR #1410, opened 2025-07-24
Prevent nativemem crash

Baraa-Hasheesh:
### Description
Currently async-profiler reads the GOT to find the original memory allocation functions to use as it's callback
This logic might fail, for example on older libc implementation there is no `aligned_alloc`

If the current lookup logic fails it will result in a hard crash.

This change checks if the symbol exists before dereferencing it to avoid dereferencing NULL
If any of the symbols is missing from the GOT entries the profiler will not run and return a proper error.

### Related issues
N/A

### Motivation and context
prevent crashes if GOT search for nativemem entries fail

### How has this been tested?
manual testing by disabling some of the CodeCache functionalites 
`make test`

---

### DIFF
--- a/src/mallocTracer.cpp
+++ b/src/mallocTracer.cpp
@@ -128,7 +128,9 @@ static void resolveMallocSymbols() {
 
 Error MallocTracer::initialize() {
     CodeCache* lib = Profiler::instance()->findLibraryByAddress((void*)MallocTracer::initialize);
-    assert(lib);
+    if (!lib) {
+        return Error("Couldn't find libasyncProfiler");
+    }
 
     resolveMallocSymbols();
 


apangin:
This does not seem the right fix to me.
`SAVE_IMPORT` macro looks in the GOT table of its own library, which we have full control of.
We should only call `SAVE_IMPORT` for the functions that we know for sure exist.

Baraa-Hasheesh:
> This does not seem the right fix to me. `SAVE_IMPORT` macro looks in the GOT table of its own library, which we have full control of. We should only call `SAVE_IMPORT` for the functions that we know for sure exist.

While I agree with this (Which is why I initially didn't add the NULL checks in the first place)
I believe that a graceful exit/stop of profiler is better than a crash in case this logic doesn't work if things go wrong, as we observed when compiling profiler on older libc that might be missing `aligned_alloc` 

Or are you suggesting doing the additional checks for only `aligned_alloc` which might be missing for older libc versions?

apangin:
The mentioned problem does not occur with the original builds of async-profiler. It only happens in the downstream version when compiled with tricky gcc options. I suggest to fix the downstream version then.

Issue #774, opened 2023-07-20
what will --alloc do when java 8?


zdyj3170101136:
seems that SetHeapSamplingInterval only work for java 11+。
so what would happen when we use java 8?
the --alloc would not useful?
so what is the default value in java 8?

apangin:
Async-profiler uses a different mechanism for allocation profiling on JDK 8. It is a bit less efficient, but still suitable for production profiling. `--alloc` is also supported, assuming the sampling granularity is not less than an average TLAB size.

zdyj3170101136:
> Async-profiler 在 JDK 8 上使用不同的分配分析机制。它的效率稍低，但仍然适合生产分析。`--alloc`假设采样粒度不小于平均 TLAB 大小，也支持。

jdk 11 use  tlab.bytes_since_last_sample_point + size_in_bytes to make 512 kb one sample.
https://github.com/openjdk/jdk11u/blob/a829804de566fb89ad0fe45d242a09db24a5867c/src/hotspot/share/gc/shared/memAllocator.cpp#L227
but in async-profiler, you are using the tlab size to  make 512 kb one sample, the tlab size is much higher than  tlab.bytes_since_last_sample_point.

zdyj3170101136:
and i also suggest implement an probablistic sampler like setheapinterval:
https://github.com/openjdk/jdk11u/blob/a829804de566fb89ad0fe45d242a09db24a5867c/src/hotspot/share/runtime/threadHeapSampler.cpp#L80

apangin:
As I mentioned, `--alloc` values are meaningful for JDK 8 when they are larger than a TLAB size. Smaller sampling intervals are not quite useful for production profiling anyway, as they would incur notable overhead.

I admit there are places for enhancement in allocation profiling on JDK 8, however, it's already good enough for production profiling (customers actually use it). With the rapidly declining popularity of JDK 8, I see little sense in investing in its further improvement.

If you have issues with allocation profiling, please open a new ticket, describing the use case in details and possibly adding a standalone reproducer. Thanks.

PR #1411, opened 2025-07-25
Add support for system wide process sampling for linux

rk-kmr:
### Description
This PR adds support for system wide process sampling for linux. Sampling implementations scans /proc directory to determine the set of processes and that reads various proc files[1] per pid to collect stats. We expose this feature by adding a `proc` command line argument. And this feature is supported for only JFR output format.

[1] Proc files read.
```
/proc/{pid}/cmdline
/proc/{pid}/stat
/proc/{pid}/status
/proc/{pid}/io
```

### Related issues
N/A

### Motivation and context
Async-profiler traditionally focuses on profiling individual Java applications, but understanding application performance often requires broader system context. This enables capturing system-wide process stats alongside application profiling, allowing users to correlate application bottlenecks with system resource contention from other processes.
comprehensive system visibility within the familiar JFR format.

### How has this been tested?
I have added tests and manually tested the different test cases under varying system load conditions.


src/jfrMetadata.cpp
```diff
@@ -256,6 +256,32 @@ JfrMetadata::JfrMetadata() : Element("root") {
                 // when encountering a T_BYTE/F_ARRAY.
                 << field("data", T_STRING, "User Data"))
 
+            << (type("profiler.ProcessSample", T_PROCESS_SAMPLE, "Process Stats Sample")
+                << category("Operating System", "Process")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
+                << field("eventThread", T_THREAD, "Event Thread", F_CPOOL)
```
apangin: Is a thread that recorded the event meaningful?
I'd assume that system-level events (like `jdk.CPULoad`) are not associated with an application thread.

src/flightRecorder.cpp
```diff
@@ -655,6 +683,98 @@ class Recording {
...
+        if (info->_cpu_percent < MIN_CPU_THRESHOLD) {
+            return false;
+        }
+
+        if (info->_mem_resident < MIN_MEMORY_THRESHOLD) {
+            return false;
+        }
+
+        return true;
+    }
```
olivergillespie: I think we should include processes which exceed either threshold, not only those which exceed both. Imagine a process using all your CPU but no memory - still interesting.

src/flightRecorder.cpp
```diff
@@ -655,6 +683,98 @@ class Recording {
...
+
+    void populateCpuPercent(ProcessInfo* info) {
+        int pid = info->_pid;
+        u64 currentCpuTotal = info->_cpu_user + info->_cpu_system;
+        u64 currentTime = info->_last_update;
+
+        ProcessHistory& history = _process_history[pid];
+        if (!history.hasHistory) {
```
olivergillespie: Does this mean the first event for every process will have the default value (`0`) for cpu usage? Is there any way to mark it as undefined or maybe skip emitting the event? Could be misleading in some circumstances, for example if a user takes the average of all the cpu values for one process this would skew it downwards. I suppose right now it would not be emitted because of the MIN_CPU_THRESHOLD, but my suggestion is to change the way that's applied so this would become an issue.

fandreuz:
Hi @rk-kmr, I noticed some of the clang-tidy checks we recently introduced in our codebase are a bit too harsh, I removed two of them in #1419.

src/flightRecorder.cpp
rohitash-kumar: I skip the first event, Its gets taken care by `flightRecorder.cpp#766` check

```
if (_last_proc_sample_time > 0 && shouldIncludeProcess(&info)) {
```

_last_proc_sample_time is 0 for first event and updated to wall time in line #774 after first event is discarded

src/flightRecorder.cpp
rohitash-kumar: Thanks for checking! I have fixed it by making the check either.

src/arguments.cpp
```diff
@@ -230,7 +231,10 @@ Error Arguments::parse(const char* args) {
                     if (_nativemem < 0) _nativemem = 0;
                 } else if (strcmp(value, EVENT_LOCK) == 0) {
                     if (_lock < 0) _lock = DEFAULT_LOCK_INTERVAL;
-                } else if (_event != NULL && !_all) {
+                } else if (strcmp(value, EVENT_PROC) == 0) {
```
apangin: Profiling events that can be specified in the `event` argument
- have stack traces;
- can be aggregated by the number of samples or by value;
- can be visualized on a flame graph.

Metrics that you collect are not "events" in this sense.

src/jfrMetadata.cpp
```diff
@@ -256,6 +256,32 @@ JfrMetadata::JfrMetadata() : Element("root") {
...
+                << field("ppid", T_INT, "Parent Process ID")
+                << field("name", T_STRING, "Process Name")
+                << field("cmdLine", T_STRING, "Command Line")
+                << field("uid", T_INT, "User ID")
+                << field("state", T_BYTE, "Process State")
+                << field("processStartTime", T_LONG, "Process Start Time")
+                << field("cpuUser", T_LONG, "User CPU Time")
+                << field("cpuSystem", T_LONG, "System CPU Time")
+                << field("cpuPercent", T_FLOAT, "CPU Percentage")
+                << field("normalizeCpuPercent", T_FLOAT, "Normalized CPU Percentage")
```
apangin: Recording already contains number of CPUs. I don't think we need to write what can be calculated.

src/jfrMetadata.cpp
```diff
@@ -256,6 +256,32 @@ JfrMetadata::JfrMetadata() : Element("root") {
...
+                << field("threads", T_INT, "Thread Count")
+                << field("memSize", T_LONG, "Virtual Memory Size")
+                << field("memResident", T_LONG, "Resident Memory")
+                << field("memShared", T_LONG, "Shared Memory")
+                << field("memText", T_LONG, "Text Memory")
+                << field("memData", T_LONG, "Data Memory")
```
apangin: I never saw these two useful.
In contrast, it's quite important to distinguish anonymous memory from file-mapped memory, but I don't see it here.

src/os_linux.cpp
```diff
@@ -430,4 +432,218 @@ bool OS::checkPreloaded() {
...
+        &parsed_pid, comm, &state, &parsed_ppid,
+        &minflt, &majflt, &utime, &stime,
+        &threads, &starttime64, &vsize, &rss);
+
+    if (parsed < 12) return false;
+
+    info->_pid           = parsed_pid;
+    info->_ppid          = parsed_ppid;
+    strncpy(info->_name, comm, sizeof(info->_name));
+    info->_name[sizeof(info->_name) - 1] = '\0';
```
apangin: Usage of `strncpy` looks suspicious to me. Why not just `memcpy`? `comm` is always NUL-terminated.

src/os_linux.cpp
```diff
@@ -430,4 +432,218 @@ bool OS::checkPreloaded() {
...
+                        "%ld "                    /*  20 num_threads                          */
+                        "%*d "                    /*  21 itrealvalue (skip)                   */
+                        "%llu "                   /*  22 starttime                            */
+                        "%lu "                    /*  23 vsize                                */
+                        "%ld",                    /*  24 rss                                  */
+        &parsed_pid, comm, &state, &parsed_ppid,
+        &minflt, &majflt, &utime, &stime,
+        &threads, &starttime64, &vsize, &rss);
+
+    if (parsed < 12) return false;
```
apangin, line 597: Why have more than 12 items in the pattern, if we only care about first 12?

src/flightRecorder.cpp
```diff
@@ -73,6 +85,16 @@ struct CpuTimes {
     CpuTime total;
 };
 
+struct ProcessHistory {
+    unsigned long prevCpuTotal;
+    unsigned long prevTimestamp;
+    bool hasHistory;
```
apangin: snake_case for struct fields.

src/flightRecorder.cpp
```diff
@@ -655,6 +685,93 @@ class Recording {
...
+
+        u64 start_time = OS::nanotime();
+        int pids[MAX_PROCESSES];
+        int pid_count = 0;
+
+        OS::getProcessIds(pids, &pid_count, MAX_PROCESSES);
+        cleanupProcessHistory(pids, pid_count);
+
+        for (int i = 0; i < pid_count; i++) {
+            if ((OS::nanotime() - start_time) > MAX_TIME_NS) {
```
apangin: Should we report incomplete cycles in the JFR file?
E.g. we can use Log::debug for this purpose.

src/main/main.cpp
```diff
@@ -75,6 +75,7 @@ static const char USAGE_STRING[] =
     "  --nofree          do not collect free calls in native allocation profiling\n"
     "  --lock duration   lock profiling threshold in nanoseconds\n"
     "  --wall interval   wall clock profiling interval\n"
+    "  --proc interval   process sampling interval\n"
     "  --all             shorthand for enabling cpu, wall, alloc, live,\n"
```
krk: We can enable proc in `--all`.

src/arguments.h
```diff
@@ -12,13 +12,15 @@
...
 const long DEFAULT_LOCK_INTERVAL = 10000;    // 10 us
+const long DEFAULT_PROC_INTERVAL = 30; // 30 seconds
 const int DEFAULT_JSTACKDEPTH = 2048;
 
 const char* const EVENT_CPU        = "cpu";
 const char* const EVENT_ALLOC      = "alloc";
 const char* const EVENT_NATIVEMEM  = "nativemem";
 const char* const EVENT_LOCK       = "lock";
-const char* const EVENT_WALL       = "wall";
+const char *const EVENT_WALL       = "wall";
```
krk: Nit: wall has a space change.

src/flightRecorder.cpp
```diff
@@ -45,6 +46,17 @@ const int RECORDING_BUFFER_LIMIT = RECORDING_BUFFER_SIZE - 4096;
 const int MAX_STRING_LENGTH = 8191;
 const u64 MAX_JLONG = 0x7fffffffffffffffULL;
 const u64 MIN_JLONG = 0x8000000000000000ULL;
+const int MAX_PROCESSES = 5000;  // Hard limit to prevent excessive work
+const u64 MAX_TIME_NS = 900000000UL; // Timeout after 900ms to guarantee runtime <1sec
+const float MIN_CPU_THRESHOLD = 0.1F;     // Minimum % cpu utilization to filter results (0 = no filtering)
+const u64 MIN_MEMORY_THRESHOLD = 1; // Minimum resident memory in pages (0 = no filtering)
```
krk: Is this bytes?

src/flightRecorder.cpp
```diff
@@ -655,6 +685,93 @@ class Recording {
         _last_gc_id = gc_id;
     }
 
+    // Helper function to determine if a process should be included based on thresholds
+    bool shouldIncludeProcess(const ProcessInfo* info) {
+        if (info->_cpu_percent >= MIN_CPU_THRESHOLD || info->_mem_resident >= MIN_MEMORY_THRESHOLD) {
+            return true;
+        }
+
+        return false;
```
krk: Can be simplified to `return info->_cpu_percent >= MIN_CPU_THRESHOLD || info->_mem_resident >= MIN_MEMORY_THRESHOLD;`

src/flightRecorder.cpp
```diff
@@ -1294,6 +1416,59 @@ class Recording {
...
+        buf->putVar32(info->_ppid);
+
+        // Limit process name length
+        size_t name_len = strlen(info->_name);
+        buf->putByteString(info->_name,
+            name_len > max_process_name_length ? max_process_name_length : name_len);
+
+        // Limit command line length (uncommented and limited)
+        size_t cmdline_len = strlen(info->_cmdline);
+        buf->putByteString(info->_cmdline,
```
krk: Do we have to report cmdline in every sample?

src/jfrMetadata.cpp
rohitash-kumar: Its not required and I wasn't populating it either. I have removed it.

src/arguments.cpp
rohitash-kumar: Got it. I have removed `proc` as event.

src/jfrMetadata.cpp
rohitash-kumar: Agreed. Removed it.

src/jfrMetadata.cpp
rohitash-kumar: Got it. I have removed these and added anonymous and files memory fields.

src/os_linux.cpp
rohitash-kumar: yeah, I wasn't sure of null termination bit. Given that scanf does ensure it. I have changed this to memcpy.

src/flightRecorder.cpp
rohitash-kumar: Thanks, I wanted this but wasn't sure on the best way to do this. Log::debug looks perfect.

src/os_linux.cpp
rohitash-kumar: Its not first 12 fields. I skip few in between thats why I have more than 12 specifiers.

src/main/main.cpp
rohitash-kumar: Fixed.

src/flightRecorder.cpp
rohitash-kumar: KB, I have pushed the commit to make units consistent for all memory fields.

src/flightRecorder.cpp
rohitash-kumar: No, We only do this for process which pass the `shouldIncludeProcess` check.

src/arguments.cpp
```diff
@@ -284,6 +289,11 @@ Error Arguments::parse(const char* args) {
                 if (_nativemem < 0) {
                     _nativemem = DEFAULT_ALLOC_INTERVAL;
                 }
+#ifdef __linux__
```
apangin: Avoid platform #ifdefs in shared code if possible.
You can often replace it with `OS::isLinux()` to make code shorter and nicer, e.g. look below at `if (_event == NULL && OS::isLinux())`, but it is even better when the code can be written in a platform-agnostic way.

src/arguments.cpp
```diff
@@ -496,6 +506,18 @@ Error Arguments::parse(const char* args) {
...
 
+#ifndef __linux__
+    // Currently we support process sampling on Linux only.
+    if (_proc >= 0) {
+        return Error("Process sampling is only supported on Linux");
+    }
+#endif
+
+    // Process sampling requires JFR output format
+    if (_proc >= 0 && _output != OUTPUT_JFR) {
```
apangin: Errors returned from `Arguments::parse` are parsing erros, but this one is more like a functional error. They are typically thrown from `Profiler::start`.
Async-profiler is tolerant to redundant parameters.

src/flightRecorder.cpp
```diff
@@ -73,6 +85,16 @@ struct CpuTimes {
...
 
+struct ProcessHistory {
+    unsigned long prev_cpu_total;
+    unsigned long prev_timestamp;
+    bool has_history;
+
+    ProcessHistory() : prev_cpu_total(0), prev_timestamp(0), has_history(false) {}
+};
+
+static std::map<int, ProcessHistory> _process_history;
```
apangin: Can be unordered_map?

src/flightRecorder.cpp
```diff
@@ -73,6 +85,16 @@ struct CpuTimes {
     CpuTime total;
 };
 
+struct ProcessHistory {
+    unsigned long prev_cpu_total;
+    unsigned long prev_timestamp;
+    bool has_history;
```
apangin: nit: `has_history` is redundant; `prev_timestamp != 0` achives the same goal

src/flightRecorder.cpp
```diff
@@ -1294,6 +1412,58 @@ class Recording {
         buf->putVar32(start, buf->offset() - start);
     }
 
+    void recordProcessSample(Buffer* buf, const ProcessInfo* info) {
+        // Estimate for fixed fields
+        const size_t process_non_string_size_limit = 200;
+        const size_t max_process_name_length = 256;
```
apangin: Isn't it 16? Or do you consider macOS too?

src/flightRecorder.cpp
```diff
@@ -1294,6 +1412,58 @@ class Recording {
...
+        const size_t max_process_cmdline_length = ASPROF_MAX_JFR_EVENT_LENGTH - process_non_string_size_limit - max_process_name_length;
```
apangin: In fact, ASPROF_MAX_JFR_EVENT_LENGTH is not a hard limit. You can have larger events if needed, as long as you flush `_proc_buf` timely. But keeping events small is a good idea anyway.

Given that `ProcessInfo` already has reasonable limits for the process name and cmdline, you can likely skip computing limits here just to simplify the code.

src/flightRecorder.cpp
```diff
@@ -1461,6 +1632,22 @@ bool FlightRecorder::timerTick(u64 wall_time, u32 gc_id) {
...
+    if (_rec->processMonitorCycle(wall_time) == true) {
+      u64 fin_time = OS::nanotime();
+      u64 delta_ns = fin_time - start_time;
+      double delta_ms = static_cast<double>(delta_ns) / 1000000.0;
+      if (delta_ms > max_ms) max_ms = delta_ms;
+      if (delta_ms < min_ms) min_ms = delta_ms;
+      sum_ms += delta_ms;
+      total += 1;
+      fprintf(stderr, "processMonitorCycle took %.3f ms\n", delta_ms);
+    };
```
apangin: 4 spaces indent
redundant `;`

src/os_linux.cpp
```diff
@@ -430,4 +432,221 @@ bool OS::checkPreloaded() {
     return dl_iterate_phdr(checkPreloadedCallback, (void*)info) == 1;
 }
 
+static inline bool isNum(const char* s) noexcept {
```
apangin: noexcept is redundant; the whole project is compiled with `-fno-exceptions`

src/os_linux.cpp
```diff
@@ -430,4 +432,221 @@ bool OS::checkPreloaded() {
...
+
+static inline int parsePid(const char* s) noexcept {
+    int pid = 0;
+    for (uint8_t c; (c = *s++); ) {
+        pid = pid * 10 + (c - '0');
+    }
+    return pid;
+}
+
+void OS::getProcessIds(int* pids, int* count, int max_pids) {
```
apangin: IMO, API would be a bit nicer if the function returned `count` instead of setting the output parameter.

src/os_linux.cpp
```diff
@@ -430,4 +432,221 @@ bool OS::checkPreloaded() {
...
+    *count = 0;
+
+    DIR* proc = opendir("/proc");
+    if (!proc) return;
+    for (dirent* de; (de = readdir(proc)) && *count < max_pids; ) {
+        if (!isNum(de->d_name)) {
```
apangin: `strtol` can replace both isNum+parsePid.

src/arguments.cpp
rohitash-kumar: Fixed.

src/arguments.cpp
rohitash-kumar: Fixed.

src/flightRecorder.cpp
rohitash-kumar: I think, I missed updating it after initial commit. Anyway I have removed this piece of logic as per comment below.

src/flightRecorder.cpp
rohitash-kumar: I like the simplification. I originally added them after looking at recordUserEvent code.

src/flightRecorder.cpp
rohitash-kumar: Fixed. I don't plan to commit this change. I am using this right now to benchmark and optimize.

src/os_linux.cpp
rohitash-kumar: I have removed these methods in favor of `strtoul`

src/os_linux.cpp
rohitash-kumar: yep, fixed.

src/os_linux.cpp
rohitash-kumar: Fixed.

src/flightRecorder.cpp
```diff
@@ -655,6 +685,86 @@ class Recording {
...
+                break;
+            }
+
+            ProcessInfo info;
+            if (OS::getBasicProcessInfo(pids[i], &info)) {
+                info._last_update = OS::nanotime();
+                populateCpuPercent(&info);
+                if (_last_proc_sample_time > 0 && shouldIncludeProcess(&info)) {
+                    OS::getDetailedProcessInfo(&info);
+                    flushIfNeeded(&_proc_buf, RECORDING_BUFFER_LIMIT - ASPROF_MAX_JFR_EVENT_LENGTH);
```
apangin: `ASPROF_MAX_JFR_EVENT_LENGTH` is for user events only, it has nothing to do with your events.
Instead, check if there is enough room to accommodate the longest possible ProcessSample.

src/flightRecorder.cpp
```diff
@@ -1294,6 +1409,41 @@ class Recording {
...
 
+    void recordProcessSample(Buffer* buf, const ProcessInfo* info) {
+        int start = buf->skip(5);
+        buf->put8(T_PROCESS_SAMPLE);
+        buf->putVar64(info->_last_update);
+
+        buf->putVar32(info->_pid);
+        buf->putVar32(info->_ppid);
+        buf->putByteString(info->_name, strlen(info->_name));
+        buf->putByteString(info->_cmdline, strlen(info->_cmdline));
```
apangin: Add an overload of `putByteString` for null-terminated strings (similar to `putUtf8`).

src/os.h
```diff
@@ -25,6 +25,48 @@ enum ThreadState {
...
+    int _ppid;
+    char _name[16];              // Process name from /proc/{pid}/stats
+    char _cmdline[2048];         // Command line from /proc/{pid}/cmdline
+    unsigned int _uid;           // User ID
+    unsigned char _state;        // Process state (R, S, D, Z, T, etc.)
+    unsigned long _start_time;   // Process start time (clock ticks since boot)
+
+    // CPU & thread stats
+    unsigned long _cpu_user;     // User CPU time (clock ticks)
+    unsigned long _cpu_system;   // System CPU time (clock ticks)
```
apangin: How would reader know what was tick frequency on the target host?
Should he values be normalized when dumping to JFR?

src/os.h
```diff
@@ -25,6 +25,48 @@ enum ThreadState {
...
 };
 
+struct ProcessInfo {
+    int _pid;
+    int _ppid;
+    char _name[16];              // Process name from /proc/{pid}/stats
+    char _cmdline[2048];         // Command line from /proc/{pid}/cmdline
+    unsigned int _uid;           // User ID
+    unsigned char _state;        // Process state (R, S, D, Z, T, etc.)
+    unsigned long _start_time;   // Process start time (clock ticks since boot)
```
apangin: Is `unsigned long` a proper type for all these values? Keep in mind that `unsigned long` is 32-bit on 32-bit systems.

src/flightRecorder.cpp
```diff
@@ -655,6 +685,86 @@ class Recording {
...
+
+        for (int i = 0; i < pid_count; i++) {
+            if ((OS::nanotime() - start_time) > MAX_TIME_NS) {
+                Log::debug("Incomplete process sampling cycle.");
+                break;
+            }
+
+            ProcessInfo info;
+            if (OS::getBasicProcessInfo(pids[i], &info)) {
+                info._last_update = OS::nanotime();
```
apangin: Sometimes, when TSC is not available/reliable, `OS::nanotime()` does an expensive syscall.
Avoid redundant calls when possible; in this case, you can reuse a timestamp taken in the beginning of the loop iteration.

rohitash-kumar: I see. I have updated as per your suggestion.

src/os.h
rohitash-kumar: Thanks! for catching it.  I am switching to unsigned long long(u64 typedef). Let me know if you’d like a different approach.

src/os.h
rohitash-kumar: good catch, I missed this. Normalization make sense to me here. I have fixed it.

src/flightRecorder.cpp
rohitash-kumar: Interesting, I interpreted `ASPROF_MAX_JFR_EVENT_LENGTH` as MAX length for "any" JFR event.  I have now defined a new constant for longest possible ProcessSample.

src/os.h
rohitash-kumar: `_start_time` has the same problem and normalization is not enough. I am exploring solution options for it.

rohitash-kumar: I have [fixed](https://github.com/async-profiler/async-profiler/pull/1411/commits/c59ae6aaa32277383cd5df09b7623269ae46f83b) _start_time as well.

src/os.h
```diff
@@ -25,6 +25,46 @@ enum ThreadState {
...
+    // Page fault stats
+    u64 _minor_faults;           // Minor page faults (no I/O required)
+    u64 _major_faults;           // Major page faults (I/O required)
+
+    // I/O stats
+    u64 _io_read;                // KB read from storage
+    u64 _io_write;               // KB written to storage
+
+    ProcessInfo() : _pid(0), _ppid(0), _uid(0), _state(0), _start_time(0),
+                    _cpu_user(0), _cpu_system(0), _cpu_percent(0.0F), _threads(0),
```
apangin: Float literals should use lowercase `f`.
I know clang-tidy bot told you the opposite, but that was a configuration mistake that we've already fixed. Sorry about that.

src/os.h
```diff
@@ -25,6 +25,46 @@ enum ThreadState {
     THREAD_SLEEPING
 };
 
+struct ProcessInfo {
+    int _pid;
```
apangin: In async-profiler code, members of data-only structs (unlike classes) usually don't start with `_`.

src/os.h
```diff
@@ -25,6 +25,46 @@ enum ThreadState {
...
+
+    // Page fault stats
+    u64 _minor_faults;           // Minor page faults (no I/O required)
+    u64 _major_faults;           // Major page faults (I/O required)
+
+    // I/O stats
+    u64 _io_read;                // KB read from storage
+    u64 _io_write;               // KB written to storage
+
+    ProcessInfo() : _pid(0), _ppid(0), _uid(0), _state(0), _start_time(0),
```
apangin: You can use initialization at declaration, e.g. `int pid = 0;`
When the project started, it did not even support C++11, but now it does.

src/jfrMetadata.cpp
```diff
@@ -256,6 +256,30 @@ JfrMetadata::JfrMetadata() : Element("root") {
...
+            << (type("profiler.ProcessSample", T_PROCESS_SAMPLE, "Process Stats Sample")
+                << category("Operating System", "Process")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
+                << field("pid", T_INT, "Process ID")
+                << field("ppid", T_INT, "Parent Process ID")
+                << field("name", T_STRING, "Process Name")
+                << field("cmdLine", T_STRING, "Command Line")
+                << field("uid", T_INT, "User ID")
+                << field("state", T_BYTE, "Process State")
+                << field("startTime", T_LONG, "Process Start Time")
```
apangin: You have two `startTime` fields. How does it even work?

src/os.h
```diff
@@ -106,6 +147,11 @@ class OS {
     static int mprotect(void* addr, size_t size, int prot);
 
     static bool checkPreloaded();
+
+    static u64 getSysBootTime();
```
apangin: This is Linux-specific time base for process counters.
I understand your feature currently works on Linux only, but it would be nice to make it more generic at least at the interface level.

src/os_linux.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <arpa/inet.h>
 #include <byteswap.h>
+#include <climits>
```
apangin: Prefer C headers (limits.h) when they are available.

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    return 0;
+}
+
+int OS::getProcessIds(int* pids, int max_pids) {
+    int count = 0;
+    DIR* proc = opendir("/proc");
+    if (!proc) return 0;
+    for (dirent* de; (de = readdir(proc)) && count < max_pids; ) {
+        unsigned long pid = strtoul(de->d_name, NULL, 10);
+        if (pid != 0 && pid != ULLONG_MAX)  {
```
apangin: How can `pid` be ULLONG_MAX?
Also, `strtoul` returns `unsigned long` whereas `ULLONG_MAX` is of `unsigned long long` type.

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    }
+    closedir(proc);
+    return count;
+}
+
+// Helper method to read command line from /proc/{pid}/cmdline
+static bool readProcessCmdline(int pid, ProcessInfo* info) {
+    char path[64];
+    snprintf(path, sizeof(path), "/proc/%d/cmdline", pid);
+    FILE* file = fopen(path, "rb");
```
apangin: When you don't need buffering and don't need to read line by line, prefer low level API (`open`/`read`) as it is slightly more efficient.

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    const size_t max_read = sizeof(info->_cmdline) - 1;
+    size_t len = fread(info->_cmdline, 1, max_read, file);
+    info->_cmdline[max_read] = '\0';
+
+    if (len > 0) {
+      // Replace null bytes with spaces (arguments are separated by null bytes)
+      for (size_t i = 0; i < len; i++) {
+          if (info->_cmdline[i] == '\0') {
+              info->_cmdline[i] = ' ';
+        }
```
apangin, line 525: Keep an eye on indents.

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
 
+    double getRssUsagePercent(const ProcessInfo &info) {
+        if (info._vm_size <= 0 || info._vm_rss <= 0) {
+            return 0.0;
+        }
+
+        return static_cast<double>(info._vm_rss) / info._vm_size * 100;
+    }
+
+    bool shouldIncludeProcess(const ProcessInfo &info) {
```
Baraa-Hasheesh: `const ProcessInfo &info` should be changed to `const ProcessInfo& info` here & everywhere better follow the async-profiler coding style

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    info->_minor_faults  = minflt;
+    info->_major_faults  = majflt;
+    info->_cpu_user      = utime;
+    info->_cpu_system    = stime;
+    info->_threads       = threads;
+    info->_start_time    = starttime;
+    // (23) vsize convert from bytes to kB
+    info->_vm_size      = vsize >> 10;
+    //(24) rss - convert from number of pages to kB
+    info->_vm_rss  = (rss * OS::page_size) >> 10;
```
apangin: Spacing should be consistent

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+            read_count++;
+            info->_rss_anon = value;
+        } else if (strcmp(key, "RssFile") == 0) {
+            read_count++;
+            info->_rss_files = value;
+        } else if (strcmp(key, "RssShmem") == 0) {
+            read_count++;
+            info->_rss_shmem = value;
+        }
+        else if (strcmp(key, "VmSize") == 0) {
```
apangin: `} else if ... {` here and below

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+        info->_io_read = -1;
+        info->_io_write = -1;
+        return false;
+    }
+
+    char line[1024];
+    while (fgets(line, sizeof(line), file)) {
+        if (strncmp(line, "read_bytes:", 11) == 0) {
+            u64 read_bytes;
+            if (sscanf(line + 11, "%llu", &read_bytes) == 1) {
```
apangin: Why not just `strtoull`?

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+            }
+        }
+    }
+
+    fclose(file);
+    return true;
+}
+
+bool OS::getBasicProcessInfo(int pid, ProcessInfo* info) {
+    *info = ProcessInfo();
```
apangin: This looks alien to me )
Is the idea to zero provided buffer but without calling `memset` (to avoid filling large `cmdline` array)?
Maybe it's the case for the _placement new_ then.

test/test/proc/ProcTests.java
```diff
@@ -0,0 +1,247 @@
...
+                .filter(e -> e.cmdLine != null && e.cmdLine.contains("BasicApp"))
+                .collect(Collectors.toList());
+
+            Assert.isEqual(appSamples.size(), 3);
+        }
+    }
+
+    @Test(mainClass = BasicApp.class, os = Os.LINUX)
+    public void processSamplingWithAllMode(TestProcess p) throws Exception {
+        Output out = p.profile("--all -d 62 -f %f.jfr", false, 65);
```
apangin: There are many long-running tests here. I don't want them to run on every commit to keep pre-integration test duration reasonable.
Make sure all tests pass then disable those that run for longer than 5 seconds. We will enable them back after #1432.

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
+            return;
+        }
+
+        u64 delta_cpu = current_cpu_total - history.prev_cpu_total;
+        u64 delta_time = sampling_time - history.prev_timestamp;
+        float cpu_percent = 0.0F;
+
+        if (delta_time > 0) {
+            float cpu_secs = static_cast<float>(delta_cpu) / OS::clock_ticks_per_sec;
+            float elapsed_secs = static_cast<float>(delta_time) / 1e9;
```
Baraa-Hasheesh: you can avoid the casting here by using `1.0e9f`

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
+            if (elapsed_secs > 0) {
+              cpu_percent = (cpu_secs / elapsed_secs) * 100.0;
```
Baraa-Hasheesh: tab size is sometimes 2 spaces & sometimes 4 spaces, can you check here & everywhere

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
+            }
+        }
+
+        info._cpu_percent = cpu_percent;
+        history.prev_cpu_total = current_cpu_total;
+        history.prev_timestamp = sampling_time;
+    }
+
+  void cleanupProcessHistory(const int* activePids, int pidCount) {
```
apangin: `active_pids`, `pid_count`

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
+        return static_cast<double>(info._vm_rss) / info._vm_size * 100;
+    }
+
+    bool shouldIncludeProcess(const ProcessInfo &info) {
+        return info._cpu_percent >= MIN_CPU_THRESHOLD || getRssUsagePercent(info) >= MIN_RSS_PERCENT_THRESHOLD;
+    }
+
+    void populateCpuPercent(ProcessInfo &info, u64 sampling_time) {
+        u64 current_cpu_total = info._cpu_user + info._cpu_system;
+        ProcessHistory& history = _process_history[info._pid];
```
Baraa-Hasheesh: While reading this logic I thought about the use case where the profiler is used on an application that triggers a fork

I'm kinda curious on how the profiler would behave when a fork happens

This is not related to this change but looks like something nice to test out in the future

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
...
+        }
+
+        return static_cast<double>(info._vm_rss) / info._vm_size * 100;
+    }
+
+    bool shouldIncludeProcess(const ProcessInfo &info) {
+        return info._cpu_percent >= MIN_CPU_THRESHOLD || getRssUsagePercent(info) >= MIN_RSS_PERCENT_THRESHOLD;
+    }
+
+    void populateCpuPercent(ProcessInfo &info, u64 sampling_time) {
```
apangin: Recorder got too much code related to process info collection and cleanup. Can we probably extract this into a separate class?

Baraa-Hasheesh: We might need to add a recommendation somewhere in our documentation to avoid enabling this feature on multiple processes in the system 

=> The system contains multiple processes each with their own async-profiler

src/flightRecorder.cpp
```diff
@@ -434,6 +469,90 @@ class Recording {
         _last_gc_id = gc_id;
     }
 
+    double getRssUsagePercent(const ProcessInfo &info) {
```
apangin: What is the physical meaning of this value? It does **not** look useful to me, as virtual memory cost nothing and a process may reserve terabytes of virtual memory thus making `vm_rss / vm_size` ratio arbitrary low.
At the same time, what _does_ matter is RSS percentage comparing to the total physical RAM on the host.

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+}
+
+int OS::getProcessIds(int* pids, int max_pids) {
+    int count = 0;
+    DIR* proc = opendir("/proc");
+    if (!proc) return 0;
+    for (dirent* de; (de = readdir(proc)) && count < max_pids; ) {
+        unsigned long pid = strtoul(de->d_name, NULL, 10);
+        if (pid != 0 && pid != ULLONG_MAX)  {
+            pids[count++] = pid;
```
Baraa-Hasheesh, line 496: you're assigning an unsigned value to a signed variable 
This might result in negative values

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    if (!file) return false;
+
+    char buffer[4096];
+    if (!fgets(buffer, sizeof(buffer), file)) {
+        fclose(file);
+        return false;
+    }
+    fclose(file);
+
+    int  parsed_pid, parsed_ppid;
```
Baraa-Hasheesh: number of spaces between variable & variable name changes from line to line

src/flightRecorder.cpp
```diff
@@ -72,6 +79,15 @@ struct CpuTimes {
     CpuTime total;
 };
 
+struct ProcessHistory {
+    u64 prev_cpu_total;
+    u64 prev_timestamp;
```
apangin: Only CPU usage is reported as delta, other counters (IO bytes, page faults) are cumulative, right? Is this intended?

src/os_linux.cpp
Baraa-Hasheesh: wouldn't it make more sense for the caller to provide his ProcessInfo,
The caller can easily declare it like `ProcessInfo info();`

src/os_linux.cpp
```diff
@@ -434,4 +437,219 @@ bool OS::checkPreloaded() {
...
+    return true;
+}
+
+bool OS::getBasicProcessInfo(int pid, ProcessInfo* info) {
+    *info = ProcessInfo();
+    info->_pid = pid;
+    return readProcessStats(pid, info);
+}
+
+bool OS::getDetailedProcessInfo(ProcessInfo* info) {\
```
Baraa-Hasheesh: I don't really like that the PID is required for `OS::getBasicProcessInfo` but you expect it inside the `ProcessInfo` for the `OS::getDetailedProcessInfo` function

src/os_macos.cpp
```diff
@@ -426,4 +427,21 @@ bool OS::checkPreloaded() {
...
+u64 OS::getSysBootTime() {
+    return 0;
+}
+
+int OS::getProcessIds(int* pids, int max_pids) {
+    return 0;
+}
+
+bool OS::getBasicProcessInfo(int pid, ProcessInfo* info) {
+    *info = ProcessInfo();
```
Baraa-Hasheesh: this isn't needed, the user should check the return code before accessing `ProcessInfo`

src/os_linux.cpp
apangin: Extra `\` at the end

test/test/proc/IoIntensiveApp.java
```diff
@@ -0,0 +1,62 @@
...
+             ByteBuffer buf = ByteBuffer.wrap(payload);
+             while (buf.hasRemaining()) {
+                 ch.write(buf);
+             }
+
+             ch.force(true);
+         }
+
+         //read
+         try(FileChannel f = FileChannel.open(tmp, ExtendedOpenOption.DIRECT)) {
```
apangin: I suggest you go through the entire PR on GitHub once again and fix all whitespace, indents, etc.
There are many minor formatting issues that hamper review on the merits.

E.g. you have `//read` and `try(` without spaces here but `// write` and `try (` just a few lines above.
`_try (FileChannel ch = FileChannel.open(` has an extra space character in front; indentation of `while (true) {` is one level off...
IDE autoformatting feature will likely help.

src/jfrMetadata.cpp
rohitash-kumar: It supposed to be processStartTime, I think introduced it [commit](https://github.com/async-profiler/async-profiler/pull/1411/commits/81345592410874d9e85382f61a755bfc56807ea3). I'll revert it back to `processStartTime`.

test/test/proc/ProcTests.java
krk: Or, could we shorten the test durations here?

src/flightRecorder.cpp
krk: Isn't this something clang-tidy bot should catch?

src/os.h
rohitash-kumar: This returns system boot time in seconds since epoch and can be generalized to other platforms.  I use the return value to calculate process start time in epoch seconds in `recordProcessSample` method, and that logic is the platform specific part. So I'll move that to `os_linux.cpp` to make it more generic.

src/flightRecorder.cpp
fandreuz: That's `clang-format` unfortunately

src/os_linux.cpp
rohitash-kumar: I'll change to c headers where available. For my curiosity, whats the is the reasoning behind this preference?

src/os_linux.cpp
rohitash-kumar: I am aware that max pid can be up to 2^22.  I added it as a defensive check because `strtoul` docs called out that it can return `ULONG_MAX`. `ULONG_MAX` vs `ULLONG_MAX` - I made copy paste mistake. I'll remove the check altogether.

src/flightRecorder.cpp
rohitash-kumar: Agreed. I'll refactor to a separate class.

test/test/proc/ProcTests.java
rohitash-kumar: There are two tests that run > 10 seconds. This tests runs with default process sampling interval (30 seconds). I would like to keep at least one test that tests with default configuration. For other test and tests < 10 seconds, I'll try to reduce the runtime.

src/flightRecorder.cpp
rohitash-kumar: Only percentage cpu usage is reported as delta, everything else is cumulative, including user and sys cpu counters. 
I wanted to keep things simple, use CPU for trend analysis and keep other counters for supplemental information.

src/flightRecorder.cpp
rohitash-kumar: I tried to come up with a memory filtering criteria based of the attributes available in ProcessInfo struct. So I went with this `vm_rss / vm_size` approach with the assumption that it should work for most cases. But if you are of opinion that second approach is more suitable, I can change it.

src/os_linux.cpp
apangin: I don't want to introduce undesired dependencies on the standard C++ library.
Initially I was thinking of [eliminating](https://github.com/async-profiler/async-profiler/issues/872) all libstdc++ dependencies, but since we've found a workaround for the issues mentioned in the ticket, I decided to leave the status quo: we'll continue using basic libstdc++ classes such as `map` and `string` but won't add new ones.

Other than that, `climits` includes `limits.h` anyway (the same holds for similar C++ headers). Why not just skip a redundant layer to reduce the total number of files involved in compilation?

BTW, HotSpot follows the same pattern. It has few usages of C++ headers, but C headers are more common.

src/flightRecorder.cpp
apangin: Sorry I missed "not" in my previous message: I wanted to say the value **does not** look useful. I suggest you replace it with percentage of the total available RAM, i.e. what is shown in `%MEM` column of the `top` output.

src/flightRecorder.cpp
apangin: OK, that's fine.

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+            return false;
+        }
+
+        const float delta_cpu = current_cpu_total - history.prev_cpu_total;
+        const u64 delta_time = sampling_time - history.prev_timestamp;
+        float cpu_percent = 0.0f;
+
+        if (delta_time > 0) {
+            float elapsed_secs = delta_time / 1.0e9f;
+            if (elapsed_secs > 0) {
```
Baraa-Hasheesh: Rather than using this if statement how about doing the following
`cpu_percent = (delta_cpu * 1.0e9f / delta_time) * 100.0;`

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+            history.prev_cpu_total = current_cpu_total;
+            history.prev_timestamp = sampling_time;
+            return false;
+        }
+
+        const float delta_cpu = current_cpu_total - history.prev_cpu_total;
+        const u64 delta_time = sampling_time - history.prev_timestamp;
+        float cpu_percent = 0.0f;
+
+        if (delta_time > 0) {
```
Baraa-Hasheesh: When is this condition false?

delta_time can never be negative as it's unsigned
So the condition is really checking for `delta_time != 0` which can be simplified as `if (delta_time)`

Now the question this should only happens if & only if `sampling_time == history.prev_timestamp`
is this even possible?

src/arguments.cpp
```diff
@@ -60,6 +60,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     live             - build allocation profile from live objects only
 //     lock[=DURATION]  - profile contended locks overflowing the DURATION ns bucket (default: 10us)
 //     wall[=NS]        - run wall clock profiling together with CPU profiling
+//     proc[=S]         - collect process stats (default: 30s)
```
Baraa-Hasheesh: IMO it would be a nice feature to create a filter similar to the thread filter
In most cases I'm assuming the user would be concerned in a subset of the system processes 
Adding a simple filter to only record information for those PIDs could be a nice addition specially for adhoc profiling via asprof command

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+
+    int fd = open(path, O_RDONLY);
+    if (fd == -1) {
+        return false;
+    }
+
+    const size_t max_read = sizeof(info->cmdline) - 1;
+    size_t len = 0;
+
+    for (;;) {
```
Baraa-Hasheesh: You can do the following here.

```
while (r = read(fd, info->cmdline + len, max_read - len)) {
    if (r > 0) {
        .....
    } else {
        .....
    }
}
```

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+    char path[64];
+    snprintf(path, sizeof(path), "/proc/%d/stat", pid);
+
+    int fd = open(path, O_RDONLY);
+    if (fd == -1) return false;
+
+    char buffer[4096];
+    size_t len = 0;
+
+    for (;;) {
```
Baraa-Hasheesh: same as above

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+        int pid = atoi(de->d_name);
+        if (pid > 0) {
+            pids[count++] = pid;
+        }
+    }
+    closedir(proc);
+    return count;
+}
+
+bool readProcessCmdline(int pid, ProcessInfo* info) {
```
Baraa-Hasheesh: is it possible to conditionally call this?

The cmd shouldn't change between samples for the same process,
If we can find an easy way to check if the PID still belongs to the same process we can skip calling this after the first call

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+    static std::unordered_map<int, ProcessHistory> _process_history;
+
+    long _sampling_interval = -1;
+    int _pids[MAX_PROCESSES];
+
+  public:
+    void cleanupProcessHistory(const int pid_count) {
+        const std::unordered_set<int> active_pid_set(_pids, _pids + pid_count);
+        for (auto it = _process_history.begin(); it != _process_history.end();) {
+            if (active_pid_set.count(it->first) == 0) {
```
Baraa-Hasheesh: what if process was destroyed and a different process was created with same PID?

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+               "%llu "                  /* 23 vsize                                 */
+               "%llu",                  /* 24 rss                                   */
+               &parsed_pid, comm, &state, &ppid, &minflt, &majflt, &utime, &stime, &threads, &starttime, &vsize, &rss);
+
+    if (parsed < 12) return false;
+
+    memcpy(info->name, comm, COMM_LEN);
+    info->pid = parsed_pid;
+    info->ppid = ppid;
+    info->state = (unsigned char)state;
```
Baraa-Hasheesh, line 602: in certain locations I'm seeing C style casts while in others I'm seeing C++ `static_cast` casts
I would prefer to use a consistent casting style,

Currently as most of async-profiler is using C style casting, I would prefer that over C++ casting style

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+    info->minor_faults = minflt;
+    info->major_faults = majflt;
+    info->cpu_user = static_cast<float>(utime) / OS::clock_ticks_per_sec;
+    info->cpu_system = static_cast<float>(stime) / OS::clock_ticks_per_sec;
+    info->threads = threads;
+    // (23) vsize convert from bytes to kB
+    info->vm_size = vsize >> 10;
+    //(24) rss - convert from number of pages to kB
```
Baraa-Hasheesh: nit: missing space after `//`

src/os_linux.cpp
```diff
@@ -434,4 +436,261 @@ bool OS::checkPreloaded() {
...
+    char line[1024];
+    char key[32];
+    u64 value;
+    while (fgets(line, sizeof(line), file) && read_count < 6) {
+        if (sscanf(line, "%31s %llu", key, &value) != 2) {
+            continue;
+        }
+
+        size_t len = strlen(key);
+        if (len && key[len - 1] == ':') {
```
Baraa-Hasheesh: why remove the ":" char here while keeping it in the `readProcessIO`

Is there a reason strncmp won't work? => `strncmp(key, "Uid", 3)`

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+};
+
+u64 ProcessSampler::_last_sample_ts = 0;
```
Baraa-Hasheesh: setting `_last_sample_ts` to an initial value of zero will guarantee that the first JFR tick always collects process samples,

Is this on purpose?

src/flightRecorder.cpp
```diff
@@ -434,6 +553,28 @@ class Recording {
         _last_gc_id = gc_id;
     }
 
+    bool processMonitorCycle(const u64 wall_time) {
```
fandreuz: Nit: Maybe this could be protected with an `#ifdef __linux__` too?

src/os.h
```diff
@@ -65,6 +96,7 @@ class OS {
   public:
     static const size_t page_size;
     static const size_t page_mask;
+    static const long clock_ticks_per_sec;
```
fandreuz, line 99: `u64`?

fandreuz: Nevermind, I see it's the return type of `sysconf(_SC_CLK_TCK)`

src/arguments.cpp
```diff
@@ -284,6 +288,11 @@ Error Arguments::parse(const char* args) {
                 if (_nativemem < 0) {
                     _nativemem = DEFAULT_ALLOC_INTERVAL;
                 }
+
+                if (_proc < 0 && OS::isLinux()) {
```
Baraa-Hasheesh, line 298: as discussed offline there is an unintended interaction between loop & proc options
Let's assume the following options `loop=2m,proc=45`

* Proc will collect PIDs & initial history at time 0 (time stamp = 0)
* Proc will collect first full sample at time 45 (Interval)
* Proc will collect second full sample at time 90 (Interval)
* Proc will collect PIDs & initial history at time 120 (time stamp = 0, profiler restarted from loop)
* Proc will collect second full sample at time 165 (Interval)

This means the sampling interval is influenced by loop,
Note this issue while exists in other profiling modes, it has much less effect due to having much lower sampling intervals for other engines

src/flightRecorder.cpp
rohitash-kumar: yes, we use first tick to establish base values. So we export to JFR from 2nd tick onwards.

src/os_linux.cpp
rohitash-kumar: That works! updated.

src/os_linux.cpp
rohitash-kumar: updated to use c style cast across the change.

src/flightRecorder.cpp
rohitash-kumar: Or should I remove `#ifdef` from `timerTick`  ?

src/flightRecorder.cpp
rohitash-kumar: I have removed it as per your suggestion.

src/arguments.cpp
rohitash-kumar: I think it can be future feature. For now, I am going to consider this outside the scope of this PR.

src/flightRecorder.cpp
apangin: The latter. It's better to avoid platform-specific ifdefs in shared code.

src/os_linux.cpp
rohitash-kumar: Updated to use open/read where applicable.

src/os_linux.cpp
rohitash-kumar: It should be possible by using PID + Process start time combination and caching cmdLine value in `ProcessHistory` struct.

src/os_linux.cpp
rohitash-kumar: This was due me being confused with C++ :)  I have removed it.

src/flightRecorder.cpp
rohitash-kumar: I think it makes sense,  What is the ideal place to put this recommendation?

src/os_linux.cpp
rohitash-kumar: I don't have any strong preference on this. I did this initially to avoid introducing extra argument. Now that I think about it, its possible to make the signature consistent for both methods i.e. with `int pid` as argument or with just `ProcessInfo* info`. Whats your preference?

src/flightRecorder.cpp
rohitash-kumar: Addressed it in [b0422ab](https://github.com/async-profiler/async-profiler/pull/1411/commits/b0422abd0bd6f5dd9ad313cf5a1fd1b21fb69d05)

src/arguments.cpp
rohitash-kumar: Thanks for catching the issue.  I have pushed a [fix](https://github.com/async-profiler/async-profiler/pull/1411/commits/b0422abd0bd6f5dd9ad313cf5a1fd1b21fb69d05) that address this issue, and start -> stop -> .. -> start scenarios as well. I tried adding the test but it looks [TestProcess](https://github.com/async-profiler/async-profiler/blob/master/test/one/profiler/test/TestProcess.java#L217) doesn't support - timestamp pattern. Let me know if there is another way to write test for loop option. I'd try that out.

src/os_linux.cpp
rohitash-kumar: I have added startTime to ProcessHistory struct as part of [b0422ab](https://github.com/async-profiler/async-profiler/pull/1411/commits/b0422abd0bd6f5dd9ad313cf5a1fd1b21fb69d05), So its possible to implement this. Its bit of tradeoff between cost of reading (with syscalls overhead) and memory cost in `ProcessHistory` struct. We do have a high threshold for CPU and memory that avoids hitting this method often.

src/arguments.cpp
Baraa-Hasheesh: I agree, I would recommend creating a follow up enhancement once this is merged

src/os.h
```diff
@@ -25,6 +25,37 @@ enum ThreadState {
     THREAD_SLEEPING
 };
 
+struct ProcessInfo {
+    int pid = 0;
```
Baraa-Hasheesh, line 29: @rk-kmr this isn't related to the change & not a blocker per say for merging.

@fandreuz can we check on how to best support this new sample in otlp format.

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+            } else {
+                ++it;
+            }
+        }
+    }
+
+    void enable(const long sampling_interval) {
+        _sampling_interval = sampling_interval;
+        u64 expected_next = _last_sample_time + sampling_interval;
+        if (_next_sample_time != expected_next) {
```
Baraa-Hasheesh: This check can be simplified to 
```
if (_sampling_interval != sampling_interval) {
       _sampling_interval = sampling_interval;
       _last_sample_time = 0;
       _process_history.clear();
}
```

Please note that the current logic  doesn't really protect against stale data,
* `start,cpu,proc=30`
* Profile for 2 minutes
* `stop`
* No profiles for for 5 minutes
* `start,cpu,proc=30`

The `_process_history` won't be cleared even by all accounts it should be cleared as it contains stale data
I would recommend changing the condition to something along the following lines 

```
if (_sampling_interval != sampling_interval || (_next_sample_time + sampling_interval) < wall_time) {
       _sampling_interval = sampling_interval;
       _last_sample_time = 0;
       _process_history.clear();
}
```

* `_sampling_interval != sampling_interval` checks for a change in the interval
* `(_next_sample_time + sampling_interval) < wall_time ` checks if the profiler was turned off for an extended duration 

Note: In the check above I'm checking for exactly one missing sample as an example, but 1 missing sample might be fine, I would recommend checking a proper safety margin at which we should consider the data in our `_process_history` stale & clear it.

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+        const int pid_count = OS::getProcessIds(_pids, MAX_PROCESSES);
+        cleanupProcessHistory(pid_count);
+        _last_sample_time = wall_time;
+        _next_sample_time = wall_time + _sampling_interval;
+        return pid_count;
+    }
+
+    bool getProcessSample(int pid_index, u64 sampling_time, ProcessInfo& info) const {
+        const int pid = _pids[pid_index];
+        if (OS::getBasicProcessInfo(pid, &info)) {
```
Baraa-Hasheesh: can be a simple return statement

```
        return OS::getBasicProcessInfo(pid, &info) && populateCpuPercent(info, sampling_time)
                    && shouldIncludeProcess(info) && OS::getDetailedProcessInfo(&info);
```

src/os_linux.cpp
Baraa-Hasheesh: I think you might need to profile both options in terms of memory usage vs CPU overhead to provide better data for this decision

src/os.h
fandreuz: As I said offline, I'm not really sure this should go into OTLP. There are other tools in OpenTelemetry to achieve what this PR is doing, and people would argue this is more of a platform monitoring than APM.

apangin: > how to best support this new sample in otlp format

Please don't.

src/os_linux.cpp
```diff
@@ -434,4 +436,252 @@ bool OS::checkPreloaded() {
...
+
+u64 OS::getSystemBootTime() {
+    static u64 system_boot_time = 0;
+
+    if (system_boot_time == 0) {
+        FILE* file = fopen("/proc/stat", "r");
+        if (!file) return 0;
+
+        char line[1024];
+        char key[6];
```
apangin: unused

src/os_linux.cpp
```diff
@@ -434,4 +436,252 @@ bool OS::checkPreloaded() {
...
 
+u64 OS::getRamSize() {
+    static u64 mem_total = 0;
+
+    if (mem_total == 0) {
+        FILE* file = fopen("/proc/meminfo", "r");
+        if (!file) return 0;
+
+        char line[1024];
+        char key[6];
```
apangin: unused

src/flightRecorder.cpp
```diff
@@ -44,6 +46,11 @@ const int RECORDING_BUFFER_LIMIT = RECORDING_BUFFER_SIZE - 4096;
 const int MAX_STRING_LENGTH = 8191;
 const u64 MAX_JLONG = 0x7fffffffffffffffULL;
 const u64 MIN_JLONG = 0x8000000000000000ULL;
+const int MAX_PROCESSES = 5000;               // Hard limit to prevent excessive work
```
apangin: Can ProcessSampler and all the related code be moved to a separate file `processSampler.h/cpp`?

src/flightRecorder.cpp
apangin: This condition looks suspicious indeed, since `_next_sample_time` is always `_last_sample_time + _sampling_interval`.

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+        }
+    }
+
+    bool shouldSample(const u64 wall_time) const {
+        return _sampling_interval > 0 && wall_time >= _next_sample_time;
+    }
+
+    static double getRssUsagePercent(const ProcessInfo& info) {
+        const u64 ram_size = OS::getRamSize();
+        if (ram_size <= 0 || info.vm_rss <= 0) return 0.0;
```
apangin: `<= 0` looks strange for unsigned values. Can be replaced with `== 0`.
I remember some static analyzer was complaining about that.

src/flightRecorder.cpp
```diff
@@ -434,6 +553,28 @@ class Recording {
...
 
+    bool processMonitorCycle(const u64 wall_time) {
+        if (!_process_sampler.shouldSample(wall_time)) return false;
+
+        const u64 deadline_ns = OS::nanotime() + MAX_TIME_NS;
+        const int process_count = _process_sampler.getSampledProcessCount(wall_time);
+        for (int pid_index = 0; pid_index < process_count; pid_index++) {
+            const u64 current_time = OS::nanotime();
+            if (current_time > deadline_ns) {
+                Log::debug("Incomplete process sampling cycle.");
```
apangin: No full stop in log messages.

src/os.h
```diff
@@ -25,6 +25,37 @@ enum ThreadState {
...
+    int ppid = 0;
+    char name[16] = {};      // Process name from /proc/{pid}/stats
+    char cmdline[2048] = {}; // Command line from /proc/{pid}/cmdline
```
apangin: Explicit zeroing of all fields does not look useful - that's what the default constructor does.
However, I think zeroing the entire large `cmdline` array is redundant.

src/flightRecorder.cpp
```diff
@@ -1068,6 +1214,41 @@ class Recording {
         buf->putVar32(start, buf->offset() - start);
     }
 
+    void recordProcessSample(Buffer* buf, const ProcessInfo* info, u64 sampling_time) {
+        int start = buf->skip(5);
+        buf->put8(T_PROCESS_SAMPLE);
+        buf->putVar64(sampling_time);
```
apangin: Event time is supposed to be in TSC ticks (and that's how JFR Metadata describes it), but you pass nanoseconds here.

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+        if (history.prev_timestamp == 0 || history.start_time != info.start_time) {
+            history.prev_cpu_total = current_cpu_total;
+            history.prev_timestamp = sampling_time;
+            history.start_time = info.start_time;
+            return false;
+        }
+
+        const float delta_cpu = current_cpu_total - history.prev_cpu_total;
+        const u64 delta_time = sampling_time - history.prev_timestamp;
+        info.cpu_percent = (delta_cpu * 1.0e9f / delta_time) * 100.0;
```
apangin: `1.0e9f` -> `1e9f`
`100.0` -> `100f`

src/flightRecorder.cpp
```diff
@@ -72,6 +79,102 @@ struct CpuTimes {
...
+
+        history.prev_cpu_total = current_cpu_total;
+        history.prev_timestamp = sampling_time;
+        return true;
+    }
+
+    int getSampledProcessCount(u64 wall_time) {
```
apangin: Not an ideal function name. It does a lot more than just returning process count.

src/jfrMetadata.cpp
```diff
@@ -256,6 +256,30 @@ JfrMetadata::JfrMetadata() : Element("root") {
...
+                << field("threads", T_INT, "Thread Count")
+                << field("vmSize", T_LONG, "Virtual Memory Size")
+                << field("vmRss", T_LONG, "Resident Memory Size")
+                << field("rssAnon", T_LONG, "Resident anonymous memory")
+                << field("rssFiles", T_LONG, "Resident file mappings")
+                << field("rssShmem", T_LONG, "Resident shared Memory")
+                << field("minorFaults", T_LONG, "Minor Page Faults")
+                << field("majorFaults", T_LONG, "Major Page Faults")
+                << field("ioRead", T_LONG, "I/O Read Bytes")
+                << field("ioWrite", T_LONG, "I/O Write Bytes"))
```
apangin: Use relevant field flags: F_UNSIGNED, F_BYTES, F_PERCENTAGE...

src/flightRecorder.cpp
apangin: In fact, you don't need to multiply by 100. Compute ratio everywhere instead of %, but annotate corresponding field in JFR metadata with `F_PERCENTAGE`. JFR viewers use these annotations to display values properly.

src/flightRecorder.cpp
Baraa-Hasheesh: @apangin  what are your thoughts about the stale data issue?

As described here https://github.com/async-profiler/async-profiler/pull/1411#discussion_r2287687979  
previously the `loop` option was indirectly modifying the sampling frequency, this was fixed in recent changes,

The fix basically converted to the engine to a static one with static data which are preserved between profiling sessions as to bypass the loop restart.

If the user uses adhoc profiling via asprof, it's quite reasonable to assume the profiler might have a long lag between profiling session.

So what should be an acceptable threshold to keep using the old historical data from previous profiling session?

src/flightRecorder.cpp
rohitash-kumar: Renamed to `sampleProcesses`

src/jfrMetadata.cpp
rohitash-kumar: Done, and I have updated the data collection code to be consistent with meta data units.

src/flightRecorder.cpp
rohitash-kumar: Updated.

src/flightRecorder.cpp
rohitash-kumar: Moved to separate files in [e2239f0](https://github.com/async-profiler/async-profiler/pull/1411/commits/e2239f0be9b219438b2c8abd0fc5fd6289094d46)

src/os.h
rohitash-kumar: Make sense, I have updated it.

src/flightRecorder.cpp
rohitash-kumar: Changed meta to use  `F_TIME_MILLIS`  and now passing milliseconds.

src/processSampler.h
```diff
@@ -0,0 +1,42 @@
...
+    void cleanupProcessHistory(int pid_count);
+
+  public:
+    void enable(long sampling_interval);
+    bool shouldSample(u64 wall_time) const;
+    int sampleProcesses(u64 wall_time);
+    bool getProcessSample(int pid_index, u64 sampling_time, ProcessInfo& info);
+};
+
+#endif // _PROCESSSAMPLER_H
```
apangin: Newline is missing.

src/processSampler.h
```diff
@@ -0,0 +1,42 @@
...
+    static double getRssUsagePercent(const ProcessInfo& info);
+    static bool shouldIncludeProcess(const ProcessInfo& info);
+    static bool populateCpuPercent(ProcessInfo& info, u64 sampling_time);
+    void cleanupProcessHistory(int pid_count);
+
+  public:
+    void enable(long sampling_interval);
+    bool shouldSample(u64 wall_time) const;
+    int sampleProcesses(u64 wall_time);
+    bool getProcessSample(int pid_index, u64 sampling_time, ProcessInfo& info);
```
apangin: `ProcessSampler::sampleProcesses` and `ProcessSampler::getProcessSample` - function names are confusing and tautological.

Maybe `ProcessSampler::sample` and `ProcessSampler::getProcessInfo`?

src/processSampler.cpp
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include "processSampler.h"
+#include <unordered_set>
+
+const float MIN_CPU_THRESHOLD = 0.05f;        // Minimum 5% cpu utilization to include results
+const float MIN_RSS_PERCENT_THRESHOLD = 5.0f; // Minimum % rss usage to include results
```
apangin: I'd also make it 0.05f (ratio instead of %) for consistency with MIN_CPU_THRESHOLD.

test/test/proc/ProcTests.java
```diff
@@ -0,0 +1,265 @@
...
+import one.jfr.JfrReader;
+import one.jfr.event.ProcessSample;
+import one.profiler.test.*;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.Comparator;
+
+public class ProcTests {
```
apangin, line 17: These tests increase build duration by 60% (+2.5 minutes). For a feature that is unlikely to be touched often, this is unacceptable. Could you please select a small subset of essential tests to be run on every commit? We'll temporary disable other until implementation of #1432.

src/arguments.cpp
apangin: Could you please also update docs/ProfilerOptions.md?

src/jfrMetadata.cpp
```diff
@@ -256,6 +256,30 @@ JfrMetadata::JfrMetadata() : Element("root") {
                 // when encountering a T_BYTE/F_ARRAY.
                 << field("data", T_STRING, "User Data"))
 
+            << (type("profiler.ProcessSample", T_PROCESS_SAMPLE, "Process Stats Sample")
+                << category("Operating System", "Process")
+                << field("startTime", T_LONG, "Start Time", F_TIME_MILLIS)
```
apangin: All other events record `startTime` in ticks.
We should use the same units so that event timestamps can be comparable to each other.
That would require additional call to `TSC::ticks()`, but that's fine.

rohitash-kumar: Done, I have reverted to timer ticks and and now using `TSC:: ticks()` in [0835595](https://github.com/async-profiler/async-profiler/pull/1411/commits/08355951ec32ca0e92230a3c1ea3b303117605ad)

src/arguments.cpp
rohitash-kumar: Updated in [0835595](https://github.com/async-profiler/async-profiler/pull/1411/commits/08355951ec32ca0e92230a3c1ea3b303117605ad)

test/test/proc/ProcTests.java
rohitash-kumar: Trimmed down to handful tests in [88df158](https://github.com/async-profiler/async-profiler/pull/1411/commits/88df158fa33380c26aa533c01e4e0d8bce6e40f2)

src/processSampler.cpp
rohitash-kumar: Addressed in [6f2f1ba](https://github.com/async-profiler/async-profiler/pull/1411/commits/6f2f1bace1dd2ccecb4622e75e307733f9b36499)

src/processSampler.h
rohitash-kumar: Agreed changed in [6f2f1ba](https://github.com/async-profiler/async-profiler/pull/1411/commits/6f2f1bace1dd2ccecb4622e75e307733f9b36499)

PR #1413, opened 2025-07-25
Add time_nanos and duration_nanos to OTLP profiles with test

adinazugravescu:
### Description
Implemented proper temporal aggregation support where consecutive dumps maintain the same time_nanos (session start) but show increasing duration_nanos, while profiler restarts generate new time_nanos values. Converted _start_time to nanosecond precision.

### Related issues
Not available

### Motivation and context
This change ensures proper temporal data interpretation across profiler sessions by providing precise timing information that allows OTLP consumers to distinguish between different profiling periods.

### How has this been tested?
Added a new test

---

### DIFF
--- a/src/otlp.h
+++ b/src/otlp.h
@@ -36,10 +36,10 @@ namespace Profile {
     const protobuf_index_t sample_type = 1;
     const protobuf_index_t sample = 2;
     const protobuf_index_t location_indices = 3;
-    const protobuf_index_t period_type = 6;
-    const protobuf_index_t period = 7;
     const protobuf_index_t time_nanos = 4;
     const protobuf_index_t duration_nanos = 5;
+    const protobuf_index_t period_type = 6;
+    const protobuf_index_t period = 7;
 }
 
 namespace ValueType {
--- a/src/otlp.h
+++ b/src/otlp.h
@@ -38,8 +38,8 @@ namespace Profile {
     const protobuf_index_t location_indices = 3;
     const protobuf_index_t period_type = 6;
     const protobuf_index_t period = 7;
-    const protobuf_index_t time_nanos = 8;
-    const protobuf_index_t duration_nanos = 9;
+    const protobuf_index_t time_nanos = 4;
+    const protobuf_index_t duration_nanos = 5;
 }
 
 namespace ValueType {
--- a/src/otlp.h
+++ b/src/otlp.h
@@ -38,6 +38,8 @@ namespace Profile {
     const protobuf_index_t location_indices = 3;
     const protobuf_index_t period_type = 6;
     const protobuf_index_t period = 7;
+    const protobuf_index_t time_nanos = 8;
+    const protobuf_index_t duration_nanos = 9;
 }
 
 namespace ValueType {
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1134,7 +1134,6 @@ Error Profiler::start(Arguments& args, bool reset) {
         _thread_names.clear();
         _thread_ids.clear();
         _first_dump = true;
-
     }
 
     // (Re-)allocate calltrace buffers
@@ -1404,11 +1403,12 @@ Error Profiler::dump(Writer& out, Arguments& args) {
             break;
         case OUTPUT_OTLP:
             dumpOtlp(out, args);
-            _first_dump = false;
             break;
         default:
             return Error("No output format selected");
     }
+    _first_dump = false;
+    
     return Error::OK;
 }
 
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1825,8 +1825,7 @@ void Profiler::stopTimer() {
 
 void Profiler::timerLoop(void* timer_id) {
     u64 current_micros = OS::micros();
-    u64 stop_micros = _stop_time;
-    u64 sleep_until = _jfr.active() ? current_micros + 1000000 : stop_micros;
+    u64 sleep_until = _jfr.active() ? current_micros + 1000000 : _stop_time;
 
     while (true) {
         {
@@ -1838,7 +1837,7 @@ void Profiler::timerLoop(void* timer_id) {
             if (_timer_id != timer_id) return;
         }
 
-        if ((current_micros = OS::micros()) >= stop_micros) {
+        if ((current_micros = OS::micros()) >= _stop_time) {
             restart(_global_args);
             return;
         }
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1751,26 +1751,26 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     out.write((const char*) otlp_buffer.data(), otlp_buffer.offset());
 }
 
-u64 Profiler::addTimeout(u64 start_micros, int timeout_sec) {
-    if (timeout_sec == 0) {
+u64 Profiler::addTimeout(u64 start_micros, int timeout) {
+    if (timeout == 0) {
         return 0x7fffffffffffffffULL;
-    } else if (timeout_sec > 0) {
-        return start_micros + (u64)timeout_sec * 1000000ULL;
+    } else if (timeout > 0) {
+        return start_micros + (u64)timeout * 1000000ULL;
     }
 
     time_t start_seconds = start_micros / 1000000ULL;
     struct tm t;
     localtime_r(&start_seconds, &t);
 
-    int hh = (timeout_sec >> 16) & 0xff;
+    int hh = (timeout >> 16) & 0xff;
     if (hh < 24) {
         t.tm_hour = hh;
     }
-    int mm = (timeout_sec >> 8) & 0xff;
+    int mm = (timeout >> 8) & 0xff;
     if (mm < 60) {
         t.tm_min = mm;
     }
-    int ss = timeout_sec & 0xff;
+    int ss = timeout & 0xff;
     if (ss < 60) {
         t.tm_sec = ss;
     }
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1133,7 +1133,6 @@ Error Profiler::start(Arguments& args, bool reset) {
         MutexLocker ml(_thread_names_lock);
         _thread_names.clear();
         _thread_ids.clear();
-        _first_dump = true;
     }
 
     // (Re-)allocate calltrace buffers
@@ -1407,7 +1406,6 @@ Error Profiler::dump(Writer& out, Arguments& args) {
         default:
             return Error("No output format selected");
     }
-    _first_dump = false;
     
     return Error::OK;
 }
@@ -1650,12 +1648,12 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
     }
 }
 
-static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const char* type, const char* units, u64 aggregation_temporality) {
+static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const char* type, const char* units) {
     using namespace Otlp;
     protobuf_mark_t sample_type_mark = otlp_buffer.startMessage(Profile::sample_type, 1);
     otlp_buffer.field(ValueType::type_strindex, strings.indexOf(type));
     otlp_buffer.field(ValueType::unit_strindex, strings.indexOf(units));
-    otlp_buffer.field(ValueType::aggregation_temporality, aggregation_temporality);
+    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative); 
     otlp_buffer.commitMessage(sample_type_mark);
 }
 
@@ -1669,9 +1667,14 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     protobuf_mark_t scope_profiles_mark = otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
     protobuf_mark_t profile_mark = otlp_buffer.startMessage(ScopeProfiles::profiles);
 
-    u64 aggregation_temporality = _first_dump ? AggregationTemporality::delta : AggregationTemporality::cumulative;
-    recordSampleType(otlp_buffer, strings, _engine->type(), "count", aggregation_temporality);
-    recordSampleType(otlp_buffer, strings, _engine->type(), _engine->units(), aggregation_temporality);
+    u64 time_nanos = (u64)_start_time * 1000000000ULL;
+    u64 duration_nanos = (u64)(time(NULL) - _start_time) * 1000000000ULL;
+
+    otlp_buffer.field(Profile::time_nanos, time_nanos);
+    otlp_buffer.field(Profile::duration_nanos, duration_nanos);
+
+    recordSampleType(otlp_buffer, strings, _engine->type(), "count");
+    recordSampleType(otlp_buffer, strings, _engine->type(), _engine->units());
 
     std::vector<CallTraceSample*> call_trace_samples;
     _call_trace_storage.collectSamples(call_trace_samples);
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1240,7 +1240,7 @@ Error Profiler::start(Arguments& args, bool reset) {
     switchThreadEvents(JVMTI_ENABLE);
 
     _state = RUNNING;
-    _start_time = time(NULL);
+    _start_time = OS::nanotime();
     _epoch++;
 
     if (args._timeout != 0 || args._output == OUTPUT_JFR) {
@@ -1406,7 +1406,6 @@ Error Profiler::dump(Writer& out, Arguments& args) {
         default:
             return Error("No output format selected");
     }
-    
     return Error::OK;
 }
 
@@ -1653,7 +1652,7 @@ static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const cha
     protobuf_mark_t sample_type_mark = otlp_buffer.startMessage(Profile::sample_type, 1);
     otlp_buffer.field(ValueType::type_strindex, strings.indexOf(type));
     otlp_buffer.field(ValueType::unit_strindex, strings.indexOf(units));
-    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative); 
+    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative);
     otlp_buffer.commitMessage(sample_type_mark);
 }
 
@@ -1667,8 +1666,8 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     protobuf_mark_t scope_profiles_mark = otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
     protobuf_mark_t profile_mark = otlp_buffer.startMessage(ScopeProfiles::profiles);
 
-    u64 time_nanos = (u64)_start_time * 1000000000ULL;
-    u64 duration_nanos = (u64)(time(NULL) - _start_time) * 1000000000ULL;
+    u64 time_nanos = _start_time;
+    u64 duration_nanos = OS::nanotime() - _start_time;
 
     otlp_buffer.field(Profile::time_nanos, time_nanos);
     otlp_buffer.field(Profile::duration_nanos, duration_nanos);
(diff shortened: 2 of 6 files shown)


src/profiler.cpp
```diff
@@ -1133,6 +1133,8 @@ Error Profiler::start(Arguments& args, bool reset) {
         MutexLocker ml(_thread_names_lock);
         _thread_names.clear();
         _thread_ids.clear();
+        _first_dump = true;
+
```
fandreuz: Newline not needed

src/profiler.cpp
```diff
@@ -1402,11 +1404,11 @@ Error Profiler::dump(Writer& out, Arguments& args) {
             break;
         case OUTPUT_OTLP:
             dumpOtlp(out, args);
+            _first_dump = false;
```
fandreuz: Let's move this below, it's good to have for all `dump` operations.

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static void main(String[] args) throws Exception {
+        AsyncProfiler profiler = AsyncProfiler.getInstance();
+        profiler.start(Events.CPU, 1_000_000);
+        
+        for (int i = 0; i < 3; i++) {
+            byte[] dump = profiler.dumpOtlp();
+            ProfilesData data = ProfilesData.parseFrom(dump);
+            Profile profile = data.getResourceProfiles(0).getScopeProfiles(0).getProfiles(0);
+            ValueType sampleType = profile.getSampleType(0);
```
fandreuz: Can you declare a `static` function which does all of this?

```java
byte[] dump = profiler.dumpOtlp();
ProfilesData data = ProfilesData.parseFrom(dump);
Profile profile = data.getResourceProfiles(0).getScopeProfiles(0).getProfiles(0);
ValueType sampleType = profile.getSampleType(0);
```
it can take a `profiler` instance.

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.ByteBuffer;
+import io.opentelemetry.proto.profiles.v1development.*;
+
+public class OtlpTemporalityTest {
+
+    public static void main(String[] args) throws Exception {
+        AsyncProfiler profiler = AsyncProfiler.getInstance();
+        profiler.start(Events.CPU, 1_000_000);
+        
+        for (int i = 0; i < 3; i++) {
```
fandreuz: If you apply the comment below, I think we don't need a for-loop anymore

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,50 @@
...
+            
+            AggregationTemporality expected = (i == 0) ? 
+                AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA : 
+                AggregationTemporality.AGGREGATION_TEMPORALITY_CUMULATIVE;
+                
+            assert sampleType.getAggregationTemporality() == expected : 
+                "After restart, dump " + i + " should be " + expected;
+        }
+        
+        System.out.println("All temporality tests passed!");
```
fandreuz: Not needed

test/test/otlp/OtlpTests.java
```diff
@@ -53,6 +53,12 @@ public void testSamples(TestProcess p) throws Exception {
         assert collapsed.contains("test/otlp/CpuBurner.main;test/otlp/CpuBurner.burn");
     }
 
+    @Test(mainClass = OtlpTemporalityTest.class, jvmArgs = "-Djava.library.path=build/lib")
+public void otlpAggregationTemporalityTest(TestProcess p) throws Exception {
+    p.waitForExit();
```
fandreuz: Please use the same indentation as other methods in this class

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class OtlpTemporalityTest {
+
+    public static void main(String[] args) throws Exception {
+        AsyncProfiler profiler = AsyncProfiler.getInstance();
+        profiler.start(Events.CPU, 1_000_000);
+
+        ValueType sampleType = getSampleType(profiler);
+        assert sampleType.getAggregationTemporality() == AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA : 
+            "First dump should be DELTA";
```
fandreuz: We don't need the assertion error messages anymore, since we're not using the for loop anymore.

```suggestion
        assert getSampleType(profiler).getAggregationTemporality() == AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA;
```

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,52 @@
...
+            "After restart, second dump should be CUMULATIVE";
+
+        sampleType = getSampleType(profiler);
+        assert sampleType.getAggregationTemporality() == AggregationTemporality.AGGREGATION_TEMPORALITY_CUMULATIVE : 
+            "After restart, third dump should be CUMULATIVE";
+
+        profiler.stop();
+    }
+
+    public static ValueType getSampleType(AsyncProfiler profiler) throws Exception {
```
fandreuz: I would rename this to something like `dumpAndGetSampleType`

fandreuz (approved):

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,36 @@
+package test.otlp;
```
fandreuz, line 5: Please copy-paste the license header from another file

test/test/otlp/OtlpTests.java
```diff
@@ -53,6 +53,12 @@ public void testSamples(TestProcess p) throws Exception {
         assert collapsed.contains("test/otlp/CpuBurner.main;test/otlp/CpuBurner.burn");
     }
 
+    @Test(mainClass = OtlpTemporalityTest.class, jvmArgs = "-Djava.library.path=build/lib")
```
fandreuz, line 56: The test is failing in GHA, could you look into it?

apangin:
It is not obvious to me how the fact that profile is dumped for the first time affects `aggregation_temporality` in the output data. IMO, these two things are not quite related; they are on different layers, and one depending on another seems to be a leaky abstraction.

Example 1:

User starts profiling (e.g. using an agent, but this does not matter), then, after a few seconds, runs `asprof dump -f test.html` just to verify that profiler is working, then after one more minute an agent gets OTLP data. In this case, OTLP profile will have `aggregation_temporality=cumulative`. If a user did not run `asprof dump`, the same profile would have `aggregation_temporality=delta`, even though the rest of the data would be identical to the first case.

Example 2:

Consider an agent that restarts profiling session and after 30 seconds calls `dumpOtlp`. Since it's a new chunk of data after restart, it will have `aggregation_temporality=delta`.
In an alternative implementation, an agents restarts profiling session and calls `dumpOtlp` immediately. It will get an empty profile with `aggregation_temporality=delta`. Then after 30 seconds the agent calls `dumpOtlp` again and gets a meaningful profile with `aggregation_temporality=cumulative`. Note that profile data itself will not be any different from a case when no immediate `dumpOtlp` call is made.
If both implementations are valid according to the proposed solution, then an agent could just emit an empty `delta` profile after every restart and let async-profiler always dump `cumulative` profile. I'm not saying this will be the right solution, this example is just to demonstrate that the logic of `_first_dump` dependency is likely flawed.

fandreuz:
Yeah, I thought about such a case. I'm not sure there's a clean solution to this problem. You could look things from this perspective: dump formats _may_ support the possibility to tell the consumer whether it's a cumulative or a delta dump. OTLP does, so we set it.  But if the first dump happens to use a format that does not support it, the information is simply lost. This seems more like misuse on the user side, which could also happen in other ways: what if the user forgets to export the packet containing the first dump after a restart?

Regarding your second example, I see just one flaw in the reasoning: profiles produced by Async-Profiler itself won't adhere to the protocol. The existence of an external agent handling the flow is implicitly assumed, which makes this solution a bit less convenient.

Other options we considered:
- set `first_dump=false` only in `dumpOtlp`: I didn't particularly like this, but it would solve the problem you highlighted in the first example
- no flag in Async-Profiler, the agent is responsible for restarting the profiler and setting `DELTA` in post-processing as necessary: this implies implementing something like `loop` in the exporter, which felt a bit like reinventing what is already there.

apangin:
My point is, `delta/cumulative` is a higher level concept, it is meaningless without baseline.
Async-profiler does not submit data itself, it does not know what was submitted previously.

If I stop profiler and then run it again after 1 hour - why it should be treated as "delta" from the previous report, if there was a long gap when no data was collected at all? Or if the agent submits 3 packets: cumulative, delta, cumulative - how will receiver recognize that the last `cumulative` report is counted from the last `delta` and not from the beginning? (without a profile timestamp which you don't currently report)

To me, a scenario which you've described as "misuse" is completely valid. Imagine I've configured an agent to report profiles to a remote server, but don't see data in the UI. To check if the issue is on the collection side, I run `asprof dump` to see if profiler is running and gathering data.

I believe the following approach makes up a simple and unambiguous solution:
1. `dumpOtlp` always produces messages with `AGGREGATION_TEMPORALITY_CUMULATIVE`;
2. Profile should include `time_nanos` and `duration_nanos` fields, where `time_nanos` is the beginning of the profiling session.
3. Consecutive dumps will have the same `time_nanos`, but when profiler restarts, `time_nanos` will change to represent the beginning of the new session. This way, we'll disambiguate when `cumulative` is aggregated from.

The example in [profiles.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/c30610041736aa5c0077b156f27b09e878b797ea/opentelemetry/proto/profiles/v1development/profiles.proto#L343-L372) describes exactly our case, if we replace "fault" with "restart".

fandreuz:
Yes, I see now that `time_nanos` + `duration_nanos` is a less ambiguous way to express the same concept. We'll amend this PR with the required changes.

I'll also check with the OTEL profiling community, maybe this aggregation temporality thing could be removed from the protocol altogether, or its semantics clarified. The eBPF profiler does not use it.

src/profiler.cpp
```diff
@@ -1406,7 +1406,7 @@ Error Profiler::dump(Writer& out, Arguments& args) {
         default:
             return Error("No output format selected");
     }
-
+
```
fandreuz: Please remove spurious spaces

src/profiler.cpp
```diff
@@ -1653,7 +1653,7 @@ static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const cha
     protobuf_mark_t sample_type_mark = otlp_buffer.startMessage(Profile::sample_type, 1);
     otlp_buffer.field(ValueType::type_strindex, strings.indexOf(type));
     otlp_buffer.field(ValueType::unit_strindex, strings.indexOf(units));
-    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative);
+    otlp_buffer.field(ValueType::aggregation_temporality, AggregationTemporality::cumulative);
```
fandreuz: Same here

src/profiler.cpp
```diff
@@ -1667,6 +1667,12 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     protobuf_mark_t scope_profiles_mark = otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
     protobuf_mark_t profile_mark = otlp_buffer.startMessage(ScopeProfiles::profiles);
 
+    u64 time_nanos = (u64)_start_time * 1000000000ULL;
```
fandreuz: I propose we change `_start_time` to contain nanoseconds (use `OS::nanotime`). Please track the usages of `_start_time` and convert to seconds as needed.

@apangin would you agree with this approach?

src/otlp.h
```diff
@@ -38,6 +38,8 @@ namespace Profile {
     const protobuf_index_t location_indices = 3;
     const protobuf_index_t period_type = 6;
     const protobuf_index_t period = 7;
+    const protobuf_index_t time_nanos = 8;
```
fandreuz: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.7.0/opentelemetry/proto/profiles/v1development/profiles.proto#L233C10-L236C28

src/profiler.cpp
```diff
@@ -1406,7 +1406,6 @@ Error Profiler::dump(Writer& out, Arguments& args) {
         default:
             return Error("No output format selected");
     }
-
```
fandreuz: Please don't remove newlines if they are not related to your changes

src/otlp.h
```diff
@@ -38,6 +38,8 @@ namespace Profile {
     const protobuf_index_t location_indices = 3;
     const protobuf_index_t period_type = 6;
     const protobuf_index_t period = 7;
+    const protobuf_index_t time_nanos = 4;
+    const protobuf_index_t duration_nanos = 5;
```
fandreuz, line 40: Let's keep the order, it makes things easier when we want to modify this file

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,46 @@
...
+        long durationNanos1 = profile1.getDurationNanos();
+        
+        Thread.sleep(100);
+        
+        Profile profile2 = dumpAndGetProfile(profiler);
+        long timeNanos2 = profile2.getTimeNanos();
+        long durationNanos2 = profile2.getDurationNanos();
+        
+        assert timeNanos1 == timeNanos2;
+        assert durationNanos2 > durationNanos1;
```
fandreuz: Let's check that this is at least 100 milliseconds bigger than `durationNanos1`

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,46 @@
...
+        
+        profiler.stop();
+        profiler.start(Events.CPU, 1_000_000);
+        
+        Profile profile3 = dumpAndGetProfile(profiler);
+        long timeNanos3 = profile3.getTimeNanos();
+        
+        assert timeNanos3 != timeNanos1;
```
fandreuz:
```suggestion
        assert timeNanos3 > timeNanos1;
```

fandreuz:
@apangin are you ok with changing `_start_time` to contain nanoseconds in `profiler.cpp`? Otherwise, we could have a separate field.

src/profiler.h
```diff
@@ -68,7 +68,7 @@ class Profiler {
     Engine* _alloc_engine;
     int _event_mask;
 
-    time_t _start_time;
+    u64 _start_time;
     time_t _stop_time;
```
apangin: `_start_time` and `_stop_time` are on the same scale. They should be measured with the same function in the same units.

src/profiler.cpp
```diff
@@ -1240,11 +1240,11 @@ Error Profiler::start(Arguments& args, bool reset) {
     switchThreadEvents(JVMTI_ENABLE);
 
     _state = RUNNING;
-    _start_time = time(NULL);
+    _start_time = OS::nanotime();
```
apangin: `time_nanos` is required to represent absolute time in nanoseconds since [epoch](https://en.wikipedia.org/wiki/Epoch_(computing)).
`OS::nanotime()` returns relative time since some undefined moment in the past. On Linux, it's OS boot time, but this does not matter. Use `OS::micros()` instead.

src/profiler.cpp
```diff
@@ -1746,7 +1752,7 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
 
 time_t Profiler::addTimeout(time_t start, int timeout) {
     if (timeout == 0) {
-        return (time_t)0x7fffffff;
+        return 0x7fffffff;
```
fandreuz: Please add the cast back

src/profiler.cpp
```diff
@@ -1819,7 +1825,7 @@ void Profiler::stopTimer() {
 
 void Profiler::timerLoop(void* timer_id) {
     u64 current_micros = OS::micros();
-    u64 stop_micros = _stop_time * 1000000ULL;
+    u64 stop_micros = _stop_time;
```
fandreuz: I guess we don't need this local variable anymore now, let's just use `_stop_time`

fandreuz (approved):

test/test/otlp/OtlpTemporalityTest.java
```diff
@@ -0,0 +1,46 @@
...
+        long durationNanos1 = profile1.getDurationNanos();
+        
+        Thread.sleep(100);
+        
+        Profile profile2 = dumpAndGetProfile(profiler);
+        long timeNanos2 = profile2.getTimeNanos();
+        long durationNanos2 = profile2.getDurationNanos();
+        
+        assert timeNanos1 == timeNanos2;
+        assert durationNanos2 - durationNanos1 >= 100_000_000L;
```
apangin, line 28: TestRunner does not currently pass `-ea` flag, so assertions do not work in the child process.
I think this should be fixed in TestRunner.

src/profiler.cpp
```diff
@@ -1240,11 +1240,11 @@ Error Profiler::start(Arguments& args, bool reset) {
     switchThreadEvents(JVMTI_ENABLE);
 
     _state = RUNNING;
-    _start_time = time(NULL);
+    _start_time = OS::micros();
     _epoch++;
 
     if (args._timeout != 0 || args._output == OUTPUT_JFR) {
-        _stop_time = addTimeout(_start_time, args._timeout);
+        _stop_time = addTimeout(_start_time / 1000000ULL, args._timeout) * 1000000ULL;
```
apangin: `addTimeout` is only called here. Perhaps its signature should be changed to accept and return time in micros instead of `time_t`.
I see you already tried this change but reverted it back, why?

fandreuz: > I see you already tried this change but reverted it back, why?

I proposed we keep the changes minimal, didn't know how you would react to the change

src/profiler.h
```diff
@@ -68,8 +68,8 @@ class Profiler {
     Engine* _alloc_engine;
     int _event_mask;
 
-    time_t _start_time;
-    time_t _stop_time;
```
apangin, line 72: There are no more usages of `time_t` in profiler.h. Does it mean `#include <time.h>` can be moved from profiler.h to .cpp?

src/profiler.h
```diff
@@ -130,7 +130,7 @@ class Profiler {
     Engine* activeEngine();
     Error checkJvmCapabilities();
 
-    time_t addTimeout(time_t start, int timeout);
+    u64 addTimeout(u64 start, int timeout);
```
apangin: Since `start` and `timeout` are now in different units, add a suffix to make it clear, e.g. `u64 start_micros, int timeout_sec`

src/profiler.cpp
```diff
@@ -1744,34 +1751,35 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     out.write((const char*) otlp_buffer.data(), otlp_buffer.offset());
 }
 
-time_t Profiler::addTimeout(time_t start, int timeout) {
-    if (timeout == 0) {
-        return (time_t)0x7fffffff;
-    } else if (timeout > 0) {
-        return start + timeout;
+u64 Profiler::addTimeout(u64 start_micros, int timeout_sec) {
```
apangin: Oops, I realized `timeout` is not always in seconds. Sorry for confusion.

src/profiler.cpp
```diff
@@ -1240,11 +1241,11 @@ Error Profiler::start(Arguments& args, bool reset) {
     switchThreadEvents(JVMTI_ENABLE);
 
     _state = RUNNING;
-    _start_time = time(NULL);
+    _start_time = OS::micros();
     _epoch++;
 
     if (args._timeout != 0 || args._output == OUTPUT_JFR) {
-        _stop_time = addTimeout(_start_time, args._timeout);
+        _stop_time = addTimeout(_start_time / 1000000ULL, args._timeout) * 1000000ULL;
```
apangin: No longer need for `/` and `*`?

src/profiler.cpp
```diff
@@ -15,6 +15,7 @@
 #include "profiler.h"
 #include "perfEvents.h"
 #include "ctimer.h"
+#include <time.h>
```
apangin: The last minor thing: do not interleave standard headers with project header files.
The recommended #include order is the following:
```
<standard c++ headers>
<standard c headers>
"currentModule.h"
"other headers"
```
all in alphabetical order.

apangin:
All good now, thanks

PR #1414, opened 2025-07-25
Per-thread flamegraph option in JFR heatmap converter

fandreuz:
### Description
In this PR I introduce a new feature in the JFR heatmap converter to have an artificial frame at the base of the flamegraph containing the thread name where the stacktrace happened.

### Related issues
#1316 

### Motivation and context
Motivation in the original issue.

### How has this been tested?
Manual testing.

image

---


src/converter/one/heatmap/MethodCache.java
```diff
@@ -97,6 +98,29 @@ public int indexForClass(int extra, byte type) {
         return method.index = methodIndex.index(method);
     }
 
+    public int indexForThread(String threadName) {
+        long methodId = (long) threadName.hashCode() << 32 | 1L << 63;
```
apangin: `threadName.hashCode()` is not unique, but `methodId` must be unique.

src/converter/one/heatmap/MethodCache.java
```diff
@@ -97,6 +98,29 @@ public int indexForClass(int extra, byte type) {
...
+        Method method = farMethods.get(methodId);
```
apangin: Copy-pasting this long non-trivial method for each synthetic method type does not seem a good idea to me.
Can we unify all these? I don't see why we need a separate `index` variant for classes too.

src/converter/one/convert/JfrToHeatmap.java
```diff
@@ -45,7 +45,7 @@ public void collect(Event event) {
                 long msFromStart = (event.time - jfr.chunkStartTicks) * 1_000 / jfr.ticksPerSec;
                 long timeMs = jfr.chunkStartNanos / 1_000_000 + msFromStart;
 
-                heatmap.addEvent(event.stackTraceId, extra, type, timeMs);
+                heatmap.addEvent(event.stackTraceId, getThreadName(event.tid), extra, type, timeMs);
```
apangin: This will create lots of garbage even when thread names are not used.

src/converter/one/convert/Frame.java
```diff
@@ -15,6 +15,7 @@ public class Frame extends HashMap<Integer, Frame> {
     public static final byte TYPE_CPP = 4;
     public static final byte TYPE_KERNEL = 5;
     public static final byte TYPE_C1_COMPILED = 6;
+    public static final byte TYPE_THREAD = 7;
```
apangin: These types should match types used in C++ code.
Fortunately, we don't need a special type or special color for thread frames - just reuse `TYPE_NATIVE` as the original FlameGraph converter does.

Baraa-Hasheesh:
@fandreuz if `record-cpu` feature is used we will have the artificial thread frame as the lowest point in the heatmap

IMO the lowest frame should be the CPU frame

image

fandreuz:
> @fandreuz if `record-cpu` feature is used we will have the artificial thread frame as the lowest point in the heatmap
> 
> IMO the lowest frame should be the CPU frame
> 
> image

@apangin what do you think? Both visualizations seems to have some sense: what @Baraa-Hasheesh suggests is probably more logical, on the other hand I kinda like that the thread (which can run on multiple CPUs at different times) is at the bottom of the graph and is not split.

src/converter/one/convert/JfrToHeatmap.java
fandreuz: Can you elaborate? `getThreadName` seems like a read-only operation

src/converter/one/heatmap/MethodCache.java
fandreuz: Technically there's no way to construct a unique ID, this frame lives in a common pool together with standard stack frame, whose ID is out of our control. We could generate a random ID until we find one which is available, am I missing something?

On the other hand, do we _really_ need a unique `methodId`? Conflicts are permitted in the pool, that's what the linked list is for. Then we disambiguate by checking the fields of the `Method` instance.

src/converter/one/convert/Frame.java
fandreuz: 30e5195fb839a2a99df3f231ab867fc13571d4b4

apangin:
Per-CPU aggregation on the outer level would make more sense; it will also match profiler's built-in flamegraph writer.
However, the problem is that CPU number is embedded as a frame in stacktrace, unlike thread ID which is recorded in a separate field. This means, CPU frame is indistinguishable from other native frames in JFR format (except by name). So, I'd prefer to avoid special cases for parsing CPU frames and treat them as normal frames. This is a very niche feature anyway - I doubt anyone is going to use CPU numbers with per-thread heatmaps, and therefore I suggest to go for a simpler approach.

src/converter/one/heatmap/MethodCache.java
fandreuz: I gave it a shot in 9266d84a9e9ced5c0def78a982b84e59bfaab1e8, but I don't particularly like it. I think the flow is harder to follow and methods get too many parameters, what do you think?

src/converter/one/convert/JfrToHeatmap.java
apangin: `getThreadName` constructs a new String in the heap every time.

fandreuz: Ah I see, fixed in 85dc551ac7b8874683dc59ddb5d2a8412d278c9b

src/converter/one/heatmap/MethodCache.java
apangin: Your `indexForThread` can return an index of a previously saved `Method` for threadName, whose hashCode is equal to the hashCode of the argument, even when their names differ.

apangin: Note: you already have thread IDs (unique among recorded threads), which are later converted to thread names and then hashed. Why not using original thread IDs instead?

fandreuz: > Your indexForThread can return an index of a previously saved Method for threadName, whose hashCode is equal to the hashCode of the argument, even when their names differ.

Yeah I understand, isn't this check enough to protect against this occurrence?

```java
if (method.originalMethodId == methodId && method.location == -1 && method.type == Frame.TYPE_NATIVE 
      && method.methodName == threadNameIdx) {
   return method.index;
}
```

> Note: you already have thread IDs (unique among recorded threads), which are later converted to thread names and then hashed. Why not using original thread IDs instead?

If you have a strong opinion I'll use the thread ID. However, correct me if I'm wrong, the problem you want to solve is potential id collisions:
1. this problem is not solved by using thread id to generate the identifier, because there are other objects in the pool
2. this problem is already taken care of by the check above, see the last piece of the AND condition.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -396,34 +396,43 @@ private static class State {
...
             }
+            if (threadName != null) {
+                cachedStackTrace[0] = methodsCache.indexForThread(threadName);
+                System.arraycopy(prototype, 0, cachedStackTrace, 1, prototype.length);
+            } else {
+                System.arraycopy(prototype, 0, cachedStackTrace, 0, prototype.length);
+            }
+            if (extra != 0) {
+                cachedStackTrace[stackSize - 1] = methodsCache.indexForClass(extra, type);
+            }
```
krk: This would fill `methodsCache.nearCache` faster than before, especially if there are many threads.

How long does it take to convert a profile with enough methods and thousands of threads to overflow the `nearCache`, with and without the per-thread option?

fandreuz: Can you elaborate? The new frames do not touch `nearCache`, they go directly into `farMethods`. The behavior is the same as what we are doing in `indexForClass`

fandreuz:
I removed `MethodCache`, these are the results for a JFR which I generated as follows:
```
% java -agentpath:./build/lib/libasyncProfiler.so=start,timeout=1000,jfr,file=shoot.jfr -jar renaissance-mit-0.16.0.jar db-shootout -t 1020
% du -sh shoot.jfr
16M	shoot.jfr
```

`master`:
```
dev-dsk-fandreuz-1d-94125930 % java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.431 s

dev-dsk-fandreuz-1d-94125930 % java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.417 s

dev-dsk-fandreuz-1d-94125930 % java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.448 s
```

Current `HEAD`:
```
% java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.457 s

% java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.48 s

% java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 0.476 s
```

We get some additional runtime by removing `MethodCache`, but I think the saving in terms of code is significant.

fandreuz:
Another test: a `4.8Gb` JFR built with async-profiler `--all`.

```
// master
% java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr shoot_master.html
Converting shoot.jfr -> shoot_master.html # 3.349 s

// per-thread-flamegraph
% java -jar build/jar/jfr-converter.jar -o heatmap shoot.jfr shoot_ptf.html
Converting shoot.jfr -> shoot_ptf.html # 3.718 s

% java -jar build/jar/jfr-converter.jar -o heatmap --threads shoot.jfr shoot_ptf.html
Converting shoot.jfr -> shoot_ptf.html # 6.106 s
```

`--threads` is relatively inefficient indeed, I guess it's because `addEvent` is doing much more stuff:
image

When `--threads` is not used, `addEvent` does almost nothing, which makes sense:
image

@apangin can we accept these numbers?

apangin:
> can we accept these numbers?

They are not too terrible, but there is a large room for improvement.
Try to avoid symbolic resolution on the fast path. That's what MethodCache did. The idea was to simplify MethodCache by using a regular HashMap, but not to get rid of it entirely. As a key in this HashMap, you may use a simple POJO of two `longs`: one is `methodId`, and another encodes `location`, `type` and `firstInStack`.

apangin:
Note that you may reuse the same HashMap for caching thread frames (by thread id) without resolving thread names prematurely.

fandreuz:
@apangin it's a bit better after a6c4a4e4b76269c833eb95b2af2070eaa5bf60fe

```
% java -jar build/jar/jfr-converter.jar --threads -o heatmap shoot.jfr
Converting shoot.jfr -> shoot.html # 4.982 s
```

fandreuz:
image

src/converter/one/heatmap/Heatmap.java
```diff
@@ -395,35 +388,55 @@ private static class State {
...
 
-        public void addEvent(int stackTraceId, int extra, byte type, long timeMs) {
+        public void addEvent(int stackTraceId, int threadId, int extra, byte type, long timeMs) {
             if (sampleList.getRecordsCount() >= LIMIT) {
                 return;
             }
 
             int prototypeId = stackTracesCache.get(stackTraceId);
-            if (extra == 0) {
+            if (extra == 0 && !arguments.threads) {
```
apangin: `args` as it is called everywhere else.
`arguments.threads` is just too verbose.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -395,35 +388,55 @@ private static class State {
...
+                cachedStackTrace[0] = threadFrameIndex;
+                System.arraycopy(prototype, 0, cachedStackTrace, 1, prototype.length);
+            } else {
+                System.arraycopy(prototype, 0, cachedStackTrace, 0, prototype.length);
+            }
+
+            if (extra != 0) {
+                long id = (long) extra << 32 | 1L << 63;
+                MethodKey key = new MethodKey(id, -1, type, false);
+                Integer threadFrameIndex = getMethodIndex(key, () -> createExtraMethod(key, extra));
```
apangin: `createExtraMethod` is a vague name.
`extra` is currently used only for classes, right? So, rename `extra` to `classId`, `createExtraMethod` to `createClassMethod` etc.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -395,35 +388,55 @@ private static class State {
...
             }
 
             int[] prototype = stackTracesRemap.get(prototypeId);
-            int stackSize = prototype.length + 1;
+            int stackSize = prototype.length + (arguments.threads ? 1 : 0) + (extra != 0 ? 1 : 0);
             if (cachedStackTrace.length < stackSize) {
                 cachedStackTrace = new int[stackSize * 2];
             }
+            if (arguments.threads) {
+                long id = (long) threadId << 32 | 1L << 63;
```
apangin: Why is `1L << 63` needed?

src/converter/one/heatmap/Heatmap.java
```diff
@@ -395,35 +388,55 @@ private static class State {
...
+                MethodKey key = new MethodKey(id, -1, Frame.TYPE_NATIVE, true);
+                Integer threadFrameIndex = getMethodIndex(key, () -> createThreadMethod(key, threadId));
+                cachedStackTrace[0] = threadFrameIndex;
+                System.arraycopy(prototype, 0, cachedStackTrace, 1, prototype.length);
+            } else {
+                System.arraycopy(prototype, 0, cachedStackTrace, 0, prototype.length);
```
apangin: I'd move `System.arraycopy` out of `if` to make the sequence more obvious:
```
// 1. add Thread frame if needed
// 2. copy frames from the prototype stack
// 3. add Class frame if needed
```

src/converter/one/heatmap/Heatmap.java
fandreuz: I couldn't figure it out. It looks like it's trying to reduce key conflict, but I don't see how it would succeed in doing so. I didn't really want to touch it, but since you don't see the reasoning either I propose to remove it: 9d947e0c51cc333b84d8718ddf348fcf0dd4b786

src/converter/one/heatmap/Heatmap.java
fandreuz: 590f60eb376f671d0529d21b34d999e27df004a0

src/converter/one/heatmap/Heatmap.java
fandreuz: e52686319563efda01bea0b18446abac2ce98f35

src/converter/one/heatmap/Heatmap.java
fandreuz: eaa733a0ed2200ff1bcada06d43af75a1ec4a092

src/converter/one/heatmap/Heatmap.java
apangin: I'm not sure - that was a genuine question, not a request to remove ) Unique IDs were needed in previous implementation of MethodCache, but I don't see this is required anymore. The same for `<< 32` btw.

Also, afaict, `originalMethodId` is no longer needed too - can you double check? And what about `hashCode` and `equals`?

src/converter/one/heatmap/Heatmap.java
```diff
@@ -395,35 +387,53 @@ private static class State {
...
+                MethodKey key = new MethodKey(threadId, -1, Frame.TYPE_NATIVE, true);
+                Integer threadFrameIndex = getMethodIndex(key, () -> createThreadMethod(key, threadId));
+                cachedStackTrace[0] = threadFrameIndex;
+            }
+
+            System.arraycopy(prototype, 0, cachedStackTrace, args.threads ? 1 : 0, prototype.length);
+
+            if (classId != 0) {
+                MethodKey key = new MethodKey((long) classId << 32, -1, type, false);
+                Integer threadFrameIndex = getMethodIndex(key, () -> createClassMethod(key, classId));
```
apangin: It's not `threadFrameIndex`. Anyhow, I'd remove a temporary variable altogether. In the above `if` too.

fandreuz: Right, thanks: 4804b5dbca15adf5d01591a09f3e8c263d756a9f

src/converter/one/heatmap/Heatmap.java
fandreuz: Right, `originalMethodId` is not needed anymore: 644d23e1c3b38d9d3ec71fd0928033255a9301a3

`hashCode` and `equals` are though, because `Method` is used inside an `Index<Method>` which is backed by an `HashMap`.

> The same for << 32 btw.

I wanted to keep it, the idea is thread ID is going to be in the range [0, 32768], and `classId` is an integer. This way, at least we now for sure there won't be key conflicts between thread frames and class frames.

apangin: Right, hashCode/equals are essential, my bad.
I see your point about ID collision, but I think it should be resolved in a different way. I have an idea, let me check something...

apangin: Alright, what I am not fully happy about in the current proposal is that you first customize parameters of the `MethodKey` depending on the frame kind (thread/class/method), then pass lambda that extracts these parameters back and creates a specialized version of `Method` object.

Problems:
1. For each frame kind, you need custom logic to create `MethodKey` and need to take care that IDs do not overlap.
2. The logic to produce `Method` from `MethodKey` is separated from the data and is passed as an additional lambda.

I propose one of two solutions:
1. Make `MethodKey` an abstract class with the only virtual method to produce `Method` from it. Create 3 descendant classes: `JavaMethodKey`, `ThreadKey` and `ClassKey` with the fields required for this particular type of frame and with its own implementation of `createMethod`.
2. Leave a single `MethodKey` but add a new field that will be used to distinguish frame kinds (Java method / thread / class). Use `switch` over this type to produce proper `Method`.

In both cases the code should be cleaner and faster than it is now.

fandreuz: I tried nr 2, but I don't particularly like it: 167bbc627f6d8550aa08c8673f845af889add9f0

- `MethodKey` now has to be non-static to refer to the fields in `Heatmap`
- `MethodKey` got a new field which encodes somewhat duplicate information as `type`, though neither can really replace the other
- more code
- it seems like we're hiding the information about the frame type, just to get it back in the next line when we call `getMethodIndex`, rather than taking a more direct approach like what I was doing with the lambda.

Maybe I got your idea wrong? nr 1 seemed too much OOP complication for the simple things we're doing here.

> In both cases the code should be cleaner and faster than it is now.

Why faster?

apangin: > but I don't particularly like it

It's good that you have this feeling :) I don't like it either, because it's not exactly what I meant. Let's take a step further and make the code nicer.

I agree that `MethodKey` should not be non-static. It is mostly a holder for frame parameters, the only place where it needs access to `JfrConverter` and `symbolTable` is `makeMethod`, so caller of `makeMethod` can pass those.

We did a good job to get rid of obscure `extra` in the last iteration, but now you've introduced it back for some reason. Each frame kind already has its unique identifier: `methodId`, `classId` or `threadId`. It can be encoded as `methodId` in `MethodKey`. Together with frame kind (or `keyType` in your implementation) it makes a globally unique frame ID.

> Why faster?

Mainly because there is no longer a capturing lambda. Now, there is a single `MethodKey` object created on the fast path, which can be likely scalar-replaced by JIT (this may require some further tricks, though).

fandreuz: Right, looks better now: 52a9689d443c577fdcab73eba6a8c76315587c5d

src/converter/one/heatmap/Heatmap.java
```diff
@@ -440,12 +447,90 @@ public void addStack(long id, long[] methods, int[] locations, byte[] types, int
...
+                MethodKey key = new MethodKey(MethodKeyType.METHOD, methodId, location, type, firstFrameInStack);
+                Integer threadFrameIndex = getMethodIndex(key);
+                cachedStackTrace[index] = threadFrameIndex;
             }
 
             stackTracesCache.put(id, stackTracesRemap.index(cachedStackTrace, size));
         }
 
+        private Integer getMethodIndex(MethodKey key) {
+            return methodCache.computeIfAbsent(key, this::makeMethod);
```
apangin: Good. Now comes the trick: instead of `this::makeMethod` pass a lambda/method reference cached in `final Function<MethodKey, Integer>` field, and you'll get 5% performance improvement out of thin air.

apangin: Alternatively, you can replace `computeIfAbsent` with `get+put` pair without lambda and get the same effect.

fandreuz: cb506dd8bdcca7c8d789ff084ec3d18d181dd3ad

I had a quick look at the flamegraph and it looks a tiny bit faster, but why can't the JVM make this optimization on its own? I don't see the difference between a `final Function<MethodKey, Integer>` and the method reference. Is it just the additional reference to the class?

src/converter/one/heatmap/Heatmap.java
```diff
@@ -440,12 +447,91 @@ public void addStack(long id, long[] methods, int[] locations, byte[] types, int
...
+
+                MethodKey key = new MethodKey(MethodKeyType.METHOD, methodId, location, type, firstFrameInStack);
+                cachedStackTrace[index] = getMethodIndex(key);
             }
 
             stackTracesCache.put(id, stackTracesRemap.index(cachedStackTrace, size));
         }
 
+        private Integer getMethodIndex(MethodKey key) {
+            Integer methodIdx = methodCache.get(key);
```
apangin: `methodCache` is never populated?

fandreuz: Right, thanks: 06fc30e30ddfdc22f8d621314721b2184af6fbc5

src/converter/one/heatmap/Heatmap.java
apangin: > but why can't the JVM make this optimization on its own

`Function` object created by `this::someMethod` depends on the instance. JVM does do per-instance caching. I can understand why: let's say there are 10 method references in the code - this would require JVM to add 10 extra synthetic fields, which may or may not be used, for _every_ instance of an otherwise small object.

So, it's up to developer to decide whether to cache non-static lambdas or not.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -440,12 +447,93 @@ public void addStack(long id, long[] methods, int[] locations, byte[] types, int
...
+                boolean firstFrameInStack = firstMethodInTrace && !args.threads;
+
+                MethodKey key = new MethodKey(MethodKeyType.METHOD, methodId, location, type, firstFrameInStack);
+                cachedStackTrace[index] = getMethodIndex(key);
             }
 
             stackTracesCache.put(id, stackTracesRemap.index(cachedStackTrace, size));
         }
 
+        private Integer getMethodIndex(MethodKey key) {
```
apangin: Why `Integer`? Return value is always used as `int`.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -440,12 +447,93 @@ public void addStack(long id, long[] methods, int[] locations, byte[] types, int
...
+            Integer methodIdx = methodCache.get(key);
+            if (methodIdx != null) return methodIdx;
+            methodIdx = makeMethod(key);
+            methodCache.put(key, methodIdx);
+            return methodIdx;
+        }
+
+        private Integer makeMethod(MethodKey key) {
```
apangin: `int`. Or just inline this function in the above method.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -440,12 +447,93 @@ public void addStack(long id, long[] methods, int[] locations, byte[] types, int
...
+            private final long methodId;
+            // 32 bits: location
+            // 8 bits: type
+            // 1 bit: firstInStack
+            private final long metadata;
+            // Used to infer what type of method to create
+            private final MethodKeyType keyType;
+
+            public MethodKey(MethodKeyType keyType, long methodId, int location, byte type, boolean firstInStack) {
+                if (type < 0) throw new IllegalArgumentException("Unexpected type: " + type);
```
apangin: Unnecessary limitation. Just replace `(long) type` with `(type & 0xffL)`.

src/converter/one/heatmap/Heatmap.java
fandreuz: 568425b5c66d23fddf9a0f7e17a1dca9183c5600

src/converter/one/heatmap/Heatmap.java
fandreuz: 568425b5c66d23fddf9a0f7e17a1dca9183c5600

src/converter/one/heatmap/Heatmap.java
fandreuz: 568425b5c66d23fddf9a0f7e17a1dca9183c5600

src/converter/one/heatmap/Heatmap.java
```diff
@@ -28,24 +24,19 @@ public class Heatmap {
...
+    public void addEvent(int stackTraceId, int threadId, int classId, byte type, long timeMs) {
+        state.addEvent(stackTraceId, threadId, classId, type, timeMs);
     }
 
     public void addStack(long id, long[] methods, int[] locations, byte[] types, int size) {
         state.addStack(id, methods, locations, types, size);
     }
 
-    public void beforeChunk() {
-        state.methodsCache.clear();
```
apangin, line 43: Seems like this corrupted frames in multi-chunk recordings.
Cleaning method cache between chunks is essential.

fandreuz: How did you find this out? I converted a 4gb JFR and the output looked ok. I'd assume it contains more than one chunk, will check

fandreuz: 14c4a2af5e282b4b7a5cdbd94433a838dc1e9e76

fandreuz: 14c4a2af5e282b4b7a5cdbd94433a838dc1e9e76

src/converter/one/convert/JfrToHeatmap.java
```diff
@@ -32,28 +32,28 @@ protected EventCollector createCollector(Arguments args) {
...
+                heatmap.addEvent(event.stackTraceId, event.tid, classId, type, timeMs);
             }
 
             @Override
             public void beforeChunk() {
-                heatmap.beforeChunk();
                 jfr.stackTraces.forEach(new Dictionary.Visitor<StackTrace>() {
                     @Override
                     public void visit(long key, StackTrace trace) {
+                        heatmap.beforeChunk();
```
apangin: for each stacktrace??

fandreuz: Dumb mistake, sorry. 9d970af314e86fa653c704609d9b5cfc5a39295b

src/converter/one/heatmap/Heatmap.java
apangin: Works like a charm now, thanks!

Issue #1412, opened 2025-07-25
Crash on SafeAccess::load


Baraa-Hasheesh:
### Describe the bug

while refactoring the `regularPeak` test, I observed a random JVM crash while running the following

```
package test.jfr;

import java.util.concurrent.ThreadLocalRandom;

public class CpuPeak {

    private static void cpuLoad() {
        double sum = 0;
        for (int i = 0; i < 1000; i++) {
            sum += Math.pow(ThreadLocalRandom.current().nextDouble(5), ThreadLocalRandom.current().nextDouble(5));
        }
    }

    private static void consistentCpuLoad() {
        long startTime = System.currentTimeMillis();

        while (System.currentTimeMillis() - startTime < 5000) {
            cpuLoad();
            try {
                Thread.sleep(20);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
        }
    }

    private static void peakCpuLoad() {
        try {
            Thread.sleep(3000);
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }
        cpuLoad();
    }

    public static void main(String[] args) throws Exception {
        Thread thread1 = new Thread(CpuPeak::consistentCpuLoad);
        thread1.start();

        peakCpuLoad();

        thread1.join();
    }
}

```

### Expected vs. actual behavior

async-profiler shouldn't crash

### Reproduction Steps

The crash happened only once randomly

### Additional Information/Context

[hs_err_pid49030.log](https://github.com/user-attachments/files/21429533/hs_err_pid49030.log)

### Async-profiler version

latest

### Environment details

macOS ARM
Java: 17.0.15-amzn

apangin:
Most often, memory access failures result in SIGBUS on macOS, but sometimes SIGSEGV is raised instead.
Async-profiler previously handled SIGBUS only. The suggested fix is to install crash handler both for SIGBUS and SIGSEGV. That's what HotSpot does too.

Baraa-Hasheesh:
Fixed in #1427

PR #1420, opened 2025-07-28
#1412: Handle both SIGBUS and SIGSEGV on macOS

apangin:
### Description

Avoid crashes on macOS by handling SIGSEGV the same way as SIGBUS.

### Related issues

#1412

### Motivation and context

Most often, memory access failures result in SIGBUS on macOS, but sometimes SIGSEGV is raised instead.
Async-profiler previously handled SIGBUS only. The suggested fix is to install crash handler both for SIGBUS and SIGSEGV. That's what HotSpot does too.

---

### DIFF
--- a/src/os_macos.cpp
+++ b/src/os_macos.cpp
@@ -97,6 +97,8 @@ JitWriteProtection::~JitWriteProtection() {
 
 
 static SigAction installed_sigaction[32];
+static SigAction orig_sigbus_handler;
+static SigAction orig_sigsegv_handler;
 
 const size_t OS::page_size = sysconf(_SC_PAGESIZE);
 const size_t OS::page_mask = OS::page_size - 1;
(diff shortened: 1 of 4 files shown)


src/os_macos.cpp
```diff
@@ -231,8 +231,19 @@ SigAction OS::replaceCrashHandler(SigAction action) {
     struct sigaction sa;
     sigaction(SIGBUS, NULL, &sa);
     SigAction old_action = sa.sa_sigaction;
+
+    // It's not well specified when macOS raises SIGBUS and when SIGSEGV.
```
Baraa-Hasheesh, line 233: The change is understood & make sense 

However, this function used to return the old signal handler to the user, in this case we replaced two signal handlers but only returned only one to the user

This makes reverting this operation not possible.

Baraa-Hasheesh: Also as a more serious issue we call the original signal handler inside profiler.cpp

If both original signal handlers are different this would result in a change in behavior in our side

apangin: See lines 238-239

Baraa-Hasheesh: Make sense
Should we add this as one of the known limitations for async profiler?

Baraa-Hasheesh: @apangin I'm just wondering here, should we even allow the profiler to run if it fails to install the crash handlers?

Won't that increase the likelihood of hard crashing the JVM during the stack walking activities?

Other than that the change looks good to me

apangin:
The problem appears more complicated.
I'm discarding this PR in favor of #1427.

PR #1422, opened 2025-07-30
Proactively check classpath in tests spawning a child process

fandreuz:
### Description
The recently introduced test `OtlpTests.otlpAggregationTemporalityTest` spawns a child process. If a `NoClassDefFoundError` happens in the child, it's harder to detect for the test runner.

### Related issues
#1413

### Motivation and context
The expected behavior is to see the test listed in the `SKIP (missing JAR)` category.

Instead, we got this:
```
INFO: Running OtlpTests.otlpAggregationTemporalityTest...
Error: Unable to initialize main class test.otlp.OtlpTemporalityTest
Caused by: java.lang.NoClassDefFoundError: com/google/protobuf/MessageOrBuilder
FAIL [1/1] OtlpTests.otlpAggregationTemporalityTest took 0.748 s
java.lang.AssertionError
        at test.otlp.OtlpTests.otlpAggregationTemporalityTest(OtlpTests.java:61)
        at one.profiler.test.Runner.run(Runner.java:124)
```

### How has this been tested?
Manual testing.

---

### DIFF
--- a/test/test/otlp/OtlpTests.java
+++ b/test/test/otlp/OtlpTests.java
@@ -55,6 +55,8 @@ public class OtlpTests {
 
     @Test(mainClass = OtlpTemporalityTest.class, jvmArgs = "-Djava.library.path=build/lib")
     public void otlpAggregationTemporalityTest(TestProcess p) throws Exception {
+        classpathCheck();
+
         p.waitForExit();
         assert p.exitCode() == 0;
     }
@@ -110,5 +112,13 @@ public class OtlpTests {
     private static void assertCloseTo(long value, long target, String message) {
         Assert.isGreaterOrEqual(value, target * 0.75, message);
         Assert.isLessOrEqual(value, target * 1.25, message);
+    }
+
+    // Simple check to make sure the classpath contains the required dependencies.
+    // If not, it will throw a NoClassDefFoundError, which will be caught by the test runner.
+    // It's useful to anticipate the same NoClassDefFoundError to happen in the child process,
+    // where it's harder to detect.
+    private static void classpathCheck() {
+        ProfilesData.newBuilder();
     }
 }


test/test/otlp/OtlpTests.java
```diff
@@ -111,4 +113,12 @@ private static void assertCloseTo(long value, long target, String message) {
         Assert.isGreaterOrEqual(value, target * 0.75, message);
         Assert.isLessOrEqual(value, target * 1.25, message);
     }
+
+    // Simple check to make sure the classpath contains the required dependencies.
+    // If not, it will throw a NoClassDefFoundError, which will be caught by the test runner.
+    // It's useful to anticipate the same NoClassDefFoundError to happen in the child process,
+    // where it's harder to detect.
+    private static void classpathCheck() {
```
Baraa-Hasheesh, line 121: Shouldn't this be done on the Runner side or inside the `TestProcess` class?

The child process is still spawned it's just reported as skipped

fandreuz: How would `TestProcess` or the runner know about what to check for?

Baraa-Hasheesh: ok, I see the issue here, this can be tricky

We can solve this by adding this to the `Test` interface

`Class<? extends Supplier<Boolean>> dependencyChecks() default TrueSupplier.class;`

However this has 2 problems: kinda unnecessarily convoluted & will only be used for this test method 

If it's between adding this extension or keep it as is I would vote to keep it as the current implementation

@apangin what do you think about this?

fandreuz: The reason why I haven't done this in the first place is that, as you said, this would only be useful in a single test method. It would add some complexity to the runner/annotations, and I remember having a similar discussion with @apangin for a very similar problem. The outcome was that if the change is only needed for a small subset of tests, a local solution is preferable.

apangin: I always said that building "architecture" or "framework" just for one instance is a futile job.
I like solution proposed in this PR as it is simple and perfectly services its purpose. We can revisit it later once we have more examples like this, but this hopefully won't be needed.

Issue #1406, opened 2025-07-23
JVM application pauses when using asprof


Baraa-Hasheesh:
### Describe the bug

When async-profiler is attached to the JVM process it will prevent any safe point from happening until it's fully loaded.

This loading includes the parsing of shared objects & unwind tables.

This means it's theoretically possible for large applications that have hundreds of shared objects to not see any safe points for a long time  

In a matter of fact this can already be seen on macOS by purging the cache via `sudo purge`& running asprof on a JVM process 

While testing I found it's possible to cause the safe points to not happen for more than ~2 seconds!! on macOS X86 (can be worse when having more shared objects)

For some of the logs I collected please refer to:
https://github.com/Baraa-Hasheesh/async-profiler/tree/debug-cold-start/logs

This in turn might cause the entire application to freeze until the async-profiler is fully loaded which (waiting for the next safe point to happen which async-profiler prevents).

### Expected vs. actual behavior

async-profiler shouldn't freeze the java application it's trying to profile

### Reproduction Steps

can be more easily observed on macOS (still relevant for Linux) as it loads more shared objects by default compared to Linux

`sudo purge`
`make test TESTS=AllocTests.alloc`

### Async-profiler version

latest

PR #1423, opened 2025-07-30
Fix issue where asprof prevents safe points for a long time

Baraa-Hasheesh:
### Description
When zinit is executed it will try to load/parse all shared objects, 
If the profiler is launched via `asprof`, this will cause the JVM to potentially not have any safe points for a long time,
This in turn might cause the entire application to freeze until the profiler is fully loaded

This change prevents that by only loading/parsing the shared objects for libjvm only inside zinit by default
This in turn makes the loading of profiler much faster for java applications 

Note1: If the async-profiler is PRELOADED the shared objects are still fully parsed via ` startProfiler` which is desired

Note2: This doesn't remove parse of all symbols when profiler runs in Agent mode, as the profiler needs to parse all that information to be able to effectually profile the startup of the process (Check VmEntry.cpp)

### Related issues
#1406

### Motivation and context
Fix issue where the profiler prevents safe points from happening for a long time when using `asprof`

### How has this been tested?
Manual testing

---


src/symbols_linux.cpp
```diff
@@ -743,6 +745,12 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
             continue;
         }
 
+        // Skip non target libs
+        const size_t file_name_len = strlen(map.file());
+        if (library && (file_name_len < library_name_len || strcmp(map.file() + (file_name_len - library_name_len), library) != 0)) {
```
fandreuz: I'd also add that `map.file()[file_name_len - library_name_len - 1] == '/'` if `file_name_len - library_name_len > 0`.

src/symbols_linux.cpp
```diff
@@ -743,6 +745,12 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
...
+            continue;
+        }
+
```
fandreuz, line 753: You can add `if (library) break;` after L781 to scrape some more ms

src/symbols_linux.cpp
```diff
@@ -743,6 +745,12 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
             continue;
         }
 
+        // Skip non target libs
+        const size_t file_name_len = strlen(map.file());
```
fandreuz: Nit: not needed when `library == nullptr`

src/symbols_macos.cpp
```diff
@@ -179,11 +179,25 @@ static std::unordered_set<const void*> _parsed_libraries;
...
 
+    if (array->count() >= MAX_NATIVE_LIBS) {
+        return;
+    }
+
+    const size_t library_name_len = library ? strlen(library) : 0;
+
+    uint32_t images = _dyld_image_count();
     for (uint32_t i = 0; i < images; i++) {
+        const char* path = _dyld_get_image_name(i);
```
fandreuz, line 194: Same 3 comments as for `symbols_linux.cpp` here

src/zInit.cpp
```diff
@@ -33,10 +33,12 @@ class LateInitializer {
 
   private:
     bool checkJvmLoaded() {
+        const char* jvm_lib = OS::isLinux() ? "libjvm.so" : "libjvm.dylib";
```
fandreuz: If I understand correctly, the fix consists in delaying parsing other shared libraries than `libjvm.so` to a subsequent call to `dlopen`, which will trigger our hook and call `updateSymbols`.

If so, do we have a guarantee that the next `dlopen` is going to happen relatively soon after `asprof` is called?

fandreuz: And assuming the above is true, do you foresee troubles if the next `dlopen` happens in a Java thread via `System.loadLibrary`? I think we would observe a similar long safepoint pause.

Baraa-Hasheesh: This fix delays the parsing of other shared objects into one of the following flows
* **VM::init**
* **Profiler::start**

No need for a subsequent call to dlopen as that would make the profiler inherently unreliable in terms of stack walking capabilities.

Regarding what's mentioned I don't believe it's a true statement, as the problem is relevant to the `jattach` process & how it loads the profiler into the JVM rather than the `System.loadLibrary` call

Now regarding the possible `System.loadLibrary` shortly after `asprof` it's not relevant to us due to order of execution 

https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1199

The `dlopen` hook is only installed after we finish parsing the symbols [here](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1193)

src/symbols_macos.cpp
Baraa-Hasheesh: I don't understand what's required/the problem here

Are you suggesting creating a common function for both?
If so it would be a function to check if a string ends with a specific sequence of characters, as that what you expect?

src/symbols_linux.cpp
Baraa-Hasheesh: sadly that's not so straight forward
Each shared object is constructed from multiple lines,
We need to read all lines to correctly update the `lib.map_end` 

While it's possible to check that library was fully parsed by checking that the inode moved, but that would be kinda convoluted so I opted against adding that check 

```
/workplace/bhashesh/async-profiler:fix-jvm-pause ✗ ➭ cat /proc/self/maps     
00400000-0040b000 r-xp 00000000 103:01 4185                              /usr/bin/cat
0060a000-0060b000 r--p 0000a000 103:01 4185                              /usr/bin/cat
0060b000-0060c000 rw-p 0000b000 103:01 4185                              /usr/bin/cat
12397000-123b8000 rw-p 00000000 00:00 0                                  [heap]
7f1425557000-7f142c126000 r--p 00000000 103:01 20726                     /usr/lib/locale/locale-archive
7f142c126000-7f142c2ca000 r-xp 00000000 103:01 3013                      /usr/lib64/libc-2.26.so
7f142c2ca000-7f142c4c9000 ---p 001a4000 103:01 3013                      /usr/lib64/libc-2.26.so
7f142c4c9000-7f142c4cd000 r--p 001a3000 103:01 3013                      /usr/lib64/libc-2.26.so
7f142c4cd000-7f142c4cf000 rw-p 001a7000 103:01 3013                      /usr/lib64/libc-2.26.so
7f142c4cf000-7f142c4d3000 rw-p 00000000 00:00 0 
7f142c4d3000-7f142c4f7000 r-xp 00000000 103:01 64490                     /usr/lib64/ld-2.26.so
7f142c6e5000-7f142c6e7000 rw-p 00000000 00:00 0 
7f142c6f6000-7f142c6f7000 r--p 00023000 103:01 64490                     /usr/lib64/ld-2.26.so
7f142c6f7000-7f142c6f8000 rw-p 00024000 103:01 64490                     /usr/lib64/ld-2.26.so
7f142c6f8000-7f142c6f9000 rw-p 00000000 00:00 0 
7ffd0d8bc000-7ffd0d8de000 rw-p 00000000 00:00 0                          [stack]
7ffd0d927000-7ffd0d92b000 r--p 00000000 00:00 0                          [vvar]
7ffd0d92b000-7ffd0d92d000 r-xp 00000000 00:00 0                          [vdso]
ffffffffff600000-ffffffffff601000 r-xp 00000000 00:00 0                  [vsyscall]

```

fandreuz: I see, missed that

src/symbols_macos.cpp
fandreuz: I meant I have the same code review comments here

src/symbols_linux.cpp
Baraa-Hasheesh: I will change the logic to be more inline with the `Profiler::findLibraryByName` function

src/symbols_macos.cpp
Baraa-Hasheesh: Handled here [5e182fb](https://github.com/async-profiler/async-profiler/pull/1423/commits/5e182fb7c7c8c18341bcf7692057b71175b6256b)

src/symbols_linux.cpp
Baraa-Hasheesh: handled here [5e182fb](https://github.com/async-profiler/async-profiler/pull/1423/commits/5e182fb7c7c8c18341bcf7692057b71175b6256b)

src/symbols_linux.cpp
Baraa-Hasheesh: handled here [5e182fb](https://github.com/async-profiler/async-profiler/pull/1423/commits/5e182fb7c7c8c18341bcf7692057b71175b6256b)

src/zInit.cpp
apangin: Yeah, that is a good callout by @fandreuz.
As part of the original change that introduced `checkJvmLoaded()`, initial library parsing was moved from `VM::init` and `Hooks::init` to `zInit`. Now, after library parsing has been removed from `zInit`, it needs to materialize back somewhere on the initialization path, otherwise it may affect OpenJ9 and native applications.

apangin: > as the problem is relevant to the jattach process & how it loads the profiler into the JVM rather than the System.loadLibrary call

@Baraa-Hasheesh What makes you think so? (I'm not challenging the statement, just want to understand what it's based on).

Baraa-Hasheesh: @apangin yeah I just realized this logic might cause issues for non hotspot JVMs

I will fix for those

Baraa-Hasheesh: > > as the problem is relevant to the jattach process & how it loads the profiler into the JVM rather than the System.loadLibrary call
> 
> @Baraa-Hasheesh What makes you think so? (I'm not challenging the statement, just want to understand what it's based on).

It's mainly based on the fact that I didn't observe the issue when using API mode when I did my testing (API mode uses System.loadLibrary).

And I specifically modified the async-profiler to take +10 seconds to load it self during the `zinit` call

Baraa-Hasheesh: Tested locally on IBM JVM J9 [b484247](https://github.com/async-profiler/async-profiler/pull/1423/commits/b484247688959cadbf4756b0c4adc6e46d4f4dad)

Baraa-Hasheesh: @apangin for native application we have 2 use cases:
* API where the user is responsible for calling the `asprof_init` which calls `Hooks::init`
* PRELOAD where the profiler is started as part of `zinit` which in turn parses all of the libraries on startup 

If desired I can move the update symbols to be among the first things done during the `start` call, 
Currently it's done after some checks are done but before checks that actually need parsing the symbols

src/symbols_linux.cpp
```diff
@@ -798,7 +808,7 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     }
 
     std::unordered_map<u64, SharedLibrary> libs;
-    collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count());
+    collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count(), library);
```
apangin: `max_count` should be 1 when parsing `libjvm.so`.
As you noticed, `max_count` logic is currently broken - let's fix it then.

src/symbols_macos.cpp
```diff
@@ -179,11 +179,27 @@ static std::unordered_set<const void*> _parsed_libraries;
...
 }
 
-void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
+void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols, const char* library) {
     MutexLocker ml(_parse_lock);
-    uint32_t images = _dyld_image_count();
 
+    if (array->count() >= MAX_NATIVE_LIBS) {
+        return;
+    }
```
apangin, line 189: Why did you add it here?

src/zInit.cpp
apangin: First, you choose library name here depending on the operating system, and then, in platform-dependent code, you look for this library name. Since it will always be libjvm, and I don't forsee need for special handling of any other library, we can probably simplify code by passing `libjvmOnly` flag instead of an arbitrary library name.

src/symbols_linux.cpp
Baraa-Hasheesh: Makes sense

src/zInit.cpp
Baraa-Hasheesh: I wanted to create something more reusable in the future in case we ever need to load a different library other than libjvm during zinit (as well)

Using a simple flag should be fine as you suggested currently as I also doubt we will need to parse any other library other than libjvm so I'm fine with using a flag as you suggested 

I will change it to using a flag as suggested 
-> if the use case even arises we would just rename the flag & use it to load the needed libs

src/symbols_macos.cpp
Baraa-Hasheesh: Adding it here will save you 2 system calls to get the image count & image path when the condition is true as well as some string operations

(Same as Linux, as this Check is also done early in that file as well)

apangin: This condition never happens in a normal application, and two system calls is not a big deal for a rare situation.

In any case, I would understand if the check is done (1) before the loop and after adding new library OR (2) before adding a new library like it was done before.

Baraa-Hasheesh: option 1 makes more sense for me compared to option 2 as it saves potentially many system calls for the non normal applications & doesn't really create overhead for the normal applications

src/zInit.cpp
Baraa-Hasheesh: updated to use a flag here [8ca44b7](https://github.com/async-profiler/async-profiler/pull/1423/commits/8ca44b70dd39be8d1079a4a4414433420d69f663)

src/symbols_macos.cpp
Baraa-Hasheesh: updated here [8ca44b7](https://github.com/async-profiler/async-profiler/pull/1423/commits/8ca44b70dd39be8d1079a4a4414433420d69f663)

src/symbols_linux.cpp
```diff
@@ -723,7 +723,7 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
     fclose(f);
 }
 
-static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count) {
+static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count, bool only_lib_jvm) {
```
Baraa-Hasheesh, line 726: Note: While rechecking this code, I found an old issue that exists in collectSharedLibraries

On the last entry, the `lib.map_end` won't be correct as it will decrement the counter on the first `lib.file == nullptr`
This means we don't read the true `map_end` for the last shared object being read,

This issue was never noticed as in normal programs we never reach the 2048 limit,

This issue makes the `findLibraryByAddress` less accurate, this needs to be fixed for this change

=> Suggested fix, don't decrement the counter, but check if an addition was done on `libs.size() == max_count`

Baraa-Hasheesh: fixed here https://github.com/async-profiler/async-profiler/pull/1423/commits/5a08516f8549a226dd9872324407facf5cf0fd86

krk:
Is there a way to optimize parsing the symbols instead? Do we have flamegraphs of the parser being slow?

Baraa-Hasheesh:
> Is there a way to optimize parsing the symbols instead

I'm open for suggestions, but either way it would be out of scope as the bug would still exist & we should minimize it's effects 

> Do we have flamegraphs of the parser being slow

Sadly this happens before the profiler runs,
The easiest way to reproduce this was to cause a large amount of cache misses for loading the shared object content into memory by purging the cache 

LD_PRELOADing a different profiler kinda warms up the cache so that's out of the window. 

A different way to reproduce this is to load hundreds of shared objects that have millions of symbols 
But doing that randomly on java isn't really safe without artificially creating those shared objects to be safe which I didn't do as the cache purge reproducer was good enough for my testing

krk:
> Sadly this happens before the profiler runs,

We can always profile the profiler, to generate profiles for the inferior profiler.

Baraa-Hasheesh:
> > Sadly this happens before the profiler runs,
> 
> We can always profile the profiler, to generate profiles for the inferior profiler.

Again I would say that this is out of scope for this change, as this change isn't about making the profiler start faster, it's more about making the `dlopen` of the profiler happen faster for Java applications

As that operation when invoked via an `asprof` agent can cause the JVM to not have any safe points until it's completed, which in turn might freeze the JVM as safe points might not happen for a long period of time.

No matter how good the start up of the profiler is, we should minimize the effect of this problem which is what's this change does 

Any enhancements on start up time can be done in a different change.
-> for example it's possible to parse multiple shared objects in parallel

apangin:
I've re-implemented the fix a bit differently in #1769.
- At load time, it parses not just libjvm, but all OpenJ9 libraries and also the async-profiler library itself. This helps to avoid an extra scan during `VM::init`.
- I tried to keep changes in `Symbols::parseLibraries` minimal: now it's only one extra `isEssentialLibrary` check.

PR #1425, opened 2025-07-31
Add `system_name` to OTEL profile output

fandreuz:
### Description
In this PR I fill the field [`system_name`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L499) for OTEL profile output.

### Related issues
#1378 

### Motivation and context
We currently don't fill this field.

### How has this been tested?
New test and GHA.

---


src/profiler.cpp
```diff
@@ -1681,16 +1691,17 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     _call_trace_storage.collectSamples(call_trace_samples);
 
     std::vector<size_t> location_indices;
-    location_indices.reserve(call_trace_samples.size());
+    location_indices.reserve(call_trace_samples.size() > 0 ?
```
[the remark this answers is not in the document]
fandreuz: Thanks!

src/profiler.cpp
```diff
@@ -1721,25 +1730,29 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+    MethodMap method_map;
+    Lookup method_lookup(&method_map, Profiler::instance()->classMap(), nullptr, &strings, OUTPUT_OTLP);
+    // Write function_table and location_table. For now, there's a 1-1 mapping between functions
+    // and locations (we don't provide address information).
+    frames.forEachOrdered([&] (size_t idx, const ASGCT_CallFrame& frame) {
+        MethodInfo* mi = method_lookup.resolveMethod(frame);
+
         protobuf_mark_t function_mark = otlp_buffer.startMessage(ProfilesDictionary::function_table, 1);
-        otlp_buffer.field(Function::name_strindex, strings.indexOf(function_name));
+        // TODO: FrameName::name produces a better looking result than Lookup
```
apangin, line 1734: What does look better exactly and can we have this functionality in Lookup?

src/lookup.cpp
```diff
@@ -150,7 +151,11 @@ bool Lookup::fillJavaMethodInfo(MethodInfo* mi, jmethodID method, bool first_tim
         mi->_name = _symbols->indexOf(method_name);
 
         if ((err = jvmti->GetClassSignature(method_class, &class_name, NULL)) == 0) {
-            mi->_class = _classes->lookup(class_name + 1, strlen(class_name) - 2);
+            std::string class_name_clean = std::string(class_name + 1, strlen(class_name) - 2);
```
apangin: This adds an unnecessary allocation even when the format is JFR.

src/profiler.cpp
fandreuz: One thing for example, `FrameName` handles different styles (signature, no semi-colon, ...)

Maybe `Lookup` could take an instance of `FrameName`?

src/lookup.cpp
fandreuz: 162293c020d71746c016dbe465ef84a3bd2a58ff

src/profiler.cpp
fandreuz: @apangin I looked into the possibility of merging `Lookup` and `FrameName`. It looks like conceptually they're doing very similar things (symbolization), but in a slightly different manner.

I'd say it's possible (not easy) to achieve something like the following:
- `Lookup` produces `MethodInfo`
- `FrameName` consumes `MethodInfo` and presents it "nicely" (e.g. user style)

So `FrameName` is concerned about how we show things to the user, and `Lookup` does the heavy lifting of resolving frames. Would that be ok for you?

Note that `MethodInfo` contains indexes to strings, rather than strings.

src/lookup.cpp
```diff
@@ -108,6 +108,7 @@ void Lookup::fillNativeMethodInfo(MethodInfo* mi, const char* name, const char*
 
     mi->_modifiers = 0x100;
 
+    mi->_system_name = _symbols->indexOf(name);
```
Baraa-Hasheesh, line 111: The mangled name is not really readable
I think it would be better to save `_symbols->indexOf(Demangle::demangle(name, true));` here

What do you think?

=> Note, I'm fine with saving the mangled name as long as it's clearly documented in our future otlp documentation, as to notify the user he needs to demangle it to get useful info


Also shouldn't this `_system_name` only be set for otlp as we do in java frames?

```
http://cplusplus.com/reference/string/string/string/
from c-string: Copies the null-terminated character sequence (C-string) pointed by s.
```

fandreuz: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L494

Baraa-Hasheesh: @fandreuz sure, but the `mi->_system_name = _symbols->indexOf(name);` should only be done for otlp no?
As this creates extra allocations that aren't needed for JFR

fandreuz: Good point, will do for the next iteration

fandreuz:
Moving to drafting while I try to reduce duplication between `frameName.cpp` and `lookup.cpp`.

PR #1427, opened 2025-08-04
Rework crash protection

apangin:
### Description

Revised implementation of the mechanism for loading from potentially unsafe locations (aka SafeAccess). Previous implementation was fragile and worked incorrectly in certain cases.

### Summary of changes

1. Previous implementation assumed that the load instruction was always within 16 bytes from the beginning of the function. This was not true for debug builds. I added a label that explicitly marks the end of the critical code.
2. Label also protects from specialization of `SafeAccess` for constant arguments. This is needed when compiling with clang, which does not support `noclone` attribute.
3. Fixed clang issue when a seemingly unused argument was optimized away.
4. If a native application did not install custom SIGSEGV handler, async-profiler could loop forever in its own handler. Now it restores SIG_DFL to pass control to the OS crash handler.
5. macOS can raise either SIGSEGV or SIGBUS. Previously, async-profiler handled SIGBUS only; now it handles both.
6. Unified `SafeAccess::load` functions. Now there are just 2 variants of `load` and a single `skipLoad`.
7. Added unit tests for SafeAccess.
8. Removed gcov, since instrumentation spoiled code in incompatible way.

### Related issues

#1412

### How has this been tested?

Added SafeAccess unit tests.
Compiled with gcc and clang with different levels of optimizations.

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -65,10 +65,10 @@ jobs:
           case "${{ matrix.runson.display }}" in
             macos*)
               brew install gcovr
-              make COMMIT_TAG=$HASH FAT_BINARY=true release coverage -j
+              make COMMIT_TAG=$HASH FAT_BINARY=true release -j
             ;;
             *)
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release coverage -j
+              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release -j
               echo "debug_archive=$(find . -type f -name "async-profiler-*-debug*" -exec basename {} \;)" >> $GITHUB_OUTPUT
             ;;
           esac
@@ -97,6 +97,7 @@ jobs:
           if-no-files-found: error
       - name: Upload coverage report
         uses: actions/upload-artifact@v4
+        if: false
         with:
           name: test-coverage-${{ matrix.runson.display }}
           path: build/test/coverage/
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -64,7 +64,6 @@ jobs:
           HASH=${GITHUB_SHA:0:7}
           case "${{ matrix.runson.display }}" in
             macos*)
-              brew install gcovr
               make COMMIT_TAG=$HASH FAT_BINARY=true release -j
               make COMMIT_TAG=$HASH FAT_BINARY=false test-cpp -j
             ;;
@@ -95,13 +94,6 @@ jobs:
         with:
           name: ${{ steps.set_artifact_name.outputs.artifact_name }}-debug
           path: ${{ steps.build.outputs.debug_archive }}
-          if-no-files-found: error
-      - name: Upload coverage report
-        uses: actions/upload-artifact@v4
-        if: false
-        with:
-          name: test-coverage-${{ matrix.runson.display }}
-          path: build/test/coverage/
           if-no-files-found: error
   integration-tests:
     needs: build-and-upload-binaries
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -66,9 +66,10 @@ jobs:
             macos*)
               brew install gcovr
               make COMMIT_TAG=$HASH FAT_BINARY=true release -j
+              make COMMIT_TAG=$HASH FAT_BINARY=false test-cpp -j
             ;;
             *)
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release -j
+              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release test-cpp -j
               echo "debug_archive=$(find . -type f -name "async-profiler-*-debug*" -exec basename {} \;)" >> $GITHUB_OUTPUT
             ;;
           esac
--- a/src/os_macos.cpp
+++ b/src/os_macos.cpp
@@ -97,6 +97,8 @@ JitWriteProtection::~JitWriteProtection() {
 
 
 static SigAction installed_sigaction[32];
+static SigAction orig_sigbus_handler;
+static SigAction orig_sigsegv_handler;
 
 const size_t OS::page_size = sysconf(_SC_PAGESIZE);
 const size_t OS::page_mask = OS::page_size - 1;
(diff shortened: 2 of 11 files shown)


.github/workflows/test-and-publish-nightly.yml
```diff
@@ -64,11 +64,11 @@ jobs:
           HASH=${GITHUB_SHA:0:7}
           case "${{ matrix.runson.display }}" in
             macos*)
-              brew install gcovr
-              make COMMIT_TAG=$HASH FAT_BINARY=true release coverage -j
+              make COMMIT_TAG=$HASH FAT_BINARY=true release -j
+              make COMMIT_TAG=$HASH FAT_BINARY=false test-cpp -j
```
fandreuz: Shouldn't we run `test-cpp` before `release`?

src/os_macos.cpp
```diff
@@ -227,13 +229,32 @@ SigAction OS::installSignalHandler(int signo, SigAction action, SigHandler handl
...
     sa.sa_sigaction = action;
+    sa.sa_flags |= SA_SIGINFO | SA_RESTART;
     sigaction(SIGBUS, &sa, NULL);
-    return old_action;
+
+    sigaction(SIGSEGV, NULL, &sa);
+    orig_sigsegv_handler = sa.sa_handler == SIG_DFL ? restoreSignalHandler : sa.sa_sigaction;
+    sa.sa_sigaction = action;
+    sa.sa_flags |= SA_SIGINFO | SA_RESTART;
+    sigaction(SIGSEGV, &sa, NULL);
```
fandreuz, line 251: Should we check the return value, and possibly log `errno`?

.github/workflows/test-and-publish-nightly.yml
apangin: I changed the order, although it does not make a big difference: `release` target only prepares a bundle.

src/os_macos.cpp
apangin: `sigaction` can only fail due to invalid arguments (invalid signal number or inaccessible address of a handler).
This is not applicable to our case - I don't think it's worth adding boilerplate to check errors that never happen.

src/safeAccess.cpp
```diff
@@ -0,0 +1,56 @@
...
+    LABEL(load32_end);
+    return ret;
+}
+
+// When a memory access error happens in SafeAccess::load/load32,
+// this function returns how many bytes to skip to advance to the next instruction
+u32 SafeAccess::skipLoad(instruction_t* pc) {
+    if ((pc >= (void*)load && pc < load_end) ||
+        (pc >= (void*)load32 && pc < load32_end)) {
+#if defined(__x86_64__) || defined(__i386__)
```
krk: What about `push [mem]; pop reg`?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -64,11 +64,11 @@ jobs:
           HASH=${GITHUB_SHA:0:7}
           case "${{ matrix.runson.display }}" in
             macos*)
-              brew install gcovr
-              make COMMIT_TAG=$HASH FAT_BINARY=true release coverage -j
+              make COMMIT_TAG=$HASH FAT_BINARY=false test-cpp -j
+              make COMMIT_TAG=$HASH FAT_BINARY=true release -j
             ;;
             *)
-              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release coverage -j
```
krk: We can keep the coverage build if we exclude the `SafeAccess` functions with `__attribute__((no_profile_instrument_function))`.

test/native/safeAccessTest.cpp
```diff
@@ -0,0 +1,26 @@
...
+TEST_CASE(SafeAccess_load) {
+    Profiler::instance()->updateSymbols(false);
+    Profiler::instance()->setupSignalHandlers();
+
+    void** bad_ptr = (void**)(uintptr_t)0xbaad0000ffffbaadULL;
+    void* good_ptr = (void*)(uintptr_t)0x123456789abcdef0ULL;
+    void* deadbeef = (void*)(uintptr_t)0xdeadbeefdeadbeefULL;
+
+    ASSERT_EQ(SafeAccess::load(bad_ptr), 0);
+    ASSERT_EQ(SafeAccess::load(&good_ptr), good_ptr);
```
krk: Can we take address of another variable in the `good_ptr` declaration, so `bad_ptr` and `good_ptr` are both `void**`? This might make these tests easier to read.

.github/workflows/test-and-publish-nightly.yml
fandreuz: Yeah it just seems more reasonable, and we don't need to prepare a bundle if `test-cpp` fails

src/safeAccess.cpp
apangin: I now generate desired instruction explicitly with inline assembler.

.github/workflows/test-and-publish-nightly.yml
apangin: Returned coverage back. Inline assembler solves the problem.

test/native/safeAccessTest.cpp
apangin: done

apangin:
OK, I gave up fighting with compiler quirks. Rewrote implementation with inline assembler for x86 and arm64. Other architectures fall back to C++ implementation which is good enough for optimized builds (debug builds probably don't work on these architectures, but we do not officially support them anyway).

src/safeAccess.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+NOINLINE
+int32_t SafeAccess::load32(int32_t* ptr, int32_t default_value) {
+#if defined(__x86_64__)
+    int32_t ret;
+    asm volatile("movl (%1), %0" : "=a"(ret) : "r"(ptr), "S"(default_value));
+#elif defined(__i386__)
+    int32_t ret;
+    asm volatile("movl (%1), %0" : "=a"(ret) : "r"(ptr), "a"(default_value));
```
krk, line 51: This can be `mov`, equivalent to `movl` x86 32 bits.

src/safeAccess.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+
+extern instruction_t load_end[];
+extern instruction_t load32_end[];
+
+NOINLINE
+void* SafeAccess::load(void** ptr, void* default_value) {
+#if defined(__x86_64__)
+    void* ret;
+    asm volatile("mov (%1), %0" : "=a"(ret) : "r"(ptr), "S"(default_value));
```
krk, line 29: In the output `default_value`/`rsi` is optimized away and it is passed in `esi` not `rsi` from the caller:

```asm
load(void**, void*):
 mov    rax,QWORD PTR [rdi]
load_end:
 ret
...
 mov    esi,0x99
...
 call load
```

https://godbolt.org/z/6nvef3zsb

It seems this could lead to incorrect `default_value` when higher bits of `rsi` is not clear, given that `StackFrame::arg1` uses `rsi`.

krk (changes requested):

src/safeAccess.cpp
apangin: > In the output default_value/rsi is optimized away

I intentionally use the register designated for the second argument by the ABI (rsi/esi), so that `SafeAccess::load` compiles to a single `mov` instruction.

> It seems this could lead to incorrect default_value when higher bits of rsi

No: `mov esi, int32` clears high bits of `rsi`.
Compiler uses this instruction because it's smaller in size.

src/safeAccess.cpp
apangin: Yes, they are the same on x86-32.

PR #1428, opened 2025-08-04
Update parseMultiModeRecording for better checks

Baraa-Hasheesh:
### Description
Modify the parseMultiModeRecording for better reliability 

There are 2 problems:
* getWaitedTime isn't not 100% accurate => Make the check more relaxed

public long getWaitedTime()
Returns the **approximate accumulated elapsed time** (in milliseconds) that the thread associated with this ThreadInfo has waited for notification since thread contention monitoring is enabled. I.e. the total accumulated time the thread has been in the [WAITING](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.State.html#WAITING) or [TIMED_WAITING](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.State.html#TIMED_WAITING) state since thread contention monitoring is enabled. This method returns -1 if thread contention monitoring is disabled.

* Not all samples are captured due to sampling interval (intended behavior) => change to `lock=0`

Furthermore to make the testing more complete, the number of locks is now checked against `getBlockedCount` rather than a static number

public long getBlockedCount()
Returns the total number of times that the thread associated with this ThreadInfo blocked to enter or reenter a monitor. I.e. the number of times a thread has been in the [BLOCKED](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.State.html#BLOCKED) state.
Returns:
the total number of times that the thread entered the BLOCKED state.

### Related issues
#1343

### Motivation and context
Remove random failures on GHA

### How has this been tested?
Test was re-ran 15 times on GHA with no failures 

---

### DIFF
--- a/test/test/jfr/JfrMultiModeProfiling.java
+++ b/test/test/jfr/JfrMultiModeProfiling.java
@@ -6,17 +6,18 @@
 package test.jfr;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Process to simulate lock contention and allocate objects.
@@ -29,7 +30,7 @@ public class JfrMultiModeProfiling {
     private static final List<byte[]> holder = new ArrayList<>();
 
     private static final ThreadMXBean tmx = ManagementFactory.getThreadMXBean();
-    private static final Map<Long, Long> threadLockTimes = new ConcurrentHashMap<>();
+    private static final Map<Long, ThreadInformation> threadInformation = new ConcurrentHashMap<>();
 
     static {
         tmx.setThreadContentionMonitoringEnabled(true);
@@ -42,8 +43,10 @@ public class JfrMultiModeProfiling {
         }
         allocate();
         executor.shutdown();
+        executor.awaitTermination(10, TimeUnit.SECONDS);
 
-        threadLockTimes.forEach((key, value) -> System.out.println(value));
+        System.out.println(threadInformation.values().stream().mapToLong(info -> info.blockedTime).sum());
+        System.out.println(threadInformation.values().stream().mapToLong(info -> info.blockedCount).sum());
     }
 
     private static void cpuIntensiveIncrement() {
@@ -54,7 +57,7 @@ public class JfrMultiModeProfiling {
         }
 
         long threadId = Thread.currentThread().getId();
-        threadLockTimes.put(threadId, tmx.getThreadInfo(threadId).getBlockedTime());
+        threadInformation.put(threadId, new ThreadInformation(tmx.getThreadInfo(threadId)));
     }
 
     private static void allocate() {
@@ -69,6 +72,16 @@ public class JfrMultiModeProfiling {
             if (holder.size() < 100_000) {
                 holder.add(new byte[1]);
             }
+        }
+    }
+
+    static class ThreadInformation {
+        long blockedTime;
+        long blockedCount;
+
+        ThreadInformation(ThreadInfo threadInfo) {
+            blockedTime = threadInfo.getBlockedTime();
+            blockedCount = threadInfo.getBlockedCount();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -42,8 +43,10 @@ public static void main(String[] args) throws InterruptedException {
         }
         allocate();
         executor.shutdown();
+        executor.awaitTermination(10, TimeUnit.SECONDS);
 
-        threadLockTimes.forEach((key, value) -> System.out.println(value));
+        System.out.println(threadInformation.values().stream().mapToLong(info -> info.blockedTime).sum());
+        System.out.println(threadInformation.values().stream().mapToLong(info -> info.blockedCount).sum());
```
apangin: I think this is unnecessary complication. The test submits 10 highly contended tasks, which should be enough to issue at least 10 MonitorEnter events.
At the same time, an acceptance interval of ±5 events does not look reasonable to me. E.g. if there were only 10 lock events, and async-profiler reported 5, there would be a huge 50% miss. It would be even more surprising If async-profiler reported more events than JMX.

test/test/jfr/JfrTests.java
```diff
@@ -90,8 +93,9 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
         }
 
         Assert.isGreater(eventsCount.get("jdk.ExecutionSample"), 50);
-        Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 10);
-        Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.90);
+        Assert.isGreaterOrEqual(eventsCount.get("jdk.JavaMonitorEnter"), totalNumberOfLocks - 5);
+        Assert.isLessOrEqual(eventsCount.get("jdk.JavaMonitorEnter"), totalNumberOfLocks + 5);
+        Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.80);
```
apangin, line 94: Will 79% mean that async-profiler works wrong?
Even though the spec says about approximation, async-profiler actually uses the same instrumentation point for lock duration computation as HotSpot, so I expect the difference to be really small (and it is in my experiments).

I'm OK with setting threshold at 80% for now just to unblock CI, but discrepancy in lock duration reported by async-profiler vs. HotSpot still worries me.

test/test/jfr/JfrTests.java
```diff
@@ -69,12 +70,14 @@ public void parseRecording(TestProcess p) throws Exception {
      * @param p The test process to profile with.
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock,quiet,jfr,file=%f", output = true)
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true)
```
apangin, line 72: It's OK to set `0` for this test as it focuses on JFR events, but the problem is that non-zero lock profiling threshold (which is the default) is not really tested anywhere.

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: In most normal runs the number of locks will be over 1000 (Linux, Mac was lower)

Now regarding the 5 range, I saw that the count will diverge by around 1~2 with over 1000samples for most cases

That difference was so small that it didn't look significant enough to chase down

test/test/jfr/JfrTests.java
Baraa-Hasheesh: Technically the `all` does trigger that 
Now if you mean the number of locks is accurate then we will kinda go in circles with that threshold, I would just suggest creating a new test

test/test/jfr/JfrTests.java
Baraa-Hasheesh: Now for anything lower than 80% I would consider something really went wrong in our profiling that would require a full proper investigation 

Now, regarding why the numbers are different we will need to check the JDK implementation 

I would do as follows if needed, reduce to 80 to unblock the CI, investigate the difference by checking the JDK implementation in depth

I plan to do that once I have the time as it looks like a fun learning/debugging experience

test/test/jfr/JfrMultiModeProfiling.java
apangin: Previous assertion checking for 50 locks often failed. I saw the number as low as 20-ish.
Considering my previous note about async-profiler and JMX using the same instrumentation point, I am surprised to see difference even by 2 samples.

test/test/jfr/JfrTests.java
apangin: OK, let's do the following:
- leave `lock=0` and 80% threshold;
- rollback `getBlockedCount` changes;
- integrate the PR and later investigate why blocked monitors reported by JVM and async-profiler differ (it's not urgent).

Baraa-Hasheesh: Done, I kept some of the changes to the test/test/jfr/JfrMultiModeProfiling.java test as they make the reported numbers more accurate & less code from before

PR #1429, opened 2025-08-06
Experiment with signal correlation

adinazugravescu:
### Description
First attempt to support signal correlation in async-profiler.Not ready for review.

### Related issues
N/A

### Motivation and context
N/A

### How has this been tested?
N/A

---


test-otel/TestApp.java
```diff
@@ -0,0 +1,39 @@
...
+            .setTracerProvider(tracerProvider)
+            .buildAndRegisterGlobal();
+        
+        Tracer tracer = GlobalOpenTelemetry.getTracer("test-app");
+        
+        Span span = tracer.spanBuilder("cpu-work").startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            System.out.println("Starting work with active span...");
+            doWork();
+            System.out.println("Work completed");
```
fandreuz: Let's remove unnecessary noise

test-otel/TestApp.java
```diff
@@ -0,0 +1,39 @@
...
+        SdkTracerProvider tracerProvider = SdkTracerProvider.builder().build();
+        OpenTelemetrySdk.builder()
+            .setTracerProvider(tracerProvider)
+            .buildAndRegisterGlobal();
+        
+        Tracer tracer = GlobalOpenTelemetry.getTracer("test-app");
+        
+        Span span = tracer.spanBuilder("cpu-work").startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            System.out.println("Starting work with active span...");
```
fandreuz: Please add a log statement which looks something like this:
```
[tid=...] trace ID: ..., span ID: ...
```

test-otel/TestApp.java
```diff
@@ -0,0 +1,39 @@
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+
+public class TestApp {
+    public static void main(String[] args) {
```
fandreuz, line 9: Can you add multiple threads? i want to see multiple different values for trace id and span id

src/profiler.cpp
```diff
@@ -692,6 +692,57 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu | 0x8000);
     }
 
+   VMThread* vm_thread = VMThread::current();
+    if (vm_thread != NULL) {
```
```suggestion
    VMThread* vm_thread;
    if (getCurrentCompileTask() == NULL && (vm_thread = VMThread::current()) != NULL) {
```

src/profiler.cpp
```diff
@@ -692,6 +692,57 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+        JNIEnv* jni = VM::jni();
+        if (jni != NULL) {
+            jni->PushLocalFrame(16);
+            
+            jclass spanClass = jni->FindClass("io/opentelemetry/api/trace/Span");
```
fandreuz: Can you move all this to somewhere else? This being here means it's done every single time `recordSample` is called, which adds overhead. For now, add what you need to `Profiler` class and fill it inside `Profiler::start`. Ideally we should only call `current()` and `getSpan(Trace)Id` in `recordSample`.

src/profiler.cpp
```diff
@@ -692,6 +692,36 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+                        if (trace_id) jni->ReleaseStringUTFChars(trace_id_str, trace_id);
+                        if (span_id) jni->ReleaseStringUTFChars(span_id_str, span_id);
+                    }
+                }
+            }
+            if (jni->ExceptionCheck()) {
+                jni->ExceptionClear();
+            }
+        }
+    }
```
krk: This is a lot of JNI calls and string allocations. What is the performance impact? Can we do without jni calls or a single jni call perhaps?

fandreuz: We plan to measure the overhead, this is one of the possible ways to solve the problem, not necessarily the one we're going to adopt eventually.

PR #1431, opened 2025-08-07
Preserve compatibility with old JFR reader

fandreuz:
### Description
In this PR I fix the incompatibility between newly written JFR files, and the previous version of Async-Profiler's JFR reader.

### Related issues
Incompatibility introduced in #1378.

### Motivation and context
JFR produced by Async-Profiler should be backward compatible.

### How has this been tested?
Manual testing.

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -784,8 +784,8 @@ class Recording {
 
         buf->putVar32(11);
 
-        Index packages;
-        Index symbols;
+        Index packages(1);
+        Index symbols(1);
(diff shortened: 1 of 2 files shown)


krk:
Why is index being 0 or 1 is an incompatibility? Is there an error message or a crash somewhere?

fandreuz:
> Why is index being 0 or 1 is an incompatibility? Is there an error message or a crash somewhere?

Yes, check the changes in `Dictionary.java` [here](https://github.com/async-profiler/async-profiler/commit/3237a0ce9bcbb458ada300fa9dd09b786e02c756). There used to be a check on the `key`. After that PR, we store symbols with ordered keys starting from zero.

PR #1433, opened 2025-08-08
Add signal correlation API for OTEL integration

adinazugravescu:
### Description
Add API methods to correlate profiling samples with OpenTelemetry trace and span IDs.Not ready for review.

### Related issues
N/A

### Motivation and context
N/A

### How has this been tested?
N/A

---

### DIFF
--- a/src/javaApi.cpp
+++ b/src/javaApi.cpp
@@ -148,6 +148,21 @@ Java_one_profiler_AsyncProfiler_filterThread0(JNIEnv* env, jobject unused, jthre
     }
 }
 
+extern "C" DLLEXPORT void JNICALL
+Java_one_profiler_AsyncProfiler_setTraceContext0(JNIEnv* env, jobject unused, jstring traceId, jstring spanId) {
+    const char* trace_id_str = env->GetStringUTFChars(traceId, NULL);
+    const char* span_id_str = env->GetStringUTFChars(spanId, NULL);
+    
+    Profiler::instance()->setTraceContext(trace_id_str, span_id_str);
+    
+    env->ReleaseStringUTFChars(traceId, trace_id_str);
+    env->ReleaseStringUTFChars(spanId, span_id_str);
+}
+
+extern "C" DLLEXPORT void JNICALL
+Java_one_profiler_AsyncProfiler_clearTraceContext0(JNIEnv* env, jobject unused) {
+    Profiler::instance()->clearTraceContext();
+}
 
 #define F(name, sig)  {(char*)#name, (char*)sig, (void*)Java_one_profiler_AsyncProfiler_##name}
 
@@ -158,6 +173,8 @@ static const JNINativeMethod profiler_natives[] = {
     F(execute1,      "(Ljava/lang/String;)[B"),
     F(getSamples,    "()J"),
     F(filterThread0, "(Ljava/lang/Thread;Z)V"),
+    F(setTraceContext0,   "(Ljava/lang/String;Ljava/lang/String;)V"),
+    F(clearTraceContext0, "()V"),
 };
 
 static const JNINativeMethod* execute0 = &profiler_natives[2];
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -65,6 +65,9 @@ static Instrument instrument;
 
 static ProfilingWindow profiling_window;
 
+// Thread-local storage for trace correlation
+thread_local std::string current_trace_id;
+thread_local std::string current_span_id;
 
 // The same constants are used in JfrSync
 enum EventMask {
@@ -609,6 +612,8 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     atomicInc(_total_samples);
 
     int tid = OS::threadId();
+    printf("[RECORD_SAMPLE] tid=%d, traceId=%s, spanId=%s\n", 
+       tid, current_trace_id.c_str(), current_span_id.c_str());
     u32 lock_index = getLockIndex(tid);
     if (!_locks[lock_index].tryLock() &&
         !_locks[lock_index = (lock_index + 1) % CONCURRENCY_LEVEL].tryLock() &&
@@ -781,6 +786,18 @@ void Profiler::writeLog(LogLevel level, const char* message) {
 
 void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
     _jfr.recordLog(level, message, len);
+}
+
+void Profiler::setTraceContext(const char* traceId, const char* spanId) {
+    current_trace_id = traceId ? traceId : "";
+    current_span_id = spanId ? spanId : "";
+    printf("[TRACE_CONTEXT] tid=%d, traceId=%s, spanId=%s\n", OS::threadId(), traceId, spanId);
+}
+
+void Profiler::clearTraceContext() {
+    current_trace_id.clear();
+    current_span_id.clear();
+    printf("[TRACE_CONTEXT] tid=%d, context cleared\n", OS::threadId());
 }
 
 void* Profiler::dlopen_hook(const char* filename, int flags) {
--- a/src/profiler.h
+++ b/src/profiler.h
@@ -213,6 +213,9 @@ class Profiler {
     void writeLog(LogLevel level, const char* message);
     void writeLog(LogLevel level, const char* message, size_t len);
 
+    void setTraceContext(const char* traceId, const char* spanId);
+    void clearTraceContext();
+
     void updateSymbols(bool kernel_symbols);
     const void* resolveSymbol(const char* name);
     const char* getLibraryName(const char* native_symbol);
--- a/src/api/one/profiler/AsyncProfiler.java
+++ b/src/api/one/profiler/AsyncProfiler.java
@@ -281,6 +281,24 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
         }
     }
 
+    /**
+     * Set trace context for signal correlation
+     * @param traceId OpenTelemetry trace identifier
+     * @param spanId OpenTelemetry span identifier
+     */
+    @Override
+    public void setTraceContext(String traceId, String spanId) {
+        setTraceContext0(traceId, spanId);
+    }
+
+    /**
+     * Clear trace context for the current thread
+     */
+    @Override
+    public void clearTraceContext() {
+        clearTraceContext0();
+    }
+
     private native void start0(String event, long interval, boolean reset) throws IllegalStateException;
 
     private native void stop0() throws IllegalStateException;
@@ -290,4 +308,8 @@ public class AsyncProfiler implements AsyncProfilerMXBean {
     private native byte[] execute1(String command) throws IllegalArgumentException, IllegalStateException, IOException;
 
     private native void filterThread0(Thread thread, boolean enable);
+
+    private native void setTraceContext0(String traceId, String spanId);
+    
+    private native void clearTraceContext0();
 }
--- a/src/api/one/profiler/AsyncProfilerMXBean.java
+++ b/src/api/one/profiler/AsyncProfilerMXBean.java
@@ -30,4 +30,7 @@ public interface AsyncProfilerMXBean {
     String dumpTraces(int maxTraces);
     String dumpFlat(int maxMethods);
     byte[] dumpOtlp();
+
+    void setTraceContext(String traceId, String spanId);
+    void clearTraceContext();
 }


src/profiler.cpp
```diff
@@ -65,6 +65,9 @@ static Instrument instrument;
 
 static ProfilingWindow profiling_window;
 
+// Thread-local storage for trace correlation
+thread_local std::string current_trace_id;
```
fandreuz: Let's use `const char*` here, creating `std::string` is one more copy

fandreuz: And remember to delete it in `clearTraceContext`

src/profiler.cpp
```diff
@@ -609,6 +612,8 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     atomicInc(_total_samples);
 
     int tid = OS::threadId();
+    printf("[RECORD_SAMPLE] tid=%d, traceId=%s, spanId=%s\n",
```
fandreuz: Please add a TODO here, to remember not to keep this line

src/profiler.cpp
```diff
@@ -783,6 +789,36 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
...
 
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    delete[] current_trace_id;
+    delete[] current_span_id;
+    
+    if (trace_id) {
+        size_t len = strlen(trace_id) + 1;
+        char* new_trace_id = new char[len];
+        strcpy(new_trace_id, trace_id);
+        current_trace_id = new_trace_id;
```
fandreuz: let's use `strdup` instead: https://en.cppreference.com/w/c/experimental/dynamic/strdup

src/profiler.cpp
```diff
@@ -783,6 +789,36 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
...
+        char* new_span_id = new char[len];
+        strcpy(new_span_id, span_id);
+        current_span_id = new_span_id;
+    } else {
+        current_span_id = nullptr;
+    }
+}
+
+void Profiler::clearTraceContext() {
+    delete[] current_trace_id;
```
fandreuz: Since we'll use `strdup`, let's replace this with `free(current_trace_id)`

src/profiler.cpp
```diff
@@ -783,6 +789,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
...
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    free((void*)current_trace_id);
+    free((void*)current_span_id);
+    
+    current_trace_id = trace_id ? strdup(trace_id) : nullptr;
+    current_span_id = span_id ? strdup(span_id) : nullptr;
+}
+
+void Profiler::clearTraceContext() {
+    free((void*)current_trace_id);
```
fandreuz: Casting shouldn't be needed, is it?

fandreuz: Same above

src/profiler.cpp
```diff
@@ -783,6 +789,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
     _jfr.recordLog(level, message, len);
 }
 
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    free((void*)current_trace_id);
+    free((void*)current_span_id);
+    
+    current_trace_id = trace_id ? strdup(trace_id) : nullptr;
```
fandreuz: Checking for `trace_id` not being empty is not needed

fandreuz: Fair enough, `strdup` does not check

src/profiler.cpp
fandreuz: Since, we are supposed to `free`, the root error is in declaring `current_trace_id ` as `const`. Let's remove `const` from the declaration

src/profiler.cpp
```diff
@@ -65,6 +65,9 @@ static Instrument instrument;
 
 static ProfilingWindow profiling_window;
 
+// Thread-local storage for trace correlation
+thread_local char* current_trace_id;
```
fandreuz: Can you make this part of the `Profiler` class?

src/profiler.cpp
```diff
@@ -783,6 +789,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
     _jfr.recordLog(level, message, len);
 }
 
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    free(current_trace_id);
+    free(current_span_id);
+    
+    current_trace_id = trace_id ? strdup(trace_id) : nullptr;
```
fandreuz: According to the [spec](https://www.w3.org/TR/trace-context/#trace-id), trace-id is 16 byte and span-id is 8 byte. Thus, we don't need to `strdup`: declaring the fields as `thread_local char trace_id[16]` and copying them with `memcpy` will be faster.

fandreuz: Actually, hold on on this. For the next phase of the implementation what you're doing now is going to simplify our life a bit. Let's discuss it on Monday

src/javaApi.cpp
```diff
@@ -148,6 +148,21 @@ Java_one_profiler_AsyncProfiler_filterThread0(JNIEnv* env, jobject unused, jthre
     }
 }
 
+extern "C" DLLEXPORT void JNICALL
+Java_one_profiler_AsyncProfiler_setTraceContext0(JNIEnv* env, jobject unused, jstring traceId, jstring spanId) {
```
Baraa-Hasheesh, line 152: Shouldn't we also decalre an equivalent API for native applications?

@fandreuz what are your thoughts on this?

Please note that if declared it should be declared as UNSTABLE

fandreuz: Eventually, maybe. Draft PR, not ready for review. This is a tinkering PR for @adinazugravescu, we'll discuss later how to make it production ready.

src/profiler.cpp
```diff
@@ -613,8 +613,9 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
 
     int tid = OS::threadId();
     // TODO: Remove this debug printf
-    printf("[RECORD_SAMPLE] tid=%d, traceId=%s, spanId=%s\n", 
-        tid, current_trace_id ? current_trace_id : "", current_span_id ? current_span_id : "");
+    // printf("[RECORD_SAMPLE] tid=%d, traceId=%s, spanId=%s\n",
```
fandreuz: We can remove this stuff now

src/profiler.h
```diff
@@ -90,6 +90,10 @@ class Profiler {
     bool _update_thread_names;
     volatile jvmtiEventMode _thread_events_state;
 
+    // Thread-local storage for trace correlation
+    static thread_local char* _current_trace_id;
```
fandreuz: Can be non-static, too

src/profiler.cpp
```diff
@@ -1710,6 +1724,14 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
         for (int j = 0; j < trace->num_frames; j++) {
             // To be written below in Profile.location_indices
             location_indices.push_back(functions.indexOf(fn.name(trace->frames[j])));
+            if (trace->frames[j].bci == BCI_TRACE_ID) {
+                const char* trace_id = (const char*)trace->frames[j].method_id;
+                printf("[dumpOtlp: ] trace_id: %s\n", trace_id ? trace_id : "NULL");
```
fandreuz: Good, now we can go to the next step: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L300

Please create `Link` messages using the protobuf writer (it's a relatively simple class, you find it in `protobuf.h`). You will also need to add more Protobuf indexes in `otlp.h`. Each `Link` should be part of the [`link_table`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L111), and the corresponding index should go in [here](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L420).

src/profiler.h
adinazugravescu: Apparently, non static thread_local members are not allowed in c++ classes

fandreuz: ouch

src/profiler.cpp
```diff
@@ -1677,9 +1709,28 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
         if (trace == NULL || excludeTrace(&fn, trace) || cts->samples == 0) continue;
 
+        const char* trace_id = nullptr;
+        const char* span_id = nullptr;
+
+        for (int j = 0; j < trace->num_frames; j++) {
+            if (trace->frames[j].bci == BCI_TRACE_ID) {
+                trace_id = (const char*)trace->frames[j].method_id;
+            }
+            if (trace->frames[j].bci == BCI_SPAN_ID) {
```
fandreuz: Can be an `else if`

src/profiler.cpp
```diff
@@ -1677,9 +1709,28 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+        }
+
         protobuf_mark_t sample_mark = otlp_buffer.startMessage(Profile::sample, 1);
         otlp_buffer.field(Sample::locations_start_index, frames_seen);
         otlp_buffer.field(Sample::locations_length, trace->num_frames);
+
+        if (trace_id && span_id) {
+            std::string link_key = std::string(trace_id) + ":" + std::string(span_id);
+            size_t link_idx = links.indexOf(link_key);
+            otlp_buffer.field(Sample::link_index, link_idx + 1);
```
fandreuz: The first idx you get out of `Index` is always going to be 1, check the implementation

src/profiler.cpp
```diff
@@ -1677,9 +1709,28 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+                span_id = (const char*)trace->frames[j].method_id;
+            }
+        }
+
         protobuf_mark_t sample_mark = otlp_buffer.startMessage(Profile::sample, 1);
         otlp_buffer.field(Sample::locations_start_index, frames_seen);
         otlp_buffer.field(Sample::locations_length, trace->num_frames);
+
+        if (trace_id && span_id) {
+            std::string link_key = std::string(trace_id) + ":" + std::string(span_id);
```
fandreuz: Let's rediscuss it after #1425 is merged, we could use the new generic `Index` instead of this additional copy. It'll be fine for now

src/profiler.cpp
fandreuz:
```suggestion
            otlp_buffer.field(Sample::link_index, link_idx);
```

src/profiler.cpp
```diff
@@ -1677,9 +1709,28 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
         CallTrace* trace = cts->acquireTrace();
         if (trace == NULL || excludeTrace(&fn, trace) || cts->samples == 0) continue;
 
+        const char* trace_id = nullptr;
+        const char* span_id = nullptr;
+
+        for (int j = 0; j < trace->num_frames; j++) {
```
fandreuz: You can move the content of this for loop to the other `for (int j = 0; j < trace->num_frames; j++) {`

src/profiler.cpp
```diff
@@ -1729,6 +1788,22 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
         otlp_buffer.commitMessage(location_mark);
     }
 
+    protobuf_mark_t default_link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+    otlp_buffer.commitMessage(default_link_mark);
+    // Write link_table
+    links.forEachOrdered([&] (size_t idx, const std::string& link_key) {
```
fandreuz, line 1773: The first string you're getting out of this is empty (see impl in `index.h`). So you can simply check if the string is empty and do what you do at L1791, otherwise handle it normally

src/profiler.cpp
```diff
@@ -1729,6 +1788,22 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+        size_t colon_pos = link_key.find(':');
```
fandreuz: You already know the index of this: trace_id is 16 byte

src/profiler.cpp
```diff
@@ -1729,6 +1788,22 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+        if (colon_pos != std::string::npos) {
```
fandreuz: Let's just assume everything is well formed

src/profiler.cpp
```diff
@@ -1729,6 +1788,22 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+            std::string trace_id = link_key.substr(0, colon_pos);
+            std::string span_id = link_key.substr(colon_pos + 1);
+            
+            protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+            otlp_buffer.field(Link::trace_id, trace_id.c_str(), trace_id.length());
```
fandreuz: What you can do here: set to zero the index where you expect the `:` to be

src/profiler.cpp
```diff
@@ -1788,18 +1786,21 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
-    protobuf_mark_t default_link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
-    otlp_buffer.commitMessage(default_link_mark);
     // Write link_table
     links.forEachOrdered([&] (size_t idx, const std::string& link_key) {
-        size_t colon_pos = link_key.find(':');
-        if (colon_pos != std::string::npos) {
-            std::string trace_id = link_key.substr(0, colon_pos);
-            std::string span_id = link_key.substr(colon_pos + 1);
+        if (link_key.empty()) {
+            protobuf_mark_t default_link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
```
fandreuz: You can move this to before the `if`, and `commitMessage` to after the `if`. The different between the two branches is only that you're filling `trace_id` and `span_id`

src/profiler.cpp
```diff
@@ -1788,18 +1786,21 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+            std::string mutable_key = link_key;
+            mutable_key[16] = '\0';
+            
+            const char* trace_id = mutable_key.c_str();
+            const char* span_id = mutable_key.c_str() + 17;
             
             protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
-            otlp_buffer.field(Link::trace_id, trace_id.c_str(), trace_id.length());
-            otlp_buffer.field(Link::span_id, span_id.c_str(), span_id.length());
+            otlp_buffer.field(Link::trace_id, trace_id, 16);
```
fandreuz: Please check if it works without mutating `link_key`. Passing the length should be enough. If yes, then you don't need a new mutable copy at all

src/profiler.cpp
```diff
@@ -1788,21 +1788,14 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
-            protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
-            otlp_buffer.field(Link::trace_id, trace_id, 16);
-            otlp_buffer.field(Link::span_id, span_id, strlen(span_id));
-            otlp_buffer.commitMessage(link_mark);
+        protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+        if (!link_key.empty()) {
+            const char* data = link_key.c_str();
+
+            otlp_buffer.field(Link::trace_id, data, 16);
+            otlp_buffer.field(Link::span_id, data + 17, link_key.length() - 17);  // after ":"
```
fandreuz: We probably don't need the colon at all

src/profiler.cpp
```diff
@@ -783,6 +794,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
     _jfr.recordLog(level, message, len);
 }
 
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    free(_current_trace_id);
```
fandreuz: Please don't free here or in `clear`, we're going to discuss how we free without having memory leaks or corrupted memory.

src/profiler.cpp
```diff
@@ -1680,16 +1712,41 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
         otlp_buffer.putVarInt(cts->samples);
         otlp_buffer.putVarInt(cts->counter);
         otlp_buffer.commitMessage(sample_value_mark);
         otlp_buffer.commitMessage(sample_mark);
 
+        const char* trace_id = nullptr;
+        const char* span_id = nullptr;
+
         for (int j = 0; j < trace->num_frames; j++) {
+            if (trace->frames[j].bci == BCI_TRACE_ID) {
```
fandreuz: Please let's have just one frame for span_id and trace_id, with one 16+8 string holding the whole thing.

src/profiler.cpp
```diff
@@ -783,6 +788,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
     _jfr.recordLog(level, message, len);
 }
 
+void Profiler::setTraceContext(const char* trace_id, const char* span_id) {
+    if (trace_id && span_id) {
+        char* combined = (char*)malloc(strlen(trace_id) + strlen(span_id) + 1);
```
fandreuz: You know the length for both

src/profiler.cpp
```diff
@@ -783,6 +788,21 @@ void Profiler::writeLog(LogLevel level, const char* message, size_t len) {
...
+        strcpy(combined, trace_id);
+        strcat(combined, span_id);
```
fandreuz: This should work too:
```suggestion
        strcpy(combined + 16, span_id);
```
`strcat` will most likely scan through the string to find the current end

fandreuz: Actually, let's use `memcpy` for both cases. I think `strcpy` does additional checks we don't really need, since we know all sizes.

src/profiler.cpp
```diff
@@ -1298,6 +1318,9 @@ Error Profiler::stop(bool restart) {
         FdTransferClient::closePeer();
     }
 
+    free(_current_trace_context);
```
fandreuz, line 1323: Could you please elaborate on why this is needed here?

src/profiler.cpp
```diff
@@ -1680,16 +1704,29 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
 
+        const char* trace_context = nullptr;
+
         for (int j = 0; j < trace->num_frames; j++) {
+            if (trace->frames[j].bci == BCI_TRACE_CONTEXT) {
+                trace_context = (const char*)trace->frames[j].method_id;
+            }
+
             // To be written below in Profile.location_indices
             location_indices.push_back(functions.indexOf(fn.name(trace->frames[j])));
```
fandreuz, line 1725: Let's not do this if it's a trace context, we don't need to see it in the samples

src/profiler.cpp
```diff
@@ -1729,6 +1766,18 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
         otlp_buffer.commitMessage(location_mark);
     }
 
+    // Write link_table
+    links.forEachOrdered([&] (size_t idx, const std::string& link_key) {
+        protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+        if (!link_key.empty()) {
+            const char* data = link_key.c_str();
+
```
fandreuz: Newline not needed here

src/profiler.cpp
```diff
@@ -1729,6 +1766,18 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+            otlp_buffer.field(Link::trace_id, data, 16);
+            otlp_buffer.field(Link::span_id, data + 16, link_key.length() - 16);
```
```suggestion
            otlp_buffer.field(Link::span_id, data + 16, 8);
```

test-otel/AsyncProfilerSpanProcessor.java
```diff
@@ -0,0 +1,33 @@
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import one.profiler.AsyncProfiler;
+
+public class AsyncProfilerSpanProcessor implements SpanProcessor {
```
fandreuz: I'd say, let's move everything under `test-otel` to the other repository. Eventually we're going to open a PR with this stuff there

src/profiler.cpp
```suggestion
        // TODO: free the memory
        char* combined = (char*)malloc(strlen(trace_id) + strlen(span_id) + 1);
```

src/profiler.cpp
fandreuz: My bad, please close

PR #1434, opened 2025-08-08
Simple smoke tests for JFR converter

fandreuz:
### Description
In this PR I add a few simple tests to verify that JFR conversion completes successfully. The output is not inspected.

### Related issues
#1378 introduced a bug (fixed in #1431) in the allocation heatmap converter.

### Motivation and context
We should have some basic protection from bugs in JFR converter.

### How has this been tested?
New tests.

---

### DIFF
--- a/test/test/jfrconverter/CpuBurner.java
+++ b/test/test/jfrconverter/CpuBurner.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.jfrconverter;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Random;
+
+public class CpuBurner {
+    private static final Random RANDOM = new Random();
+    private static final Duration TEST_DURATION = Duration.ofSeconds(1);
+
+    static void burn() {
+        long n = RANDOM.nextLong();
+        if (Long.toString(n).hashCode() == 0) {
+            System.out.println(n);
+        }
+    }
+
+    public static void main(String[] args) {
+        Instant start = Instant.now();
+        while (Duration.between(start, Instant.now()).compareTo(TEST_DURATION) < 0) {
+            burn();
+        }
+    }
+}
--- a/test/test/jfrconverter/JfrconverterTests.java
+++ b/test/test/jfrconverter/JfrconverterTests.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.jfrconverter;
+
+import one.convert.*;
+import one.profiler.test.*;
+
+// Simple smoke tests for JFR converter. The output is not inspected for errors,
+// we only verify that the conversion completes successfully.
+public class JfrconverterTests {
+
+    private static final String TMP_DIR = System.getProperty("java.io.tmpdir");
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")
+    public void testAllocationHeatmapConversion(TestProcess p) throws Exception {
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("alloc"));
+    }
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")
+    public void testCpuHeatmapConversion(TestProcess p) throws Exception {
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("cpu"));
+    }
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")
+    public void testFlamegraphConversion(TestProcess p) throws Exception {
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+        JfrToFlame.convert(p.getFilePath("%f"), "/dev/null", new Arguments());
+    }
+}
--- a/test/test/jfrconverter/JfrconverterTests.java
+++ b/test/test/jfrconverter/JfrconverterTests.java
@@ -18,14 +18,14 @@ public class JfrconverterTests {
     public void testAllocationHeatmapConversion(TestProcess p) throws Exception {
         Output out = p.waitForExit("%f");
         assert p.exitCode() == 0;
-        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("alloc"));
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("--alloc"));
     }
 
     @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")
     public void testCpuHeatmapConversion(TestProcess p) throws Exception {
         Output out = p.waitForExit("%f");
         assert p.exitCode() == 0;
-        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("cpu"));
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("--cpu"));
     }
 
     @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")


fandreuz:
@apangin do you have a reproducer for the alloc heatmap crash? I tried the following, but I don't get a crash:
```
% git checkout jfrconv-tests -- test
% make test-java -j TESTS=jfrconverter
[...]
Total test duration: 3.594 s
Results Summary:
PASS: 3 (100.0%)
FAIL: 0
```

fandreuz:
> @apangin do you have a reproducer for the alloc heatmap crash? I tried the following, but I don't get a crash:
> 
> ```
> % git checkout jfrconv-tests -- test
> % make test-java -j TESTS=jfrconverter
> [...]
> Total test duration: 3.594 s
> Results Summary:
> PASS: 3 (100.0%)
> FAIL: 0
> ```

Got it, adding `all` did the trick.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -0,0 +1,37 @@
...
+package test.jfrconverter;
+
+import one.convert.*;
+import one.profiler.test.*;
+
+// Simple smoke tests for JFR converter. The output is not inspected for errors,
+// we only verify that the conversion completes successfully.
+public class JfrconverterTests {
+
+    private static final String TMP_DIR = System.getProperty("java.io.tmpdir");
```
apangin: Unused

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,all,file=%f")
+    public void testAllocationHeatmapConversion(TestProcess p) throws Exception {
```
apangin: Do not prefix test names with `test`. Annotation is more than enough to mark tests.
Also, `Conversion` is redundant, because all tests in this folder test JFR conversion by design.
Keep in mind that these method names are used for:
- filtering tests via the command line options;
- naming test logs directories.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -0,0 +1,37 @@
...
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("--alloc"));
+    }
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,all,file=%f")
+    public void testCpuHeatmapConversion(TestProcess p) throws Exception {
```
apangin: All tests here run exactly the same program with exactly the same profiler arguments. I think there is no need to do this 3 times. Just call 3 different conversions for the same output file.

fandreuz: c1c29039f3c1f69a157bebe1343db03e1a77523b

test/test/jfrconverter/JfrconverterTests.java
fandreuz: Bad habits from my Python times: https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery

c1c29039f3c1f69a157bebe1343db03e1a77523b

test/test/jfrconverter/JfrconverterTests.java
fandreuz: c1c29039f3c1f69a157bebe1343db03e1a77523b

test/test/jfrconverter/CpuBurner.java
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.jfrconverter;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Random;
+
+public class CpuBurner {
```
krk: We have a CPU burner at `test/cpu/CpuBurner.java`, is it reusable here?

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -0,0 +1,31 @@
...
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=profile.jfr,all,file=%f")
+    public void allocationHeatmapConversion(TestProcess p) throws Exception {
+        p.waitForExit();
+        assert p.exitCode() == 0;
+        JfrToHeatmap.convert("profile.jfr", "/dev/null", new Arguments("--alloc"));
+    }
+
+    @Test(sh = "")
+    public void cpuHeatmapConversion(TestProcess p) throws Exception {
+        JfrToHeatmap.convert("profile.jfr", "/dev/null", new Arguments("--cpu"));
```
krk: Looks like this has dependency to an artifact generated in a different test. Can we make the tests independent from each other and runnable in any order?

krk (changes requested):

test/test/jfrconverter/JfrconverterTests.java
fandreuz: Yeah that was the initial idea: bf304f10333ad9f787374870c1d815f132ddb932

But I got this comment: https://github.com/async-profiler/async-profiler/pull/1434#discussion_r2263469986

Do you have a better idea?

test/test/jfrconverter/CpuBurner.java
fandreuz: Right, we don't need anything special: 6d9efe7ae00720e3e313648e47985d0577d3f28a

test/test/jfrconverter/JfrconverterTests.java
apangin: Yes, tests must not depend on each other.
My idea was just to merge them in one test.

fandreuz: 90b93f93576b50eb62a1a9a92a07ddf563d35100

krk: You can generate the jfr for each test separately.

apangin: I'm particularly against producing the same JFR 3 times in a row just to run converter with slightly different options. This is suboptimal wrt. test run time and resource usage.

fandreuz: I'm with @krk here, having three conceptually different tests doing very different things all in one test looks confusing

fandreuz: How about it's not the same JFR? No need to create it with `all`

apangin: > conceptually different tests doing very different things

I don't fully agree. These tests run exactly the same program with exactly the same arguments. The only difference is parameters of conversion. OK, maybe FlameGraph conversion and Heatmap conversion can be split into separate tests, but rerunning the test program from scratch and reprofiling it again with the same parameters just to change one heatmap option does not make much sense to me. It's a waste of time and compute resources, not to mention lots of copy-pasted lines of test code. Coverage will not get any bigger.

apangin: Alright, we can have a meaningful compromise. What about making two tests: one for Heatmap conversion and another one for FlameGraph conversion? This will allow keeping the test run time short and the amount of test code small, while increasing code coverage by adding more conversions with different arguments.

fandreuz: 3b433e5555671a25a14f20a973943666bb6137c4

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -0,0 +1,30 @@
...
+        assert p.exitCode() == 0;
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("--alloc"));
+        JfrToHeatmap.convert(p.getFilePath("%f"), "/dev/null", new Arguments("--cpu"));
+    }
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,all,file=%f")
+    public void flamegraphConversion(TestProcess p) throws Exception {
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+        JfrToFlame.convert(p.getFilePath("%f"), "/dev/null", new Arguments());
```
apangin, line 28: While you are here, adding a similar conversion with `--alloc` option for `JfrToFlame` would also make sense.

fandreuz: 6088fb7abe09b2a2a492202f880c74c1a32888e0

PR #1436, opened 2025-08-15
Add signal correlation API with thread-local buffer approach

adinazugravescu:
### Description
Add API methods to correlate profiling samples with OpenTelemetry trace and span IDs.Not ready for review.

### Related issues
N/A

### Motivation and context
N/A

### How has this been tested?
N/A

---


src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        if (buffer == null) return;
+        
+        buffer.position(0);
+        for (int i = 0; i < 49; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
+        if (str == null) str = "";
```
fandreuz: Let's throw an exception here

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        byte[] bytes = str.getBytes(StandardCharsets.UTF_8);
+        int copyLen = Math.min(bytes.length, len);
+        buffer.put(bytes, 0, copyLen);
+        for (int i = copyLen; i < len; i++) buffer.put((byte) 0);
```
fandreuz: Let's just check that `len <= str.length()`, otherwise throw

src/profiler.cpp
```diff
@@ -692,6 +693,12 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu | 0x8000);
     }
 
+    if (_trace_buffer[0] != '\0') {
+        char* trace_copy = (char*)malloc(49);
```
fandreuz: Let's have a `constexpr` for `49` in `otel.h`

src/api/one/profiler/TraceContext.java
```diff
@@ -36,10 +36,10 @@ public static void clearTraceContext() {
     }
 
     private static void writeString(ByteBuffer buffer, String str, int len) {
-        if (str == null) str = "";
+        if (str == null) throw new IllegalArgumentException("String cannot be null");
+        if (str.length() > len) throw new IllegalArgumentException("String too long");
```
fandreuz, line 40: This case would be fine. The bad case is `str.length() < len`

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        
+        buffer.position(0);
+        for (int i = 0; i < 49; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
+        if (str == null) throw new IllegalArgumentException("String cannot be null");
+        if (str.length() > len) throw new IllegalArgumentException("String too long");
```
fandreuz: This case would be fine. The bad case is str.length() < len

src/profiler.cpp
```diff
@@ -692,6 +694,12 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu | 0x8000);
     }
 
+    if (_trace_buffer[0] != '\0') {
+        char* trace_copy = (char*)malloc(Otlp::TRACE_CONTEXT_BUFFER_SIZE);
```
fandreuz: Let's put a TODO here to remember freeing, similar to what we have in #1433

src/javaApi.cpp
```diff
@@ -158,6 +162,7 @@ static const JNINativeMethod profiler_natives[] = {
     F(execute1,      "(Ljava/lang/String;)[B"),
     F(getSamples,    "()J"),
     F(filterThread0, "(Ljava/lang/Thread;Z)V"),
+    F(getThreadLocalBuffer, "()Ljava/nio/ByteBuffer;"),
```
fandreuz, line 166: Please fix the indentation of the lines above

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        
+        buffer.position(0);
+        for (int i = 0; i < 49; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
+        if (str == null) throw new IllegalArgumentException("String cannot be null");
+        if (str.length() < len) throw new IllegalArgumentException("String too long");
```
fandreuz:
```suggestion
        if (str.length() < len) throw new IllegalArgumentException("String too short");
```

src/profiler.cpp
```diff
@@ -1680,17 +1690,31 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+            }
+
             // To be written below in Profile.location_indices
             location_indices.push_back(functions.indexOf(fn.name(trace->frames[j])));
         }
-        frames_seen += trace->num_frames;
+
+        if (trace_context) {
+            size_t link_idx = links.indexOf(trace_context);
+            otlp_buffer.field(Sample::link_index, link_idx);
```
fandreuz: This should happen before `otlp_buffer.commitMessage(sample_mark)`, otherwise it won't be found by the Protobuf reader on the other side (i.e. Pyroscope)

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -290,4 +311,5 @@ private void filterThread(Thread thread, boolean enable) {
     private native byte[] execute1(String command) throws IllegalArgumentException, IllegalStateException, IOException;
 
     private native void filterThread0(Thread thread, boolean enable);
+
```
fandreuz: Spurious newline

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+    public static void setTraceContext(String traceId, String spanId) {
+        ByteBuffer buffer = BUFFER.get();
+        if (buffer == null) return;
+        
+        buffer.position(0);
+        writeString(buffer, traceId, 32);
+        writeString(buffer, spanId, 16);
+        buffer.put((byte) 0);
```
fandreuz: Is this needed? We know the size

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
+        if (str == null) throw new IllegalArgumentException("String cannot be null");
+        if (str.length() < len) throw new IllegalArgumentException("String too short");
+        byte[] bytes = str.getBytes(StandardCharsets.UTF_8);
+        buffer.put(bytes, 0, bytes.length);
+        for (int i = bytes.length; i < len; i++) buffer.put((byte) 0);
+    }
+}
```
fandreuz: Missing newline at the end of the file

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        if (buffer == null) return;
+        
+        buffer.position(0);
+        for (int i = 0; i < 49; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
+        if (str == null) throw new IllegalArgumentException("String cannot be null");
```
fandreuz: Let's remove this check, anyway there'd be a NPE just below

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,45 @@
...
+        ByteBuffer buffer = BUFFER.get();
+        if (buffer == null) return;
+        
+        buffer.position(0);
+        for (int i = 0; i < 49; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void writeString(ByteBuffer buffer, String str, int len) {
```
fandreuz: Let's remove the `len` parameter altogether, I don't think it's useful for us now

src/javaApi.cpp
```diff
@@ -148,16 +148,21 @@ Java_one_profiler_AsyncProfiler_filterThread0(JNIEnv* env, jobject unused, jthre
     }
 }
 
+extern "C" DLLEXPORT jobject JNICALL
+Java_one_profiler_AsyncProfiler_getThreadLocalBuffer(JNIEnv* env, jclass unused) {
+    return env->NewDirectByteBuffer(Profiler::traceBuffer(), 49);
```
fandreuz: 48 should be enough. Can you use the constant from `otlp.h` here?

src/otlp.h
```diff
@@ -11,11 +11,13 @@
 namespace Otlp {
 
 const u32 OTLP_BUFFER_INITIAL_SIZE = 5120;
+const u32 TRACE_CONTEXT_BUFFER_SIZE = 49;
```
fandreuz: 48 should be enough

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,38 @@
...
+        writeString(buffer, traceId);
+        writeString(buffer, spanId);
+    }
+
+    public static void clearTraceContext() {
+        ByteBuffer buffer = BUFFER.get();
+        if (buffer == null) return;
+        
+        buffer.clear();
+        }
```
fandreuz, line 45: Weird indentation

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private static void writeString(ByteBuffer buffer, String str) {
```
fandreuz: Let's just inline this, we don't need a new function now: `buffer.put(str.getBytes(StandardCharsets.UTF_8))`

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -9,6 +9,7 @@
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
```
fandreuz, line 12: Can we have integration tests for this feature? Something similar to what you did in #1413. That will tell us if the feature is broken just by running `make -j test-java TESTS=otlp`

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,79 @@
...
+        Profile profile = data.getResourceProfiles(0).getScopeProfiles(0).getProfiles(0);
+        int samplesWithLinks = 0;
+        for (Sample sample : profile.getSampleList()) {
+            if (sample.getLinkIndex() > 0) {
+                samplesWithLinks++;
+            }
+        }
+        assert samplesWithLinks > 0;
+        
+        profiler.stop();
```
fandreuz, line 121: This can be moved right after `dump`

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private static void burnCpu() {
+        Instant start = Instant.now();
+        while (Duration.between(start, Instant.now()).compareTo(TEST_DURATION) < 0) {
+            CpuBurner.burn();
+        }
+    }
+
+    public static ProfilesData dumpAndGetProfile(AsyncProfiler profiler) throws Exception {
```
fandreuz, line 131: Can you use the method from `OtlpTests` instead of defining a new one?

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assert data.getDictionary().getLinkTableCount() >= 3;
+        
+        Profile profile = data.getResourceProfiles(0).getScopeProfiles(0).getProfiles(0);
+        int samplesWithLinks = 0;
+        for (Sample sample : profile.getSampleList()) {
+            if (sample.getLinkIndex() > 0) {
+                samplesWithLinks++;
+            }
+        }
+        assert samplesWithLinks > 0;
```
fandreuz: Let's check that `link_table` contains exactly the three values above

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+    };
+
+    public static void setTraceContext(String traceId, String spanId) {
+        ByteBuffer buffer = BUFFER.get();
+        if (buffer == null) return;
+
+        buffer.position(0);
+        byte[] traceBytes = traceId.getBytes(StandardCharsets.UTF_8);
```
fandreuz: I had some more thinking about this, I think this is wrong. We should have a `16+8` buffer as we had initially. And we should convert `trace_id` and `span_id` to appropriately sized `byte[]` arrays.

Each pair of chars encodes a byte (`[0, 255]`) in hex form:
- Use something like `Character.isDigit(char c, int radix`) you can convert each character of the string to a number.
- data[i] = (byte) ((high << 4) + low)

Why I want to fix this:
- We use two times as much space as we should
- The spec [clearly states](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L297) the size of the array

src/profiler.cpp
```diff
@@ -1680,17 +1690,30 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+        }
+        otlp_buffer.commitMessage(sample_mark);
+        frames_seen += trace->num_frames - (trace_context ? 1 : 0);
```
fandreuz: I don't really like this, especially because in future we may have many types of artificial frames which should not be counted here.

Can you move the increment inside the loop, and increment by `1` every time the loop completes without a `continue`?

fandreuz: Also why was this moved here? No need

src/api/one/profiler/TraceContext.java
```diff
@@ -20,20 +22,27 @@ public static void setTraceContext(String traceId, String spanId) {
...
         
         buffer.position(0);
-        for (int i = 0; i < 48; i++) {
+        for (int i = 0; i < 24; i++) {
             buffer.put((byte) 0);
         }
     }
+
+    private static void putHexString(ByteBuffer buffer, String hex) {
+        int startPos = buffer.position();
```
fandreuz, line 40: Unused

src/api/one/profiler/TraceContext.java
```diff
@@ -0,0 +1,48 @@
...
+        if (buffer == null) return;
+        
+        buffer.position(0);
+        for (int i = 0; i < 24; i++) {
+            buffer.put((byte) 0);
+        }
+    }
+
+    private static void putHexString(ByteBuffer buffer, String hex) {
+        int startPos = buffer.position();
```
fandreuz: Unused

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final Duration TEST_DURATION = Duration.ofSeconds(3);
+
+    public static void main(String[] args) throws Exception {
+        AsyncProfiler profiler = AsyncProfiler.getInstance();
+        profiler.start(Events.CPU, 1_000_000);
+
+        Thread[] threads = new Thread[3];
+        
+        threads[0] = new Thread(() -> {
+            TraceContext.setTraceContext("09a6c61f1181ce4fc439e5728c5fef75", "b98c89ad5d208dcc");
```
fandreuz: Can you store all trace ids and span ids in constants so we can reuse them?

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        
+        for (Thread t : threads) {
+            t.join();
+        }
+
+        ProfilesData data = dumpAndGetProfile(profiler);
+        profiler.stop();
+        
+        assert data.getDictionary().getLinkTableCount() >= 3;
```
fandreuz: It should be exactly 4

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,108 @@
...
+            String linkTrace = bytesToHex(link.getTraceId());
+            String linkSpan = bytesToHex(link.getSpanId());
+            
+            if (!expectedPairs.containsKey(linkTrace) || 
+                !expectedPairs.get(linkTrace).equals(linkSpan)) {
+                allLinksValid = false;
+                break;
+            }
+        }
+        assert allLinksValid;
```
fandreuz: Please add `expectedPairs.put("", "")` so we also check for the empty link.
```suggestion
        // TODO Remove debugOutput
        StringBuilder debugOutput = new StringBuilder();
        for (int i = 0; i < data.getDictionary().getLinkTableCount(); i++) {
            Link link = data.getDictionary().getLinkTable(i);
            
            String linkTrace = bytesToHex(link.getTraceId());
            String linkSpan = bytesToHex(link.getSpanId());
            
            assert linkSpan.equals(expectedPairs.get(linkTrace));
        }
```

src/profiler.cpp
```diff
@@ -692,6 +694,13 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu | 0x8000);
     }
 
+    if (_trace_buffer[0] != 0) {
+        // TODO: free
+        uint8_t* trace_copy = (uint8_t*)malloc(Otlp::TRACE_CONTEXT_BUFFER_SIZE);
```
fandreuz: As we discussed, `malloc` is not safe in `recordSample`, we should find an alternative

src/profiler.h
```diff
@@ -90,6 +91,13 @@ class Profiler {
     bool _update_thread_names;
     volatile jvmtiEventMode _thread_events_state;
 
+    // Thread-local storage for trace correlation
+    static thread_local uint8_t _trace_context_buffer[Otlp::TRACE_CONTEXT_BUFFER_SIZE];
+    static thread_local u32 _last_trace_context_idx;
+    // TODO: Is this enough? We can ask the user to restart every x seconds
+    uint8_t _trace_contexts[Otlp::TRACE_CONTEXT_BUFFER_SIZE * 1000];
```
fandreuz: TODO: let's make this configurable by the user?

fandreuz: Also, consider conditionally allocating this memory in `Profiler::start` if it's needed (i.e. output is OTLP).

src/profiler.cpp
```diff
@@ -692,6 +695,16 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+    if (_trace_context_buffer[Otlp::TRACE_CONTEXT_CONTROL_BYTE] == 1) {
+        _trace_context_buffer[Otlp::TRACE_CONTEXT_CONTROL_BYTE] = 2;
+        _last_trace_context_idx = atomicInc(_trace_context_count);
+        u32 buffer_idx = _last_trace_context_idx * Otlp::TRACE_CONTEXT_BUFFER_SIZE;
+        memcpy(_trace_contexts + buffer_idx, _trace_context_buffer,
+                Otlp::TRACE_CONTEXT_BUFFER_SIZE);
+    }
+    if (_trace_context_buffer[Otlp::TRACE_CONTEXT_CONTROL_BYTE] == 2) {
+        num_frames += makeFrame(frames + num_frames, BCI_TRACE_CONTEXT, (jmethodID) (u64) _last_trace_context_idx);
+    }
```
fandreuz, line 722: Newline here

src/profiler.cpp
```diff
@@ -1729,6 +1752,18 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
     }
 
+    // Write link_table
+    // Start with empty link
+    protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+    otlp_buffer.commitMessage(link_mark);
+    for (size_t link_idx = 0; link_idx < trace_context_count; ++link_idx) {
+        protobuf_mark_t link_mark = otlp_buffer.startMessage(ProfilesDictionary::link_table, 1);
+        size_t start = TRACE_CONTEXT_BUFFER_SIZE * link_idx;
+        otlp_buffer.field(Link::trace_id, &_trace_contexts[start], 16);
```
fandreuz: `_trace_contexts` is basically a ring buffer. A few improvements to be considered
- Restart the index once we reach the end of the buffer (at the moment we just ignore the case, but we have to account for it)
- Properly handle old samples which got a link index, and after some time that same link index had to be overwritten by a more recent span ID, since the buffer index restarted from zero. I think we could account for this case by letting the ring buffer index grow indefinitely, and access the buffer with `idx % buffer_size`. All samples having `link_index` smaller than the last multiple of the buffer size are to be considered invalid.

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,139 @@
...
+        for (Sample sample : profile.getSampleList()) {
+            if (sample.getLinkIndex() > 0) {
+                samplesWithLinks++;
+            }
+        }
+        assert samplesWithLinks > 4;
+        
+        Map<String, Map<String, String>> threadToExpectedPairs = new HashMap<>();
+        
+        Map<String, String> thread1Pairs = new HashMap<>();
```
fandreuz: Relying on maps now just makes things more cumbersome. Please treat all the cases in dedicated if statements (`if thread1 -> expect ... else if thread2 ....`)

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,139 @@
...
+                    String linkTrace = bytesToHex(link.getTraceId());
+                    String linkSpan = bytesToHex(link.getSpanId());
+                    
+                    Map<String, String> expectedPairs = threadToExpectedPairs.get(threadName);
+                    if (expectedPairs != null) {
+                        assert expectedPairs.containsKey(linkTrace);
+                        assert expectedPairs.get(linkTrace).equals(linkSpan);
+                    } else {
+                        assert linkTrace.equals("") && linkSpan.equals("");
+                        System.out.println("OK: System thread " + threadName);
```
fandreuz: No output please

test/test/otlp/OtlpTests.java
```diff
@@ -75,6 +75,14 @@ public void aggregationTemporality(TestProcess p) throws Exception {
         assert p.exitCode() == 0;
     }
 
+    @Test(mainClass = OtlpSignalCorrelationTest.class, jvmArgs = "-Djava.library.path=build/lib")
+    public void otlpSignalCorrelationTest(TestProcess p) throws Exception {
```
fandreuz:
```suggestion
    public void signalCorrelation(TestProcess p) throws Exception {
```

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,139 @@
...
+        thread2Pairs.put(TRACE_ID_2, SPAN_ID_2);
+        threadToExpectedPairs.put("CorrelationThread2", thread2Pairs);
+
+        Map<String, String> thread3Pairs = new HashMap<>();
+        thread3Pairs.put("", "");
+        thread3Pairs.put(TRACE_ID_3, SPAN_ID_3);
+        threadToExpectedPairs.put("CorrelationThread3", thread3Pairs);
+        
+        for (Sample sample : profile.getSampleList()) {
+            if (sample.getLinkIndex() >= 0) {
```
fandreuz, line 78: What happens if this condition evaluates to `false` for all samples? The test will pass, but that's obviously an undesirable situation

adinazugravescu: Already checked samplesWithLinks > 4 above, so we're guaranteed to have samples with linkIndex >= 0

test/test/otlp/OtlpTests.java
```diff
@@ -123,7 +131,7 @@ private static Profile getFirstProfile(ProfilesData profilesData) {
         return scopeProfiles.getProfiles(0);
     }
 
-    private static Optional<AnyValue> getAttribute(Sample sample, ProfilesDictionary dictionary, String name) {
+    public static Optional<AnyValue> getAttribute(Sample sample, ProfilesDictionary dictionary, String name) {
```
fandreuz: Package-private is enough:
```suggestion
    static Optional<AnyValue> getAttribute(Sample sample, ProfilesDictionary dictionary, String name) {
```

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,125 @@
...
+        
+        for (Thread t : threads) {
+            t.join();
+        }
+
+        ProfilesData data = dumpAndGetProfile(profiler);
+        profiler.stop();
+        
+        assert data.getDictionary().getLinkTableCount() == 5 : 
+            data.getDictionary().getLinkTableCount();
```
fandreuz: Please add some context here, otherwise we'll get `AssertionError + someNumber` which is not really informative

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,125 @@
...
+                Optional<AnyValue> threadNameOpt = OtlpTests.getAttribute(sample, data.getDictionary(), "thread.name");
+                if (threadNameOpt.isPresent()) {
+                    String threadName = threadNameOpt.get().getStringValue();
+                    
+                    Link link = data.getDictionary().getLinkTable(sample.getLinkIndex());
+                    String linkTrace = bytesToHex(link.getTraceId());
+                    String linkSpan = bytesToHex(link.getSpanId());
+                    
+                    if (threadName.equals("CorrelationThread1")) {
+                        assert linkTrace.equals("") && linkSpan.equals("") ||
```
fandreuz: Can you test that we found at least one trace_id for **each** branch? including trace_id_1_alt

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,125 @@
...
+                            linkTrace.equals(TRACE_ID_1) && linkSpan.equals(SPAN_ID_1) ||
+                            linkTrace.equals(TRACE_ID_1_ALT) && linkSpan.equals(SPAN_ID_1_ALT);
+                    } else if (threadName.equals("CorrelationThread2")) {
+                        assert linkTrace.equals("") && linkSpan.equals("") ||
+                            linkTrace.equals(TRACE_ID_2) && linkSpan.equals(SPAN_ID_2);
+                    } else if (threadName.equals("CorrelationThread3")) {
+                        assert linkTrace.equals("") && linkSpan.equals("") ||
```
fandreuz: Please add some parentheses

src/arch.h
```diff
@@ -43,6 +43,10 @@ static inline u64 loadAcquire(u64& var) {
     return __atomic_load_n(&var, __ATOMIC_ACQUIRE);
 }
 
+static inline u32 loadAcquire(u32& var) {
```
fandreuz: This is probably not needed anymore

src/profiler.cpp
```diff
@@ -1699,15 +1715,25 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
+                u32 key = (u32) (u64) trace->frames[j].method_id;
+                size_t link_idx = links.indexOf(trace_contexts_map[key]);
+                otlp_buffer.field(Sample::link_index, link_idx);
+                continue;
+            }
+            
             // To be written below in Profile.location_indices
             location_indices.push_back(functions.indexOf(fn.name(trace->frames[j])));
-            ++frames_seen;
+            frames_seen++;
```
fandreuz: Is this change needed?

src/profiler.cpp
```diff
@@ -1699,15 +1715,25 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
         }
+
+        otlp_buffer.field(Sample::locations_start_index, locations_start_index);
+        otlp_buffer.field(Sample::locations_length, frames_seen - locations_start_index);
+        
         if (thread_name_idx != 0) {
             otlp_buffer.field(Sample::attribute_indices, thread_name_idx);
         }
-
+
```
fandreuz: Spurious space

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,138 @@
...
+                    }
+                }
+            }
+        }
+
+        assert foundTrace1 : "TRACE_ID_1 not found";
+        assert foundTrace1Alt : "TRACE_ID_1_ALT not found";
+        assert foundTrace2 : "TRACE_ID_2 not found";
+        assert foundTrace3 : "TRACE_ID_3 not found";
+
```
fandreuz, line 113: Additional newline

test/test/otlp/OtlpSignalCorrelationTest.java
```diff
@@ -0,0 +1,138 @@
...
+        
+        for (Thread t : threads) {
+            t.join();
+        }
+
+        ProfilesData data = dumpAndGetProfile(profiler);
+        profiler.stop();
+        
+        assert data.getDictionary().getLinkTableCount() == 5 : 
+            "Expected 5 links in link_table, but got: " + data.getDictionary().getLinkTableCount();
```
fandreuz: I think it does not hurt to print the whole dictionary here:
```suggestion
        assert data.getDictionary().getLinkTableCount() == 5 : data.getDictionary().getLinkTable().toString();
```
double check that it compiles

src/profiler.h
```diff
@@ -163,6 +170,7 @@ class Profiler {
         _thread_filter(),
         _call_trace_storage(),
         _jfr(),
+        _trace_context_allocator(1024 * 1024),
```
fandreuz: Initial size should be a multiplier of 24 (use the constant from `otlp.h` please).
Also, 1Mb is probably too much to start with, I think we could even start with zero: anyway, at some point we'll have to resize. And it's better to have consistent performance than "all good and then suddenly it's bad".

src/profiler.cpp
```diff
@@ -1116,6 +1131,7 @@ Error Profiler::start(Arguments& args, bool reset) {
         _class_map.clear();
         _thread_filter.clear();
         _call_trace_storage.clear();
+        _trace_context_allocator.clear();
```
fandreuz, line 1156: We should also handle the `thread_local` ptr somehow. We cannot set it to null here, but we also can't just let it be read at the next `recordSample`. Is it handled and tested?

src/profiler.cpp
```diff
@@ -692,6 +695,18 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         num_frames += makeFrame(frames + num_frames, BCI_CPU, java_ctx.cpu | 0x8000);
     }
 
+    if (_trace_context_buffer[Otlp::TRACE_CONTEXT_CONTROL_BYTE] == 1) {
+        _trace_context_buffer[Otlp::TRACE_CONTEXT_CONTROL_BYTE] = 2;
+        void* trace_ptr = _trace_context_allocator.alloc(24);
```
fandreuz: Use the constant from `otlp.h`

src/profiler.cpp
```diff
@@ -692,6 +695,18 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+        if (trace_ptr != NULL) {
```
fandreuz, line 713: We should set `_last_trace_context_ptr` to null if this check is unsuccessful

test/test/otlp/OtlpTests.java
```diff
@@ -75,6 +75,14 @@ public void aggregationTemporality(TestProcess p) throws Exception {
         assert p.exitCode() == 0;
     }
 
+    @Test(mainClass = OtlpSignalCorrelationTest.class, jvmArgs = "-Djava.library.path=build/lib")
+    public void signalCorrelation(TestProcess p) throws Exception {
```
fandreuz, line 79: If you have time, please add a test `signalCorrelationWithRestart`. We don't really need to `loop`, you can restart manually. Bonus points if we check the case with spans which are still running while the profiler is restarted. It's fine to restart those spans for now, but we should handle that situation gracefully

test/test/otlp/OtlpSignalCorrelationWithRestartTest.java
```diff
@@ -0,0 +1,40 @@
...
+            for (int i = 0; i < 3; i++) {
+                burnCpu();
+            }
+            
+        }, "LongSpan");
+        
+        profiler.execute("start,otlp,threads");
+        longSpanThread.start();
+        
+        Thread.sleep(1000);
```
fandreuz: You can't really be sure this is going to wake up before your thread dies. I suggest doing something like this in the thread, so you have deterministic control over how long it's going to be alive:
https://github.com/async-profiler/async-profiler/blob/d97a7d334331383c40a062572fa683d46c954492/test/test/instrument/CpuBurner.java#L15-L21

fandreuz: You can sleep 500ms and let the thread run for 1s, that should be enough.

test/test/otlp/OtlpSignalCorrelationWithRestartTest.java
```diff
@@ -0,0 +1,40 @@
...
+        profiler.stop();
+        profiler.execute("start,otlp,threads");
```
fandreuz, line 33: How about `profiler.execute("restart")`?

test/test/otlp/OtlpSignalCorrelationWithRestartTest.java
```diff
@@ -0,0 +1,40 @@
...
+        
+        longSpanThread.join();
+        
+        ProfilesData data = dumpAndGetProfile(profiler);
+        profiler.stop();
+        
+        assert data.getDictionary().getLinkTableCount() == 1 : data.getDictionary().getLinkTableList().toString();
+
```
fandreuz: Additional newline

test/test/otlp/OtlpSignalCorrelationWithRestartTest.java
```diff
@@ -0,0 +1,40 @@
...
+        Thread.sleep(1000);
+        profiler.stop();
+        profiler.execute("start,otlp,threads");
+        
+        longSpanThread.join();
+        
+        ProfilesData data = dumpAndGetProfile(profiler);
+        profiler.stop();
+        
+        assert data.getDictionary().getLinkTableCount() == 1 : data.getDictionary().getLinkTableList().toString();
```
fandreuz, line 40: This is unexpected(?) Where is the link table coming from, since we restarted the profiler and the span context info is gone?

src/profiler.cpp
```diff
@@ -614,6 +618,11 @@ void Profiler::fillFrameTypes(ASGCT_CallFrame* frames, int num_frames, NMethod*
 u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Event* event) {
     atomicInc(_total_samples);
 
+    if (_last_epoch != _epoch) {
```
fandreuz, line 621: I'm pretty sure this should be read with some level of synchronization, but you can leave it as it is for now.

test/test/otlp/OtlpSignalCorrelationWithRestartTest.java
adinazugravescu: We should still have the empty link for samples without trace context

fandreuz: Right, thanks

apangin:
C++ `thread_local` is not async-signal-safe.
We will use a different approach to pass and store SpanId/TraceId.

PR #1440, opened 2025-08-18
Remove unused included headers

fandreuz:
### Description
In this PR I'm cleaning a couple of unused inclusions. 

I'm also adding `misc-include-cleaner` to our `.clang-tidy`. We got the new configuration option `misc-include-cleaner.MissingIncludes` in the most recent build of `clang-tidy`, which is hardcoded to `true` in older versions. Having `MissingIncludes` enabled would generate too much noise, thus why we need to disable it.

### Related issues
n/a

### Motivation and context
`misc-include-cleaner` can clean up some of our unused include statements.

### How has this been tested?
n/a

---

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -132,3 +132,5 @@ CheckOptions:
     value: UPPER_CASE
   - key: misc-include-cleaner.MissingIncludes
     value: false
+  - key: misc-include-cleaner.IgnoreHeaders
+    value: "sys/time.h" # https://github.com/llvm/llvm-project/issues/157876
--- a/src/ctimer_linux.cpp
+++ b/src/ctimer_linux.cpp
@@ -11,6 +11,9 @@
 #include <unistd.h>
 #include "ctimer.h"
 #include "j9StackTraces.h"
+#include "log.h"
+#include "os.h"
+#include "vmEntry.h"
 
 
 #ifndef SIGEV_THREAD_ID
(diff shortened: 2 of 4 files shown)


docker/code-check.Dockerfile
```diff
@@ -2,7 +2,22 @@
...
+    libc++ \
+    libc++-dev \
+    make \
+    git \
+    python3 \
+    py3-pip \
+    bash
+RUN apk add \
+    --no-cache \
+    --repository=https://dl-cdn.alpinelinux.org/alpine/edge/testing \
```
fandreuz: This is coming from the `testing` repository since `clang21` is a release candidate, it will be available in the mainline repo in a couple of weeks

.clang-tidy
```diff
@@ -128,3 +128,5 @@ CheckOptions:
     value: '(KB|Thread|setDaemon|klassOop|nVMs|loadLibrary|getTicksFrequency|counterTime|System|M|R|s_)'
   - key: readability-identifier-naming.PrivateMemberPrefix
     value: _
+  - key: misc-include-cleaner.MissingIncludes
```
fandreuz, line 133: Unfortunately, this setting is not available in previous versions. Thus anybody running locally without the Docker image is most likely going to receive some more warnings than they should. Not sure if this is a problem

fandreuz:
Closing for now, too many "special cases".

PR #1443, opened 2025-08-20
only measure blocked time around the synchronized lock

Baraa-Hasheesh:
### Description
Only account for the blocked time around the synchronized lock to avoid increments happening `ObjectMonitor::wait` & `ObjectMonitor::INotify` JDK methods

### Related issues
N/A

### Motivation and context
Fix potential test fragileness 

### How has this been tested?
Test was ran 100 times here https://github.com/Baraa-Hasheesh/async-profiler/actions/runs/17096889460
And here https://github.com/Baraa-Hasheesh/async-profiler/actions/runs/17098823202/job/48489764236 

---

### DIFF
--- a/test/test/jfr/JfrMultiModeProfiling.java
+++ b/test/test/jfr/JfrMultiModeProfiling.java
@@ -6,12 +6,15 @@
 package test.jfr;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -29,7 +32,7 @@ public class JfrMultiModeProfiling {
     private static final List<byte[]> holder = new ArrayList<>();
 
     private static final ThreadMXBean tmx = ManagementFactory.getThreadMXBean();
-    private static final Map<Long, Long> threadLockTimes = new ConcurrentHashMap<>();
+    private static final Map<Long, Boolean> threadIds = new ConcurrentHashMap<>();
 
     static {
         tmx.setThreadContentionMonitoringEnabled(true);
@@ -37,29 +40,29 @@ public class JfrMultiModeProfiling {
 
     public static void main(String[] args) throws InterruptedException {
         ExecutorService executor = Executors.newFixedThreadPool(2);
+        List<CompletableFuture<Void>> futures = new ArrayList<>();
+
         for (int i = 0; i < 10; i++) {
-            executor.submit(JfrMultiModeProfiling::cpuIntensiveIncrement);
+            futures.add(CompletableFuture.runAsync(JfrMultiModeProfiling::cpuIntensiveIncrement, executor));
         }
         allocate();
+        futures.forEach(CompletableFuture::join);
+
+        long[] ids = threadIds.keySet().stream().mapToLong(Long::longValue).toArray();
+        Arrays.stream(tmx.getThreadInfo(ids)).mapToLong(ThreadInfo::getBlockedTime).forEach(System.out::println);
+
         executor.shutdown();
         executor.awaitTermination(10, TimeUnit.SECONDS);
-
-        threadLockTimes.values().forEach(System.out::println);
     }
 
     private static void cpuIntensiveIncrement() {
-        long threadId = Thread.currentThread().getId();
-        long totalBlockedTime = threadLockTimes.getOrDefault(threadId, 0L);
+        threadIds.putIfAbsent(Thread.currentThread().getId(), Boolean.TRUE);
 
         for (int i = 0; i < 100_000; i++) {
-            long previousBlockedTime = tmx.getThreadInfo(threadId).getBlockedTime();
             synchronized (lock) {
-                totalBlockedTime += tmx.getThreadInfo(threadId).getBlockedTime() - previousBlockedTime;
                 count += System.getProperties().hashCode();
             }
         }
-
-        threadLockTimes.put(threadId, totalBlockedTime);
     }
 
     private static void allocate() {
--- a/test/test/jfr/JfrTests.java
+++ b/test/test/jfr/JfrTests.java
@@ -91,7 +91,7 @@ public class JfrTests {
 
         Assert.isGreater(eventsCount.get("jdk.ExecutionSample"), 50);
         Assert.isGreater(eventsCount.get("jdk.JavaMonitorEnter"), 10);
-        Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.90);
+        Assert.isGreater(jfrTotalLockDurationMillis / totalLockDurationMillis, 0.80);
         Assert.isGreater(eventsCount.get("jdk.ObjectAllocationInNewTLAB"), 50);
     }
 


test/test/jfr/JfrMultiModeProfiling.java
```diff
@@ -48,14 +48,18 @@ public static void main(String[] args) throws InterruptedException {
     }
 
     private static void cpuIntensiveIncrement() {
+        long threadId = Thread.currentThread().getId();
+        long totalBlockedTime = threadLockTimes.getOrDefault(threadId, 0L);
+
         for (int i = 0; i < 100_000; i++) {
+            long previousBlockedTime = tmx.getThreadInfo(threadId).getBlockedTime();
             synchronized (lock) {
+                totalBlockedTime += tmx.getThreadInfo(threadId).getBlockedTime() - previousBlockedTime;
```
apangin: `getThreadInfo` adds an expensive native call inside a loop, dramatically changing the way how the loop is compiled and executed. The code being tested is now quite different from what it was before.

Since there are no other contended locks inside the loop, can we hoist `getThreadInfo` calls outside?

test/test/jfr/JfrTests.java
```diff
@@ -102,7 +102,7 @@ public void parseMultiModeRecording(TestProcess p) throws Exception {
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
     @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,file=%f.jfr", nameSuffix = "noOverride")
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,alloc=100,file=%f.jfr", nameSuffix = "overrideAlloc")
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,all,alloc=262143,file=%f.jfr", nameSuffix = "overrideAlloc")
```
apangin, line 105: Why is this change?

Baraa-Hasheesh: alloc=100 was causing the test to be much slower than need be as it was capture many allocation samples 
I noticed this when I added `tmx.getThreadInfo(threadId)` into the loop which caused the test run duration to jump from 5seconds to +30 seconds in some cases

The test doesn't really check that the new interval is effecting the overall profiling it just checks that profiler run normally 

I didn't see any harm of changing this value to make the test faster

test/test/jfr/JfrMultiModeProfiling.java
Baraa-Hasheesh: The problem here if we hoist the `getThreadInfo` out of the loop we would go back to having `ObjectMonitor::wait` & `ObjectMonitor::INotify` side effects on our values

Adding the `getBlockedTime` inside the loop minimizes those side effects 

Regarding the expensive native call, yes I did notice that it did effect the overall run time 
But the difference wasn't that bad/severe, so I didn't consider it a big issue

apangin: I'm not talking about performance. I mean the test is changed in a way that tries to workaround failure by changing _what_ is executed instead of _how_ it is measured. This is similar to adding `sleep` for "fixing" concurrency issues.

Where is `wait` called inside the loop?

Baraa-Hasheesh: I did quick check to see where it's called 

It's usually called outside the loop

image

I will bench mark hoisting it outside the loop to check it it's stable, during my previous benchmarks for outside the loop I did see higher difference than using inside the loop

Baraa-Hasheesh: Just to clarify the problem is in the `ObjectMonitor::INotify` function rather than the `ObjectMonitor::wait` function

as the counter is started in the notify function & disabled in the wait, I'm not sure where the INotify is called, but it's called in such a way that it pollutes the timer if we hoist above the method, 

An alternative **hacky** way to fix this is to make the CPU sampling more frequency which in turn will interrupt any wait the JDK might enforce between  `ObjectMonitor::INotify` & `ObjectMonitor::wait` this in turn would result in less interference as the counter is disabled early 

I will try to use perf/begin-end profiling to capture the ObjectMonitor::INotify call & understand where/how it's called to better understand why it's having this kind of effect

Baraa-Hasheesh: To confirm, hoisting the statement outside the loop isn't stable, it will fail if we do 100 runs back to back

Baraa-Hasheesh: small update about current state 

I added some custom samples via traps & got the following 

image

image

while I can confirm that some timers is called during random times during the thread lifecycle, I'm not able to fully confirm which timer is being incremented here (I'm speculating that it's lock timer)

My theory is that these small samples happening are snow balling to cause the issue, 

Furthermore what's kinda funny in a bad way is that the ThreadBean it self is contributing to the counter 

image

I'm checking possible ways to minimize these.
What I did find interesting is that if we do small interval for cpu (for example 1us) the test stabilizes which let's me to believe the counter is started in some locations & followed by a wait statement. (Still trying to confirm this)

Baraa-Hasheesh: @apangin while they are not fully equivalent, in this particular case Java Flight recorder does offer a more controlled base line compared to the ThreadBean objects (Less location where the timers might be incremented)

```
Recording recording = new Recording();
recording.enable("jdk.JavaMonitorEnter").with("threshold", "0 ms");
recording.setDestination(new File(args[0]).toPath());
recording.start();
```

I did check the src code, the flight recorder starts counting time a little before `post_monitor_contended_enter`
=> Which we use to initialize our timer in async-profiler

It also stops the timer shortly after calling `post_monitor_contended_entered`
=> which is used to record the profiler event

I did run against the 0.80 margin 100 times & it's stable, now the margin will still have some cases where it's lower than 90% due to the following two problems:
* async-profiler unwind overhead effects the JDK flight recorder time 
* async-profiler might interrupt a lock sample in the middle of it's unwinding (Due to cpu profiling) which in turn will add to the flight recorder time 

I would say if the margin drops below 80% one of the following cases should hold true:
* CPU scheduler was behaving badly (Can theoretically happen in any implementation that we choose so I don't think we should consider it as deal breaker)
* Profiler overhead was significant enough to break the test (We should check if we can optimize)

I don't think it's worth chasing where the problem happens in the JDK side (As it seems to have been solved in newer JDKs as I only observed the instability in older JDKs), & would recommend just moving the logic to use the flight recorder as a base line

What are your thoughts?

apangin: JFR is heavyweight (in every respect). Introducing Flight Recorder and JFR parser just for the purpose of having a baseline seems an overkill to me. The fact that JFR event duration includes async-profiler's stack walking time is also a point against it.

I wonder why you gave up ThreadMXBean approach if you already found the cause of discrepancy? If wait/notify comes from class linking/initialization, you can probably call the corresponding methods once ahead of time to make sure all initialization is complete by the time you start benchmarking.

Baraa-Hasheesh: while it's possible to use the ThreadBean by minimizing calls that may cause the timer to be effected, 
I wanted to replace the `ThreadMXBean` approach as it had more vectors of control,

While JFR is indeed heavy, it has less vectors of control which makes look better for the purpose,

Now regarding the ThreadMXBean, the best way IMO is basically remove the ThreadMXBean & any other weird calls from the picture, this can be done by changing the Thread strategy from a simple Thread to a completable future & with small touches on the main method

Basically we run the contented function as a completable future & wait for all to complete 
After which we collect all thread information in the main Thread via a single call to `tmx.getThreadInfo`

This should reduce the noise we have,

The mentioned has been implemented in this commit [d877229](https://github.com/async-profiler/async-profiler/pull/1443/commits/d8772295396eb9e9efdd9b7e679e5bb2e00ed39d)

The changed code has been run 100 times with no failures https://github.com/Baraa-Hasheesh/async-profiler/actions/runs/17289300838

Issue #1402, opened 2025-07-22
Incorrect stack trace for Math.pow() on Arm


WadeWalker:
### Describe the bug

Hi guys,

When I profiled a workload that's heavy on `Math.pow()` (inspired by DaCapo biojava-large), I noticed that the stacks are resolved correctly on Intel and AMD, but not on Arm. The bad Arm stacks are of the form `all/break_interpreted/unknown/SharedRuntime::dpow`.

The `Math.pow()` function doesn't seem to have an intrinsic on Arm, but it does on x86. So the problem may really be with unwinding stacks that originate in functions that live in a shared library.

This looks like a different phenomenon than I reported in https://github.com/async-profiler/async-profiler/issues/1325 and https://github.com/async-profiler/async-profiler/issues/1385.

### Expected vs. actual behavior

Expected flame graph, like this one from m7a:

<img width="1199" height="154" alt="Image" src="https://github.com/user-attachments/assets/2241598e-75df-4e34-b9de-eb273f923a74" />

Actual flame graph on m8g:

<img width="1186" height="150" alt="Image" src="https://github.com/user-attachments/assets/9a3c80f3-4e40-4fa7-8274-4f387d167c1f" />

### Reproduction Steps

Note: instead of the below, you can also look for the `Math.pow()` stack traces in DaCapo biojava, if you already have them available. Otherwise...

Build this little benchmark by unzipping, cd-ing into it, and typing `make`:

[wbench.tar.gz](https://github.com/user-attachments/files/21375149/wbench.tar.gz)

Profile it like this:

`java -agentpath:async-profiler-4.1-linux-arm64/lib/libasyncProfiler.so=start,cstack=vm,event=cpu,wall,file=profile.jfr -jar wbench.jar math-pow`

Convert JFR to flame graph with this command:

`java -jar jfr-converter.jar --inverted --state default -o html profile.jfr profile.html`

Observe the bad stack frames :)

### Additional Information/Context

JFR traces for the benchmark:

[wbench-math-pow-m7a.metal-48xl-profile.jfr.gz](https://github.com/user-attachments/files/21375280/wbench-math-pow-m7a.metal-48xl-profile.jfr.gz)
[wbench-math-pow-m8g.metal-24xl-profile.jfr.gz](https://github.com/user-attachments/files/21375279/wbench-math-pow-m8g.metal-24xl-profile.jfr.gz)

Flame graphs for the benchmark:

[...]

PR #1445, opened 2025-08-21
Prevent the user from selecting end & begin as the same symbol

Baraa-Hasheesh:
### Description
While debugging other parts of the profiler I noticed that if the user choses begin & end functions as the same function that would result in an infinite loop

The infinite loop happens due to the following:
* Profiler installs Breakpoint at begin function
* Function is called, which causes a breakpoint which issues a SIGTRAP signal & that is captured by the profiler trap handler
* Profiler uninstalls begin trap
* Profiler installs end trap by adding a Breakpoint at end function which is the same as begin function
* Profiler reverts the PC to the start of the function as to allow the function to start normally 
* Function is re-executed from start which triggers the breakpoint again
* Profiler uninstalls end trap
* profiler installs begin trap
* Profiler reverts the PC to the start of the function as to allow the function to start normally 
* repeat 

### Related issues
N/A

### Motivation and context
prevent bad inputs on profiler from breaking the user application 

### How has this been tested?
a new test was added 
`make test`

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -810,7 +810,7 @@ Error Profiler::installTraps(const char* begin, const char* end, bool nostop) {
         return Error("End address not found");
     }
 
-    // begin & end shouldn't have the same address, as that would result in an infinite loop
+    // Having 'begin' and 'end' traps at the same address would result in an infinite loop
     if (begin_addr && begin_addr == end_addr) {
         return Error("begin and end symbols should not resolve to the same address");
     }


src/profiler.cpp
```diff
@@ -810,6 +810,11 @@ Error Profiler::installTraps(const char* begin, const char* end, bool nostop) {
         return Error("End address not found");
     }
 
+    // begin & end shouldn't have the same address, as that would result in an infinite loop
+    if (begin_addr && begin_addr == end_addr) {
+        return Error("begin & end should be 2 different symbols");
```
krk: I propose `begin and end symbols should not resolve to the same address`. They can be different symbols with the same address.

Baraa-Hasheesh: sure, handled here [1153784](https://github.com/async-profiler/async-profiler/pull/1445/commits/115378422280f3ffc73f45e93641c244e598d72b)

test/test/c/begin_end.c
```diff
@@ -0,0 +1,57 @@
...
+#else
+#define EXT ".dylib"
+#endif
+
+static void fail(const char* msg) {
+    fprintf(stderr, "%s\n", msg);
+    exit(1);
+}
+
+extern void testMethod() {
```
apangin: Any reason why this can't be tested with exiting functions in the standard C library or in libjvm?
It's better to avoid creating a separate C program for testing every little thing,

Baraa-Hasheesh: It's possible to reuse the `Ttsp` test `begin=SafepointSynchronize::begin,end=SafepointSynchronize::begin`
would that be acceptable?

apangin: Reuse in what sense? If I understand intention of the test correctly, any native function that always exists will be sufficient.

Baraa-Hasheesh: That's true, we need a native function that's always exists & is actually called during the application lifespan (make sure everything is still working)

In this test I'm also checking the stdout to validate that the application is still working normally 

regarding the reuse, running the application with the mentioned args should achieve the same purpose as here

Baraa-Hasheesh: @apangin just to confirm, is the issue here the custom function or the new test program 

I understood the comment as let's reuse one of the old tests that we already have

apangin: It's about new C program, which has lots of boilerplate code to load and start profiler and which needs to be compiled separately. Since the problem is not specific to native API, I suggest to have a regular Java test instead. It can be an existing program or a new one, I don't mind. Just keep it simple.

Baraa-Hasheesh: handled [e12d9a8](https://github.com/async-profiler/async-profiler/pull/1445/commits/e12d9a81b9e7c5a8779490081932d0c922fa0bae)

test/test/jfr/JfrTests.java
```diff
@@ -198,6 +198,14 @@ public void ttspNostop(TestProcess p) throws Exception {
         assert containsSamplesOutsideWindow(p) : "Expected to find samples outside of ttsp window";
     }
 
+    @Test(mainClass = Ttsp.class, agentArgs = "start,begin=SafepointSynchronize::begin,end=SafepointSynchronize::begin,file=%f.jfr", output = true)
+    public void beginEnd(TestProcess p) throws Exception {
+        Output out = p.waitForExit(TestProcess.STDOUT);
```
apangin, line 204: No need to wait for several seconds until the whole test completes just to check one warning.
This can be any HelloWorld-like program that exits immediately.

Baraa-Hasheesh: makes sense updated [f5ef57c](https://github.com/async-profiler/async-profiler/pull/1445/commits/f5ef57cab03d9b825457d24bce740bc386b92489)

Please note I switched to using `jni_CallStaticVoidMethod` to keep the implicit infinite loop check (Process finishes normally)

Baraa-Hasheesh: Failure in GHA seems related to #1402
Unwind didn't happen correctly from SharedRuntime::dpow

apangin: OK, although the choice of a function is a bit strange: `jni_CallStaticVoidMethod` is not the most obvious one, and it's a private symbol that does not exist in some JDK distributions.
Can it be a basic function from libc, for example, `write`? It's an exported symbol and `write` is certainly called in HelloWorld.

Baraa-Hasheesh: `write` should work find for our purposes, 
=> "jni_CallStaticVoidMethod" looked like a good candidate for me as it's called as part of the `main` function flow for hotspot JDKs

Baraa-Hasheesh: updated [efa0ea1](https://github.com/async-profiler/async-profiler/pull/1445/commits/efa0ea1a9b1d43b2ae28b4f71c52f4f53d11c3cf)

PR #1448, opened 2025-08-22
Thread name in OpenTelemetry output

fandreuz:
### Description
`thread.name` is an optional `Sample` attribute. We can provide it when it is available.

### Related issues
#1436

### Motivation and context
Will simplify testing of #1436.

### How has this been tested?
New test.

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1749,7 +1749,7 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     // Write attribute_table (only threads for now)
     strings.forEachOrdered([&] (size_t idx, const std::string& s) {
         protobuf_mark_t attr_mark = otlp_buffer.startMessage(ProfilesDictionary::attribute_table);
-        otlp_buffer.field(Key::key, OTLP_THREAD_NAME, OTLP_THREAD_NAME_LEN);
+        otlp_buffer.field(Key::key, OTLP_THREAD_NAME);
         protobuf_mark_t value_mark = otlp_buffer.startMessage(Key::value);
         otlp_buffer.field(AnyValue::string_value, s.data(), s.length());
         otlp_buffer.commitMessage(value_mark);
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1651,6 +1651,8 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     ProtoBuffer otlp_buffer(OTLP_BUFFER_INITIAL_SIZE);
     Index strings;
     Index functions;
+    // Eventually this is going to be Index<Attribute>, for now we keep it simple
+    Index thread_names;
 
     protobuf_mark_t resource_profiles_mark = otlp_buffer.startMessage(ProfilesData::resource_profiles);
     protobuf_mark_t scope_profiles_mark = otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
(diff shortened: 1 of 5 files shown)


src/profiler.cpp
```diff
@@ -1680,17 +1682,27 @@ void Profiler::dumpOtlp(Writer& out, Arguments& args) {
...
-        protobuf_mark_t sample_value_mark = otlp_buffer.startMessage(Sample::value, 1);
-        otlp_buffer.putVarInt(cts->samples);
-        otlp_buffer.putVarInt(cts->counter);
-        otlp_buffer.commitMessage(sample_value_mark);
-        otlp_buffer.commitMessage(sample_mark);
 
+        u32 thread_name_idx = 0;
         for (int j = 0; j < trace->num_frames; j++) {
+            if (trace->frames[j].bci == BCI_THREAD_ID) {
+                thread_name_idx = thread_names.indexOf(fn.name(trace->frames[j]));
```
apangin: `FrameName` returns a decorated name. Shall OTLP record the orignal name instead?

test/one/profiler/test/Output.java
```diff
@@ -40,6 +40,10 @@ public boolean contains(String regex) {
         return stream(regex).findAny().isPresent();
     }
 
+    public boolean containsExact(String string) {
+        return Arrays.stream(lines).anyMatch(s -> s.contains(string));
```
apangin: `Arrays.stream(lines)` -> `stream()`

src/profiler.cpp
fandreuz: Yeah that makes sense: bce03feb9e2a4094cbafc4cda0a33732ff50d00d

There's also `thread.id`: https://opentelemetry.io/docs/specs/semconv/registry/attributes/thread/

But that's not what we have, so I'll leave it there

test/one/profiler/test/Output.java
fandreuz: ebbf22c5b2ceabe4fc1b27a183a87a0f9304747c

test/test/otlp/OtlpTests.java
```diff
@@ -42,19 +40,36 @@ public void testSampleType(TestProcess p) throws Exception {
...
+        Profile profile = getFirstProfile(profilesData);
+        assert profile.getSampleTypeList().size() == 2;
+
+        Set<String> threadNames = new HashSet<>();
+        for (Sample sample: profile.getSampleList()) {
+            Optional<AnyValue> threadName = getAttribute(sample, profilesData.getDictionary(), "thread.name");
+            if (!threadName.isPresent()) continue;
+            threadNames.add(threadName.get().getStringValue());
+        }
+        boolean cpuBurnerFound = threadNames.stream().anyMatch("CpuBurnerWorker"::equals);
```
apangin: Must-read post: https://medium.com/@tagir_valeev/use-stream-api-simpler-or-dont-use-it-at-all-ea0a44a4b1ff
Notably §1.6

src/profiler.cpp
apangin: We have [Java Thread IDs](https://github.com/async-profiler/async-profiler/blob/5454c9bf7f3628c2510b832de3676a51669b4bd7/src/profiler.h#L60), even though async-profiler uses OS tid as primary thread ID.

test/test/otlp/OtlpTests.java
fandreuz: Ouch, this made a bit more sense before 7adc34426ce9c7db4aa9b5b7756ca4a78dff905d

Now it doesn't indeed: 82a084283bd8c0db215d21f3740830076619c950

Issue #1441, opened 2025-08-19
RecoveryTests.intrinsics/VM fails sometimes


fandreuz:
### Describe the bug

Fails on MacOS and AL2: 
- https://github.com/fandreuz/async-profiler/actions/runs/17064291180/job/48377804145#step:11:357
- https://github.com/fandreuz/async-profiler/actions/runs/17064291180/job/48377804076#step:10:326

### Async-profiler version

master

Baraa-Hasheesh:
For macOS it might be linked to #1418

apangin:
No, this is not macOS-specific and not architecture specific, even though the test is more likely to fail on AArch64.
The test highlights a real stack walking problem that I've summarized in #1444.

apangin:
Fixed in #1449

PR #1449, opened 2025-08-22
Special handling of prologue and epilogue of compiled methods with cstack=vm

### LINKED ISSUE #1444: cstack=vm should collect stacktraces at method epilogue
### Describe the feature

Improve stack walking to account for the epilogue of JIT-compiled methods.

The epilogue of a compiled methods removes the top frame off the stack, performs a safepoint poll and returns. During execution of this sequence, stack pointer is incremented, and the invariant `previous_sp = current_ + frame_size` no longer holds true. This results in stack unwinding failures when sampling hits the epilogue. `StackWalker::walkVM` needs to handle such situations to reduce the number of broken stacks with `unknown` frames.

The epilogue is different across various HotSpot versions.

#### JDK 8

```
  0x00007feaa1415030: add    $0x40,%rsp
  0x00007feaa1415034: pop    %rbp
  0x00007feaa1415035: test   %eax,0x16ca8fc5(%rip)        ;   {poll_return}
  0x00007feaa141503b: ret    
```

#### JDK 11

```
  0x00007cfe0d9c249a: add    $0x40,%rsp
  0x00007cfe0d9c249e: pop    %rbp
  0x00007cfe0d9c249f: mov    0x108(%r15),%r10
  0x00007cfe0d9c24a6: test   %eax,(%r10)        ;   {poll_return}
  0x00007cfe0d9c24a9: ret    
```

#### JDK 17-25

```
  0x00007ceff427eb72:   add    $0x40,%rsp
  0x00007ceff427eb76:   pop    %rbp
  0x00007ceff427eb77:   cmp    0x348(%r15),%rsp             ;   {poll_return}
  0x00007ceff427eb7e:   ja     0x00007ceff427ebcc
  0x00007ceff427eb84:   ret    
```

Note: the size of `cmp` instruction can be either 4 bytes (`cmp 0x28(%r15),%rsp`) or 7 bytes (`cmp 0x348(%r15),%rsp`).

#### JDK 8

```
  0x0000ffff88247058: ldp	x29, x30, [sp, #32]
  0x0000ffff8824705c: add	sp, sp, #0x30
  0x0000ffff88247060: adrp	x8, 0x0000ffff97b66000       ;   {poll_return}
  0x0000ffff88247064: ldr	wzr, [x8]       ;   {poll_return}
  0x0000ffff88247068: ret
```

### JDK 11

```
  0x0000ffff9fedb468: ldp	x29, x30, [sp, #32]
  0x0000ffff9fedb46c: add	sp, sp, #0x30
  0x0000ffff9fedb470: ldr	x8, [x28, #264]
  0x0000ffff9fedb474: ldr	wzr, [x8]       ;   {poll_return}
  0x0000ffff9fedb478: ret
```

### JDK 17-25

```
  0x0000ffffa940efa0:   ldp	x29, x30, [sp, #32]
  0x0000ffffa940efa4:   add	sp, sp, #0x30
  0x0000ffffa940efa8:   ldr	x8, [x28, #832]             ;   {poll_return}
  0x0000ffffa940efac:   cmp	sp, x8
  0x0000ffffa940efb0:   b.hi	0x0000ffffa940efd4
  0x0000ffffa940efb4:   ret
```

### Use Case

- Reduce the number of `unknown` frames.
- Fix sporadic failures of `recovery` tests with `cstack=vm` option.

### Proposed Solution

Use pattern matching to detect the epilogue.

### Acknowledgements

apangin:
### Description

Add special cases in VMStructs-based stack walker to unwind partially constructed top frame in the method prologue or epilogue.

This PR focues on x86-64 and AArch64 only.

### Related issues

Resolves #1444, #1441

### Motivation and context

This is an important milestone towards making `cstack=vm` the default unwinding mode.

In HotSpot JVM, the majority of JIT compiled methods have common prologue and epilogue. Prologue includes stack overflow check, frame setup and nmethod entry barrier (JDK 21+). Epilogue pops frame off the stack, performs safepoint poll or stack watermark check and returns from the method. Prologue and epilogue are usually most problematic for stack unwinding, since they deal with transitional stack states where the topmost frame is in the middle of construction or destruction. This PR aims to detect when a sample falls in the method prologue or epilogue and, depending on the currently executed instruction, unwind the top frame accordingly.

### How has this been tested?

- Added `--cstack vm` option to all "recovery" tests.
- Manually ran spring-petclinic and watched the number of "unknown" and "break_compiled" frames.

---


src/stackFrame_aarch64.cpp
```diff
@@ -224,6 +224,107 @@ bool StackFrame::unwindCompiled(NMethod* nm, uintptr_t& pc, uintptr_t& sp, uintp
...
+    // Native wrappers:
+    //   {stack_bang}
+    //   stp  x29, x30, [sp, #-16]!
+    //   mov  x29, sp
+    //   sub  sp, sp, #0x50
+    //
+    instruction_t* ip = (instruction_t*)pc;
+    instruction_t* entry = (instruction_t*)nm->entry();
+    if (ip <= entry) {
+        pc = link();
```
Baraa-Hasheesh, line 254: I've been wondering about this as I've seen `link()` recovery in many locations,
Shouldn't we add a restriction to only allow this recovery strategy on the top frame (depth == 0)?

It's not a blocker for this as this is already done in many other locations, I'm just wondering if we should add this additional restriction in future changes

src/stackFrame_aarch64.cpp
```diff
@@ -224,6 +224,107 @@ bool StackFrame::unwindCompiled(NMethod* nm, uintptr_t& pc, uintptr_t& sp, uintp
...
+        // ldr x8, preceded by add sp
+        return true;
+    } else if ((ip[0] & 0x9f00001f) == 0x90000008 && (ip[-1] & 0xff8003ff) == 0x910003ff) {
+        // adrp x8, preceded by add sp
+        return true;
+    } else if (ip[0] == 0xeb2863ff && ip[2] == 0xd65f03c0) {
+        // cmp sp, x8, followed by ret
+        return true;
+    } else if ((ip[0] & 0xff000010) == 0x54000000 && ip[1] == 0xd65f03c0) {
+        // b.cond, followed by ret
```
Baraa-Hasheesh, line 306: Nit: in the comment on JDK 17+ is written as `b.hi offset` but here  it's `b.cond`

src/stackFrame_aarch64.cpp
```diff
@@ -224,6 +224,107 @@ bool StackFrame::unwindCompiled(NMethod* nm, uintptr_t& pc, uintptr_t& sp, uintp
...
+    //   ldr  wzr, [x8]
+    //   ret
+    //
+    if ((ip[0] & 0xffc003ff) == 0xf9400388 && (ip[-1] & 0xff8003ff) == 0x910003ff) {
+        // ldr x8, preceded by add sp
+        return true;
+    } else if ((ip[0] & 0x9f00001f) == 0x90000008 && (ip[-1] & 0xff8003ff) == 0x910003ff) {
+        // adrp x8, preceded by add sp
+        return true;
+    } else if (ip[0] == 0xeb2863ff && ip[2] == 0xd65f03c0) {
```
Baraa-Hasheesh, line 302: should we also check for the `b.cond` at ip[1]?

test/test/recovery/RecoveryTests.java
```diff
@@ -44,6 +52,11 @@ public void numbers(TestProcess p) throws Exception {
 
         out = p.profile("-d 3 -e cpu -o collapsed --safe-mode 31");
         Assert.isGreater(out.ratio("unknown_Java"), 0.05);
+
+        out = p.profile("-d 2 -e cpu -i 1ms --cstack vm -o collapsed");
+        Assert.isGreater(out.ratio("Numbers.main;test/recovery/Numbers.loop"), 0.8);
+        Assert.isGreater(out.ratio("Numbers.main;test/recovery/Numbers.loop;test/recovery/Numbers.avg"), 0.5);
+        Assert.isLess(out.ratio("unknown|break_compiled"), 0.002);
```
Baraa-Hasheesh, line 59: what do you think about extracting this `unknown|break_compiled` to a constant & add other errors to it such as 
* break_entry_frame
* break_interpreted
* break_deopt

src/stackFrame_aarch64.cpp
apangin: That's a fair point - link register is only used to unwind the topmost frame. But the corresponding instruction sequence never occurs in the intermediary frames.

src/stackFrame_aarch64.cpp
apangin: True. But there is no contradiction: the expression here checks for any conditional branch, not only `b.hi`.

src/stackFrame_aarch64.cpp
apangin: Checking `ret` is enough: these two instructions unambiguously identify the epilogue sequence. Checking `ip[1]` would be also fine, but I prefer to keep the number of comparisons minimal.

test/test/recovery/RecoveryTests.java
apangin: This PR only improves unwinding of JIT compiled frames, it does not touch interpreted or entry frames. Some interpreter edge cases are not covered yet - they will be addressed separately.

As to the constant, I think it won't make code more maintainable here.

PR #1454, opened 2025-08-27
Profiler dumps in native threads with loaded JVM

Baraa-Hasheesh:
### Description
For native apps, it's possible for the user to call the `dump` command using a native thread that is not attached to the JVM
This in turn will cause the stacks/methods to resolved incorrectly 

This change allows the user to correctly dump the profiler data with proper JDK information, even if the JVM isn't attached the the thread in question.

### Related issues
N/A

### Motivation and context
Correctly dump data for native apps when a JVM is present

### How has this been tested?
`make test` 
A new test was added that fails without the changes but passes with them

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1278,6 +1278,7 @@ Error Profiler::stop(bool restart) {
 
     _engine->stop();
 
+    SafeJvmContext safe_jvm_context(_dynamic_jvm);
     switchLibraryTrap(false);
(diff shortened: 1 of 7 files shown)


apangin:
Thank you for your desire to improve native application support in async-profiler. I have some doubts about this PR though. Let me leave it for our later discussion.

Baraa-Hasheesh:
As discussed offline this will be closed as won't fix
& user should handle the attaching of such native threads to the JVM instance before dumping the output

PR #1457, opened 2025-08-29
Add `manual-test` make target

fandreuz:
### Description
Sometimes I want to gdb into the `mainClass` of a test. I thought adding a dedicated `Makefile` target to set up the classpath would be useful.

### Related issues
n/a

### Motivation and context
Before:
```
java -Djava.library.path=build/test/lib -cp build/test.jar:build/jar/*:test/deps/*:test/gen/* 
-agentpath:./build/lib/libasyncProfiler.so=start,threads,event=test.instrument.CpuBurner.burn,latency=100ms,collapsed 
test.instrument.CpuBurner
```

After:
```
java $(make -s manual-test) -agentpath:./build/lib/libasyncProfiler.so=start,threads,event=test.instrument.CpuBurner.burn,latency=100ms,collapsed 
test.instrument.CpuBurner
```

### How has this been tested?
Manual testing

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,9 @@ coverage: clean-coverage
 
 test: test-cpp test-java
 
+manual-test: build-test-java
+	echo "-Djava.library.path=$(TEST_LIB_DIR)" -ea -cp "build/$(TEST_JAR):build/jar/*:$(TEST_DEPS_DIR)/*:$(TEST_GEN_DIR)/*"
+
 $(TEST_DEPS_DIR):
 	mkdir -p $@
 
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,8 @@ test-cpp: build-test-cpp
 
 test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -257,7 +257,8 @@ test-cpp: build-test-cpp
 
 test-java: build-test-java
 	echo "Running tests against $(LIB_PROFILER)"
-	$(JAVA) "-Djava.library.path=$(TEST_LIB_DIR)" $(TEST_FLAGS) -ea -cp "build/$(TEST_JAR):build/jar/*:build/lib/*:$(TEST_DEPS_DIR)/*:$(TEST_GEN_DIR)/*" one.profiler.test.Runner $(subst $(COMMA), ,$(TESTS))
+	PATH_ARGS=$$($(MAKE) -s manual-test) && \
```
apangin, line 260: Calling `make` recursively just to build a part of the command line seems... not perfect.
Creating a local variable with the desired value would be better. But I have doubts about the use case in the first place. I'd solve the problem by running `make` with an option that causes Runner print all executed commands, then just copy-paste the command of interest.

fandreuz: That could solve the problem too. Should the test _really_ run in that case, or should it count as a dry run?

apangin: I thought of it as of a logging option that just makes runner verbose. In fact, there is already such an option: `LOG_LEVEL=FINE`. But I admit it's not quite obvious - maybe we can replace it with `VERBOSE=true` (or create an alias for it).

fandreuz: `LOG_LEVEL=FINE` solves my problem, let me know if you think we should create another PR to have `VERBOSE`. For now, I'll close this PR.

Issue #1408, opened 2025-07-24
VM/VMX sometimes fails to unwind from __exp1 on ARM


Baraa-Hasheesh:
### Describe the bug

While testing #1405 to validate the full scope of the issue it was noticed when running the `StackwalkerTests.largeFrame` test that the stack walker fails to unwind from the `__exp1` method

This can be observed by simply running the test & checking the collected samples 

```
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask 12
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;pow 7
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;pow;__ieee754_pow 52
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;pow;__ieee754_pow;__exp1 176
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;unknown;__ieee754_pow 2
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;__kernel_standard 25
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;__sqrt 2
Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;pow@plt 1
Java_test_stackwalker_StackGenerator_largeFrame;unknown;__kernel_standard 2
Java_test_stackwalker_StackGenerator_largeFrame;__kernel_standard 1
__exp1 3
```

### Reproduction Steps

`make test TESTS=StackwalkerTests.largeFrame`

### Async-profiler version

latest

### Environment details

aarch64
AL2

fandreuz:
Possibly dup of #1402 ?

Baraa-Hasheesh:
@fandreuz could be the same problem manifesting on a different method, as I'm analyzing a different issue, I didn't look into this.

fandreuz:
@Baraa-Hasheesh did you manage to check if this is the same issue?

Baraa-Hasheesh:
@fandreuz I didn't look into this yet

Baraa-Hasheesh:
@apangin while checking stacks across multiple systems I did observe that async events such as CPU & WALL have a higher error rate on the top frame

For example 
* Java_test_stackwalker_StackGenerator_largeFrame;doCpuTask;unknown;__ieee754_pow 
Fails to unwind correctly from **__ieee754_pow** to **doCpuTask**
[...]

PR #1459, opened 2025-09-01
Enhance arm unwinding

Baraa-Hasheesh:
### Description
If a sample is captured in a native function prologue before the execution of the `stp x29, x30...` command, it would result in a failure of the unwinder.

That happens as we try to read the new PC by de-referencing the SP (with offset), but the since the PC was yet to be stored into the frame, the read we do would result in garbage values which would either result in:
* immediate failure (PC is in deadzone)
* [Unkown] frame on the top of the reported stack

This change fixes the issue by reading the instructions from the PC to the function base address & checking if `stp x29, x30...` was executed, if yes we assume dwarf worked as expected, if not we assume dwarf failed & use `lr` register to get the new PC

```

000000000002f540 <__exp1>:
   2f540:	d10083ff                                     	sub	sp, sp, #0x20
   2f544:	90000481                                     	adrp	x1, bf000 <__FRAME_END__+0x14e44>
   2f548:	9e660004                                     	fmov	x4, d0
   2f54c:	52b86e00                                     	mov	w0, #0xc3700000            	// #-1016070144
   2f550:	52840026                                     	mov	w6, #0x2001                	// #8193
   2f554:	d2800005                                     	mov	x5, #0x0                   	// #0
   2f558:	a9017bfd                                     	stp	x29, x30, [sp, #16]
   2f55c:	910043fd                                     	add	x29, sp, #0x10
   2f560:	72a07ec6                                     	movk	w6, #0x3f6, lsl #16
```

We might still see the occasional weird frames of
```
test/stackwalker/StackGenerator.main_[0];test/stackwalker/StackGenerator.leafFrame_[0];Java_test_stackwalker_StackGenerator_leafFrame;__kernel_standard 1
```
This seems to be happening because in some cases the assembly code shows a branch rather than a function call `b vs bl`

```
    92d4:	a8c27bfd                                     	ldp	x29, x30, [sp], #32
    92d8:	140001ea                                     	b	9a80 <__kernel_standard>
```

GDB reports the same stacks as async-profiler

```
#5  0x0000fffff6c21ad8 in __kernel_standard () from /lib64/libm.so.6
#6  0x0000fffff583fa38 in Java_test_stackwalker_StackGenerator_leafFrame () from /workplace/bhashesh/async-profiler/build/test/lib/libjninativestacks.so
#7  0x0000ffffd87d1cfc in ?? ()
#8  0x00000000fcb7e1c8 in ?? 
```

Check was made using ARM documentation on bit values [here](https://developer.arm.com/documentation/ddi0602/2025-06/Base-Instructions/STP--Store-pair-of-registers-?lang=en#Wt1OrWZR)

### Related issues
#1408

### Motivation and context
make ARM64 stackwalking more accurate 

### How has this been tested?
manual testing

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -291,9 +291,9 @@ CodeCache* Profiler::findLibraryByAddress(const void* address) {
     return NULL;
 }
 
-const char* Profiler::findNativeMethod(const void* address) {
+const char* Profiler::findNativeMethod(const void* address, const void** start_address) {
     CodeCache* lib = findLibraryByAddress(address);
-    return lib == NULL ? NULL : lib->binarySearch(address);
+    return lib == NULL ? NULL : lib->binarySearch(address, start_address);
 }
 
 CodeBlob* Profiler::findRuntimeStub(const void* address) {
@@ -340,9 +340,10 @@ int Profiler::getNativeTrace(void* ucontext, ASGCT_CallFrame* frames, EventType 
 int Profiler::convertNativeTrace(int native_frames, const void** callchain, ASGCT_CallFrame* frames, EventType event_type) {
     int depth = 0;
     jmethodID prev_method = NULL;
+    const void* start_addr;
 
     for (int i = 0; i < native_frames; i++) {
-        const char* current_method_name = findNativeMethod(callchain[i]);
+        const char* current_method_name = findNativeMethod(callchain[i], &start_addr);
         char mark;
         if (current_method_name != NULL && (mark = NativeFunc::mark(current_method_name)) != 0) {
             if (mark == MARK_VM_RUNTIME && event_type >= ALLOC_SAMPLE) {
--- a/src/stackFrame_aarch64.cpp
+++ b/src/stackFrame_aarch64.cpp
@@ -402,4 +402,23 @@ bool StackFrame::isSyscall(instruction_t* pc) {
     return (*pc & 0xffffefff) == 0xd4000001;
 }
 
+bool StackFrame::completeNativeFrame(uintptr_t start_addr, uintptr_t pc) {
+    // {stack_bang}
+    // stp	x29, x30, [sp, #offset]
+
+    if (pc >= start_addr + 100) { // if PC has progressed enough distance from base, assume frame is complete
+        return true;
+    }
+
+    instruction_t* ip = (instruction_t*)pc - 1;
+    while (ip >= (instruction_t*)start_addr) {
+        if ((*ip & 0xff407fff) == 0xa9007bfd) { // stp	x29, x30, [sp, #offset]
+            return true;
+        }
+        ip--;
+    }
+
+    return false;
+}
+
 #endif // __aarch64__
--- a/src/codeCache.cpp
+++ b/src/codeCache.cpp
@@ -130,7 +130,7 @@ CodeBlob* CodeCache::findBlobByAddress(const void* address) {
     return NULL;
 }
 
-const char* CodeCache::binarySearch(const void* address) {
+const char* CodeCache::binarySearch(const void* address, const void** start_address) {
     int low = 0;
     int high = _count - 1;
 
@@ -141,6 +141,7 @@ const char* CodeCache::binarySearch(const void* address) {
         } else if (_blobs[mid]._start > address) {
             high = mid - 1;
         } else {
+            *start_address = _blobs[mid]._start;
             return _blobs[mid]._name;
         }
     }
@@ -148,8 +149,10 @@ const char* CodeCache::binarySearch(const void* address) {
     // Symbols with zero size can be valid functions: e.g. ASM entry points or kernel code.
     // Also, in some cases (endless loop) the return address may point beyond the function.
     if (low > 0 && (_blobs[low - 1]._start == _blobs[low - 1]._end || _blobs[low - 1]._end == address)) {
+        *start_address = _blobs[low - 1]._start;
         return _blobs[low - 1]._name;
     }
+    *start_address = NULL;
     return _name;
 }
 
--- a/src/codeCache.h
+++ b/src/codeCache.h
@@ -198,7 +198,7 @@ class CodeCache {
 
     CodeBlob* findBlob(const char* name);
     CodeBlob* findBlobByAddress(const void* address);
-    const char* binarySearch(const void* address);
+    const char* binarySearch(const void* address, const void** start_address);
     const void* findSymbol(const char* name);
     const void* findSymbolByPrefix(const char* prefix);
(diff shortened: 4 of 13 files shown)


src/profiler.cpp
```diff
@@ -340,9 +340,10 @@ int Profiler::getNativeTrace(void* ucontext, ASGCT_CallFrame* frames, EventType
 int Profiler::convertNativeTrace(int native_frames, const void** callchain, ASGCT_CallFrame* frames, EventType event_type) {
     int depth = 0;
     jmethodID prev_method = NULL;
+    const void* start_addr;
 
     for (int i = 0; i < native_frames; i++) {
-        const char* current_method_name = findNativeMethod(callchain[i]);
+        const char* current_method_name = findNativeMethod(callchain[i], &start_addr);
```
fandreuz, line 346: `start_addr` seems to be ignored here. If that's the case, we should have a default value set to `nullptr` in `binarySearch`, and `binarySearch` should ignore the parameter if it's `nullptr`.

src/stackFrame_aarch64.cpp
```diff
@@ -402,4 +402,23 @@ bool StackFrame::isSyscall(instruction_t* pc) {
...
+bool StackFrame::completeNativeFrame(uintptr_t start_addr, uintptr_t pc) {
+    // {stack_bang}
+    // stp	x29, x30, [sp, #offset]
+
+    if (pc >= start_addr + 100) { // if PC has progressed enough distance from base, assume frame is complete
+        return true;
+    }
+
+    instruction_t* ip = (instruction_t*)pc - 1;
+    while (ip >= (instruction_t*)start_addr) {
```
fandreuz, line 414: This could be a `for` loop

apangin:
I'm afraid this is a wrong solution to a wrong problem.

Async-profiler knows what prologue and epilogue HotSpot compiler generates, so it can use this knowledge for unwinding. But when it comes to native code, async-profiler cannot trust it at all: this can be an arbitrary machine code that does not follow any common pattern. There is an infinite number of variations of prologue/eplilogue depending on the compiler version, compiler flags, and particular function being compiled, not to mention that a native function can be coded in a hand-written assembly.

`.eh_frame` / `.debug_frame` is supposed to provide unwinding instructions for any part of a function.

Baraa-Hasheesh:
@apangin 

The elf file has the following unwind details

```
000045f8 0000000000000024 000045fc FDE cie=00000000 pc=000000000002f540..000000000002f870
   LOC           CFA      x29   ra    
000000000002f540 sp+0     u     u     
000000000002f544 sp+32    u     u     
000000000002f55c sp+32    c-16  c-8   
000000000002f560 x29+16   c-16  c-8   
000000000002f5ec sp+0     u     u     
000000000002f5f0 x29+16   c-16  c-8   

 [  45f8] FDE length=36 cie=[     0]
   CIE_pointer:              17916
   initial_location:         +0x000000000002f540 <__exp1> (offset: 0x2f540)
   address_range:            0x330 (end offset: 0x2f870)

   Program:
     advance_loc 1 to 0x2f544
     def_cfa_offset 32
     advance_loc 6 to 0x2f55c
     offset r29 (x29) at cfa-16
     offset r30 (x30) at cfa-8
     advance_loc 1 to 0x2f560
     def_cfa r29 (x29) at offset 16
     advance_loc 35 to 0x2f5ec
     remember_state
     restore r29 (x29)
     restore r30 (x30)
     def_cfa r31 (sp) at offset 0
     advance_loc 1 to 0x2f5f0
     restore_state
     nop
     nop
```

The following is async-profiler representation of the unwinding info for this function

```
loc = 0x2f540, cfa_reg = 0x1f, cfa_off = 0, fp_off = 0x80000000, pc_off = 0
loc = 0x2f544, cfa_reg = 0x1f, cfa_off = 32, fp_off = 0x80000000, pc_off = 0
loc = 0x2f55c, cfa_reg = 0x1f, cfa_off = 32, fp_off = 0xfffffff0, pc_off = -8
loc = 0x2f560, cfa_reg = 0x1d, cfa_off = 16, fp_off = 0xfffffff0, pc_off = -8
loc = 0x2f5ec, cfa_reg = 0x1f, cfa_off = 0, fp_off = 0x80000000, pc_off = -8
loc = 0x2f5f0, cfa_reg = 0x1d, cfa_off = 16, fp_off = 0xfffffff0, pc_off = -8
```

Let's assume loc = 0x2f54c, this would result in capturing  FrameDesc
`loc = 0x2f544, cfa_reg = 0x1f, cfa_off = 32, fp_off = 0x80000000, pc_off = 0`

The CFA is indeed correct as it' effected by the "sub	sp, sp, #0x20" command at loc `0x2f540`
& `pc_off = 0` make sense as the PC is still not written on the frame which happens at loc `0x2f558` which is also represented in the dwarf info ` offset r30 (x30) at cfa-8`

So the profiler will try to do this when unwinding the function, 
`pc = stripPointer(*(void**)(sp + f->pc_off));` ==> `pc = *(void**)sp;` which fails to get correct data

Sadly using `pc = link` always at depth == 1 without extra checks is also wrong & would result in many unwinding issues,
So the thing that made most sense for me is to use link register conditionally only if it was never stored on the stack.

However as you mentioned it might be problamatic to search for `stp` as the user might compile with `-fomit-frame-pointer` which would indeed remove the stp command from the generated binaries 

=> debug_frame support is still work on progress

=> async-profiler looks to be using dwarf correctly as far as I'm observing from the info we have, it's that the unwinding fails when reading the PC via using the SP

An alternative is to do the following as it should address outright unwinding failures 
```
if (inDeadZone(pc) && depth == 1)) {
    pc = (const void*)frame.link();
}
```
Note in this option Some [Unkown] frames would remain on top in some cases

Baraa-Hasheesh:
ok I think I see the issue in async-profiler FrameDesc representation 

```
000045f8 0000000000000024 000045fc FDE cie=00000000 pc=000000000002f540..000000000002f870
   LOC           CFA      x29   ra    
000000000002f540 sp+0     u     u     
000000000002f544 sp+32    u     u     
000000000002f55c sp+32    c-16  c-8   
000000000002f560 x29+16   c-16  c-8   
000000000002f5ec sp+0     u     u     
000000000002f5f0 x29+16   c-16  c-8 
```

For location **000000000002f544** the return Address is unknown, as such profiler should use link register,
However, the RA is only considered unknown if the following conditions are true 
* cfa_off == 0
* fp_off == DW_SAME_FP

I the above assembly we would end up in the `pc = stripPointer(*(void**)(sp + f->pc_off));` branch which is wrong
We should have went for the `else if (depth <= 1) { pc = (const void*)frame.link();` as RA is unknown currently.

PR #1463, opened 2025-09-03
Encode a default invalid RA for ARM FrameDesc 

Baraa-Hasheesh:
### Description
For the following FDE in DWARF, if a sample was captured between instructions at locations **0x2f544** & **0x2f55c**,
The profiler will use the following FrameDesc `loc = 0x2f544, cfa_reg = 0x1f, cfa_off = 32, fp_off = 0x80000000, pc_off = 0`

In this case the DWARF denotes the RA as invalid for these instructions so we should use the `lr` register ,
However the encoded FrameDesc would result in reading the return address from the `sp` rather than using `lr`
That results in reading random values, which results in `unknown` frame or just fail to unwind

```
000045f8 0000000000000024 000045fc FDE cie=00000000 pc=000000000002f540..000000000002f870
   LOC           CFA      x29   ra    
000000000002f540 sp+0     u     u     
000000000002f544 sp+32    u     u     
000000000002f55c sp+32    c-16  c-8   
000000000002f560 x29+16   c-16  c-8   
000000000002f5ec sp+0     u     u     
000000000002f5f0 x29+16   c-16  c-8 
```

This change encodes those cases with  `0x80000000` for the `pc_off` as to mark these cases to use the `lr` register

### Related issues
#1408

### Motivation and context
Fix unwinding failures for ARM that happen on top frame

### How has this been tested?
new test added
manual testing on X86 & ARM using dacapo benchmarks 

---

### DIFF
--- a/src/dwarf.h
+++ b/src/dwarf.h
@@ -28,7 +28,7 @@ const int DW_REG_SP = 7;
 const int DW_REG_PC = 16;
 const int EMPTY_FRAME_SIZE = DW_STACK_SLOT;
 const int LINKED_FRAME_SIZE = 2 * DW_STACK_SLOT;
-const int DEFAULT_PC_OFF = -EMPTY_FRAME_SIZE;
+const int INITIAL_PC_OFFSET = -EMPTY_FRAME_SIZE;
 
 #elif defined(__i386__)
 
@@ -39,7 +39,7 @@ const int DW_REG_SP = 4;
 const int DW_REG_PC = 8;
 const int EMPTY_FRAME_SIZE = DW_STACK_SLOT;
 const int LINKED_FRAME_SIZE = 2 * DW_STACK_SLOT;
-const int DEFAULT_PC_OFF = -EMPTY_FRAME_SIZE;
+const int INITIAL_PC_OFFSET = -EMPTY_FRAME_SIZE;
 
 #elif defined(__aarch64__)
 
@@ -50,7 +50,7 @@ const int DW_REG_SP = 31;
 const int DW_REG_PC = 30;
 const int EMPTY_FRAME_SIZE = 0;
 const int LINKED_FRAME_SIZE = 0;
-const int DEFAULT_PC_OFF = DW_LINK_REGISTER;
+const int INITIAL_PC_OFFSET = DW_LINK_REGISTER;
 
 #else
 
@@ -61,7 +61,7 @@ const int DW_REG_SP = 1;
 const int DW_REG_PC = 2;
 const int EMPTY_FRAME_SIZE = 0;
 const int LINKED_FRAME_SIZE = 0;
-const int DEFAULT_PC_OFF = DW_LINK_REGISTER;
+const int INITIAL_PC_OFFSET = DW_LINK_REGISTER;
 
 #endif
 
--- a/src/dwarf.h
+++ b/src/dwarf.h
@@ -15,7 +15,7 @@ const int DW_REG_INVALID = 255;  // denotes unsupported configuration
 
 const int DW_PC_OFFSET = 1;
 const int DW_SAME_FP = 0x80000000;
-const int DW_PC_RA = 0x80000000;
+const int DW_LINK_REGISTER = 0x80000000;
 const int DW_STACK_SLOT = sizeof(void*);
 
 
@@ -50,7 +50,7 @@ const int DW_REG_SP = 31;
 const int DW_REG_PC = 30;
 const int EMPTY_FRAME_SIZE = 0;
 const int LINKED_FRAME_SIZE = 0;
-const int DEFAULT_PC_OFF = DW_PC_RA;
+const int DEFAULT_PC_OFF = DW_LINK_REGISTER;
 
 #else
 
@@ -61,7 +61,7 @@ const int DW_REG_SP = 1;
 const int DW_REG_PC = 2;
 const int EMPTY_FRAME_SIZE = 0;
 const int LINKED_FRAME_SIZE = 0;
-const int DEFAULT_PC_OFF = DW_PC_RA;
+const int DEFAULT_PC_OFF = DW_LINK_REGISTER;
 
 #endif
 
--- a/src/dwarf.h
+++ b/src/dwarf.h
@@ -15,6 +15,7 @@ const int DW_REG_INVALID = 255;  // denotes unsupported configuration
 
 const int DW_PC_OFFSET = 1;
 const int DW_SAME_FP = 0x80000000;
+const int DW_PC_RA = 0x80000000;
 const int DW_STACK_SLOT = sizeof(void*);
 
 
@@ -27,6 +28,7 @@ const int DW_REG_SP = 7;
 const int DW_REG_PC = 16;
 const int EMPTY_FRAME_SIZE = DW_STACK_SLOT;
 const int LINKED_FRAME_SIZE = 2 * DW_STACK_SLOT;
+const int DEFAULT_PC_OFF = -EMPTY_FRAME_SIZE;
 
 #elif defined(__i386__)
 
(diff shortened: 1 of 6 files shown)


Baraa-Hasheesh:
Please note that sometimes you would still get broken stacks such as 
`Java_test_stackwalker_StackGenerator_largeInnerFrame;largeInnerFrameFinal 1`

But sadly the async-profiler can't really do much for those, as when checked via GDB all SP & PC adjustments are happening correctly per dwarf info, but we end up hitting the `SP` sanity checks

GDB reports the same broken stacks with a possible corrupted stack warning (It doesn't reach the first java PC while it unwind)

```
// Check if the next frame is below on the current stack
if (sp < prev_sp || sp >= prev_sp + MAX_FRAME_SIZE || sp >= bottom) {
      break;
}
```

The Java part unwinding can be handled once the `anchor` unwinding get implemented

src/stackWalker.cpp
```diff
@@ -172,24 +172,26 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
         if (f->fp_off & DW_PC_OFFSET) {
             pc = (const char*)pc + (f->fp_off >> 1);
         } else {
+            // Adjust FP
```
fandreuz: Is this comment needed? It does not really give more details than what I can get by looking at the branches of the `if` (same for the `if`s below)

Baraa-Hasheesh: removed [e13bd6b](https://github.com/async-profiler/async-profiler/pull/1463/commits/e13bd6bf86c76874df4848544d38944b8b50395a)

src/stackWalker.cpp
```diff
@@ -422,21 +421,20 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
             if (f->fp_off != DW_SAME_FP && f->fp_off < MAX_FRAME_SIZE && f->fp_off > -MAX_FRAME_SIZE) {
                 fp = *(uintptr_t*)(sp + f->fp_off);
             }
-            if (EMPTY_FRAME_SIZE > 0 || cfa_off != 0) {
-                // x86 or AArch64 non-default frame
-                pc = stripPointer(*(void**)(sp + f->pc_off));
-            } else if (f->fp_off != DW_SAME_FP) {
-                // AArch64 default_frame
+
+            if (EMPTY_FRAME_SIZE == 0 && depth == 1 && f->pc_off == DW_PC_RA) {
```
apangin: `EMPTY_FRAME_SIZE == 0` is not necessary here anymore. `DW_PC_RA` should only be set on platforms with link register.

src/stackWalker.cpp
```diff
@@ -175,21 +175,20 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
...
+            if (EMPTY_FRAME_SIZE == 0 && cfa_off == 0 && f->fp_off != DW_SAME_FP) {
                 sp = defaultSenderSP(sp, fp);
                 if (sp < prev_sp || sp >= bottom || !aligned(sp)) {
                     break;
                 }
-            } else if (depth <= 1) {
-                pc = (const void*)frame.link();
-            } else {
-                // Stack bottom
-                break;
```
apangin, line 192: Why was this removed?

src/stackWalker.cpp
```diff
@@ -175,21 +175,20 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
...
+
+            if (EMPTY_FRAME_SIZE == 0 && depth == 1 && f->pc_off == DW_PC_RA) {
+                pc = (const void*)frame.link();
+            } else if (f->pc_off != DW_PC_RA) {
                 pc = stripPointer(SafeAccess::load((void**)(sp + f->pc_off)));
+            } else {
+                break;
+            }
+
+            if (EMPTY_FRAME_SIZE == 0 && cfa_off == 0 && f->fp_off != DW_SAME_FP) {
```
apangin, line 187: After deletion of the comment about AArch64 default_frame, it's hard to understand what this condition means.

src/dwarf.h
```diff
@@ -15,6 +15,7 @@ const int DW_REG_INVALID = 255;  // denotes unsupported configuration
 
 const int DW_PC_OFFSET = 1;
 const int DW_SAME_FP = 0x80000000;
+const int DW_PC_RA = 0x80000000;
```
apangin: Looks like a bunch of random letters )
`DW_LINK_REGISTER` would be better.

src/stackWalker.cpp
Baraa-Hasheesh: True, but the reason I added the `EMPTY_FRAME_SIZE == 0` is to allow the compilers to optimize the branches out for X86

`EMPTY_FRAME_SIZE == 0` is either always true or always false depending if it's ARM or X86 

My idea was to add it such that the compiler would optimize it out & simplify the condition to the following on X86

```
if (f->pc_off != DW_PC_RA) {
    ....
} else {
    break;
}
````

File: src/stackWalker.cpp (hunk 3 above)
Comment (Baraa-Hasheesh) [reply src/stackWalker.cpp:192]: Not really removed, 
This branche would only be ever executed on ARM if you get an invalid return address (empty frame) in Frame that's not at depth == 1

This is technically covered in the new PC operation block of the code (break branch)
If you get  f->pc_off == DW_PC_RA @ depth != 1 the break will execute 
It's not exactly the same check but it should cover the same cases per my understanding of the flow

File: src/dwarf.h (hunk 5 above)
Comment (Baraa-Hasheesh) [reply src/dwarf.h]: renamed here https://github.com/async-profiler/async-profiler/pull/1463/commits/fa2758d06ff94066798fc6c6251992e127efd465

File: src/stackWalker.cpp (hunk 4 above)
Comment (Baraa-Hasheesh) [reply src/stackWalker.cpp:187]: re-added the comment here https://github.com/async-profiler/async-profiler/commit/fa2758d06ff94066798fc6c6251992e127efd465 

I thought about making it more explicit (Adjusting SP from FP), but the old comment should be good enough

File: src/stackWalker.cpp (hunk 2 above)
Comment (apangin) [reply src/stackWalker.cpp]: In the way how you wrote it, there is always at least one check on x86, with or without `EMPTY_FRAME_SIZE == 0`.
If you wanted to optimize it, you could do like this:

```
if (EMPTY_FRAME_SIZE > 0 || f->pc_off != DW_LINK_REGISTER) {
    ....
} else if (depth == 1) {
    pc = frame.link();
} else {
    break;
}
```

Comment (Baraa-Hasheesh) [reply src/stackWalker.cpp]: The suggested implementation is indeed better, I will apply it

Comment (Baraa-Hasheesh) [reply src/stackWalker.cpp]: changed here [2daf923](https://github.com/async-profiler/async-profiler/pull/1463/commits/2daf923f5015af2b2298880b29a99151168421cc)

File: src/dwarf.h [hunk 6]
```diff
@@ -47,6 +50,7 @@ const int DW_REG_SP = 31;
 const int DW_REG_PC = 30;
 const int EMPTY_FRAME_SIZE = 0;
 const int LINKED_FRAME_SIZE = 0;
+const int DEFAULT_PC_OFF = DW_LINK_REGISTER;
```
apangin: nit: it's not the _default_ (to avoid confusion with the `default_frame`). I'd call it `INITIAL_PC_OFFSET`

Baraa-Hasheesh: make sense [3ada4c0](https://github.com/async-profiler/async-profiler/pull/1463/commits/3ada4c0c41ce2b2306976c8b805fba9e56cc7bb5)

apangin:
Looks good now. I'll run some sanity checks manually, and if everything works as expected, will merge it soon.

PR #1464, opened 2025-09-04
Change SmokeTests.cpu test to be more deterministic

Baraa-Hasheesh:
### Description
In old implementation of the test both method1 & method2 could end up taking different time depending on various factors (Faster CPU cores is one example)

It's possible that under bad circumstances that method1/2 timing could coordinate with method3 leading to one of the method1/2 not being sampled 

This change guarantees a sample is collected in each method by running each method for at least 20ms at each cycle which is higher than the default sampling interval 

### Related issues
N/A

### Motivation and context
Make tests less likely to fail

### How has this been tested?
manual testing & test was re-run 100 times on GHA

---

### DIFF
--- a/test/test/smoke/Cpu.java
+++ b/test/test/smoke/Cpu.java
@@ -6,25 +6,32 @@
 package test.smoke;
 
 import java.io.File;
+import java.util.concurrent.ThreadLocalRandom;
 
 public class Cpu {
     private static volatile int value;
 
     private static void method1() {
-        for (int i = 0; i < 1000000; i++) {
-            value++;
+        long startTime = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


test/test/smoke/Cpu.java
```diff
@@ -6,25 +6,32 @@
...
+import java.util.concurrent.ThreadLocalRandom;
 
 public class Cpu {
     private static volatile int value;
 
     private static void method1() {
-        for (int i = 0; i < 1000000; i++) {
-            value++;
+        long startTime = System.currentTimeMillis();
+        while (System.currentTimeMillis() - startTime < 20) {
```
fandreuz, line 16: `20ms` is enough to get a sample on all three methods?

PR #1466, opened 2025-09-08
Add Liberica to the CI on Alpaquita musl

krk:
### Motivation and context

We can consider including more distributions in the GHA CI.

### How has this been tested?
CI

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -60,6 +60,9 @@ jobs:
         include:
           - test-platform: linux-x64-alpine
             container-image: alpine:corretto-11
+            use-builtin-jdk: true
+            java-distribution: corretto
+            java-version: 11
           - test-platform: linux-x64-AL2
             container-image: amazonlinux:2
             # GHA provides Node.js by attaching a volume to the container. The container path is
@@ -71,6 +74,11 @@ jobs:
             container-image: amazonlinux:2023
             java-version: 11
             java-distribution: corretto
+          - test-platform: linux-x64-alpaquita
+            container-image: alpaquita:x86_64-liberica-21
+            use-builtin-jdk: true
+            java-distribution: liberica
+            java-version: 21
     uses: ./.github/workflows/integ.yml
     with:
       platform: linux-x64
@@ -80,6 +88,7 @@ jobs:
       container-volumes: ${{ matrix.container-volumes || '' }}
       java-version: ${{ matrix.java-version }}
       java-distribution: ${{ matrix.java-distribution }}
+      use-builtin-jdk: ${{ matrix.use-builtin-jdk || false }}
 
   integ-linux-arm64:
     name: integ / linux-arm64
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -28,6 +28,10 @@ on:
       container-volumes:
         type: string
         required: false
+      use-builtin-jdk:
+        type: boolean
+        required: false
+        default: false
 
 jobs:
   integration-test:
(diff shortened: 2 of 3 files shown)


apangin:
What value does this particular configuration add? Do you know any errors that other configurations don't catch?

krk:
Testing a commit that could potentially fail.

krk:
Please see https://github.com/async-profiler/async-profiler/actions/runs/17581662920/job/49939353504?pr=1466 for failures about malloc tracer with the gcc version in alpaquita.

krk:
All 4 failures are about `calloc` in `nativemem` tests.

```
NativememTests.canAgentTraceMallocCalloc (args: once) (agentArgs: start,nativemem,total,collapsed,file=%f)
NativememTests.canAgentTraceMallocCalloc/VM (args: once) (agentArgs: start,nativemem,total,cstack=vm,collapsed,file=%f)
NativememTests.canAsprofTraceAllNoLeak
NativememTests.canAsprofTraceMallocCalloc
```

Looks like the same root cause as https://github.com/async-profiler/async-profiler/pull/1226#issuecomment-2797294532, cannot sample `calloc`.

Note that the [`calloc_hook_dummy`](https://github.com/async-profiler/async-profiler/blob/2c188fe4908c1d8736707da815e08e6636f8fafb/src/mallocTracer.cpp#L52) workaround not applicable in this case.

Baraa-Hasheesh:
@krk I've checked the issue here & the root cause for the failures is that on Alpaquita musl `calloc` doesn't call `malloc` internally

```
        if (OS::isMusl()) {
            // On musl, calloc() calls malloc() internally, and posix_memalign() calls aligned_alloc().
            // Use dummy hooks to prevent double-accounting. Dummy frames from AP are introduced
            // to preserve the frame link to the original caller (see #1226).
            cc->patchImport(im_calloc, (void*)calloc_hook_dummy);
            cc->patchImport(im_posix_memalign, (void*)posix_memalign_hook_dummy);
        } else {
            cc->patchImport(im_calloc, (void*)calloc_hook);
            cc->patchImport(im_posix_memalign, (void*)posix_memalign_hook);
        }
```

We assume that if MUSL that `calloc` would always call `malloc` internally which seems to be incorrect in this case

Just change the `cc->patchImport(im_calloc, (void*)calloc_hook_dummy);` to `cc->patchImport(im_calloc, (void*)calloc_hook);` & it should work (Not a proper fix, but you can try it)

apangin:
It turns out that in this Alpaquita docker image musl is statically linked with mimalloc (!) which behaves differently from musl's built-in malloc implementation. So, it's not enough to check if the current process runs with glibc or musl, because libc itself can be built with an alternative implementation :-/

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -172,6 +172,17 @@ jobs:
...
             asprof-binaries-job: linux-x64
+          - runson:
+              display: alpqauita-liberica-alpine
+              name: ubuntu-latest
+            java-version: 21
+            java-distribution: liberica
+            image: public.ecr.aws/bellsoft/alpaquita-linux-gcc:15.2-musl
+            asprof-binaries-job: linux-x64
+            before-checkout: |
+              apk add liberica21-jdk util-linux-misc
```
apangin: Instead of installing these every time, wouldn't it be better to prepare an image with all the tools in place? That's what we do for other OS distributions.
This is especially relevant in the light of HTTP 504 failures during `apk add` I saw in the [previous runs](https://github.com/async-profiler/async-profiler/actions/runs/17947783828/job/51038531987?pr=1466).

krk: Will do after https://github.com/async-profiler/async-profiler/pull/1502 is merged.

krk:
Force-pushed to resolve the conflict.

PR #1468, opened 2025-09-09
Reduce assert threshold for `GetProcessIds_returns_valid_pids` test

rk-kmr:
### Description
`GetProcessIds_returns_valid_pids` fails when number of process discovered are less than 5. In this change, I am relaxing the threshold to 1.

### Related issues
N/A

### Motivation and context
N/A

### How has this been tested?
Tested before and after running the same steps manually from docker initialization [here](https://github.com/async-profiler/async-profiler/blob/master/docker/amazonlinux2023.Dockerfile)

---

### DIFF
--- a/test/native/processInfoTest.cpp
+++ b/test/native/processInfoTest.cpp
@@ -36,7 +36,7 @@ TEST_CASE(GetProcessIds_returns_valid_pids) {
     int pids[100];
     int count = OS::getProcessIds(pids, 100);
     ASSERT_GT(count, 0);
-    ASSERT_GT(count, 5);
+    ASSERT_GT(count, 1);
     ASSERT_LTE(count, 100);
 
     for (int i = 0; i < count; i++) {


fandreuz (approved):

test/native/processInfoTest.cpp
```diff
@@ -36,7 +36,7 @@ TEST_CASE(GetProcessIds_returns_valid_pids) {
     int pids[100];
     int count = OS::getProcessIds(pids, 100);
     ASSERT_GT(count, 0);
```
apangin: This one is redundant as you check `count > 1` in the very next line.

test/native/processInfoTest.cpp
```diff
@@ -36,7 +36,7 @@ TEST_CASE(GetProcessIds_returns_valid_pids) {
...
-    ASSERT_GT(count, 5);
+    ASSERT_GT(count, 1);
```
apangin, line 38: There is a similar check in `GetProcessIds_respects_max_limit`

test/native/processInfoTest.cpp
fandreuz: I think the reason for this is to have a more informative output when `count == 0`? Maybe we could achieve the same by improving the macro

test/native/processInfoTest.cpp
rohitash-kumar: `GetProcessIds_respects_max_limit` shouldn't be an issue because assert used is `ASSERT_LTE` with 5 as upper limit. 
```
    ASSERT_LTE(count, 5);
```

test/native/processInfoTest.cpp
apangin: Failure message is already informative enough:
```
Assertion failed: (count > 5),
	actual values: count = 4 (0x4), 5 = 5 (0x5)
	at test/native/processInfoTest.cpp:39
```

rohitash-kumar: I agree with Andrei, that check doesn't make sense. I'll remove it.

test/native/processInfoTest.cpp
apangin: Overlooked that, sorry.

PR #1469, opened 2025-09-09
GC sections while building asprof too

krk:
### Motivation and context

There is no specific reason not to do section gc while building `asprof`. 

One reason against would be increasing the file size because of having more sections and not enough garbage to be collected, which, in this case, not applicable: The file size of `asprof` is reduced by a total of `8 bytes` to `47240`.

### How has this been tested?
Can run `build/bin/asprof`.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ else
   ifeq ($(MERGE),true)
     CXXFLAGS += -fwhole-program
   endif
+  CFLAGS += -fdata-sections -ffunction-sections -Wl,--gc-sections
   LIBS += -lrt
   INCLUDES += -I$(JAVA_HOME)/include/linux
   SOEXT=so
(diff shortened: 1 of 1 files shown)


apangin:
Sorry, motivation section is not clear enough to me. What is the purpose of this change?

krk:
This change removes unused sections in `asprof,` mirroring a similar optimization in `libasyncProfiler.so`. While this doesn't significantly reduce the binary size today, it ensures that these sections can be garbage-collected in the future as the codebase changes.

apangin:
This optimization is mostly useful when linking some or all of the libraries statically. That's why it is used for compiling libasyncProfiler.so (`-static-libstdc++ -static-libgcc`) and when asprof is linked statically (`STATIC_BINARY=true`).

There is no particular harm in always adding this option, but there is no benefit either. Things being equal, I prefer to keep it as is: collocating `--gc-sections` option with `-static*` options make its intention clearer.

krk:
There still could be unused, unexported functions and data that asprof references, that could be gc'd, without `-static`.

Issue #1162, opened 2025-03-06
jvmt->GetEnv can result in signal unsafe memory allocation


pnf:
### Describe the bug

Calls to `JavaVM::GetEnv` result in access of a `__thread Thread* Thread::_thr_current` in `jni.cc`  (for gcc builds of the jvm at least) resulting in a call to `__tls_get_addr`, which might need to update slots, calling `realloc` in the process.

We recently observed a deadlock due to a perf event that occurred in the middle of jemalloc extent management, giving us a stack like the below.  Note that, while we are using jemalloc as an allocator, we are not using its memory profiling here.  

```
_pthread_mutex_lock
je_malloc_mutex_lock_slow
extent_recycle
...
do_rallocx
realloc
dl_resize_dtv
dl_update_slotinfo
__tls_get_addr
jni_GetEnv
Profiler::getJavaTraceAsync
Profiler::recordSample
PerfEvents::signalHandler
__restore_rt
je_eset_remove
extent_recycle
...
```

While the deadlock is infrequent, our QA plant sampling turns up quite a few examples of the stack segment from `__tls_get_addr` to `realloc`.

 

### Reproduction Steps

This happens rarely, and I cannot give you a repro at this point. 

### Async-profiler version

3.0 with some unrelated (so I claim) modifications.

### Environment details

The problem has occurred so far only on azul openjdk 20.0.2 and rh 8.

apangin:
Does the problem happen with the original `libasyncProfiler.so`?
Async-profiler [has protection](https://github.com/async-profiler/async-profiler/blob/e27198e324c429e733b34fa7842fdc3f66e58067/src/profiler.cpp#L392-L397) against the situation you describe.

Also, you can use `cstack=vm` that walks Java stack without calling `GetEnv` or `AsyncGetCallTrace`.

pnf:
We have that protection in place, but __tls_get_addr is still called occasionally from GetEnv.  I’m not sure if it’s relevant, but we sometimes see __tls_get_addr invoked first via
     __tls_get_addr
     Thread::current()
     Mutex::lock_without_safepoint_check
     HandshakeState::has
     Unsafe_Park
     LockTracer::UnsafeParkHook
triggering dl_resize, then a realloc, and then getting a perf signal in the middle of extent_recycle, leading to a reentrant call to __tls_get_addr. 

[...]

PR #1470, opened 2025-09-09
Vtable

apangin:
### Description

Add `VMThread::isJavaThread()` function that is semantically equivalent to HotSpot's `Thread::is_Java_thread()`.

Implement `VMThread::jni()` without calling JNI GetEnv function which is [unsafe](https://github.com/async-profiler/async-profiler/issues/1162) insde a signal handler.

### Related issues

#1162, #1329

### Motivation and context

Differentiating between Java and non-Java threads is necessary to safely access JavaThread's fields through VMStructs. In particular, this is needed by VMStructs-based stack walker to discover the topmost Java frame using [`JavaThread::_anchor`](https://github.com/async-profiler/async-profiler/pull/1329/files#r2191214920).

Another use case is a cheap and signal-safe way to get `JNIEnv*` for the current thread. Previously used JNI GetEnv function could deadlock when called inside a signal hander.

### How has this been tested?

Proposed implementation uses heuristics which was verified on different JDK versions (from 8 to 25): release, fastdebug and slowdebug builds. However, this heuristics may break if new virtual methods are inserted in `Thread` class in future JDK versions. In order to detect such cases early, I introduced new developer-only option `jnienv` that validates proposed heuristics against reference implementation in HotSpot. Added new integration tests that make use of this option.

---

### DIFF
--- a/src/vmStructs.h
+++ b/src/vmStructs.h
@@ -118,6 +118,7 @@ class VMStructs {
     static jfieldID _tid;
     static jfieldID _klass;
     static int _tls_index;
+    static void* _java_thread_vtbl[6];
 
     typedef void (*LockFunc)(void*);
     static LockFunc _lock_func;
@@ -351,6 +352,21 @@ class VMThread : VMStructs {
     static int nativeThreadId(JNIEnv* jni, jthread thread);
 
     int osThreadId();
+
+    const void** vtable() {
+        return *(const void***)this;
+    }
+
+    // Carefully chosen heuristics: we consider this thread a JavaThread
+    // if at least 2 of the selected 3 vtable entries match those of a known
+    // JavaThread (which is either application thread or AttachListener).
+    // Verified on OpenJDK 8 to 25: product, fastdebug and slowdebug builds.
+    bool isJavaThread() {
+        const void** vtbl = vtable();
+        return (vtbl[1] == _java_thread_vtbl[1]) +
+               (vtbl[3] == _java_thread_vtbl[3]) +
+               (vtbl[5] == _java_thread_vtbl[5]) >= 2;
+    }
 
     int state() {
         return _thread_state_offset >= 0 ? *(int*) at(_thread_state_offset) : 0;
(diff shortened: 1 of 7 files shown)


src/vmStructs.h
```diff
@@ -352,6 +354,22 @@ class VMThread : VMStructs {
...
+    JNIEnv* jni();
+
+    const void** vtable() {
+        return *(const void***)this;
+    }
+
+    // This thread is considered a JavaThread if at least 2 of the selected 3 vtable entries
+    // match those of a known JavaThread (which is either application thread or AttachListener).
+    // Indexes were carefully chosen to work on OpenJDK 8 to 25, both product an debug builds.
+    bool isJavaThread() {
```
krk, line 366: Hypothetically, is it possible to detect `isJavaThread` if we have never seen a java thread before and `_java_thread_vtbl` were unavailable.

apangin: Profiler is always initialized on a Java thread. It can be
1. an application thread that loads profiler using its Java API;
2. AttachListener thread when libasyncProfiler is loaded via Dynamic Attach Mechanism (i.e. using asprof CLI);
3. JVM main (bootstrap) thread when profiler is loaded via `-agentpath`.

We copy vtable of this thread when profiler initializes, but only on HotSpot JVM. On OpenJ9, profiler uses JNI GetEnv as before.

fandreuz:
@apangin I could not reproduce #1162 anymore after the change, I let it run for some time already. I'll wait a bit more, but it looks good, it usually takes less than two minutes.

src/vmStructs.h
```diff
@@ -352,6 +354,22 @@ class VMThread : VMStructs {
...
+        const void** vtbl = vtable();
+        return (vtbl[1] == _java_thread_vtbl[1]) +
```
Baraa-Hasheesh, line 368: Curious about if this is only valid for hotspot JDKs or is also true for J9 & other JDK types

apangin: This holds only for HotSpot. J9 will continue using JNI GetEnv as before.

PR #1421, opened 2025-07-29
Enable recording JVM method exit via bytecode instrumentation

fandreuz:
### Description
In this PR I enable recording JVM method exit via bytecode instrumentation.

### Related issues
n/a

### Motivation and context
n/a

### How has this been tested?
`java -agentpath:./build/lib/libasyncProfiler.so=start,timeout=1,event=test.cpu.CpuBurner.burn,collapsed,file=x.txt -cp test test.cpu.CpuBurner`

I still want to test special cases like when a jump is converted from narrow to wide.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #1471, opened 2025-09-10
Optimize method tracing when the function is not profiled

fandreuz:
### Description
In the current implementation, we always make a JNI call to `recordExit` regardless of whether the current function call breaks the latency threshold or not. We could move the check from C++ to Java, so we could save a JNI call when the method is not going to be profiled.

### Related issues
#1421 

### Motivation and context
This simple class shows a significant improvement in terms of throughput:
```java
public class Test {
    private static volatile long x = 0;

    public static void doStuff() throws Exception {
        ++x;
    }

        long start = System.nanoTime();
        while (System.nanoTime() - start < 4 _000_000_000l) {
            doStuff();
        }
        System.out.println(x);
    }
}
```

| Branch              | Value      |
|---------------------|------------|
| `master`              | 83084633  |
| `latency-jni-optimize` | 119116380  |

So we get approximately 40% more method calls, when the method is never profiled and the workload is negligible.

### How has this been tested?
The current set of tests is good enough to validate this change.

---


src/helper/one/profiler/Instrument.java
```diff
@@ -14,5 +14,13 @@ private Instrument() {
     }
 
     public static native void recordEntry();
-    public static native void recordExit(long startTimeNanos);
+
+    public static void recordExit(long startTimeNs, long minLatency) {
```
fandreuz, line 18: I thought we could save `minLatency` as a `static` field in `Instrument`, maybe even `final` if we accept having some processing of the class bytecode before we register it. This would spare us the `ldc2_w`.

But then I realized this would break simultaneous instrumentation with differing latencies, which should be supported now. Is it something we want to keep?

Baraa-Hasheesh: What do you mean by "simultaneous instrumentation with differing latencies"
Do you mean multiple profilers?

Was this tested? 
=> I have a strong feeling the multiple profilers won't play nice with one another

fandreuz: > What do you mean by "simultaneous instrumentation with differing latencies"

- `latency=x,event=ClassA.method`
- `latency=y,event=ClassB.method`
- `x != y`

> Was this tested?

No

> => I have a strong feeling the multiple profilers won't play nice with one another

Can you elaborate?

Baraa-Hasheesh: Historically from my own experience when multiple profilers are involved weird behaviors happen which include but not limited to:
* infinite recursive loops between the profilers
* One profiler losing it's sampling abilities
* One profiler getting double sampling of both profilers 

If the justification is, this is for potentially multiple profilers running in parallel 
Then ideally this should be tested to get what actually will happen

I'm not asking to fix issues you may find, but what you might find might need to be documented/understood to know what the limitation are & if there are things we can fix

fandreuz: I see your point, but I'm asking a different question here. 

> But then I realized this would break simultaneous instrumentation with differing latencies, which should be supported now. Is it something we want to keep?

If the answer is no, the solution I have now is suboptimal and should be improved. If the answer is yes, then the solution is ok.

Baraa-Hasheesh: Honestly I would be more interested in testing a single profiler with the following flow
* Start with Latency of 100
* stop & dump output
* re start profiler with latency of 50

Regarding the multiple profiler subject, I think we should at least guarantee that a crash won't happen via some manual testing.

As if we should support this at all currently in this PR if it's not already supported, my answer would be no 
But I believe @apangin would be more of help here on this subject

src/instrument.cpp
```diff
@@ -341,6 +350,12 @@ class BytecodeRewriter {
             case Result::JUMP_OVERFLOW:
```
Baraa-Hasheesh, line 350: can you try running on macOS, I'm getting this when running async-profiler (master)
-> some warnings here

```
./src/instrument.cpp:334:17: warning: enumeration values 'OK' and 'CLASS_DOES_NOT_MATCH' not handled in switch [-Wswitch]
  334 |         switch (res) {
      |                 ^~~
./src/instrument.cpp:334:17: note: add missing switch cases                       
  334 |         switch (res) {
```

fandreuz: - `OK` is handled earlier, I'm not doing it in the `switch` to save allocations in he happy path
- `CLASS_DOES_NOT_MATCH` is handled, checkout the branch in this PR

src/helper/one/profiler/Instrument.java
fandreuz: As discussed with @apangin, the optimization I propose in this comment may not be worth the additional complications it introduces.

src/profiler.cpp
```diff
@@ -674,7 +674,18 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+        case INSTRUMENTED_METHOD:
+            start_depth = 1;
+            break;
+        case METHOD_TRACE:
+            start_depth = 2;
+            break;
+        default:
+            start_depth = 0;
+            break;
+        }
```
Baraa-Hasheesh: The switch style here isn't consistent with other locations in the profiler

```suggestion
        switch (event_type) {
            case INSTRUMENTED_METHOD:
                start_depth = 1;
                break;
            case METHOD_TRACE:
                start_depth = 2;
                break;
            default:
                start_depth = 0;
                break;
        }
```

fandreuz: Thanks, b7fec953dcff0f60efce20e0cfedfae7f0f519b7

src/instrument.cpp
```diff
@@ -1085,14 +1112,13 @@ void JNICALL Instrument::recordEntry(JNIEnv* jni, jobject unused) {
     }
 }
 
-void JNICALL Instrument::recordExit(JNIEnv* jni, jobject unused, jlong startTimeNanos) {
+void JNICALL Instrument::recordExit0(JNIEnv* jni, jobject unused, jlong durationNs) {
     if (!_enabled) return;
 
     u64 now_ticks = TSC::ticks();
```
apangin: What worries me most in this change is that it introduces inaccuracy in reported timings.
There can be a potentially large delay between the last call to `System.nanoTime` and this TSC::ticks()`, especially if a thread is de-scheduled from CPU in this time window. This delay can cause shift in the reported event start time, which may in turn break correlation with other events.

src/profiler.cpp
```diff
@@ -674,7 +674,18 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
...
+            case INSTRUMENTED_METHOD:
+                start_depth = 1;
+                break;
+            case METHOD_TRACE:
+                start_depth = 2;
+                break;
+            default:
+                start_depth = 0;
+                break;
+        }
```
apangin: You've replaced one line with 12 for such a small thing.
Could you just use `?:` chain?

fandreuz: 454e908fa5f6b082dee1b3ecff444e50701e76ad

src/instrument.cpp
fandreuz: I see, how about 027ba68d375caa949863bb840d2adf1b625b6c02 ?

We'll record a slightly higher `durationNs` than what we use to test the latency. But if anything, the measurement is going to be more realistic, since it'll also include the JNI penalty.

src/instrument.cpp
```diff
@@ -17,7 +17,10 @@
 #include "vmEntry.h"
 #include "instrument.h"
 
-INCLUDE_HELPER_CLASS(INSTRUMENT_NAME, INSTRUMENT_CLASS, "one/profiler/Instrument")
+#define PROFILER_PACKAGE "one/profiler/"
+static constexpr u32 PROFILER_PACKAGE_LEN = 13;
+
+INCLUDE_HELPER_CLASS(INSTRUMENT_NAME, INSTRUMENT_CLASS, PROFILER_PACKAGE "Instrument")
```
apangin: It's OK to have PROFILER_PACKAGE/PROFILER_PACKAGE_LEN constants, but please don't break class names this way: this is less readable and not searchable.

src/instrument.cpp
```diff
@@ -273,6 +275,12 @@ class BytecodeRewriter {
         put16(ref2);
     }
 
+    void putLongConstant(u32 high_bytes, u32 low_bytes) {
```
apangin: I'd expect a function with this name accept one `u64` or `jlong` constant and do word splitting inside.

src/instrument.cpp
```diff
@@ -341,6 +350,12 @@ class BytecodeRewriter {
             case Result::JUMP_OVERFLOW:
                 Log::warn("Jump overflow: %s.%s", class_name.c_str(), method_name.c_str());
                 break;
+            case Result::CLASS_DOES_NOT_MATCH:
+                Log::debug("Skipping instrumentation of %s: class does not match", class_name.c_str());
+                break;
+            case Result::PROFILER_CLASS:
+                Log::trace("Skipping instrumentation of %s: internal profiler class", class_name.c_str());
+                break;
```
apangin: Why `debug` in one case and `trace` in another?
But in my opinion, neither of these logs is useful.

apangin: Also, add
```
default:
    break;
```
to suppress compiler warning.

src/instrument.cpp
```diff
@@ -1085,12 +1112,12 @@ void JNICALL Instrument::recordEntry(JNIEnv* jni, jobject unused) {
...
 
-void JNICALL Instrument::recordExit(JNIEnv* jni, jobject unused, jlong startTimeNanos) {
+void JNICALL Instrument::recordExit0(JNIEnv* jni, jobject unused, jlong startTimeNs) {
     if (!_enabled) return;
 
     u64 now_ticks = TSC::ticks();
-    u64 duration_ns = OS::nanotime() - (u64) startTimeNanos;
-    if (duration_ns >= _latency && shouldRecordSample()) {
+    u64 duration_ns = OS::nanotime() - (u64) startTimeNs;
+    if (shouldRecordSample()) {
```
apangin, line 1122: Now that you don't check latency here, it's OK to move `now_ticks` and `duration_ns` insde the `if` block.

src/instrument.cpp
apangin: That's better. Now timings should be more accurate, for the price of an extra `nanotime` call.
I know how to improve this further by leveraging `jdk.jfr.internal.JVM.counterTime()` intrinsic in Java code, but this will require lots of changes. We can discuss it and probably leave for a separate PR.

src/instrument.cpp
fandreuz: e87c71552a44f6ec2f20b77bebc000fd91a5cf1d

src/instrument.cpp
fandreuz: e87c71552a44f6ec2f20b77bebc000fd91a5cf1d

src/instrument.cpp
fandreuz: e87c71552a44f6ec2f20b77bebc000fd91a5cf1d

src/instrument.cpp
fandreuz: > But in my opinion, neither of these logs is useful.

I agree, that's why I made them invisible by default. The idea is not to skip quietly any decision I take during the flow, and let this information be available upon closer inspection.

> to suppress compiler warning.

e87c71552a44f6ec2f20b77bebc000fd91a5cf1d

apangin: But these messages [get recorded](https://github.com/async-profiler/async-profiler/blob/145fc2dd28255f3f4d299fdd09d0cef4dcd8f78a/src/log.cpp#L98-L101) in .jfr anyway.
I suggest to remove them completely or change them both to `trace` and adjust logging not write `trace` level logs to jfr.

fandreuz: I see, I didn't know they go there. 74053fc809e4b928e4692dcb1ec658fc672efb22

src/flightRecorder.cpp
```diff
@@ -1020,8 +1020,8 @@ class Recording {
 
     void writeLogLevels(Buffer* buf) {
         buf->putVar32(T_LOG_LEVEL);
-        buf->putVar32(LOG_ERROR - LOG_TRACE + 1);
-        for (int i = LOG_TRACE; i <= LOG_ERROR; i++) {
+        buf->putVar32(LOG_ERROR - FlightRecorder::MIN_LOG_LEVEL + 1);
+        for (int i = FlightRecorder::MIN_LOG_LEVEL; i <= LOG_ERROR; i++) {
```
apangin: Leave the code as it was. There's nothing bad in listing all available levels, especially since there is a (theoretical) possibility to call `writeLog` directly bypassing checks in `Log::log`.

src/log.cpp
```diff
@@ -85,6 +86,11 @@ void Log::writeRaw(LogLevel level, const char* msg, size_t len) {
 }
 
 void Log::log(LogLevel level, const char* msg, va_list args) {
+    if (_level > level && FlightRecorder::MIN_LOG_LEVEL > level) {
```
apangin: nit: I'd reverse the check: when an input argument participates in some condition, it better reads when an argument is on the left side. `CONSTANT > arg` looks unnatural to me.

`if (level < _level && level < FlightRecorder::MIN_LOG_LEVEL)`

src/log.cpp
```diff
@@ -95,8 +101,7 @@ void Log::log(LogLevel level, const char* msg, va_list args) {
     }
     buf[prefix_len + msg_len] = '\n';
 
-    // Write all messages to JFR, if active
-    if (level < LOG_ERROR) {
+    if (level >= FlightRecorder::MIN_LOG_LEVEL) {
```
apangin: Have you removed `< LOG_ERROR` intentionally?
Error message typically indicates a situation when profiler is unable to execute a user command. Errors don't happen out of the blue at runtime and don't need to be recorded.

src/helper/one/profiler/Instrument.java
```diff
@@ -14,5 +14,12 @@ private Instrument() {
...
+        if (System.nanoTime() - startTimeNs >= minLatency) {
+            recordExit0(startTimeNs);
```
apangin, line 20: If `latency == 0`, you can call `recordExit0` directly from the generated code. Even the name of the method suits well :)

src/flightRecorder.cpp
fandreuz: I see, reverted in 148e6b3e46bb8699e32baa831a11101f6740809f

src/log.cpp
fandreuz: Got it, 148e6b3e46bb8699e32baa831a11101f6740809f

src/log.cpp
fandreuz: 148e6b3e46bb8699e32baa831a11101f6740809f

src/helper/one/profiler/Instrument.java
apangin: Ah, this will complicate decision how many frames to skip in the stacktrace.
But I think the optimization to skip redundant `nanoTime` call is useful. Maybe we can add another variant of `recordExit` call? wdyt?

fandreuz: > Maybe we can add another variant of recordExit call?

Yeah I think this is the easiest way: a9a8604e8b0f87d66a32d98b1b04c200d9a99beb 

Or we add something like `EventType::METHOD_TRACE_ZERO`.

Issue #1332, opened 2025-06-09
`WallTests.cpuWall` fails sometimes


fandreuz:
### Describe the bug

```
java.lang.AssertionError
	at test.wall.WallTests.cpuWall(WallTests.java:27)
	at one.profiler.test.Runner.main(Runner.java:207)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

### Expected vs. actual behavior

Expected test to pass.

### Reproduction Steps

GHA

### Additional Information/Context

https://github.com/fandreuz/async-profiler/actions/runs/15534127445/job/43729312239

### Async-profiler version

master

### Environment details

GHA macOS

Baraa-Hasheesh:
@apangin on macOS when using CPU event we would end up using wall profiling

```
if (event_name == NULL) {
        return &noop_engine;
    } else if (strcmp(event_name, EVENT_CPU) == 0) {
        if (FdTransferClient::hasPeer() || PerfEvents::supported()) {
            return &perf_events;
        } else if (CTimer::supported()) {
            return &ctimer;
        } else {
            return &wall_clock; <-------------
        }
    }
```

This means that technically speaking we're profiling wall twice on macOS in this test (different modes)

If my understanding is correct this test is meant to show the difference samplings done by WALL & CPU

for macOS It's checking WALL_CPU mode vs WALL_BATCH mode, is this intentional?

apangin:
@Baraa-Hasheesh Yes, CPU profiling is implemented using the Wall clock engine on macOS. The main difference is that, unlike wall clock sampling that signals all threads, `WALL_CPU` only samples threads in `RUNNABLE` state.

fandreuz:
@Baraa-Hasheesh should we expect this to be fixed after #1417 ?

Baraa-Hasheesh:
@fandreuz no
There are 2 issues here
* Sleep isn't that accurate on macOS GHA 
* Number of samples randomly deviates between threads

We've yet to fully confirm why the second issue happens here

apangin:
Let's see how it behaves after #1520.

PR #1473, opened 2025-09-10
disable wall-test for macOS

Baraa-Hasheesh:
### Description
As discussed internally wall test is currently not stable on macOS GHA runners
One of the contributing reasons is that sleep functions aren't per say stable on macOS GHA runners which the wall engine depend on to have accurate profiling results

For now we will disable the test to prevent random failures on the pipeline
& in the background we will try to find a way to re-enable the test for macOS

### Related issues
#1332 

### Motivation and context
reduce noise of random failures on GHA runners

### How has this been tested?
N/A

---

### DIFF
--- a/test/test/wall/WallTests.java
+++ b/test/test/wall/WallTests.java
@@ -5,6 +5,7 @@
 
 package test.wall;
 
+import one.profiler.test.Os;
 import one.profiler.test.Output;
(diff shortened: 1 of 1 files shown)


apangin:
Let's see if #1520 helps.
I'll close the PR for now. Can be reopened if failures continue.

PR #1474, opened 2025-09-11
Make comp-task matching more generic

Baraa-Hasheesh:
### Description
testCompTask checks that the compfeature is working, by checking for the synthetic frames async-profiler adds
These synthetic frames can be for classes outside of the `java/lang` package 

For example

```
;C2Compiler::compile_method;sun/util/locale/LocaleUtils.caseIgnoreMatch
;C2Compiler::compile_method;jdk/internal/loader/URLClassPath.getLoader
;Compiler::compile_method;sun/net/www/protocol/jar/Handler.parseContextSpec
;Compiler::compile_method;jdk/internal/loader/URLClassPath$JarLoader$1.run
```

This change allows for a more generic matching reducing the possibility of failures 

### Related issues
N/A

### Motivation and context
Reduce possibility for test failure

### How has this been tested?
manual testing

---

### DIFF
--- a/test/test/comptask/ComptaskTests.java
+++ b/test/test/comptask/ComptaskTests.java
@@ -17,7 +17,7 @@ public class ComptaskTests {
     public void testCompTask(TestProcess p) throws Exception {
         Output out = p.waitForExit("%f");
         assert p.exitCode() == 0;
-        assert out.contains(";Compiler::compile_method;(java|sun|jdk)/[^;]+\\.[^;]+;");
-        assert out.contains(";C2Compiler::compile_method;(java|sun|jdk)/[^;]+\\.[^;]+;");
+        assert out.contains(";Compiler::compile_method;(java|sun|jdk)/[^;]+[^;/]\\.[^;/]+;");
+        assert out.contains(";C2Compiler::compile_method;(java|sun|jdk)/[^;]+[^;/]\\.[^;/]+;");
     }
 }


test/test/comptask/ComptaskTests.java
```diff
@@ -17,7 +17,7 @@ public class ComptaskTests {
     public void testCompTask(TestProcess p) throws Exception {
         Output out = p.waitForExit("%f");
         assert p.exitCode() == 0;
-        assert out.contains(";Compiler::compile_method;java/lang/\\w+\\.");
-        assert out.contains(";C2Compiler::compile_method;java/lang/\\w+\\.");
+        assert out.contains(";Compiler::compile_method;(java|sun|jdk)/.*");
```
fandreuz: I wonder if we really need to be explicit here. Do we really care what package the compiled method is coming from?

Baraa-Hasheesh: are you suggesting checking against  `\\w+/.*`to match any packages that might be external?

In this case I don't really see the need as the test is plain JDK test which should always start from the above packages (sun,java,jdk)

fandreuz: > are you suggesting checking against \\w+/.*to match any packages that might be external?

Yes. But I'm not saying we should target external packages, just that perhaps we don't care about the package we catch here.

Baraa-Hasheesh: I think we should care about the base package name as that acts as a simple sanity test that async-profiler is adding logical frames to the generated stacks

fandreuz: Fair enough

apangin: The idea of the test is to verify that a synthetic frame that represents an existing Java method (not just an arbitrary character sequence) is placed above `compile_method`.

I don't mind making condition more generic, but I wonder in which circumstances it may happen that no `java/lang/` methods appear in the compilation profile?

Baraa-Hasheesh: The `java/lang` methods appear they just don't match the pattern in the regex

For example  `java/lang/ref/...` does appear in the profile

But the check is too strict

apangin: Thanks, but my question is different. How come that none of basic java/lang classes (String, Thread, Class, etc.) were seen in the compilation profile? Weren't these methods compiled, or they were but async-profiler did not catch any? How can I reproduce the issue? (Haven't seen it so far)

Baraa-Hasheesh: I'm attaching some flame graphs, when checked both show that both C1 & C2 compiler are working fine & are being sampled as we expect them to,

image

It's just that C1 compiler is simply fast & sometimes is not sampled for the classes/methods directly under `java/lang` package
=> Reducing the sampling interval increases the stability of the test.

If we check the available samples for C1 we can see the synthetic frames & we can see that each method was sampled only once which is what we expect to see

Due to the high performance of C1 it was possible for certain runs to not capture samples for the methods we currently expect, as such changing the pattern was needed,

I didn't see a value in picking one package over another as we may fall into the same issue in the future, so I opted for a pattern that should match any java synthetic frame

image

The `java/lang` can be still be seen on C2 compiler, and each method is only sampled once which is what we expect to see.
However to keep the check uniform, I opted to change the pattern for C2 as well.

[flamegraph.html](https://github.com/user-attachments/files/22276504/flamegraph.html)
[flamegraph2.html](https://github.com/user-attachments/files/22276580/flamegraph2.html)

apangin: Thanks, that makes sense.
The pattern, however, does not make difference between a class name and a method name. I'd like the test verify that it's a Java method that is inserted as a synthetic frame, i.e. frame name contains one dot.

Baraa-Hasheesh: Sure, that make sense, I will update the pattern

Baraa-Hasheesh: Pattern update to better match java methods 

New pattern `(java|sun|jdk)/[^;]+[^;/]\.[^;/]+`
* (java|sun|jdk) => top package sanity check 
* /[^;]+ => checks for nested packages
* [^;/]\.[^;/]+ => checks a method name that shouldn't contain "/;" characters & should have a "." in it

PR #1475, opened 2025-09-11
Add CPUTimeSample event support to jfrconv

krk:
This PR adds support for the `jdk.CPUTimeSample` event, an *experimental* feature introduced in OpenJDK 25.

This new mode provides a more accurate CPU profiling alternative to the default `ExecutionSample` in the JDK. It uses thread CPU time, which is particularly useful for analyzing CPU-bound applications.

### Motivation and context

`CPUTimeSample` event support was added to OpenJDK in https://github.com/openjdk/jdk/pull/25654, we can extend the JFR converter to support this new event type.

### How has this been tested?
By manually generating flamegraphs and collapsed output.

---

### DIFF
--- a/docs/ConverterUsage.md
+++ b/docs/ConverterUsage.md
@@ -45,8 +45,8 @@ Conversion options:
 
 
 JFR options:
-    --cpu              Generate only CPU profile during conversion.
-    --cpu-time         Generate only CPU profile during conversion, using CPUTimeSample events.
+    --cpu              Generate only CPU profile during conversion
+    --cpu-time         Generate only CPU profile, using CPUTimeSample events
     --wall             Generate only Wall clock profile during conversion
     --alloc            Generate only Allocation profile during conversion
     --live             Build allocation profile from live objects only during conversion
--- a/src/converter/one/convert/JfrConverter.java
+++ b/src/converter/one/convert/JfrConverter.java
@@ -75,7 +75,7 @@ public abstract class JfrConverter extends Classifier {
             threadStates = getThreadStates(false);
         } else if (args.cpuTime) {
             threadStates = new BitSet();
-            threadStates.set(ExecutionSample.CpuTimeSample);
+            threadStates.set(ExecutionSample.CPU_TIME_SAMPLE);
         }
 
         long startTicks = args.from != 0 ? toTicks(args.from) : Long.MIN_VALUE;
--- a/src/converter/one/jfr/event/ExecutionSample.java
+++ b/src/converter/one/jfr/event/ExecutionSample.java
@@ -6,7 +6,10 @@
 package one.jfr.event;
 
 public class ExecutionSample extends Event {
-    public static final int CpuTimeSample = 254;
+    // Synthetic thread state to distinguish samples converted from jdk.CPUTimeSample event.
+    // A small constant suitable for BitSet, does not clash with any existing thread state.
+    public static final int CPU_TIME_SAMPLE = 254;
+
     public final int threadState;
     public final int samples;
 
@@ -25,5 +28,4 @@ public class ExecutionSample extends Event {
     public long value() {
         return samples;
     }
-
 }
--- a/src/converter/one/jfr/JfrReader.java
+++ b/src/converter/one/jfr/JfrReader.java
@@ -250,10 +250,10 @@ public class JfrReader implements Closeable {
         long time = getVarlong();
         int stackTraceId = getVarint();
         int tid = getVarint();
-        boolean _failed = getBoolean();
-        long _samplingPeriod = getVarlong();
-        boolean _biased = getBoolean();
-        return new ExecutionSample(time, tid, stackTraceId, ExecutionSample.CpuTimeSample, 1);
+        boolean failed = getBoolean();
+        long samplingPeriod = getVarlong();
+        boolean biased = getBoolean();
+        return new ExecutionSample(time, tid, stackTraceId, ExecutionSample.CPU_TIME_SAMPLE, 1);
     }
 
     private MallocEvent readMallocEvent(boolean hasSize) {
@@ -653,7 +653,7 @@ public class JfrReader implements Closeable {
     }
 
     public boolean getBoolean() {
-        return getByte() != 0;
+        return buf.get() != 0;
     }
 
     public String getString() {


docs/ConverterUsage.md
```diff
@@ -45,7 +45,7 @@ Conversion options:
 
 
 JFR options:
-    --cpu              Generate only CPU profile during conversion
+    --cpu [EVENT]      Generate only CPU profile during conversion. EVENT is CPUTimeSample if specified, otherwise ExecutionSample
```
Baraa-Hasheesh: Writing `--cpu CPUTimeSample` isn't per say convenient 
How about just adding a new option like `--cpu-time` 

Also if this is introduced as a full new option a lot of aspects in this PR can be simplified 
For example we will no longer need the threadState changes

It's my opinion that we can deal with this sample as we do other sample types that don't have thread states such as Lock & Malloc samples.

What do you think?

apangin: 100% agree. `--cpu CPUTimeSample` is not only unusable, it also makes code hacky and breaks backward compatibility.
Btw, I like `--cpu-time`, but `Arguments` does not support dashes (we can fix this or come up with another alternative).

src/converter/one/convert/Arguments.java
```diff
@@ -76,6 +76,18 @@ public Arguments(String... args) {
...
+                    try {
+                        f.set(this, hasNextArg ? Enum.valueOf((Class<Enum>) type, args[++i]) : CpuSampleType.ExecutionSample);
+                    } catch (IllegalArgumentException e) {
+                        f.set(this, CpuSampleType.ExecutionSample);
+
+                        // Next arg was not a valid value for the enum
+                        if (hasNextArg) {
+                            i--;
+                        }
+                    }
```
apangin: It's a hack. Let's not change behavior of an existing option.
I think `--cpu` can be a "smart" option that automatically detects what event to use. To force using a specific event, there can be a separate option.

src/converter/one/jfr/event/IThreadState.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.jfr.event;
+
+public interface IThreadState {
```
apangin: No. There are many ways to do it without this artificial interface.

src/converter/one/jfr/event/CPUTimeSample.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.jfr.event;
+
+import one.jfr.JfrReader;
+
+public class CPUTimeSample extends Event implements IThreadState {
```
apangin: Why a new event? Can you reuse existing `ExecutionSample`?
The only conceptually new field here is `biased`, which is not used anywhere.
It's OK to ignore the field. Or we can added it to `ExecutionSample` if needed.

Baraa-Hasheesh: I have a small concern about using ExecutionSample for this new event. ExecutionSample would end up referring to 3 different events being wall, cpu, & cpu time.

That would increase the confusion that is currently happening when a user does a simple `jfrconv file.jfr file.html` command as the user would potentially end up with 3 different sample types in the generated flamegraph

Baraa-Hasheesh: On a somewhat related subject should this event be disabled when using the`JfrSync` feature?

apangin: > ExecutionSample would end up referring to 3 different events being wall, cpu, & cpu time.

Yes, and this is fine, as long as all 3 events carry similar information about program execution. They mostly differ by the collection mechanism, completeness and accuracy.
What `jfrconv` displays by default is an orthogonal problem, irrespective of representation. That's why I mentioned above that `--cpu` should be ideally a smart option that automatically selects the best available source of CPU profile data.

> should this event be disabled when using the JfrSync feature

Yes.

docs/ConverterUsage.md
krk: Adding `--cpu-time`.

src/converter/one/convert/Arguments.java
krk: We won't need this with `--cpu-time`. As for making `--cpu` a smart option, leaving that out of the current PR.

src/converter/one/jfr/event/CPUTimeSample.java
krk: It will be confusing to read the code but works. `CPUTimeSample` in the JFR file is read into `ExecutionSample` objects, with a synthetic thread state.

docs/ConverterUsage.md
Baraa-Hasheesh: let's revert the `--cpu [EVENT]` mention here  & replace it with the correct new `cpu-time`

src/converter/one/convert/Arguments.java
```diff
@@ -106,6 +107,20 @@ private static String alias(char c) {
         }
     }
 
+    private static String toCamelCase(String name) {
```
Baraa-Hasheesh, line 110: nit, you can simplify this to `return name.replaceAll("-", "");` & just make the name not camel case
-> Doing this logic just to make the name camel case seems like unneeded complexity IMO

src/converter/one/convert/ThreadStateFilter.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.convert;
+
+public enum ThreadStateFilter {
+    Wall, Cpu, CpuTime, Default
+}
```
Baraa-Hasheesh: whitespace

src/converter/one/convert/JfrConverter.java
```diff
@@ -63,24 +63,28 @@ protected void collectEvents() throws IOException {
...
         } else if (args.cpu) {
-            threadStates = getThreadStates(true);
+            threadStates = getThreadStates(ThreadStateFilter.Cpu);
+        } else if (args.cpuTime) {
+            threadStates = getThreadStates(ThreadStateFilter.CpuTime);
         } else if (args.wall) {
-            threadStates = getThreadStates(false);
+            threadStates = getThreadStates(ThreadStateFilter.Wall);
+        } else {
+            threadStates = getThreadStates(ThreadStateFilter.Default);
```
Baraa-Hasheesh: let's remove this branch & keep `threadStates =  null`, I don't see why it would be needed
You can do 

```
!(event instanceof ExecutionSample) || threadStates == null || threadStates.get(((ExecutionSample) event).threadState)
```

src/converter/one/convert/JfrConverter.java
```diff
@@ -63,24 +63,28 @@ protected void collectEvents() throws IOException {
...
         }
 
         long startTicks = args.from != 0 ? toTicks(args.from) : Long.MIN_VALUE;
         long endTicks = args.to != 0 ? toTicks(args.to) : Long.MAX_VALUE;
 
         for (Event event; (event = jfr.readEvent(eventClass)) != null; ) {
             if (event.time >= startTicks && event.time <= endTicks) {
-                if (threadStates == null || threadStates.get(((ExecutionSample) event).threadState)) {
+                if (!(event instanceof ExecutionSample) || threadStates.get(((ExecutionSample) event).threadState)) {
```
Baraa-Hasheesh: yeah this was a risky code :) 
If you do `--alloc --cpu` it would have failed previously

src/converter/one/convert/Arguments.java
apangin: That works, but better `replace` instead of `replaceAll`.

As for the camel case translation, implementation could be simpler:
```
for (int i; (i = name.lastIndexOf('-', name.length() - 2)) >= 0; ) {
    name = name.substring(0, i) + Character.toUpperCase(name.charAt(i + 1)) + name.substring(i + 2);
}
return name;
```

src/converter/one/convert/JfrConverter.java
apangin: Right, `threadStates == null` was an intentional micro-optimization for the common case when no filtering is needed.
BTW, `threadStates == null` can also substitute `instanceof` check: just don't set `threadStates` for event classes other than `ExecutionSample`.

src/converter/one/convert/JfrConverter.java
```diff
@@ -103,14 +107,22 @@ protected int toThreadState(String name) {
         throw new IllegalArgumentException("Unknown thread state: " + name);
     }
 
-    protected BitSet getThreadStates(boolean cpu) {
+    protected BitSet getThreadStates(ThreadStateFilter filter) {
```
apangin: I'm confused about this extra enum and all the related complications.
Will it work if you just add `if (args.cpuTime)` branch above and leave this method intact?

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: Currently if you do bad usages such as `--alloc --cpu` it would exit with a casting exception 

The correct solution IMO would adding a validation to prevent such cases

The simplest would be throwing a clear exception when the threadStates aren't null & the sample class is anything other than ExecutionSample

src/converter/one/convert/Arguments.java
krk: Let's keep camel-caseing.

Why is that implementation simpler though? I find it harder to read the intent. 

Not that performance matters in this choice, I got curious and run jmh on both, `lastIndexOf` suggestion is slower too.

src/converter/one/convert/ThreadStateFilter.java
krk: Thanks, didn't we have a linter for this?

src/converter/one/convert/JfrConverter.java
krk: We need to make the distinction between `cpu` and `cpuTime` and lack of both thereof, `boolean` wasn't cutting it.

Since now we are parsing `CPUTimeSample`s as `ExecutionSample`s, we need to exclude `CPUTimeSample`s with `getThreadState` if mode is `cpuTime`. Both sample types can be in the jfr at the same time.

apangin: No, you don't need to call this function for `cpuTime` - that's the point.
Your `ThreadStateFilter` serves as unnecessary bridge to connect two pieces of logic that should not be separate in the first place.

E.g., instead of
```
if (a) {
    filter = F1;
} else if (b) {
    filter = F2;
} else if (c) {
    filter = F3;
}

if (filter == F1) {
    // do_A
} else if (filter == F2) {
    // do_B
} else if (filter == F3) {
    // do_C
}
```
you could just write
```
if (a) {
    // do_A
} else if (b) {
    // do_B
} else if (c) {
    // do_C
}
```

src/converter/one/convert/Arguments.java
apangin: Because it's shorter and easier to follow - it literaly says: "while there is a dash, concatenate two parts of the string before and after the dash".
And it does not do unnecessary copying and allocations for the most common case when there are no dashes.

src/converter/one/convert/ThreadStateFilter.java
Baraa-Hasheesh: that's clang-format which is different from clang-tidy

src/converter/one/jfr/event/ExecutionSample.java
```diff
@@ -6,6 +6,7 @@
 package one.jfr.event;
 
 public class ExecutionSample extends Event {
+    public static final int CpuTimeSample = 254;
```
Baraa-Hasheesh: can you add some context behind why this value was selected,
It does look out of place when first checking the code

apangin: I added comment

PR #1499, opened 2025-09-19
Latency Profiling enhancements

fandreuz:
### Description
- Removed `--latency` in favor of `--trace`
- Format examples: 
   - `--trace A.B`: latency=0, wildcard signature
   - `--trace A.B(JJ)V`: latency=0
   - `--trace A.B(JJ)V+10ms`: latency=10ms
   - `--trace A.B+10ms`: latency=10ms, wildcard signature
- `--trace` can appear multiple times in the arguments
- `--trace` can target multiple methods in the same or different classes
- `--trace` takes one independent latency threshold for each target
- `event=A.B` will always perform "old" instrumentation
- Latency profiling can run in parallel with other profiling modes
- `args._include/exclude` now use `std::vector`

### Related issues
#1421 

### Motivation and context
This feature is nice to have for users, latency profiling should not exclude other profiling modes.

### How has this been tested?
New tests.

---

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -77,7 +77,7 @@ const size_t EXTRA_BUF_SIZE = 512;
 //     interval=N              - sampling interval in ns (default: 10'000'000, i.e. 10 ms)
 //     jstackdepth=N           - maximum Java stack depth (default: 2048)
 //     signal=N                - use alternative signal for cpu or wall clock profiling
-//     features=LIST           - advanced stack trace features (vtable, comptask, pcaddr)"
+//     features=LIST           - advanced stack trace features (mixed, vtable, comptask, pcaddr)"
 //     safemode=BITS           - disable stack recovery techniques (default: 0, i.e. everything enabled)
 //     file=FILENAME           - output file name for dumping
 //     log=FILENAME            - log warnings and errors to the given dedicated stream
@@ -320,6 +320,7 @@ Error Arguments::parse(const char* args) {
                     if (strstr(value, "stats"))    _features.stats = 1;
                     if (strstr(value, "jnienv"))   _features.jnienv = 1;
                     if (strstr(value, "probesp"))  _features.probe_sp = 1;
+                    if (strstr(value, "mixed"))    _features.mixed = 1;
                     if (strstr(value, "vtable"))   _features.vtable_target = 1;
                     if (strstr(value, "comptask")) _features.comp_task = 1;
                     if (strstr(value, "pcaddr"))   _features.pc_addr = 1;
@@ -406,7 +407,9 @@ Error Arguments::parse(const char* args) {
                     } else if (strcmp(value, "vm") == 0) {
                         _cstack = CSTACK_VM;
                     } else if (strcmp(value, "vmx") == 0) {
-                        _cstack = CSTACK_VMX;
+                        // cstack=vmx is a shorthand for cstack=vm,features=mixed
+                        _cstack = CSTACK_VM;
+                        _features.mixed = 1;
                     } else {
                         _cstack = CSTACK_NO;
                     }
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -61,8 +61,7 @@ enum SHORT_ENUM CStack {
     CSTACK_FP,       // walk stack using Frame Pointer links
     CSTACK_DWARF,    // use DWARF unwinding info from .eh_frame section
     CSTACK_LBR,      // Last Branch Record hardware capability
-    CSTACK_VM,       // unwind using HotSpot VMStructs
-    CSTACK_VMX       // same as CSTACK_VM but with intermediate native frames
+    CSTACK_VM        // unwind using HotSpot VMStructs
 };
 
 enum SHORT_ENUM Clock {
@@ -106,7 +105,7 @@ enum EventMask {
 constexpr int EVENT_MASK_SIZE = 6;
 
 struct StackWalkFeatures {
-    // Stack recovery techniques used to workaround AsyncGetCallTrace flaws
+    // Deprecated stack recovery techniques used to workaround AsyncGetCallTrace flaws
     unsigned short unknown_java  : 1;
     unsigned short unwind_stub   : 1;
     unsigned short unwind_comp   : 1;
@@ -115,19 +114,16 @@ struct StackWalkFeatures {
     unsigned short gc_traces     : 1;
 
     // Common features
-    unsigned short stats         : 1;
+    unsigned short stats         : 1;  // collect stack walking duration statistics
 
     // Additional HotSpot-specific features
-    unsigned short jnienv        : 1;
-    unsigned short probe_sp      : 1;
-    unsigned short vtable_target : 1;
-    unsigned short comp_task     : 1;
-    unsigned short pc_addr       : 1;
-    unsigned short _reserved     : 4;
-
-    StackWalkFeatures() : unknown_java(1), unwind_stub(1), unwind_comp(1), unwind_native(1), java_anchor(1), gc_traces(1),
-                          stats(0), jnienv(0), probe_sp(0), vtable_target(0), comp_task(0), pc_addr(0), _reserved(0) {
-    }
+    unsigned short jnienv        : 1;  // verify JNIEnv* obtained using VMStructs
+    unsigned short probe_sp      : 1;  // when AsyncGetCallTrace fails, adjust SP and retry
+    unsigned short mixed         : 1;  // mixed stack traces with Java and native frames interleaved
+    unsigned short vtable_target : 1;  // show receiver classes of vtable/itable stubs
+    unsigned short comp_task     : 1;  // display current compilation task for JIT threads
+    unsigned short pc_addr       : 1;  // record exact PC address for each sample
+    unsigned short _padding      : 3;  // pad structure to 16 bits
 };
 
 
@@ -271,7 +267,7 @@ class Arguments {
         _fdtransfer_path(NULL),
         _target_cpu(-1),
         _style(0),
-        _features(),
+        _features{1, 1, 1, 1, 1, 1},
         _cstack(CSTACK_DEFAULT),
         _clock(CLK_DEFAULT),
         _output(OUTPUT_NONE),
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -59,7 +59,7 @@ static jmethodID _start_method;
 static jmethodID _stop_method;
 static jmethodID _box_method;
 
-static const char* const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr", "vm", "vmx"};
+static const char* const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr", "vm"};
 
 
 struct CpuTime {
@@ -539,7 +539,7 @@ class Recording {
     }
 
     static const char* getFeaturesString(char* str, size_t size, StackWalkFeatures& f) {
-        snprintf(str, size, "%s %s %s %s %s %s %s %s %s %s %s %s",
+        snprintf(str, size, "%s %s %s %s %s %s %s %s %s %s %s %s %s",
                  f.unknown_java  ? "unknown_java"  : "-",
                  f.unwind_stub   ? "unwind_stub"   : "-",
                  f.unwind_comp   ? "unwind_comp"   : "-",
@@ -549,6 +549,7 @@ class Recording {
                  f.stats         ? "stats"         : "-",
                  f.jnienv        ? "jnienv"        : "-",
                  f.probe_sp      ? "probesp"       : "-",
+                 f.mixed         ? "mixed"         : "-",
                  f.vtable_target ? "vtable"        : "-",
                  f.comp_task     ? "comptask"      : "-",
                  f.pc_addr       ? "pcaddr"        : "-");
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -59,7 +59,7 @@ static const char USAGE_STRING[] =
     "  -I include          output only stack traces containing the specified pattern\n"
     "  -X exclude          exclude stack traces with the specified pattern\n"
     "  -L level            log level: debug|info|warn|error|none\n"
-    "  -F features         advanced stack trace features: vtable, comptask, pcaddr\n"
+    "  -F features         advanced stack trace features: mixed, vtable, comptask, pcaddr\n"
     "  -v, --version       display version string\n"
     "\n"
     "  --title string      FlameGraph title\n"
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -317,7 +317,7 @@ int Profiler::getNativeTrace(void* ucontext, ASGCT_CallFrame* frames, EventType 
     // Use PerfEvents stack walker for execution samples, or basic stack walker for other events
     if (event_type == PERF_SAMPLE) {
         native_frames = PerfEvents::walk(tid, ucontext, callchain, MAX_NATIVE_FRAMES, java_ctx);
-    } else if (_cstack >= CSTACK_VM) {
+    } else if (_cstack == CSTACK_VM) {
         return 0;
     } else if (_cstack == CSTACK_DWARF) {
         native_frames = StackWalker::walkDwarf(ucontext, callchain, MAX_NATIVE_FRAMES, java_ctx);
@@ -645,11 +645,11 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
         }
     }
 
-    if (_cstack == CSTACK_VMX) {
-        num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_EXPERT);
+    if (_features.mixed) {
+        num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, _features, event_type);
     } else if (event_type <= MALLOC_SAMPLE) {
         if (_cstack == CSTACK_VM) {
-            num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, VM_NORMAL);
+            num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, _features, event_type);
         } else {
             int java_frames = getJavaTraceAsync(ucontext, frames + num_frames, _max_stack_depth, &java_ctx);
             if (java_frames > 0 && java_ctx.pc != NULL && VMStructs::hasMethodStructs()) {
@@ -663,7 +663,7 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     } else if (event_type >= ALLOC_SAMPLE && event_type <= ALLOC_OUTSIDE_TLAB && _alloc_engine == &alloc_tracer) {
         VMThread* vm_thread;
         if (VMStructs::hasStackStructs() && (vm_thread = VMThread::current()) != NULL) {
-            num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, vm_thread->anchor());
+            num_frames += StackWalker::walkVM(ucontext, frames + num_frames, _max_stack_depth, vm_thread->anchor(), event_type);
         } else {
             num_frames += getJavaTraceAsync(ucontext, frames + num_frames, _max_stack_depth, &java_ctx);
         }
@@ -1177,7 +1177,7 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("DWARF unwinding is not supported on this platform");
     } else if (_cstack == CSTACK_LBR && _engine != &perf_events) {
         return Error("Branch stack is supported only with PMU events");
-    } else if (_cstack >= CSTACK_VM && VM::loaded() && !VMStructs::hasStackStructs()) {
+    } else if (_cstack == CSTACK_VM && VM::loaded() && !VMStructs::hasStackStructs()) {
         return Error("VMStructs stack walking is not supported on this JVM/platform");
     }
 
@@ -1356,7 +1356,7 @@ Error Profiler::check(Arguments& args) {
             return Error("DWARF unwinding is not supported on this platform");
         } else if (args._cstack == CSTACK_LBR && _engine != &perf_events) {
             return Error("Branch stack is supported only with PMU events");
-        } else if (args._cstack >= CSTACK_VM && VM::loaded() && !VMStructs::hasStackStructs()) {
+        } else if (args._cstack == CSTACK_VM && VM::loaded() && !VMStructs::hasStackStructs()) {
             return Error("VMStructs stack walking is not supported on this JVM/platform");
         }
     }
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -48,6 +48,11 @@ static inline void fillFrame(ASGCT_CallFrame& frame, ASGCT_CallFrameType type, c
     frame.method_id = (jmethodID)name;
 }
 
+static inline void fillFrame(ASGCT_CallFrame& frame, ASGCT_CallFrameType type, u32 class_id) {
+    frame.bci = type;
+    frame.method_id = (jmethodID)(uintptr_t)class_id;
+}
+
 static inline void fillFrame(ASGCT_CallFrame& frame, FrameTypeId type, int bci, jmethodID method) {
     frame.bci = FrameType::encode(type, bci);
     frame.method_id = method;
@@ -201,18 +206,19 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
     return depth;
 }
 
-int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, StackDetail detail) {
+int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
+                        StackWalkFeatures features, EventType event_type) {
     if (ucontext == NULL) {
-        return walkVM(&empty_ucontext, frames, max_depth, detail,
+        return walkVM(&empty_ucontext, frames, max_depth, features, event_type,
                       callerPC(), (uintptr_t)callerSP(), (uintptr_t)callerFP());
     } else {
         StackFrame frame(ucontext);
-        return walkVM(ucontext, frames, max_depth, detail,
+        return walkVM(ucontext, frames, max_depth, features, event_type,
                       (const void*)frame.pc(), frame.sp(), frame.fp());
     }
 }
 
-int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, JavaFrameAnchor* anchor) {
+int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, JavaFrameAnchor* anchor, EventType event_type) {
     uintptr_t sp = anchor->lastJavaSP();
     if (sp == 0) {
         return 0;
@@ -228,11 +234,13 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
         pc = ((const void**)sp)[-1];
     }
 
-    return walkVM(ucontext, frames, max_depth, VM_BASIC, pc, sp, fp);
+    StackWalkFeatures no_features{};
+    return walkVM(ucontext, frames, max_depth, no_features, event_type, pc, sp, fp);
 }
 
 int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
-                        StackDetail detail, const void* pc, uintptr_t sp, uintptr_t fp) {
+                        StackWalkFeatures features, EventType event_type,
+                        const void* pc, uintptr_t sp, uintptr_t fp) {
     StackFrame frame(ucontext);
     uintptr_t bottom = (uintptr_t)&frame + MAX_WALK_SIZE;
 
@@ -257,8 +265,11 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
         }
     }
 
+    // Show extended frame types and stub frames for execution-type events
+    bool details = event_type <= MALLOC_SAMPLE || features.mixed;
+
     JavaFrameAnchor* anchor = NULL;
-    if (detail > VM_BASIC && vm_thread != NULL && vm_thread->isJavaThread()) {
+    if (details && vm_thread != NULL && vm_thread->isJavaThread()) {
         anchor = vm_thread->anchor();
     }
 
@@ -287,7 +298,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
 
             if (nm->isNMethod()) {
                 int level = nm->level();
-                FrameTypeId type = detail != VM_BASIC && level >= 1 && level <= 3 ? FRAME_C1_COMPILED : FRAME_JIT_COMPILED;
+                FrameTypeId type = details && level >= 1 && level <= 3 ? FRAME_C1_COMPILED : FRAME_JIT_COMPILED;
                 fillFrame(frames[depth++], type, 0, nm->method()->id());
 
                 if (nm->isFrameCompleteAt(pc)) {
@@ -301,7 +312,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                         ScopeDesc scope(nm);
                         do {
                             scope_offset = scope.decode(scope_offset);
-                            if (detail != VM_BASIC) {
+                            if (details) {
                                 type = scope_offset > 0 ? FRAME_INLINED :
                                        level >= 1 && level <= 3 ? FRAME_C1_COMPILED : FRAME_JIT_COMPILED;
                             }
@@ -368,7 +379,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
 
                 fillFrame(frames[depth++], BCI_ERROR, "break_interpreted");
                 break;
-            } else if (detail < VM_EXPERT && nm->isEntryFrame(pc)) {
+            } else if (nm->isEntryFrame(pc) && !features.mixed) {
                 JavaFrameAnchor* next_anchor = JavaFrameAnchor::fromEntryFrame(fp);
                 if (next_anchor == NULL) {
                     fillFrame(frames[depth++], BCI_ERROR, "break_entry_frame");
@@ -385,11 +396,20 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                 }
                 continue;
             } else {
+                if (features.vtable_target && nm->isVTableStub() && depth == 0) {
+                    uintptr_t receiver = frame.jarg0();
+                    if (receiver != 0) {
+                        VMSymbol* symbol = VMKlass::fromOop(receiver)->name();
+                        u32 class_id = profiler->classMap()->lookup(symbol->body(), symbol->length());
+                        fillFrame(frames[depth++], BCI_ALLOC, class_id);
+                    }
+                }
+
                 CodeBlob* stub = profiler->findRuntimeStub(pc);
                 const void* start = stub != NULL ? stub->_start : nm->code();
                 const char* name = stub != NULL ? stub->_name : nm->name();
 
-                if (detail != VM_BASIC) {
+                if (details) {
                     fillFrame(frames[depth++], BCI_NATIVE_FRAME, name);
                 }
 
@@ -405,7 +425,22 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                 }
             }
         } else {
-            fillFrame(frames[depth++], BCI_NATIVE_FRAME, profiler->findNativeMethod(pc));
+            const char* method_name = profiler->findNativeMethod(pc);
+            char mark;
+            if (method_name != NULL && (mark = NativeFunc::mark(method_name)) != 0) {
+                if (mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
+                    // Skip all internal frames above malloc_hook functions, leave the hook itself
+                    depth = 0;
+                } else if (mark == MARK_COMPILER_ENTRY && features.comp_task && vm_thread != NULL) {
+                    // Insert current compile task as a pseudo Java frame
+                    VMMethod* method = vm_thread->compiledMethod();
+                    jmethodID method_id = method != NULL ? method->id() : NULL;
+                    if (method_id != NULL) {
+                        fillFrame(frames[depth++], FRAME_JIT_COMPILED, 0, method_id);
+                    }
+                }
+            }
+            fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
         }
 
         uintptr_t prev_sp = sp;
--- a/src/stackWalker.h
+++ b/src/stackWalker.h
@@ -7,6 +7,8 @@
 #define _STACKWALKER_H
 
 #include <stdint.h>
+#include "arguments.h"
+#include "event.h"
 #include "vmEntry.h"
 
 
@@ -25,23 +27,17 @@ struct StackContext {
     }
 };
 
-// Detail level of VMStructs stack walking
-enum StackDetail {
-    VM_BASIC,   // only basic Java frames similar to what AsyncGetCallTrace provides
-    VM_NORMAL,  // include frame types and runtime stubs
-    VM_EXPERT   // all features: frame types, runtime stubs, and intermediate native frames
-};
-
 class StackWalker {
   private:
     static int walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
-                      StackDetail detail, const void* pc, uintptr_t sp, uintptr_t fp);
+                      StackWalkFeatures features, EventType event_type,
+                      const void* pc, uintptr_t sp, uintptr_t fp);
 
   public:
     static int walkFP(void* ucontext, const void** callchain, int max_depth, StackContext* java_ctx);
     static int walkDwarf(void* ucontext, const void** callchain, int max_depth, StackContext* java_ctx);
-    static int walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, StackDetail detail);
-    static int walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, JavaFrameAnchor* anchor);
+    static int walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, StackWalkFeatures features, EventType event_type);
+    static int walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, JavaFrameAnchor* anchor, EventType event_type);
 
     static void checkFault();
 };
--- a/src/vmStructs.h
+++ b/src/vmStructs.h
@@ -489,7 +489,7 @@ class NMethod : VMStructs {
             return *(void**) at(-_nmethod_entry_offset);
         }
     }
-    
+
     bool contains(const void* pc) {
         return pc >= this && pc < at(size());
     }
@@ -519,6 +519,11 @@ class NMethod : VMStructs {
     bool isStub() {
         const char* n = name();
         return n != NULL && strncmp(n, "StubRoutines", 12) == 0;
+    }
+
+    bool isVTableStub() {
+        const char* n = name();
+        return n != NULL && strcmp(n, "vtable chunks") == 0;
     }
 
     VMMethod* method() {
--- a/test/test/vtable/VtableTests.java
+++ b/test/test/vtable/VtableTests.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.vtable;
+
+import one.profiler.test.*;
+import test.recovery.Numbers;
+import test.recovery.Suppliers;
+
+public class VtableTests {
+
+    @Test(mainClass = Numbers.class, jvm = Jvm.HOTSPOT)
+    public void vtableStubs(TestProcess p) throws Exception {
+        Output out = p.profile("-d 3 -e cpu -i 1ms --cstack vm -F vtable -o collapsed");
+        assert out.contains("Numbers.avg;vtable stub;java.lang.Integer_\\[i]");
+        assert out.contains("Numbers.avg;vtable stub;java.lang.Long_\\[i]");
+    }
+
+    @Test(mainClass = Suppliers.class, jvm = Jvm.HOTSPOT)
+    public void itableStubs(TestProcess p) throws Exception {
+        Output out = p.profile("-d 3 -e cpu -i 1ms --cstack vm -F comptask,vtable -o collapsed");
+        assert out.contains("Suppliers.loop;itable stub;test.recovery.Suppliers[^_]+_\\[i]");
+    }
+}


src/profiler.cpp
```diff
@@ -74,6 +74,7 @@ enum EventMask {
     EM_LOCK  = 4,
     EM_WALL  = 8,
     EM_NATIVEMEM = 16,
+    EM_INSTRUMENT = 32,
```
fandreuz: The comment above says these constants are the same we use in `JfrSync`. Is this correspondence something we want to keep @apangin ?

src/arguments.h
```diff
@@ -162,6 +162,7 @@ class Arguments {
     Action _action;
     Counter _counter;
     const char* _event;
+    const char* _instrument;
```
fandreuz: This change makes `latency` and `interval` a bit uglier because it's going to be the same for `instrument` and `event`. Should we think about a different solution?

Makefile
```diff
@@ -311,10 +311,10 @@ cpp-lint-diff:
 		clang-tidy-diff.py -p1 $(CLANG_TIDY_ARGS_EXTRA) -- -x c++ $(CXXFLAGS) $(INCLUDES) $(DEFS) $(LIBS)
 
 check-md:
-	prettier -c README.md "docs/**/*.md"
+	prettier -c README.md "docs/**/*.md" "!docs/ProfilerOptions.md"
```
apangin: No. Either remove prettier from CI altogether, or make the checks pass.

docs/ProfilerOptions.md
```diff
@@ -37,6 +37,8 @@ The below options are `action`s for async-profiler and common for both `asprof`
...
 | `--latency N`      | `latency=N`       | Duration threshold for Java method latency profiling. Should be a time interval (see `interval` docs for more details).                                                                                                                                                                                                                                                                                                                                                                                                                     |
+| `--instrument C.M` | `instrument=C.M` | Class (`C`) and method (`M`) to be instrumented. Equivalent to `event=C.M`. |
```
apangin: `--trace METHOD` would be better.

src/arguments.cpp
```diff
@@ -60,6 +60,7 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
 //     live               - build allocation profile from live objects only
 //     nativemem[=BYTES]  - profile native allocations with BYTES interval
 //     nofree             - do not collect free calls in native allocation profiling
+//     instrument=C.M     - class (C) and method (M) to be instrumented, equivalent to event=C.M
```
apangin: Since you are adding a new parameter anyway, let's take this opportunity to improve UX.
1. IMO, `trace` is a better name: it is shorter and better reflects the purpose of the option.
2. I want to make it possible to specify multiple methods: either by listing them separated by `;` or by adding multiple `trace` options: `trace=Class1.method1,trace=Class2.method2,...`
3. `latency` option could be probably merged with `trace` parameter - to be able to specify different latencies for different methods, e.g. `trace=Class1.method1:5ms,trace=Class2.method2:10ms,...` Syntax is discussible.
4. We can handle `trace` option completely differently from `-e`. It's OK to leave `-e` for backward compatible method entry profiling only, while moving all the advanced tracing logic to a separate option.

I'm happy to discuss more alternatives if you have other ideas.

src/profiler.cpp
apangin: Why not?

src/arguments.cpp
Baraa-Hasheesh: Adding here, `interval` is now confusing & not really usable when both engines are active
=> `start,event=cpu,trace=Class1.method1:5ms,interval=100ms`

I like option3 & IMO, we should consider encoding the frequency into the `trace` option as well (Not just the duration)

src/arguments.cpp
```diff
@@ -260,6 +261,9 @@ Error Arguments::parse(const char* args) {
             CASE("nofree")
                 _nofree = true;
 
+            CASE("instrument")
```
Baraa-Hasheesh: If I understand the flow correctly `start,instrument=Class.method` would behave the same as doing `start,event=Class.method`, doesn't this mean we should update the JfrSync event mask?
```
int event_mask = (args._event != NULL ? 1 : 0) |
                     (args._alloc >= 0 ? 2 : 0) |
                     (args._lock >= 0 ? 4 : 0) |
                     (args._latency >= 0 ? 8 : 0) |
                     ((args._jfr_options ^ JFR_SYNC_OPTS) << 4);

env->CallStaticVoidMethod(_jfr_sync_class, _start_method, jfilename, jsettings, event_mask);
```

src/instrument.cpp
```diff
@@ -1012,7 +1012,7 @@ Error Instrument::start(Arguments& args) {
         return error;
     }
 
-    setupTargetClassAndMethod(args._event);
+    setupTargetClassAndMethod(args._instrument != NULL ? args._instrument : args._event);
```
Baraa-Hasheesh: I noticed that the `JfrSync.class` was updated but the `JfrSync.java` wasn't
is there a reason for that?

src/profiler.cpp
fandreuz: This was settled in #1507

fandreuz:
Hi team, thanks for the feedback. This PR currently does a bit more than what I was expecting:
- enable method tracing in parallel with other engines
- enable multiple methods to be traced

I'll need some more changes to get the following:
1. enable multiple methods to be traced **in the same class** (won't work at the moment)
2. per-method latency
3. per-method interval

I see two ways to move forward:
- keep this PR open a bit more and add all the required features (I think 1# is needed, 2# is nice, 3# is lower-priority);
- we add checks to report an appropriate failure when user tries to do 1#, merge this PR, and follow-up later.

Any thoughts @apangin ?

src/arguments.cpp
```diff
@@ -46,79 +46,81 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
...
+//     status                  - print profiling status (inactive / running for X seconds)
+//     meminfo                 - print profiler memory stats
+//     list                    - show the list of available profiling events
+//     version                 - display the agent version
+//     event=EVENT             - which event to trace (cpu, wall, cache-misses, etc.)
+//     alloc[=BYTES]           - profile allocations with BYTES interval
+//     live                    - build allocation profile from live objects only
+//     nativemem[=BYTES]       - profile native allocations with BYTES interval
+//     nofree                  - do not collect free calls in native allocation profiling
+//     trace=METHOD[:DURATION] - method to be traced with optional latency,
```
Baraa-Hasheesh: nit: can we add a larger space here to avoid changing the space on every new arg we may add

src/instrument.cpp
```diff
@@ -973,13 +974,29 @@ Result BytecodeRewriter::rewriteClass() {
...
+std::vector<std::string> Instrument::_targets;
 bool Instrument::_instrument_class_loaded = false;
 u64 Instrument::_interval;
 long Instrument::_latency;
 volatile u64 Instrument::_calls;
 volatile bool Instrument::_running;
 
+bool oneDotPerTarget(const char* target) {
+    bool found_dot = false;
+    for (const char* s = target; *s != 0; ++s) {
```
Baraa-Hasheesh: nit, `s != '\0'` is more readable IMO

docs/ProfilerOptions.md
```diff
@@ -37,6 +37,8 @@ The below options are `action`s for async-profiler and common for both `asprof`
...
 | `--latency N`      | `latency=N`       | Duration threshold for Java method latency profiling. Should be a time interval (see `interval` docs for more details).                                                                                                                                                                                                                                                                                                                                                                                                                     |
+| `--trace METHOD[:DURATION]` | `trace=METHOD[:DURATION]` | Method to be traced with optional latency treshold, supports multiple values separated by `+`. |
```
Baraa-Hasheesh: This isn't not really user friendly If I want to profile multiple options via asprof, I would have to do the following

```
--trace Class1.Method1:10ms+Class2.Method1:20ms+Class2.Method4:50ms
```

I think @apangin previous suggestion is cleaner which would result in the following

```
--trace Class1.Method1:10ms \
--trace Class2.Method1:20ms \
--trace Class2.Method4:15ms 
```

src/arguments.cpp
```diff
@@ -495,7 +500,7 @@ Error Arguments::parse(const char* args) {
         return Error(msg);
     }
 
-    if (_event == NULL && _alloc < 0 && _lock < 0 && _wall < 0 && _nativemem < 0) {
+    if (_event == NULL && _alloc < 0 && _lock < 0 && _wall < 0 && _nativemem < 0 && _latency < 0) {
```
Baraa-Hasheesh: This should be `trace == NULL`

src/instrument.cpp
```diff
@@ -1026,24 +1051,41 @@ Error Instrument::start(Arguments& args) {
...
-    *strrchr(new_class, '.') = 0;
-
-    for (char* s = new_class; *s; s++) {
-        if (*s == '.') *s = '/';
+    _targets.clear();
+    std::string new_event = event;
+    size_t current_start = 0;
+    long last_dot = -1;
+
+#define HANDLE_TARGET(end) /* end non-inclusive */                            \
```
Baraa-Hasheesh: I think it would be cleaner as a function
If for performance I would recommend moving the definition to the top of the file

src/main/main.cpp
```diff
@@ -505,7 +507,7 @@ int main(int argc, const char** argv) {
             format << "," << (arg.str() + 2);
 
         } else if (arg == "--alloc" || arg == "--nativemem" || arg == "--lock" || arg == "--wall" ||
-                   arg == "--latency" || arg == "--chunksize" || arg == "--chunktime" ||
+                   arg == "trace" || arg == "--latency" || arg == "--chunksize" || arg == "--chunktime" ||
```
Baraa-Hasheesh: shouldn't this be "--trace"?

src/arguments.h
```diff
@@ -311,7 +313,8 @@ class Arguments {
                (_alloc     >= 0    ? EM_ALLOC                                   : 0) |
                (_lock      >= 0    ? EM_LOCK                                    : 0) |
                (_wall      >= 0    ? EM_WALL                                    : 0) |
-               (_nativemem >= 0    ? EM_NATIVEMEM                               : 0);
+               (_nativemem >= 0    ? EM_NATIVEMEM                               : 0) |
+               (_trace     != NULL ? EM_METHOD_TRACE                            : 0);
```
Baraa-Hasheesh: this isn't always true, if latency is not encoded/selected, then you should disable the EM_CPU
Also I think you need to clean the first line of this function `EM_METHOD_TRACE`
`        return (_event     != NULL ? (_latency >= 0 ? EM_METHOD_TRACE : EM_CPU) : 0) |`

src/arguments.cpp
```diff
@@ -260,6 +262,9 @@ Error Arguments::parse(const char* args) {
             CASE("nofree")
                 _nofree = true;
 
+            CASE("trace")
+                _trace = value;
```
Baraa-Hasheesh: if latency isn't encoded this can technically will enable Instrumentation which records the samples as ExecutionSample

```
case INSTRUMENTED_METHOD:
     _rec->recordExecutionSample(buf, tid, call_trace_id, (ExecutionEvent*)event);
     break;
```

This in turn will create confusion which CPU samples are instrumentation & which are actual CPU samples.

src/instrument.cpp
```diff
@@ -1003,6 +1020,14 @@ Error Instrument::check(Arguments& args) {
     if (args._latency < 0 && args._interval < 0) {
         return Error("interval must be positive");
     }
+    const char* target = args._trace != NULL ? args._trace : args._event;
```
Baraa-Hasheesh: This might be bad as the user can do the following `event=Class1.Method1,trace=Class2.Method2:10ms`

I think it might be a safe approach to check if the user is selecting old instrumentation via old event option but is also selecting tracing via the new trace option

src/instrument.cpp
fandreuz: It's only used in two places in the function where I define it

src/instrument.cpp
apangin: Let me repeat [this comment](https://github.com/async-profiler/async-profiler/pull/1507#discussion_r2379972776) once again: focus on functional problems in your code reviews, not on subjective style preferences.

fandreuz:
I propose the following action table:

| trace=A.B(50ms) | event=NULL | event=cpu | event=C.D | latency=100ms | Result |
|:-----------------:|:------------:|:-----------:|:-----------:|:--------------:|:--------|
| yes             | yes        | no        | no        | yes          |  Error: unexpected argument `latency`  |
| yes             | no         | yes       | no        | yes          |  Error: unexpected argument `latency`  |
| yes             | no         | no        | yes       | yes          |  Trace `A.B` with latency=50ms, trace `C.D` with latency=100ms  |
| yes             | yes        | no        | no        | no           |   Trace `A.B` with latency=50ms  |
| yes             | no         | yes       | no        | no           |   Trace `A.B` with latency=50ms, and cpu sampling  |
| yes             | no         | no        | yes       | no           |   Trace `A.B` with latency=50ms, old instrumentation on `C.D`  |
| no              | yes        | no        | no        | yes          |   Error: unexpected argument `latency`  |
| no              | no         | yes       | no        | yes          |   Error: unexpected argument `latency`  |
| no              | no         | no        | yes       | yes          |   Trace `C.D` with latency=100ms   |
| no              | yes        | no        | no        | no           |   Cpu sampling  |
| no              | no         | yes       | no        | no           |    Cpu sampling   |
| no              | no         | no        | yes       | no           |   Old instrumentation on `C.D`  |

Q:
- Is it ok to let `--latency` refer only to `event=C.D`, not to `--trace`?
    - Consequence: `--trace=A.B` will fail, `--trace=A.B(50ms)` is expected
    - Alternative: `--trace=A.B` will do old instrumentation rather than method tracing
          - If we do this, then we should probably reconsider the name `--trace` as it won't _only_ do method tracing
- Do we want to support row 3# ?

fandreuz:
Discussed offline, we'll remove `--latency`. New table:
| trace=A.B(50ms) | event=NULL | event=cpu | event=C.D | Result                                             |
|:---------------:|:----------:|:----------:|:----------:|:--------------------------------------------------|
| yes             | yes        | no         | no         | Trace `A.B` with latency=50ms                      |
| yes             | no         | yes        | no         | Trace `A.B` with latency=50ms, and cpu sampling    |
| yes             | no         | no         | yes        | Trace `A.B` with latency=50ms, old instrumentation on `C.D` |
| no              | yes        | no         | no         | Cpu sampling                                       |
| no              | no         | yes        | no         | Cpu sampling                                       |
| no              | no         | no         | yes        | Old instrumentation on `C.D`                       |

src/instrument.cpp
fandreuz: Added a new check for this:
```java
    if (!args._trace.empty() && args._event != NULL && strchr(args._event, '.') != NULL) {
        return Error("Running method tracing and Java method sampling in parallel is not supported");
    }
```
so old instrumentation + method tracing is not allowed

src/instrument.cpp
```diff
@@ -1000,20 +1047,26 @@ Error Instrument::check(Arguments& args) {
         _instrument_class_loaded = true;
     }
 
-    if (args._latency < 0 && args._interval < 0) {
-        return Error("interval must be positive");
+    if (args._trace.empty() && args._event == NULL) {
+        return Error("Missing target class for instrumentation");
```
Baraa-Hasheesh: When is this branch ever executed?

I can't find a single way to ever reach this line

src/instrument.cpp
```diff
@@ -967,19 +1014,19 @@ Result BytecodeRewriter::rewriteClass() {
...
-char* Instrument::_target_class = NULL;
+Targets Instrument::_targets;
 bool Instrument::_instrument_class_loaded = false;
 u64 Instrument::_interval;
-long Instrument::_latency;
 volatile u64 Instrument::_calls;
 volatile bool Instrument::_running;
 
+
 Error Instrument::check(Arguments& args) {
```
Baraa-Hasheesh, line 1062: you need to update `Profiler::check` method in `profiler.cpp` 
Currently it's assumed that the only way to select this engine is via `event` option
But this is no longer true after these changes

src/instrument.cpp
```diff
@@ -1000,20 +1047,26 @@ Error Instrument::check(Arguments& args) {
...
 
-    if (args._latency < 0 && args._interval < 0) {
-        return Error("interval must be positive");
+    if (args._trace.empty() && args._event == NULL) {
+        return Error("Missing target class for instrumentation");
     }
+    if (!args._trace.empty() && args._event != NULL && strchr(args._event, '.') != NULL) {
+        return Error("Running method tracing and Java method sampling in parallel is not supported");
+    }
+    if (args._trace.empty() && strchr(args._event, '+') != NULL) {
```
Baraa-Hasheesh: `args._trace.empty() ` is not needed here

`args._event && strchr(args._event, '+') != NULL`

src/instrument.cpp
```diff
@@ -1000,20 +1047,26 @@ Error Instrument::check(Arguments& args) {
         _instrument_class_loaded = true;
     }
 
-    if (args._latency < 0 && args._interval < 0) {
-        return Error("interval must be positive");
+    if (args._trace.empty() && args._event == NULL) {
+        return Error("Missing target class for instrumentation");
     }
+    if (!args._trace.empty() && args._event != NULL && strchr(args._event, '.') != NULL) {
+        return Error("Running method tracing and Java method sampling in parallel is not supported");
```
Baraa-Hasheesh: I think the better place to check this is the `profiler.cpp` file 
=> Check how the wall engine is checked against the event engine

you can do the same there, `!args._trace.empty()  && _engine == &instrument`

src/instrument.cpp
```diff
@@ -1026,24 +1079,107 @@ Error Instrument::start(Arguments& args) {
...
+    if (class_name.find_first_of("+()") != std::string::npos) {
+        // E.g. wrong signature
+        // the.package.name.ClassName.MethodName(Jjava.lang.String;)V
+        return Error("Unexpected format for tracing target");
+    }
+
+    // Replace all '.' in package name with '/'
+    size_t dot_idx;
+    size_t pos = 0;
+    while ((dot_idx = class_name.find('.', pos)) != std::string::npos) {
```
Baraa-Hasheesh, line 1138: nit: Wouldn't just iterating over the string length be simpler than using find?

fandreuz: How would this make things simpler?

You would still have all the branches you see in this function, what I like from the current impl is that we can "jump around" using `strchr`, rather than checking each symbol manually, which would require additional branching. Happy to hear alternatives.

src/instrument.cpp
fandreuz: 252643a7c821b4128564118b09aecea874d18607

src/instrument.cpp
fandreuz: 252643a7c821b4128564118b09aecea874d18607

src/instrument.cpp
fandreuz: 3e29d069a88525e93e7da295d6c7d57be9335d2a

src/instrument.cpp
fandreuz: 1c815f809c5c4f4bc57b0a59b1d722c006e1e45a

src/profiler.cpp
```diff
@@ -1166,6 +1168,8 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("target-cpu is only supported with perf_events");
     } else if (_engine != &perf_events && args._record_cpu) {
         return Error("record-cpu is only supported with perf_events");
+    } else if (_engine == &instrument && !args._trace.empty()) {
+        return Error("Running method tracing and Java method sampling in parallel is not supported");
```
Baraa-Hasheesh, line 1172: @apangin This might be out of scope for this PR, but these checks for engine type should be moved under the `Profiler::check` method as that is a more proper place for these kind of checks

What are your thoughts about this?

src/profiler.cpp
```diff
@@ -1329,6 +1343,9 @@ Error Profiler::check(Arguments& args) {
     if (!error && args._lock >= 0) {
         error = lock_tracer.check(args);
     }
+    if (!error && !args._trace.empty()) {
+        error = instrument.check(args);
```
Baraa-Hasheesh, line 1347: Sorry, I just realized this isn't needed
The `_engine == &instrument && !args._trace.empty()` is good enough on it's own

docs/ProfilingModes.md
```diff
@@ -191,9 +191,8 @@ The massive CodeCache flush doesn't occur if attaching async-profiler as an agen
 
 ### Latency profiling
 
-In the context of _Java method profiling_, async-profiler supports latency profiling. Users may
-specify a threshold latency with the parameter `latency`. Only the calls exceeding this threshold
-in total runtime will be profiled. Method calls resulting in an exception are ignored.
+Please refer to our blog post on [latency profiling](https://github.com/async-profiler/async-profiler/discussions/1497)
```
Baraa-Hasheesh, line 194: The content of the blog post are no longer 100% valid, 
specially how to invoke this profiling mode

src/profiler.cpp
apangin: You gave me an idea: #1518

src/instrument.h
```diff
@@ -7,16 +7,26 @@
...
+typedef long Latency;
+typedef std::unordered_map<MethodName, std::unordered_map<MethodSignature, Latency>> MethodTargets;
+typedef std::unordered_map<ClassName, MethodTargets> Targets;
+
+const MethodTargets EMPTY_METHOD_TARGETS;
 
 class Instrument : public Engine {
   private:
-    static char* _target_class;
+    static Targets _targets;
```
Baraa-Hasheesh, line 24: what would happen if I have 2 targets referring to the same method?
`trace=*.method1+50ms,trace=class1.method1+20ms`

fandreuz: Non-wildcard has the priority. Wildcard is also not documented, so users won't even need to ask this question.

src/arguments.h
```diff
@@ -166,16 +172,15 @@ class Arguments {
...
     static long long hash(const char* arg);
     static Output detectOutputFormat(const char* file);
-    static long parseUnits(const char* str, const Multiplier* multipliers);
     static int parseTimeout(const char* str);
 
   public:
     Action _action;
     Counter _counter;
     const char* _event;
+    std::vector<const char*> _trace;
```
apangin, line 175: Well... For a long time, I tried to avoid STL classes in `Arguments`, hoping to get rid of libstdc++ completely some day. That also helped to keep all content of the class static (with `_buf` being the only dynamic field that is allocated just once).

I think I'm not that categorical anymore, but nonetheless, what I can't accept is nonuniformity in argument handling. Let's agree on how to store lists and stick with that approach for all similar arguments. Currently, we handle `include`/`exclude` lists differently - see `Arguments::appendToEmbeddedList`. I admit it's ingenious but hacky :) I'm open to changing it, but my point is that `include`, `exclude` and `trace` lists should follow the same pattern.

src/instrument.h
```diff
@@ -7,16 +7,26 @@
...
+#include <vector>
 #include "arch.h"
 #include "engine.h"
 
+typedef std::string ClassName;
+typedef std::string MethodName;
+typedef std::string MethodSignature;
+typedef long Latency;
+typedef std::unordered_map<MethodName, std::unordered_map<MethodSignature, Latency>> MethodTargets;
+typedef std::unordered_map<ClassName, MethodTargets> Targets;
```
apangin: No to 3-level maps.
I can understand why 2 levels may be necessary (because you have to make an early decision based on a class name), but 3 is overkill.

src/arguments.cpp
```diff
@@ -46,78 +40,78 @@ static const Multiplier UNIVERSAL[] = {{'n', 1}, {'u', 1000}, {'m', 1000000}, {'
...
+//     status                  - print profiling status (inactive / running for X seconds)
+//     meminfo                 - print profiler memory stats
+//     list                    - show the list of available profiling events
+//     version                 - display the agent version
+//     event=EVENT             - which event to trace (cpu, wall, cache-misses, etc.)
+//     alloc[=BYTES]           - profile allocations with BYTES interval
+//     live                    - build allocation profile from live objects only
+//     nativemem[=BYTES]       - profile native allocations with BYTES interval
+//     nofree                  - do not collect free calls in native allocation profiling
+//     trace=METHOD[+DURATION] - method to be traced with optional latency threshold
```
apangin: Why did you choose `+`?
`:` would look more natural.
Is it because of [condition](https://github.com/async-profiler/async-profiler/blob/6453ccca4364f4d8e0655945e0edc98e9ea0924c/src/profiler.cpp#L1008) in `selectEngine`? It can be changed to something smarter.

src/instrument.cpp
Baraa-Hasheesh: doing the following looks simpler to me, & should have better performance 

```
for (size_t i = 0; i <= class_name_size; i++) {
     if (class_name[i] == '.') class_name[i] = '/';
}
```

fandreuz: > & should have better performance

I'm not sure about that.

> doing the following looks simpler to me

Clunkier syntax, why complicate things if we have handy methods from `std::string` (which we use anyway due to the `std::unordered_map`?

src/arguments.h
fandreuz: My personal preference would be changing all of them to use std::vector :)

What do you think? But also, maybe it's not a problem we need to settle in this PR, I changed `_trace` to use the common pattern: 388115571924afbd3b3927ae8a31aed5f0140c07

src/instrument.h
fandreuz: I think things are a bit uglier this way: 6ff4fc6fb0a6e1f03349b02179941869460044ba

Am I missing some obvious simplification?

src/arguments.h
fandreuz: All vector now: f4df48f2144994f9a2c3b9fa167b1d2f51097558

src/arguments.cpp
fandreuz: No real reason actually: a5a787daf3cae44580324225399ba46098710178

src/instrument.cpp
```diff
@@ -1026,24 +1076,107 @@ Error Instrument::start(Arguments& args) {
...
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_CLASS_FILE_LOAD_HOOK, NULL);
+
+    _running = false;
+    _targets.clear();
 }
 
-void Instrument::setupTargetClassAndMethod(const char* event) {
-    char* new_class = strdup(event);
-    *strrchr(new_class, '.') = 0;
+Error handleTarget(Targets& targets, const char* s, long default_latency) {
```
fandreuz: @apangin can I use `<regex>` here?

fandreuz:
Should we have a per-engine `interval` as a follow-up to this PR?

Baraa-Hasheesh:
> Should we have a per-engine `interval` as a follow-up to this PR?

Is the question about doing that at all, or just doing it later in a different PR?

If it's the former, I believe that should it be done as the interval is confusing & not really usable when both engines are active,

If it's the latter, that should be fine IMO, but I would vote to remove the interval logic from the method tracing for now 

What are your thoughts here @fandreuz

src/arguments.h
```diff
@@ -190,8 +193,8 @@ class Arguments {
     const char* _unknown_arg;
     const char* _server;
     const char* _filter;
-    int _include;
-    int _exclude;
+    std::vector<const char*> _include;
+    std::vector<const char*> _exclude;
```
apangin, line 193: Should the initializer be changed?
```
    _include(0),
    _exclude(0),
```

src/flightRecorder.cpp
```diff
@@ -675,9 +675,9 @@ class Recording {
             writeIntSetting(buf, T_MONITOR_ENTER, "lock", args._lock);
         }
 
-        writeBoolSetting(buf, T_METHOD_TRACE, "enabled", args._latency >= 0);
-        if (args._latency >= 0) {
-            writeIntSetting(buf, T_METHOD_TRACE, "latency", args._latency);
+        writeBoolSetting(buf, T_METHOD_TRACE, "enabled", !args._trace.empty());
+        if (!args._trace.empty()) {
```
apangin: Redundant condition.

src/instrument.cpp
```diff
@@ -867,6 +876,27 @@ Result BytecodeRewriter::rewriteCodeAttributes(const u16* relocation_table, int
     return Result::OK;
 }
 
+bool BytecodeRewriter::findLatency(long& latency, u16 method_name_index, u16 method_desc_index) {
+    auto name_it = _target_methods.find(_cpool[method_name_index]->as_string());
+    if (name_it == _target_methods.end()) {
+        name_it = _target_methods.find("*");
+    }
+    if (name_it == _target_methods.end()) return false;
```
apangin: There is no longer possibility to specify a partial name?
E.g. `read*` to match `readByte`, `readInt`, etc..

src/instrument.cpp
```diff
@@ -1000,20 +1050,20 @@ Error Instrument::check(Arguments& args) {
         _instrument_class_loaded = true;
     }
 
-    if (args._latency < 0 && args._interval < 0) {
-        return Error("interval must be positive");
+    if (args._event != NULL && strchr(args._event, ':') != NULL) {
+        return Error("Use 'trace' for latency profiling");
```
apangin: Can we really reach here? I'd assume `Instrument` engine should not be selected for such events.

src/instrument.cpp
```diff
@@ -1026,24 +1076,107 @@ Error Instrument::start(Arguments& args) {
...
+Error handleTarget(Targets& targets, const char* s, long default_latency) {
+    // Expected formats:
+    // - the.package.name.ClassName.MethodName
+    // - the.package.name.ClassName.MethodName+50ms
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V+50ms
+
+    const char* last_dot = strrchr(s, '.');
+    if (last_dot == NULL) {
+        return Error("Unexpected format for tracing target");
```
apangin, line 1125: I'd expect we never reach here. Maybe `assert`?

src/instrument.cpp
apangin: What for?
There are no prohibited APIs - it's always a question of benefits/price tradeoff.
If regex makes the code much smaller and nicer without compromising compatibility, performance or footprint - then I'll be all for it. But something tells me it won't come for free.

src/instrument.cpp
fandreuz: I think we can, there's no check on `trace` before we reach here. We could have this check in `check()`, but I think that would be a bit wasteful since we're using the value here.

src/instrument.cpp
fandreuz: Right, d11ae22fe99ab3a90c81561012a5bfeb0ade723e

src/flightRecorder.cpp
fandreuz: d11ae22fe99ab3a90c81561012a5bfeb0ade723e

src/arguments.h
fandreuz: d11ae22fe99ab3a90c81561012a5bfeb0ade723e

src/instrument.cpp
fandreuz: I brought it back in 5119d2a444195911d2349eb96d245801023a04d8

But we have to iterate potentially all map entries every time. I thought of having an hybrid approach (first try exact match, then wildcards) but there are some drawbacks:
- I think `unordered_map` is less efficient
- direct access to the map requires a copy to form the `std::string` if we have a `const char*`

Probably not worth the effort.

src/instrument.cpp
```diff
@@ -37,6 +37,20 @@ static inline u16 alignUp4(u16 i) {
...
 }
 
+static bool matchesPattern(const char* value, size_t len, const std::string& pattern) {
+    if (len == 0 || pattern.empty()) return false;
+    return (
+        // wildcard match
+        (
+            pattern[pattern.length() - 1] == '*' &&
+            len >= pattern.length() - 1 &&
+            memcmp(pattern.c_str(), value, pattern.length() - 1) == 0
```
Baraa-Hasheesh, line 49: why `memcmp` rather than `strncmp`?
I'm aware that both will work for this case, I'm just curious why

src/instrument.cpp
```diff
@@ -1026,24 +1075,107 @@ Error Instrument::start(Arguments& args) {
...
+    } else {
+        signature = "*";
+        colon = strchr(last_dot, ':');
+    }
+
+    // Latency
+    long latency = default_latency;
+    if (colon != NULL) {
+        latency = Arguments::parseUnits(colon + 1, NANOS);
+        if (latency == -1) {
```
Baraa-Hasheesh: let's make this check for any negative value

test/test/instrument/InstrumentTests.java
```diff
@@ -12,6 +12,7 @@
 import java.time.Duration;
```
Baraa-Hasheesh, line 12: I would suggest adding an additional test here for jfrsync

According to the current implementation it's possible to use the JDK25+ MethodTracing in the flight recorder in parallel with profiler MethodTracing
Can we add a test that collects samples from both profiler & flight recorder

I'm not asking to test flight recorder, I'm asking to make sure that our instrumentation doesn't remove theirs & vice versa

This wasn't a concern previously as we disabled the MethodTracing previously in jfrsync which is no longer the case

src/instrument.cpp
fandreuz: Because I know the length of both strings here

src/instrument.cpp
fandreuz: `-1` is the value returned by `parseUnits` when the string cannot be parsed

Baraa-Hasheesh: You can try `start,threads,trace=test.instrument.CpuBurner.burn:-100ms,collapsed,file=%f` 

It will hit the `(target.latency >= 0), function rewriteClass, file instrument.cpp, line 990.` inst

fandreuz: 265ae642bde905c77377080a3353ac5853098a83

src/instrument.cpp
```diff
@@ -437,7 +441,18 @@ Result BytecodeRewriter::rewriteCode(u16 access_flags, u16 descriptor_index) {
...
+        put8(JVM_OPC_invokestatic);
+        put16(_recordEntry_cpool_idx);
+        // nop ensures that tableswitch/lookupswitch needs no realignment
+        put8(JVM_OPC_nop);
+
+        // The rest of the code is unchanged
+        put(code, code_length);
+        for (u16 i = 0; i <= code_length; ++i) relocation_table[i] = EXTRA_BYTECODES_SIMPLE_ENTRY;
+    } else {
+        assert(latency >= 0);
```
Baraa-Hasheesh, line 465: This assert is redundant & can be safely removed

The new check that returns error for negative values which was added in latest commit already checks for this

src/instrument.cpp
```diff
@@ -1026,24 +1075,107 @@ Error Instrument::start(Arguments& args) {
...
 
-void Instrument::setupTargetClassAndMethod(const char* event) {
-    char* new_class = strdup(event);
-    *strrchr(new_class, '.') = 0;
+Error handleTarget(Targets& targets, const char* s, long default_latency) {
+    // Expected formats:
+    // - the.package.name.ClassName.MethodName
+    // - the.package.name.ClassName.MethodName+50ms
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V+50ms
```
Baraa-Hasheesh: Same here

src/instrument.cpp
```diff
@@ -1026,24 +1075,107 @@ Error Instrument::start(Arguments& args) {
...
+    _targets.clear();
 }
 
-void Instrument::setupTargetClassAndMethod(const char* event) {
-    char* new_class = strdup(event);
-    *strrchr(new_class, '.') = 0;
+Error handleTarget(Targets& targets, const char* s, long default_latency) {
+    // Expected formats:
+    // - the.package.name.ClassName.MethodName
+    // - the.package.name.ClassName.MethodName+50ms
```
Baraa-Hasheesh: Comment needs to be updated with latest syntax

src/instrument.cpp
```diff
@@ -1026,24 +1075,107 @@ Error Instrument::start(Arguments& args) {
...
+    }
+
+    // We need to know what's the next special symbol (either '(' or ':')
+    // to figure the method name.
+    std::string method_name;
+
+    const char* colon;
+
+    // Signature
+    const char* bracket = strchr(last_dot, '(');
```
Baraa-Hasheesh: What do you think about doing this & moving the checks below into this chain?
It would be easier to follow the cases if they are defined in a single location rather than 2

```
const char* bracket = strchr(last_dot, '(');
const char* colon = strchr(bracket ? bracket : last_dot, ':'):

if (bracket & colon)
else if (bracket)
else if (colon)
else
```

->I'm referring to the checks related to method_name & signature 
I think it's better to keep the latency parsing a separate check as you're doing now

src/instrument.cpp
```diff
@@ -1059,23 +1191,33 @@ void Instrument::retransformMatchedClasses(jvmtiEnv* jvmti) {
...
+
+        for (const auto& target : _targets) {
+            if (matchesPattern(signature + 1, len - 2, target.first)) {
+                jboolean modifiable;
+                if (
+                    target.first[target.first.length() - 1] != '*' ||
+                    // Some classes are not modifiable. With wildcard matching we skip
+                    // them quietly; when the class is specifically selected by the user
+                    // we let JVMTI fail loudly.
+                    (jvmti->IsModifiableClass(classes[i], &modifiable) == 0 && modifiable)
```
Baraa-Hasheesh, line 1213: What does "fail loudly" mean

Does it mean an error message, an assertion, a crash

If it means the user program would be killed I think it's better to just skip them even if specified & log an error message for the user

src/instrument.cpp
```diff
@@ -159,14 +173,19 @@ class BytecodeRewriter {
...
+    u16 _recordEntry_cpool_idx;
+    // one/profiler/Instrument.recordExit(JJ)V
+    u16 _recordExit_cpool_idx;
+    // one/profiler/Instrument.recordExit(J)V
+    u16 _recordExit_latency0_cpool_idx;
+    // java/lang/System.nanoTime()J
+    u16 _nanoTime_cpool_idx;
+
+    // Maps latency to the index in the constant pool
+    std::unordered_map<long, long> _latency_constant_idx;
```
apangin: Shall it be `<long, u16>`?

apangin: nit: `_latency_cpool_idx` for consistency with above variables.

src/instrument.cpp
```diff
@@ -919,25 +956,45 @@ Result BytecodeRewriter::rewriteClass() {
...
+    u16 new_cpool_len = _cpool_len + 19;
+    for (const auto& name_it : _target_methods) {
+        for (const MethodTarget& target : name_it.second) {
+            if (_latency_constant_idx.find(target.latency) != _latency_constant_idx.end()) continue;
+            if (target.latency == MethodTarget::NO_LATENCY) continue;
+            assert(target.latency >= 0);
+
+            putLongConstant(target.latency);
+            _latency_constant_idx[target.latency] = new_cpool_len;
+            new_cpool_len += 2;
```
apangin: Shorter and more accurate way to express the same:
```java
if (target.latency > 0 && _latency_cpool_idx[target.latency] == 0) {
    _latency_cpool_idx[target.latency] = new_cpool_len;
    putLongConstant(target.latency);
    new_cpool_len += 2;
}
```
Note that you don't need to put latency `0` in the constant pool.

src/instrument.cpp
apangin: `(` is usually called "parenthesis" (or "paren") in technical context - to avoid confusion with `[`.

src/instrument.cpp
```diff
@@ -1026,24 +1075,107 @@ Error Instrument::start(Arguments& args) {
...
+            method_name = std::string(last_dot + 1);
+        } else {
+            method_name = std::string(last_dot + 1, colon - last_dot - 1);
+        }
     }
 
-    char* old_class = _target_class;
-    _target_class = new_class;
-    free(old_class);
+    targets[std::move(class_name)][std::move(method_name)].emplace_back(std::move(signature), latency);
```
apangin: Implementation of this function looks quite intricate.
Wouldn't it be easier if you parse a string from the end? For example, like `getBreakpoint` does (breakpoint syntax is even more complex).

src/instrument.cpp
```diff
@@ -906,7 +942,8 @@ Result BytecodeRewriter::rewriteClass() {
     put32(version);
 
     _cpool_len = get16();
-    put16(_cpool_len + EXTRA_CONSTANTS);
+    u32 cpool_len_idx = _dst_len;
+    put16(_cpool_len); // to be patched later
```
apangin: If it is patched anyway, why bother writing something other than `0`?

src/flightRecorder.cpp
```diff
@@ -675,10 +675,8 @@ class Recording {
             writeIntSetting(buf, T_MONITOR_ENTER, "lock", args._lock);
         }
 
-        writeBoolSetting(buf, T_METHOD_TRACE, "enabled", args._latency >= 0);
-        if (args._latency >= 0) {
-            writeIntSetting(buf, T_METHOD_TRACE, "latency", args._latency);
-        }
+        writeBoolSetting(buf, T_METHOD_TRACE, "enabled", !args._trace.empty());
+        writeListSetting(buf, T_METHOD_TRACE, "targets", args._trace);
```
apangin: Should be "target", since one setting contains only one target.
But I'd better call it "trace" to reflect argument name, similarly to other settings.

src/instrument.cpp
```diff
@@ -1100,10 +1242,21 @@ void JNICALL Instrument::ClassFileLoadHook(jvmtiEnv* jvmti, JNIEnv* jni,
     // Do not retransform if the profiling has stopped
     if (!_running) return;
 
-    bool wildcard_class = _target_class[0] == '*' && strlen(_target_class) == 1;
-    if (name == NULL || wildcard_class || strcmp(name, _target_class) == 0) {
-        BytecodeRewriter rewriter(class_data, class_data_len, _target_class, _latency);
+    if (name == NULL) {
+        // Make sure the class is rewritten, constant pool reconstitution
+        // doesn't always work (JDK-8216547)
+        BytecodeRewriter rewriter(class_data, class_data_len, EMPTY_METHOD_TARGETS);
```
apangin: Does `EMPTY_METHOD_TARGETS` mean that no methods will be rewritten?

src/instrument.h
```diff
@@ -7,16 +7,44 @@
...
+        latency(latency) {}
+
+    MethodTarget(const MethodTarget&) = delete;
+    MethodTarget& operator=(const MethodTarget&) = delete;
+    MethodTarget(MethodTarget&& other)
+        : signature(std::move(other.signature)), latency(other.latency) {}
+};
+
+typedef std::map<MethodName, std::vector<MethodTarget>> MethodTargets;
+typedef std::map<ClassName, MethodTargets> Targets;
```
apangin: You still have a 3-level structure. I think an algorithm to build a list of targets and to perform lookup could be simpler if you chose a different structure. One obvious simplification is to flatten MethodTargets:
```
map<ClassName, vector<MethodTarget>>
```
where `MethodTarget` contains both name and signature. Most often, a vector will contain only 1 element anyway, so performance will not suffer.

Another alternative is
```
map<ClassName, map<Method, Latency>>
```
where `ClassName` and `Method` are strings, and `Latency` is `long`.
Ordered map allows to navigate directly to the required method. Search key is a method name concatenated with a signature. Found item "matches" search key iff the item is a full prefix of the key.

src/instrument.cpp
fandreuz: Yes, thanks: 61ac6aecc8e0065c4abd2782322aad0cc89d7e0a

src/instrument.cpp
fandreuz: 91bf880a83b22d010db1470d8ec2c13b644e6551

src/instrument.cpp
fandreuz: Yeah makes sense: 329867407c345c875371e9578b402a7e3e91d16d

src/flightRecorder.cpp
fandreuz: 06485f2d92e7d4c4240a61f2e6ca33c191753b07

src/instrument.cpp
fandreuz: Yes, it's defined as an empty map in `instrument.h`. Removing this branch causes segfaults in jdk8 which I think are related to the ticket in the comment.

src/instrument.cpp
fandreuz: Thanks, applied in 0aceb33c669fbd052aa6bd52bf44e17001184a81

> more accurate way

Why? Just for the case `latency=0`?

src/instrument.cpp
apangin: Right, but the feature will not work as expected in this case.
You still need to rewrite matched methods, even if the class name can be only obtained from the constant pool.

src/instrument.cpp
fandreuz: The check is in another part of the code, so I'd rather keep the assert. It expresses the preconditions I need to go through here.

src/instrument.cpp
fandreuz: cfbb396f7226d9e5c6b4a68c45c8631e7579a705

src/instrument.cpp
fandreuz: cfbb396f7226d9e5c6b4a68c45c8631e7579a705

src/instrument.h
fandreuz: Yeah indeed merging method name and signature resulted in lots of removed lines. Thanks: 82538c86dc531fe407be65129d0cce57c940b631

src/instrument.cpp
fandreuz: It's a bit simpler now that I don't need to account for signature separately than name

src/instrument.cpp
fandreuz: Fixed in d1ba93579389b0d1e4837e8419d4149bdc6c2963

src/instrument.cpp
fandreuz: Please check the logic. I call `RetransformClasses` and I check if an error happened. If yes, I log. If no, nothing happens.

src/instrument.cpp
fandreuz: Outdated, please reopen if this is still valid.

src/frameName.cpp
```diff
@@ -88,8 +88,8 @@ FrameName::FrameName(Arguments& args, int style, int epoch, Mutex& thread_names_
     // Require printf to use standard C format regardless of system locale
     _saved_locale = uselocale(newlocale(LC_NUMERIC_MASK, "C", (locale_t)0));
 
-    buildFilter(_include, args._buf, args._include);
-    buildFilter(_exclude, args._buf, args._exclude);
+    for (const char* s : args._include) _include.push_back(s);
```
Baraa-Hasheesh, line 91: why not use `_include.insert(_include.end(), args. _include.begin(), args._include.end())` here & below?

Baraa-Hasheesh:
@fandreuz I think the [commit](https://github.com/async-profiler/async-profiler/pull/1499/commits/12b4ce7e44a0a6265479cb12ba4974a74b930146) should be pushed into a separate PR 

it's failing on master & I think it's better to handle in a separate PR

fandreuz:
> @fandreuz I think the [commit](https://github.com/async-profiler/async-profiler/pull/1499/commits/12b4ce7e44a0a6265479cb12ba4974a74b930146) should be pushed into a separate PR
> 
> it's failing on master & I think it's better to handle in a separate PR

I disagree.

1. it's a test failure resulting from an unrealistic test environment (people should not use `*.*` as a selector)
2. It's not even a crash: it's a warning on stderr.

There's no rush in having it in `master`.

krk:
This PR is already very large and not easy to review as is. I think we should take every chance to refactor bits out that can be merged in smaller PRs.

> @fandreuz I think the [commit](https://github.com/async-profiler/async-profiler/pull/1499/commits/12b4ce7e44a0a6265479cb12ba4974a74b930146) should be pushed into a separate PR

Are there other candidates to refactor out into new smaller PRs, in addition to this?

fandreuz:
> This PR is already very large and not easy to review as is. I think we should take every chance to refactor bits out that can be merged in smaller PRs.
> 
> > @fandreuz I think the [commit](https://github.com/async-profiler/async-profiler/pull/1499/commits/12b4ce7e44a0a6265479cb12ba4974a74b930146) should be pushed into a separate PR
> 
> Are there other candidates to refactor out into new smaller PRs, in addition to this?

Not really, all these changes should go together as they don't really make sense without each other now. I proposed to split them [here](https://github.com/async-profiler/async-profiler/pull/1499#issuecomment-3347335519), but my comment did not receive any reaction.

I reverted 12b4ce7e44a0a6265479cb12ba4974a74b930146 in d49d41ac506c262a13a180dfe787235ee97a0666.

src/instrument.cpp
```diff
@@ -64,9 +78,14 @@ class Constant {
     }
 
     const char* utf8() const {
+        assert(_tag == JVM_CONSTANT_Utf8);
         return (const char*) (_info + 2);
     }
 
+    std::string as_string() const {
```
apangin: It does not follow general naming convention. Maybe `toString()` ?

src/instrument.cpp
```diff
@@ -97,14 +116,11 @@ class Constant {
...
-        return _tag == JVM_CONSTANT_Utf8 && info() == len && memcmp(utf8(), value, len) == 0;
+        return info() == len && memcmp(utf8(), value, len) == 0;
     }
 
-    bool matches(const char* value, u16 len) const {
-        if (len > 0 && value[len - 1] == '*') {
-            return _tag == JVM_CONSTANT_Utf8 && info() >= len - 1 && memcmp(utf8(), value, len - 1) == 0;
-        }
-        return equals(value, len);
+    bool matches(const std::string& pattern) const {
```
apangin: I can't find usages of this function.

src/instrument.cpp
```diff
@@ -159,14 +173,22 @@ class BytecodeRewriter {
...
+    u16 _nanoTime_cpool_idx;
+
+    // Maps latency to the index in the constant pool
+    std::unordered_map<u64, u16> _latency_cpool_idx;
 
-    long _latency;
+    // '_targets' will be used when the name of the class is not
+    // known beforehand. Only one shall be non-empty.
+    const Targets* _targets;
+    const MethodTargets* _target_methods;
```
apangin: I'm not sure why you need to differentiate these two cases. In the end, no matter where you get the class name from, the algorithm is the same.

src/instrument.cpp
```diff
@@ -283,41 +305,31 @@ class BytecodeRewriter {
 
     // BytecodeRewriter
 
-    Result rewriteCode(u16 access_flags, u16 descriptor_index);
-    Result rewriteCodeForLatency(const u8* code, u16 code_length, u8 start_time_loc_index, u16* relocation_table);
+    Result rewriteCode(u16 access_flags, u16 descriptor_index, long latency);
+    Result rewriteCodeForLatency(const u8* code, u16 code_length, u8 start_time_loc_index, u16* relocation_table, u64 latency);
```
apangin: Why is latency sometimes `long` and sometimes `u64`?

src/instrument.cpp
```diff
@@ -518,18 +527,20 @@ Result BytecodeRewriter::rewriteCodeForLatency(const u8* code, u16 code_length,
...
             put8(start_time_loc_index);
 
-            if (_latency > 0) {
+            if (latency > 0) {
                 put8(JVM_OPC_ldc2_w);
-                put16(_cpool_len + 16);
+
+                auto latency_it = _latency_cpool_idx.find(latency);
+                assert(latency_it != _latency_cpool_idx.end());
+                put16(latency_it->second);
```
apangin: Since C++ is a "write-only" language, the less C++ you write, the easier it is for everyone :)
```
put16(_latency_cpool_idx[latency]);
```
Assertion is not that much useful in this case to sacrifice brevity. I mean, in the worst case, class redefinition will fail anyway.

src/instrument.cpp
```diff
@@ -867,6 +893,31 @@ Result BytecodeRewriter::rewriteCodeAttributes(const u16* relocation_table, int
     return Result::OK;
 }
 
+static bool findLatency(const MethodTargets* target_methods, const std::string&& method_name,
+                        const std::string&& method_desc, long& latency) {
+    if (target_methods->empty()) return false;
```
apangin: Can we even get here if `target_methods` is empty?

docs/ProfilingModes.md
Baraa-Hasheesh: @fandreuz any update on this?

src/arguments.h
```diff
@@ -161,21 +166,19 @@ class Arguments {
...
-    static long parseUnits(const char* str, const Multiplier* multipliers);
     static int parseTimeout(const char* str);
 
   public:
     Action _action;
     Counter _counter;
     const char* _event;
+    std::vector<const char*> _trace;
     int _timeout;
     long _interval;
```
Baraa-Hasheesh, line 177: Raising this again as I have yet to receive a confirmation on the subject 
What's the plan for interval usage when both engines `cpu` & `trace` are selected 

Using the same in both engines isn't really user friendly nor usable 
`start,event=cpu,trace=class.method:10ms,interval=100ms`

This would result in `interval = 100 * 1000000`

This means we would only collect a single trace sample for every 100000000 invocations!!

Please either delete the interval logic from method tracing or declare a separate interval option for it

src/instrument.cpp
```diff
@@ -867,6 +893,31 @@ Result BytecodeRewriter::rewriteCodeAttributes(const u16* relocation_table, int
...
+                memcmp(method_name.c_str(), it->first.c_str(), it->first.length()) == 0)
+        ) {
+            latency = it->second;
+            return true;
+        }
+
+        if (it == target_methods->begin()) return false;
+        --it;
+    }
+}
```
apangin, line 920: Why is a loop? We should be able to find the correct entry with a single lookup.
With a small trick, we can even get rid of wildcard vs. full match logic and reduce all that to a single "starts-with" check.

| method argument | what we store in a tree |
|-----------------|-------------------------|
| read*           | read                    |
| read            | read(                   |
| readInt()I      | readInt()I              |

src/instrument.cpp
```diff
@@ -918,26 +972,61 @@ Result BytecodeRewriter::rewriteClass() {
     const u8* cpool_end = _src;
     put(cpool_start, cpool_end - cpool_start);
 
+    if (_target_methods == nullptr) {
+        for (int i = 1; i < _cpool_len; i += _cpool[i]->slots()) {
+            if (_cpool[i]->tag() != JVM_CONSTANT_Class) continue;
```
apangin: Um.. you look for the first random `Class` constant and treat it as a class name? This is not gonna work.
Why don't you just take `class_name_index`?

src/instrument.cpp
```diff
@@ -1026,24 +1107,80 @@ Error Instrument::start(Arguments& args) {
...
+    // - the.package.name.ClassName.MethodName:50ms
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V
+    // - the.package.name.ClassName.MethodName(Jjava/lang/String;)V:50ms
+
+    const char* last_dot = strrchr(s, '.');
+    if (last_dot == NULL) {
+        return Error("Unexpected format for tracing target");
+    }
+    size_t class_name_size = last_dot - s;
+    std::string class_name(s, class_name_size);
```
apangin:
```
std::string class_name(s, last_dot - s);
```

src/instrument.cpp
```diff
@@ -1026,24 +1107,80 @@ Error Instrument::start(Arguments& args) {
...
     retransformMatchedClasses(jvmti);  // undo transformation
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_CLASS_FILE_LOAD_HOOK, NULL);
+
+    _targets.clear();
 }
 
-void Instrument::setupTargetClassAndMethod(const char* event) {
-    char* new_class = strdup(event);
-    *strrchr(new_class, '.') = 0;
+Error handleTarget(Targets& targets, const char* s, long default_latency) {
```
apangin: "handle"? Maybe "addTarget"?

src/instrument.cpp
```diff
@@ -1100,10 +1247,21 @@ void JNICALL Instrument::ClassFileLoadHook(jvmtiEnv* jvmti, JNIEnv* jni,
     // Do not retransform if the profiling has stopped
     if (!_running) return;
 
-    bool wildcard_class = _target_class[0] == '*' && strlen(_target_class) == 1;
-    if (name == NULL || wildcard_class || strcmp(name, _target_class) == 0) {
-        BytecodeRewriter rewriter(class_data, class_data_len, _target_class, _latency);
+    if (name == NULL) {
+        // Make sure the class is rewritten, constant pool reconstitution
+        // doesn't always work (JDK-8216547)
```
apangin: How is JDK-8216547 related here?

src/instrument.cpp
```diff
@@ -918,26 +972,61 @@ Result BytecodeRewriter::rewriteClass() {
...
     put(cpool_start, cpool_end - cpool_start);
 
+    if (_target_methods == nullptr) {
+        for (int i = 1; i < _cpool_len; i += _cpool[i]->slots()) {
+            if (_cpool[i]->tag() != JVM_CONSTANT_Class) continue;
+
+            u16 name_idx = _cpool[i]->info();
+            const auto& it = _targets->find(_cpool[name_idx]->as_string());
+            if (it == _targets->end()) {
+                _target_methods = &EMPTY_METHOD_TARGETS;
```
apangin: `EMPTY_METHOD_TARGETS` does not make sense to me.
If there is nothing to rewrite, return immediately.
If returning NULL from `ClassFileLoadHook` does not work for some reason (I still don't understand why), you can just copy `_src` to `_dst`, and that's it.

src/instrument.cpp
```diff
@@ -918,26 +972,61 @@ Result BytecodeRewriter::rewriteClass() {
     const u8* cpool_end = _src;
     put(cpool_start, cpool_end - cpool_start);
 
+    if (_target_methods == nullptr) {
+        for (int i = 1; i < _cpool_len; i += _cpool[i]->slots()) {
+            if (_cpool[i]->tag() != JVM_CONSTANT_Class) continue;
+
+            u16 name_idx = _cpool[i]->info();
+            const auto& it = _targets->find(_cpool[name_idx]->as_string());
```
apangin: This lookup logic differs from the lookup logic in `ClassFileLoadHook`.

test/test/instrument/InstrumentTests.java
```diff
@@ -236,12 +449,31 @@ public void recursive(TestProcess p) throws Exception {
         assert out.samples(MAIN_METHOD_SEGMENT + RECURSIVE_METHOD_SEGMENT + " ") >= duration.toNanos();
     }
 
+    @Test(
+        mainClass = MethodTracingStop.class,
+        jvmArgs   = "-Djava.library.path=build/lib -Xverify:all -XX:+IgnoreUnrecognizedVMOptions --enable-native-access=ALL-UNNAMED",
```
apangin, line 454: JVM warnings should be fixed in a centralized way. Not in this PR, though.

src/instrument.cpp
fandreuz: When I have `if (name == NULL) return` in the `ClassFileLoadHook` i get a segfault with this stacktrace:
```
V  [libjvm.so+0xa9a409]  Symbol::as_C_string() const+0x9
V  [libjvm.so+0x4e6cd1]  ConstantPool::copy_cpool_bytes(int, SymbolHashMap*, unsigned char*)+0x191
V  [libjvm.so+0x75d83c]  JvmtiClassFileReconstituter::write_class_file_format()+0x2cc
V  [libjvm.so+0x7b4f5f]  JvmtiEnv::RetransformClasses(int, _jclass* const*)+0x78f
V  [libjvm.so+0x7600a6]  jvmti_RetransformClasses+0x136
C  [libasyncProfiler.so+0xea53]  VM::RetransformClassesHook(_jvmtiEnv*, int, _jclass* const*)+0x23
C  [libasyncProfiler.so+0x1ea6d]  Instrument::retransformMatchedClasses(_jvmtiEnv*) [clone .constprop.0]+0xcd
C  [libasyncProfiler.so+0x1ecbf]  Instrument::stop()+0x2f
C  [libasyncProfiler.so+0x4b61d]  Profiler::stop(bool)+0xbd
C  [libasyncProfiler.so+0x5468f]  Profiler::runInternal(Arguments&, Writer&)+0x14f
C  [libasyncProfiler.so+0x55326]  Profiler::run(Arguments&)+0xc6
C  [libasyncProfiler.so+0x5605e]  VM::VMDeath(_jvmtiEnv*, JNIEnv_*)+0x4e
V  [libjvm.so+0x7c1b69]  JvmtiExport::post_vm_death()+0x1f9
V  [libjvm.so+0x69dfbf]  before_exit(JavaThread*)+0x12f
V  [libjvm.so+0xae450b]  Threads::destroy_vm()+0x34b
V  [libjvm.so+0x7110cf]  jni_DestroyJavaVM+0x7f
C  [libjli.so+0x8618]  JavaMain+0x5c8
C  [libc.so.6+0x891f5]
```

My impression is that the class data is corrupted somehow and needs to be copied from the original version, which is more or less what is happening in the ticket. I could investigate more if you want to understand the root cause, but it only affects jdk8 and I figured it's not a priority.

src/instrument.cpp
fandreuz: `-1` means no latency profiling, which is the old instrumentation profiling we had before #1421.

src/instrument.cpp
fandreuz: Yes, when we receive `EMPTY_METHOD_TARGET`. Let's discuss that below, I'll remove this once we get rid of `EMPTY_METHOD_TARGET`.

src/instrument.cpp
fandreuz: 2542eb5c98b00c86733af4ed91a2cdf0f5540692

src/instrument.cpp
fandreuz: 2542eb5c98b00c86733af4ed91a2cdf0f5540692

src/instrument.cpp
fandreuz: 2542eb5c98b00c86733af4ed91a2cdf0f5540692

src/instrument.cpp
fandreuz: Leftover, 2542eb5c98b00c86733af4ed91a2cdf0f5540692

src/instrument.cpp
fandreuz: 2542eb5c98b00c86733af4ed91a2cdf0f5540692

src/instrument.cpp
fandreuz: This works fine:
```
if (name == NULL) {
    VM::jvmti()->Allocate(class_data_len, new_class_data);
    memcpy(*new_class_data, class_data, class_data_len);
    *new_class_data_len = class_data_len;
    return;
}
```
but does not support the request you had the other day, that we should support the feature even when `name == NULL` by getting the name from `class_data`.

src/instrument.cpp
Baraa-Hasheesh: I was more concerned about the fact that this is using a sorted map so it's technically possible to match a different target depending if you start from the start or the end 

`trace=class.method*:1000ms,trace=class.methodInt*:100ms,trace=class.methodIntEnhanced*:20ms`

I can potentially see this matching the method with the incorrect latency

src/instrument.cpp
```diff
@@ -867,6 +886,31 @@ Result BytecodeRewriter::rewriteCodeAttributes(const u16* relocation_table, int
     return Result::OK;
 }
 
+static bool findLatency(const MethodTargets* target_methods, const std::string&& method_name,
+                        const std::string&& method_desc, long& latency) {
+    if (target_methods->empty()) return false;
+
+    const std::string method = method_name + method_desc;
+    auto it = target_methods->lower_bound(method);
```
Baraa-Hasheesh: I don't like this usage of `lower_bound` 
It makes the code nearly unreadable for me,

I understand that you're relaying on the sorting of the `map` structure but that is not obvious for one at a first glance

It will take a developer a long time to understand how the matching is done & why is `lower_bound` is even needed & how it contributes to the matching algorithm

fandreuz: I don't understand. Are you saying we should not use advanced features provided by a data structure because they are not obvious enough? I strongly disagree.

https://en.cppreference.com/w/cpp/container/map/lower_bound.html

docs/ProfilingModes.md
fandreuz: PR is not merged, so blog post is not updated. Also, not relevant for the PR, closing.

src/arguments.h
fandreuz: I agree on the usability part, but I've already received complaints that the scope of this PR is too large. I'd rather tackle it in a separate PR, since as you said there wasn't much engagement on your previous comment.

src/instrument.cpp
Baraa-Hasheesh: The code works & the advance feature is indeed useful, but I think it should be a cost benefit, 

In this case you will require multiple mental steps to understand what `lower_bound` even does & how it contributes to the matching algorithm being used here

Also I doubt anyone can do key comparison of strings mentally, I had to double check that the matching algorithm will behave as I expect it to in regards to wildcard characters for example

again this is my personal opinion but the code obfuscation this advance feature does is doing more harm than good & I would vote to remove it,

I would add @apangin for his opinion here

fandreuz: https://github.com/async-profiler/async-profiler/pull/1499#discussion_r2430445309

src/instrument.cpp
fandreuz: > If returning NULL from ClassFileLoadHook does not work for some reason (I still don't understand why),

See [here](https://github.com/async-profiler/async-profiler/pull/1499#discussion_r2431607679)

fandreuz: > just copy _src to _dst, and that's it.

Check `Result::ABORTED`: 5696577a796f340d9ba2ce6ddd8a3f6bf12986c7

src/instrument.cpp
fandreuz: Fixed in 5696577a796f340d9ba2ce6ddd8a3f6bf12986c7

src/instrument.cpp
fandreuz: Fixed in 5696577a796f340d9ba2ce6ddd8a3f6bf12986c7

src/instrument.cpp
fandreuz: Fixed in 5696577a796f340d9ba2ce6ddd8a3f6bf12986c7. I had to keep the matching both in `ClassFileLoadHook` and `rewriteClass`: we probably don't want to pass all classes in `ClassFileLoadHook` .

src/instrument.cpp
fandreuz: Are you sure about this? What if we have the following:
```
tree = {re*, read2}
argument = "read1"
```
`map::lower_bound` will return `read2`, which won't match the method.

src/frameName.cpp
fandreuz: a85ddd5fe15905318658664aa79d199327174e38

apangin: A like previous version better: it was shorter and clearer, and it was explicit about _adding_ elements rather than _inserting_. Batch version would make sense if the elements were copied verbatim, but in this case, each element is translated.
It's of course a subjective judgment, and that's why I asked you many times _not_ to focus on coding style and other preference based things in your reviews. Leave these judments to me.

fandreuz: Reverted in 0a65747e1cd5b19042c01a11ff24d66bd1cf609d

src/instrument.cpp
apangin: Yes, I am. I did not mean `lower_bound` though. What you need is something like `TreeMap.floorEntry` in Java, but it seems there is no direct equivalent in C++. `upper_bound` + one step back can be a replacement.

src/instrument.cpp
apangin: This is again an opinion-based discussion. In fact, I can agree with both points. It can be an efficient algorithm that requires a bit more mental efforts, or it can be a simple straightforward loop iteration, given that it's not a performance sensitive code. But in either case, the code should be concise and _nice_. Current implementation does not satisfy this requirement :)

src/arguments.h
apangin: Sure, let's leave per-method intervals for a future PR. However, where I agree with @Baraa-Hasheesh, is that one `interval` argument should not take effect on both CPU profiling and tracing _at the same time_.
I propose to use `interval` only when tracer is run exclusively, that is, when `(_event == NULL) ^ _trace.empty()`

src/instrument.cpp
apangin: Yes, but why `u64` in another case then? The only place I know is where you put a constant in a constant pool.

src/arguments.h
Baraa-Hasheesh: I was thinking more of a global interval for all methods, along the lines of 

`start,event=cpu,interval=100ms,trace=class1.method1:10ms,trace=class1.method2:20ms,trace-interval=100`

I don't really see the value or the use case of a per method interval, a global interval as we do on other engines should be good enough for most use cases.

=> I think it's better to keep it simple & avoid the complexity a per method interval might introduce 

What do you think @apangin

src/instrument.cpp
fandreuz: In `rewriteCodeForLatency` we know `latency` cannot be `-1`, so it made sense to me to make this knowledge explicit in the method signature.

src/arguments.h
apangin: Not in this PR. We need to merge it today.

src/instrument.cpp
fandreuz: Hmm how about this?
```
tree = {re*, read1, read2}
argument = "readd"
```
`upper_bound` will return `iterator::end` since there's no `key > readd`. But I guess we want to match with `re*`.

src/arguments.h
fandreuz: cccbd14bd36f91a41471f09f49fb1407dca1d30f

src/instrument.cpp
```diff
@@ -919,43 +973,70 @@ Result BytecodeRewriter::rewriteClass() {
...
-        return Result::CLASS_DOES_NOT_MATCH;
-    }
     // We should not instrument classes from one.profiler.* to avoid infinite recursion
     if (_class_name->info() >= PROFILER_PACKAGE_LEN &&
-        strncmp(_class_name->utf8(), PROFILER_PACKAGE, PROFILER_PACKAGE_LEN) == 0) {
+    strncmp(_class_name->utf8(), PROFILER_PACKAGE, PROFILER_PACKAGE_LEN) == 0) {
         return Result::PROFILER_CLASS;
     }
 
+    for (const auto& target : *_targets) {
```
fandreuz: This is kind of redundant, redoing the search is the alternative to taking a `const MethodTargets*` in the constructor. Let me know which one you prefer @apangin 

We cannot remove the logic from `ClassFileLoadHook` because it would be inefficient

src/instrument.cpp
```diff
@@ -146,6 +158,9 @@ enum PatchConstants {
 
 class BytecodeRewriter {
   private:
+    const u8* _class_data;
+    const jint _class_data_len;
```
apangin: These don't need to be fields, since they are only needed in the scope of `rewrite` function.
But I hope they will go away completely if shutdown fix helps.

src/instrument.cpp
apangin: I'm not convinced, sorry. It is confusing to see two sister methods with the same `latency` argument but with a different type. Furthermore, you already have `Latency` typedef, why not use that?

src/instrument.cpp
apangin: You're right, sorry for confusion.

src/instrument.cpp
apangin: OK, let's have it in the constructor, especially since you still have both fields.
But I'd suggest to extract this lookup in a separate function to reduce code duplication and to ensure it's in sync with ClassFileLoadHook.

src/instrument.cpp
fandreuz: I didn't want to use `Latency` exactly because of this change in signature from one method to another, but I see how it can be confusing. I'm just going to use `Latency`

fandreuz: be3cd3655ad8c35c59dc2f13f25c0d122cdce2a0

src/instrument.cpp
fandreuz: I'll remove them once we merge #1538

fandreuz: 7d6e477b6f93136713340f2c184244a7a7ab3dbe

src/instrument.cpp
fandreuz: 99bc73433e0fb25d09843fdfcbfd78f1482a5903

PR #1500, opened 2025-09-21
Detect if calloc calls malloc

apangin:
### Description

Use temporary malloc hook to detect if `calloc` implementation calls `malloc` under the hood. This is required for proper `nativemem` profiling.

### Related issues

#1466

### Motivation and context

Various malloc implementations may or may not do cross-calls. For example, default `calloc` in musl libc calls `malloc` internally. Async-profiler handles this case specially to prevent double-counting of native memory alocations. Previously, it simply checked `OS::isMusl`, but this was inaccurate, since musl libc could be built with different malloc libraries. For example, in Alpaquita Linux, musl is linked with mimalloc, where `calloc` does not calls `malloc`.

### How has this been tested?

`nativemem` tests now pass on Alpaquita Linux as well as on vanilla Alpine.

---

### DIFF
--- a/src/mallocTracer.cpp
+++ b/src/mallocTracer.cpp
@@ -106,25 +106,25 @@ bool MallocTracer::_initialized = false;
 volatile bool MallocTracer::_running = false;
 
 static pthread_t _current_thread;
-static bool _recursive_malloc = false;
+static bool _nested_malloc = false;
 
 // Test if calloc() implementation calls malloc()
-static void* recursive_malloc_hook(size_t size) {
+static void* nested_malloc_hook(size_t size) {
     if (pthread_self() == _current_thread) {
-        _recursive_malloc = true;
+        _nested_malloc = true;
     }
     return _orig_malloc(size);
 }
 
 // In some implementations, specifically on musl, calloc() calls malloc() internally,
 // and posix_memalign() calls aligned_alloc(). Detect such cases to prevent double-accounting.
-static void detectRecursiveMalloc() {
+static void detectNestedMalloc() {
     CodeCache* libc = Profiler::instance()->findLibraryByAddress((void*)_orig_calloc);
     if (libc == NULL) {
         return;
     }
 
-    libc->patchImport(im_malloc, (void*)recursive_malloc_hook);
+    libc->patchImport(im_malloc, (void*)nested_malloc_hook);
 
     _current_thread = pthread_self();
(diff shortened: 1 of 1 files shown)


Baraa-Hasheesh:
While thinking about this issue, the solution that made most sense to me was something along the lines of what is implented here

However I have some doubts
It's assumeed that if `calloc` calls `malloc`, then the  `posix_memalign` would also call `aligned_alloc` 
I think it's better to check for that separately, as user might change `calloc` via LD_PRELOAD but keep `posix_memalign` the same 

On a similar note, do we want to support other patterns the user might introduce via an LD_PRELOAD?  (For example a user might change malloc to call calloc)

apangin:
@Baraa-Hasheesh Checking only one type of relation (calloc calling malloc) was a conscious choice. This relation is a distinctive characteristic of musl's own memory allocator. Other popular allocators (glibc's malloc, jemalloc, tcmalloc, mimalloc) do not do cross-calls via PLT.

Checking relation between posix_memalign and aligned_alloc would require writing more code in vain (with no effect for the above allocators). At the same time, it would not be a complete solution for a custom user-provided allocator anyway.

Support for custom allocators that hook arbitrary `malloc` functions is out of scope of `nativemem` feature. Doing LD_PRELOAD hacks you've mentioned while running `nativemem` profiling sounds extraordinary. I think it is unreasonable to build a complicated solution for a theoretical problem that noone have probably seen.

src/mallocTracer.cpp
```diff
@@ -104,6 +105,32 @@ int MallocTracer::_patched_libs = 0;
 bool MallocTracer::_initialized = false;
 volatile bool MallocTracer::_running = false;
 
+static pthread_t _current_thread;
+static bool _recursive_malloc = false;
+
+// Test if calloc() implementation calls malloc()
+static void* recursive_malloc_hook(size_t size) {
```
krk: Not sure if "recursive" is the right naming, as function is not calling itself.

apangin: Yeah, I had doubts about that too, but decided it's still recursive in the sense the malloc algorithm (as the whole) calls to itself via PLT. Do you have a suggestion how to name it better?

apangin: Renamed to `nested`

krk: "Nested" sounds good, as we are calling all malloc-family functions malloc here.

PR #1502, opened 2025-09-23
Fix integration test dependencies to build and cosmetic changes

krk:
Before this PR, all integration tests depended on all builds, causing e.g. linux-x64 integration tests to wait for macos build.

This is now fixed with some cosmetic changes to make the UI look like attached:

image

### Previous look and structure

image

### Motivation and context
Integration tests should not wait for builds on unrelated OS/archs.

### How has this been tested?
By the current PR github action.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,8 +17,8 @@ on:
         description: "Container image to use"
 
 env:
-  java_default_distribution: corretto
-  java_default_version: 11
+  build_java_distribution: corretto
+  build_java_version: 11
 
 jobs:
   build:
@@ -33,8 +33,8 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
-          distribution: ${{ env.java_default_distribution }}
-          java-version: ${{ env.java_default_version }}
+          distribution: ${{ env.build_java_distribution }}
+          java-version: ${{ env.build_java_version }}
       - name: Checkout sources
         uses: actions/checkout@v4
       - name: Build and unit test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,15 +6,12 @@ on:
       platform:
         type: string
         required: true
-        description: "Platform identifier (linux-x64, linux-arm64, macos)"
       runner:
         type: string
         required: true
-        description: "GitHub runner to use"
       container-image:
         type: string
         required: false
-        description: "Container image to use"
 
 env:
   build_java_distribution: corretto
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,83 @@
+name: build-template
+
+on:
+  workflow_call:
+    inputs:
+      platform:
+        type: string
+        required: true
+        description: "Platform identifier (linux-x64, linux-arm64, macos)"
+      runner:
+        type: string
+        required: true
+        description: "GitHub runner to use"
+      container-image:
+        type: string
+        required: false
+        description: "Container image to use"
+
+env:
+  java_default_distribution: corretto
+  java_default_version: 11
+
+jobs:
+  build:
+    runs-on: ${{ inputs.runner }}
+    container:
+      image: ${{ inputs.container-image }}
+    name: "build and unit test (${{ inputs.platform }})"
+    steps:
+      - name: Run container setup
+        if: inputs.container-image != ''
+        run: "[ ! -f /root/setup.sh ] || /root/setup.sh"
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: ${{ env.java_default_distribution }}
+          java-version: ${{ env.java_default_version }}
+      - name: Checkout sources
+        uses: actions/checkout@v4
+      - name: Build and unit test
+        id: build
+        run: |
+          set -x
+          HASH=${GITHUB_SHA:0:7}
+          case "${{ inputs.platform }}" in
+            macos*)
+              brew install gcovr
+              make COMMIT_TAG=$HASH FAT_BINARY=true release coverage -j
+            ;;
+            *)
+              make COMMIT_TAG=$HASH CC=/usr/local/musl/bin/musl-gcc release coverage -j
+              echo "debug_archive=$(find . -type f -name "async-profiler-*-debug*" -exec basename {} \;)" >> $GITHUB_OUTPUT
+            ;;
+          esac
+          echo "archive=$(find . -type f -name "async-profiler-*" -not -name "*-debug*" -exec basename {} \;)" >> $GITHUB_OUTPUT
+        shell: bash
+        env:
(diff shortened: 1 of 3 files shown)


Baraa-Hasheesh:
This looks nice,

But it might introduce some undesired behaviour
For example if you break unit tests on macOS, the Linux integration tests would still run

https://github.com/Baraa-Hasheesh/async-profiler/actions/runs/17956941491

.github/workflows/build.yml
```diff
@@ -0,0 +1,83 @@
+name: build-template
+
+on:
+  workflow_call:
+    inputs:
+      platform:
+        type: string
+        required: true
+        description: "Platform identifier (linux-x64, linux-arm64, macos)"
```
apangin: What are these definitions for? Are they even visible anywhere?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -27,231 +27,107 @@ jobs:
           name: async-profiler-jars
           path: build/jar/*
           if-no-files-found: error
-  build-and-upload-binaries:
+  build-linux-arm64:
+    name: build / linux-arm64
+    uses: ./.github/workflows/build.yml
```
apangin, line 29: Why do all paths start with `./`? Doesn't it work without it?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -27,231 +27,107 @@ jobs:
...
-            hs_err*.log
+    uses: ./.github/workflows/integ.yml
+    with:
+      platform: macos
+      test-platform: ${{ matrix.test-platform }}
+      runner: ${{ matrix.runner }}
+      java-version: ${{ matrix.java-version }}
+      architecture: ${{ matrix.architecture || '' }}
+
+  integ-containers:
```
apangin: What do you mean by "containers" here?
Why is it separate from `integ-linux-x64` and `integ-linux-arm64` that also run in containers?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -27,231 +27,107 @@ jobs:
...
+    with:
+      platform: linux-x64
+      test-platform: linux-x64
+      runner: ubuntu-latest
+      container-image: public.ecr.aws/async-profiler/asprof-builder-x86:latest
+      java-version: ${{ matrix.java-version }}
+
+  integ-linux-arm64:
+    name: integration test / linux-arm64
+    needs: [build-jars, build-linux-arm64]
```
apangin: Integration tests previously did not depend on `build-jars`? What has changed?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -27,231 +27,107 @@ jobs:
...
+    needs: [build-jars, build-linux-x64]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test-platform: linux-x64-alpine
+            container-image: public.ecr.aws/async-profiler/asprof-builder-alpine:corretto-11
+          - test-platform: linux-x64-AL2
+            container-image: public.ecr.aws/async-profiler/asprof-builder-amazonlinux:2
+            container-volumes: '["/tmp/node20:/__e/node20"]'
```
fandreuz: There used to be a comment here to explain why this cryptic config is needed. Can we keep it?

https://github.com/async-profiler/async-profiler/blob/d97a7d334331383c40a062572fa683d46c954492/.github/workflows/test-and-publish-nightly.yml#L163-L166

.github/workflows/integ.yml
```diff
@@ -0,0 +1,115 @@
...
+      options: --privileged
+      volumes: ${{ fromJSON(inputs.container-volumes || '[]') }}
+    name: "${{ inputs.test-platform }}, ${{ inputs.java-distribution }} ${{ inputs.java-version }}"
+    steps:
+      - name: Run container setup
+        if: inputs.container-image != ''
+        run: "[ ! -f /root/setup.sh ] || /root/setup.sh"
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        if: ${{ !contains(inputs.test-platform, 'alpine') }}
```
fandreuz, line 47: Can we keep the comment here to explain why Alpine is excluded?

https://github.com/async-profiler/async-profiler/blob/d97a7d334331383c40a062572fa683d46c954492/.github/workflows/test-and-publish-nightly.yml#L186-L187

.github/workflows/build.yml
krk: These are for the developer, not visible in UI afaik.

.github/workflows/test-and-publish-nightly.yml
krk: I meant "non-default", considering `public.ecr.aws/async-profiler/asprof-builder-x86` and `public.ecr.aws/async-profiler/asprof-builder-arm` default. Will find a better name.

.github/workflows/test-and-publish-nightly.yml
krk: Good catch, removing.

.github/workflows/test-and-publish-nightly.yml
krk: Doesn't work without.

krk:
>  For example if you break unit tests on macOS, the Linux integration tests would still run

This is a feature. No point in cancelling linux tests when macos tests are failing. Linux tests may fail or pass and that could give the developer an idea of what may be wrong.

apangin:
On your screenshots, previous workflow ran for 4:41, whereas the new "optimized" one ran for more than a minute longer. Why is that?

.github/workflows/test-and-publish-nightly.yml
krk: Merged in to `integ-linux-x64`.

krk:
> On your screenshots, previous workflow ran for 4:41, whereas the new "optimized" one ran for more than a minute longer. Why is that?

It may have to do with waiting for a GHA runner, as I was triggering runs in quick succession before I took the screenshot.

.github/workflows/test-and-publish-nightly.yml
apangin: There is nothing special about these containers. The only difference is that `integ-linux-*` run in the same containers that are used for building artifacts, but this does not really matter. My points is, division by "default" and "non-default" is not justitied, let's keep it uniform.

.github/workflows/build.yml
apangin: For developers, you mean, for me and you? :) But this looks like some formal API description, not quite human-friendly. In other words, it's just too much boilerplate to explain trivial things.

.github/workflows/test-and-publish-nightly.yml
krk: Yes, removed the "default" vs. "non-default" separation.

apangin:
@krk Can job titles be shortened? This is unreadable:

image

PR #1503, opened 2025-09-23
Converter can skip logging to system out with quiet flag set.

avrecko:
### Description
Added quiet flag to converter.

### Motivation and context
Using converter inside a larger application. It is polluting the log system with `Converting ... -> ...` messages to system out. It is very annoying to see. 

### How has this been tested?
Ran it with and without quiet flag.

---

### DIFF
--- a/src/converter/Main.java
+++ b/src/converter/Main.java
@@ -39,14 +39,18 @@ public class Main {
             String input = args.files.get(i);
             String output = isDirectory ? new File(lastFile, replaceExt(input, args.output)).getPath() : lastFile;
 
-            System.out.print("Converting " + getFileName(input) + " -> " + getFileName(output) + " ");
-            System.out.flush();
+            if (!args.quiet) {
+                System.out.print("Converting " + getFileName(input) + " -> " + getFileName(output) + " ");
+                System.out.flush();
+            }
 
             long startTime = System.nanoTime();
             convert(input, output, args);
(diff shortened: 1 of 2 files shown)


apangin:
`Main.main()` method is intended for CLI use.
If you embed converter in a larger application, why don't you call `convert()` directly?

avrecko:
> `Main.main()` method is intended for CLI use. If you embed converter in a larger application, why don't you call `convert()` directly?

Great point. I'll change my code to call convert.

I started using main in the previous version (v3.0) as it automatically selected the appropriate converter. Wasn't paying attention that convert now does the selection.

This change might nevertheless be helpful to somebody who does CLI from bash scripts or similar.

Feel free to close this pull request.

apangin:
> This change might nevertheless be helpful to somebody who does CLI from bash scripts or similar.

I thought about scripts too, but in that case, status messages can be easily suppressed with `> /dev/null`.

If you wish to add an option anyway, usage and documentation also need to be updated.

avrecko:
> > This change might nevertheless be helpful to somebody who does CLI from bash scripts or similar.
> 
> I thought about scripts too, but in that case, status messages can be easily suppressed with `> /dev/null`.
> 
> If you wish to add an option anyway, usage and documentation also need to be updated.

Great point. Closing this. With your explanations I think the extra lines of code don't add any added value.

PR #1504, opened 2025-09-23
Don't display kernel warning for non-root users when quiet flag is set.

avrecko:
### Description
Don't print this warning if quiet flag is set:

```
Log::warn("Kernel symbols are unavailable due to restrictions. Try\n"
                  "  sysctl kernel.perf_event_paranoid=1\n"
                  "  sysctl kernel.kptr_restrict=0");
```

### Motivation and context
Using async profiler inside another application. It is distracting to see the warning in the logs with quiet flag set.

### How has this been tested?

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -821,7 +821,7 @@ Error PerfEvents::start(Arguments& args) {
 
     _alluser = args._alluser;
     _kernel_stack = !_alluser && _cstack != CSTACK_NO;
-    if (_kernel_stack && !Symbols::haveKernelSymbols()) {
+    if (!args._quiet && _kernel_stack && !Symbols::haveKernelSymbols()) {
         Log::warn("Kernel symbols are unavailable due to restrictions. Try\n"
                   "  sysctl kernel.perf_event_paranoid=1\n"
                   "  sysctl kernel.kptr_restrict=0");


src/perfEvents_linux.cpp
```diff
@@ -821,7 +821,7 @@ Error PerfEvents::start(Arguments& args) {
 
     _alluser = args._alluser;
     _kernel_stack = !_alluser && _cstack != CSTACK_NO;
-    if (_kernel_stack && !Symbols::haveKernelSymbols()) {
+    if (!args._quiet && _kernel_stack && !Symbols::haveKernelSymbols()) {
         Log::warn("Kernel symbols are unavailable due to restrictions. Try\n"
                   "  sysctl kernel.perf_event_paranoid=1\n"
                   "  sysctl kernel.kptr_restrict=0");
```
krk, line 827: It would be better if we can wrap the log with a new `if (!args._quiet)`. We still need the rest of this block.

src/perfEvents_linux.cpp
```diff
@@ -821,7 +821,7 @@ Error PerfEvents::start(Arguments& args) {
 
     _alluser = args._alluser;
     _kernel_stack = !_alluser && _cstack != CSTACK_NO;
-    if (_kernel_stack && !Symbols::haveKernelSymbols()) {
+    if (!args._quiet && _kernel_stack && !Symbols::haveKernelSymbols()) {
```
apangin, line 824: This is not the right fix for the problem you've mentioned, because:
1. `quiet` option serves a different purpose, it's not intended to suppress logs;
2. there are many more places with `Log::warn`, it's not clear why this one should be handled specially.

If you wish to suppress async-profiler warnings, the standard way of doing so is to pass `-L`/`loglevel` option.

But this warning is really useful. It means that you've chosen to profile kernel stacks but kernel symbols are disabled on the OS level. If you do not intend to profile kernel stacks, use `ctimer` mode instead.

avrecko: > This is not the right fix for the problem you've mentioned, because:
> 
> 1. `quiet` option serves a different purpose, it's not intended to suppress logs;
> 2. there are many more places with `Log::warn`, it's not clear why this one should be handled specially.
> 
> If you wish to suppress async-profiler warnings, the standard way of doing so is to pass `-L`/`loglevel` option.
> 
> But this warning is really useful. It means that you've chosen to profile kernel stacks but kernel symbols are disabled on the OS level. If you do not intend to profile kernel stacks, use `ctimer` mode instead.

Makes sense. What are your thoughts of using cstack=no instead?

apangin: `cstack=no` will disable all native frames completely, leaving out GC and Compiler stacks, VM internals and all JNI code.

Ability to show native frames is one of the async-profiler's distinguishing features, why hide this important piece of information? If warning is your only concern, just add `loglevel=error`.

avrecko:
@apangin Thank you for the explanations. Closing this.

PR #1507, opened 2025-09-24
Clean up `EventMask` usages

fandreuz:
### Description
- `EventMask` correspondence between `Profiler` class and `flightRecorder` is now explicit
- Minor removal of duplicated code
- `main.cpp` was missing handling of `--jfropts`
- I introduced some constants instead of magic numbers to improve clarity and ease finding related pieces of code with global search
- Added a new `jfrsync` smoke test

### Related issues
#1499

### Motivation and context
This settles some discussion which originated under #1499.

### How has this been tested?
GHA

---

### DIFF
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -7,7 +7,6 @@
 #define _ARGUMENTS_H
 
 #include <stddef.h>
-#include "arch.h"
 
 
 const long DEFAULT_INTERVAL = 10000000;      // 10 ms
@@ -103,7 +102,7 @@ enum EventMask {
     EM_NATIVEMEM    = 16,
     EM_METHOD_TRACE = 32
 };
-constexpr u8 EVENT_MASK_SIZE = 6;
+constexpr int EVENT_MASK_SIZE = 6;
 
 struct StackWalkFeatures {
     // Stack recovery techniques used to workaround AsyncGetCallTrace flaws


test/test/jfr/JfrTests.java
```diff
@@ -170,6 +170,12 @@ public void allModeEventOverride(TestProcess p) throws Exception {
         assert events.contains("profiler.Free"); // nativemem profiling
     }
 
+    // Simple smoke test, nothing in particular is tested
+    @Test(mainClass = JfrCpuProfiling.class)
+    public void jfrSyncSmoke(TestProcess p) throws Exception {
+        p.profile("-d 3 --jfrsync default --jfropts 4 -f %f.jfr");
```
Baraa-Hasheesh: if this is a smoke test you don't need to do profiling for 3 seconds
1 should be enough no?

Also can we add an assert that the output file is created & not empty

src/profiler.h
```diff
@@ -47,6 +47,24 @@ enum State {
...
 
+enum EventMask {
+    EM_CPU          = 1,
+    EM_ALLOC        = 2,
+    EM_LOCK         = 4,
+    EM_WALL         = 8,
+    EM_NATIVEMEM    = 16,
+    EM_METHOD_TRACE = 32,
+};
+constexpr u8 EVENT_MASK_SIZE = 6;
```
Baraa-Hasheesh: people will forget to update this.
How about adding an `EM_MAX = 64` 

You can then assign `EVEN_MASK_SIZE = __builtin_ctz(EM_MAX)`

Baraa-Hasheesh: alternatively if you're able to add a clear assertion that gets triggered if someone forgets to update this that should be good enough

fandreuz: > How about adding an EM_MAX = 64

I agree that this is not optimal, but how is this harder to forget than updating `EVENT_MASK_SIZE`?

> alternatively if you're able to add a clear assertion

I looked for something like that, unfortunately I could not find anything. Happy to hear suggestions on this.

test/test/jfr/JfrTests.java
fandreuz: 300b5d5c4d1060c5d38ff054aa4c735429d8e4ff

src/profiler.h
Baraa-Hasheesh: > I agree that this is not optimal, but how is this harder to forget than updating `EVENT_MASK_SIZE`?

If someone touches the enum he will notice EM_MAX
Currently I can see someone updating the enum & missing the `EVENT_MASK_SIZE`

EVEN_MASK_SIZE = __builtin_ctz(EM_MAX)

fandreuz: I don't like polluting the enum with a value which should not be used to construct a mask. Also, what you suggest is one more constant than what is needed, in exchange for no additional guarantee at all.

src/flightRecorder.cpp
```diff
@@ -1382,11 +1382,9 @@ Error FlightRecorder::startMasterRecording(Arguments& args, const char* filename
...
     jobject jfilename = env->NewStringUTF(filename);
     jobject jsettings = args._jfr_sync == NULL ? NULL : env->NewStringUTF(args._jfr_sync);
-    int event_mask = (args._event != NULL ? 1 : 0) |
-                     (args._alloc >= 0 ? 2 : 0) |
-                     (args._lock >= 0 ? 4 : 0) |
-                     (args._latency >= 0 ? 8 : 0) |
-                     ((args._jfr_options ^ JFR_SYNC_OPTS) << 4);
+
+    int event_mask = make_event_mask(args) |
+                     ((args._jfr_options ^ JFR_SYNC_OPTS) << (EVENT_MASK_SIZE - 1));
```
krk: Why not `JFR_OPTS_SHIFT`?

src/helper/one/profiler/JfrSync.java
```diff
@@ -21,6 +21,22 @@
  * Synchronize async-profiler recording with an existing JFR recording.
  */
 class JfrSync implements FlightRecorderListener {
+    // Keep in sync with EventMask
+    private static final int EM_CPU            = 1;
+    private static final int EM_ALLOC          = 2;
+    private static final int EM_LOCK           = 4;
+    private static final int EM_METHOD_TRACE   = 32;
+
+    // Keep in sync with EVENT_MASK_SIZE - 1
```
krk: Let's add a static assert in that case?

src/main/main.cpp
```diff
@@ -529,6 +529,10 @@ int main(int argc, const char** argv) {
             params << ",jfrsync=" << args.next();
             output = "jfr";
 
+        } else if (arg == "--jfropts") {
```
krk: Do we need a documentation update?

src/profiler.h
```diff
@@ -47,6 +47,24 @@ enum State {
...
+
+static int make_event_mask(const Arguments& args) {
+    return (args._event     != NULL ? (args._latency >= 0 ? EM_METHOD_TRACE : EM_CPU) : 0) |
+           (args._alloc     >= 0    ? EM_ALLOC        : 0) |
```
krk: nit: If we are doing the `?` alignment, can we align the last `:` characters as well?

test/test/jfr/JfrTests.java
```diff
@@ -170,6 +170,26 @@ public void allModeEventOverride(TestProcess p) throws Exception {
         assert events.contains("profiler.Free"); // nativemem profiling
     }
 
+    // Simple smoke test, nothing in particular is tested
```
krk, line 173: > nothing in particular is tested

This is not a good argument for adding a test.

For example, can we test `--jfropts 9999`, `0`, `-1` etc. magic values?

krk (changes requested):

src/main/main.cpp
fandreuz: No, it's already there:
https://github.com/async-profiler/async-profiler/blob/d97a7d334331383c40a062572fa683d46c954492/src/main/main.cpp#L92

src/profiler.h
krk: We have some examples e.g. `NUM_IMPORTS` or `NUM_` as the last element of enums in the codebase.

Letting the compiler do the work would be better than remembering to update an associated declaration.

test/test/jfr/JfrTests.java
fandreuz: > This is not a good argument for adding a test.

We have some smoke test already.

Ideally I'd like to test that setting `jfropts` has "some" effect, but it does not seem to be possible. If an event is disabled for JFR, async-profiler will provide it, and viceversa, so there's no observable effect I can use to write the test.

> For example, can we test --jfropts 9999, 0, -1 etc. magic values?

We don't really do this kind of tests in general, do we?

src/helper/one/profiler/JfrSync.java
fandreuz: Can you clarify on this? I'm missing how this could work

src/profiler.h
fandreuz: > Letting the compiler do the work would be better than remembering to update an associated declaration.

`EM_MAX` would be set to `EM_METHOD_TRACE + 1`, unless we set it manually to something else. Am I missing something?

test/test/jfr/JfrTests.java
```diff
@@ -170,6 +170,26 @@ public void allModeEventOverride(TestProcess p) throws Exception {
...
+
+        Map<String, Integer> eventsCount = new HashMap<>();
+        try (RecordingFile recordingFile = new RecordingFile(p.getFile("%f").toPath())) {
+            while (recordingFile.hasMoreEvents()) {
+                RecordedEvent event = recordingFile.readEvent();
+                eventsCount.compute(event.getEventType().getName(), (key, old) -> old == null ? 1 : (old + 1));
+            }
+        }
+
+        assert eventsCount.get("jdk.OSInformation") != null : eventsCount;
```
Baraa-Hasheesh: how about using `assert eventsCount.contains`

src/helper/one/profiler/JfrSync.java
krk: I was thinking of the other file but I see now this is not possible, we can ignore this comment.

src/profiler.h
krk: I don't follow, can't we calculate `EVENT_MASK_SIZE` instead of setting to const number?

fandreuz: > I don't follow, can't we calculate EVENT_MASK_SIZE instead of setting to const number?

How do you suggest doing this? I could not find a way which does not involve some manual work to keep it up to date.

Baraa-Hasheesh: > How do you suggest doing this? I could not find a way which does not involve some manual work to keep it up to date.

Assuming you will keep the default `EM_MAX` (EM_METHOD_TRACE + 1)
You can do this `EVEN_MASK_SIZE = __builtin_ctz(EM_MAX - 1) + 1`

Even if you need manual change for EM_MAX, it would be the same amount of effort but much less likely to miss something as currently is

fandreuz: 6556e70b3329453d2704f37b75e1a95bc761bdfb

src/flightRecorder.cpp
fandreuz: I changed both places to use the same variable naming: 8c8cfae74661ccc23706d0e612e98c35cc9b0b70

src/profiler.h
fandreuz: 3931b663afad75a7b9a03ca40ad4e80e56547eb3

test/test/jfr/JfrTests.java
fandreuz: cff2d22138ccb4aaf924f88c24bf45a14c22a4ba

src/helper/one/profiler/JfrSync.java
```diff
@@ -21,6 +21,22 @@
  * Synchronize async-profiler recording with an existing JFR recording.
  */
 class JfrSync implements FlightRecorderListener {
+    // Keep in sync with EventMask
+    private static final int EM_CPU            = 1;
+    private static final int EM_ALLOC          = 2;
+    private static final int EM_LOCK           = 4;
+    private static final int EM_METHOD_TRACE   = 32;
```
Baraa-Hasheesh: What do you think about including the wall & nativemem values but not use them to make the mirror between cpp & java more obvious 

I'm fine with either

src/profiler.h
```diff
@@ -47,6 +47,25 @@ enum State {
     TERMINATED
 };
 
+enum EventMask {
```
Baraa-Hasheesh: Let's add a comment here to update the java file when changed

src/helper/one/profiler/JfrSync.java
fandreuz: I don't like unused constants.

Baraa-Hasheesh: Fair, would a comment be acceptable?
// EM_WALL = 8
// EM_NATIVE_MEM = 16

src/profiler.h
fandreuz: 5b7bc80a1f5283d1e7613762630a941458c08393

test/test/jfr/JfrTests.java
apangin: > We don't really do this kind of tests in general, do we?

We don't.
It's disappointing to see debates in the comments on how to unit-test `a + b` or how to name a constant for zero. I exaggerate, but it's roughly how it looks from the outside.
Please guys focus on logical errors and design flaws in your reviews; leave comments about coding style to me :)

src/flightRecorder.cpp
apangin: This looks wrong. Either `EVENT_MASK_SIZE` is not a mask size, or you should not subtract 1.

src/helper/one/profiler/JfrSync.java
```diff
@@ -72,40 +88,43 @@ public static void stop() {
...
-        if ((eventMask & 4) != 0) {
+        if ((eventMask & EM_LOCK) != 0) {
             recording.disable("jdk.JavaMonitorEnter");
             recording.disable("jdk.ThreadPark");
         }
-        if ((eventMask & 8) != 0) {
+        // No built-in event related to EM_WALL
+        // No built-in event related to EM_NATIVEMEM
+        if ((eventMask & EM_METHOD_TRACE) != 0) {
             recording.disable("jdk.MethodTrace");
```
apangin: I don't see a reason to disable built-in `jdk.MethodTrace` events.
1. They are not enabled by default. When they are, it means user enabled them explicitly.
2. They don't conflict with async-profiler's events. Moreover, it is possible that JFR and async-profiler trace different groups of methods.

src/main/main.cpp
apangin: Can be merged with `--jfrsync` branch.

src/profiler.h
```diff
@@ -47,6 +47,26 @@ enum State {
...
+    EM_CPU          = 1,
+    EM_ALLOC        = 2,
+    EM_LOCK         = 4,
+    EM_WALL         = 8,
+    EM_NATIVEMEM    = 16,
+    EM_METHOD_TRACE = 32,
+    EM_MAX
+};
+constexpr u8 EVENT_MASK_SIZE = __builtin_ctz(EM_MAX - 1) + 1;
```
apangin: 😵 I know you were asked to write this, but please undo.

How will a normal developer find what's the _actual_ size of the mask in bits by looking at this expression? (like whether it fits in one byte or not)
Then, an extra `EM_MAX` constant with some meaningless value just to calculate this weird expression.
The cherry on top is Java counterpart with `EVENT_MASK_SIZE = 6` ruining all arguments about "forgetting to update the value".
Finally, who said EVENT_MASK_SIZE needs to be equal to the number of constants? It can be an arbitrary value that splits 32 bits between EventMask and JfrOption, e.g., 16 to forget about updating the constant forever.

TL;DR put a number.

src/profiler.h
```diff
@@ -47,6 +47,26 @@ enum State {
...
+
+static int make_event_mask(const Arguments& args) {
```
apangin: Top level global functions are discouraged.
It mostly deals with Arguments' fields - why don't you make it a member function of Arguments then?

src/profiler.h
```diff
@@ -47,6 +47,26 @@ enum State {
...
+    return (args._event     != NULL ? (args._latency >= 0 ? EM_METHOD_TRACE : EM_CPU) : 0) |
```
apangin: Let's leave `event` for ExecutionSamples only.
Separate tracing engine will be only enabled with `trace` option.

src/helper/one/profiler/JfrSync.java
```diff
@@ -72,40 +88,43 @@ public static void stop() {
...
         }
 
+        eventMask >>= (EVENT_MASK_SIZE - 1);
```
apangin: `>>>=` and no `-1`

src/helper/one/profiler/JfrSync.java
fandreuz: Makes sense. I'll keep the spot in the mask for consistency: 509595af34f6fc426ca9d968228ce3490d01f275

src/main/main.cpp
fandreuz: 509595af34f6fc426ca9d968228ce3490d01f275

src/profiler.h
fandreuz: Agreed, 509595af34f6fc426ca9d968228ce3490d01f275

src/profiler.h
fandreuz: a528a696d667dde37a577317a32c0c5c9abcd301

src/profiler.h
fandreuz: What's the advantage? I understand its spot won't be used due to [this](https://github.com/async-profiler/async-profiler/pull/1507#discussion_r2380564893), but it would be nice if this mask delivered a faithful representation of the currently active engines.

src/flightRecorder.cpp
fandreuz: Right, this was coming from the fact that flightRecorder does not use all spots in the mask. But we shouldn't keep that: ff8b4447a684a3d915c3eb2243425589c295c7b1

src/helper/one/profiler/JfrSync.java
fandreuz: ff8b4447a684a3d915c3eb2243425589c295c7b1

src/arguments.h
```diff
@@ -7,6 +7,7 @@
 #define _ARGUMENTS_H
 
 #include <stddef.h>
+#include "arch.h"
```
apangin: Is it only for `u8`? EVENT_MASK_SIZE can be int

src/arguments.h
```diff
@@ -295,6 +307,14 @@ class Arguments {
         return (_jfr_options & option) != 0;
     }
 
+    int make_event_mask() const {
```
apangin: `eventMask()`

src/arguments.h
fandreuz: 5333f605fb86aa89b25c96fd8372551a02cdac3e

src/arguments.h
fandreuz: Ops, 3f738de897e5cf7e43761afaf06373395e96c252

src/profiler.h
apangin: OK, I hope this will be changed in the follow-up PR for running CPU + tracing engines together.

Issue #1496, opened 2025-09-17
Add JDK17,21 ARM & 25 x86/ARM on GHA


Baraa-Hasheesh:
### Describe the feature

as mentioned in title Add JDK17 & 21 for ARM
Add JDK25 for both x86 & ARM

### Use Case

early detection of failures on ARM

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
Done #1508

PR #1508, opened 2025-09-24
Adding more GHA runners to cover JDK versions on ARM

Baraa-Hasheesh:
### Description
Adding GHA runners to cover JDKs 17,21, & 25 on ARM & Also replacing JDK24 with 25 on x86

we recently found some issues on more recent JDK versions on ARM that would have been found by simply running the integration tests

### Related issues
#1496

### Motivation and context
Increase the coverage & early detection of bugs on ARM

### How has this been tested?
GHA runners & `make test`

---

### DIFF
--- a/.github/workflows/test-and-publish-nightly.yml
+++ b/.github/workflows/test-and-publish-nightly.yml
@@ -110,7 +110,7 @@ jobs:
           - display: linux-x64
             # Using "latest" here as the build and test will any ways run inside a container which we control
             name: ubuntu-latest
-        java-version: [8, 11, 17, 21, 24]
+        java-version: [8, 11, 17, 21, 25]
         java-distribution: [corretto]
         image: ["public.ecr.aws/async-profiler/asprof-builder-x86:latest"]
         include:
@@ -146,6 +146,27 @@ jobs:
               # There is no "latest" tag available (yet) as ARM runners are still in public preview
               name: ubuntu-24.04-arm
             java-version: 11
+            java-distribution: corretto
+            image: public.ecr.aws/async-profiler/asprof-builder-arm:latest
+          - runson:
+              display: linux-arm64
+              # There is no "latest" tag available (yet) as ARM runners are still in public preview
+              name: ubuntu-24.04-arm
+            java-version: 17
+            java-distribution: corretto
+            image: public.ecr.aws/async-profiler/asprof-builder-arm:latest
+          - runson:
+              display: linux-arm64
+              # There is no "latest" tag available (yet) as ARM runners are still in public preview
+              name: ubuntu-24.04-arm
+            java-version: 21
+            java-distribution: corretto
+            image: public.ecr.aws/async-profiler/asprof-builder-arm:latest
+          - runson:
+              display: linux-arm64
+              # There is no "latest" tag available (yet) as ARM runners are still in public preview
+              name: ubuntu-24.04-arm
+            java-version: 25
             java-distribution: corretto
             image: public.ecr.aws/async-profiler/asprof-builder-arm:latest
           - runson:
--- a/test/test/comptask/ComptaskTests.java
+++ b/test/test/comptask/ComptaskTests.java
@@ -12,7 +12,8 @@ public class ComptaskTests {
         mainClass = Main.class,
         agentArgs = "start,features=comptask,collapsed,interval=1ms,file=%f",
         jvmArgs = "-Xcomp",
-        jvm = Jvm.HOTSPOT
+        jvm = Jvm.HOTSPOT,
+        jvmVer = {8, 24} // TODO: Enable on next JDK25 is released JDK-8367689
     )
     public void testCompTask(TestProcess p) throws Exception {
         Output out = p.waitForExit("%f");
--- a/test/test/recovery/RecoveryTests.java
+++ b/test/test/recovery/RecoveryTests.java
@@ -30,7 +30,15 @@ public class RecoveryTests {
         Assert.isLess(out.ratio("unknown|break_compiled"), 0.002);
     }
 
-    @Test(mainClass = StringBuilderTest.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.ARM32})
+    @Test(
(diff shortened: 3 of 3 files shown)


fandreuz:
@Baraa-Hasheesh I think you're adding "GHA jobs", not runners

test/test/comptask/ComptaskTests.java
```diff
@@ -12,7 +12,8 @@ public class ComptaskTests {
         mainClass = Main.class,
         agentArgs = "start,features=comptask,collapsed,interval=1ms,file=%f",
         jvmArgs = "-Xcomp",
-        jvm = Jvm.HOTSPOT
+        jvm = Jvm.HOTSPOT,
+        jvmVer = {8, 24} // TODO: Enable on next JDK25 is released JDK-8367689
```
fandreuz: Unclear comment, what does it mean?

test/test/recovery/RecoveryTests.java
```diff
@@ -30,7 +30,15 @@ public void stringBuilder(TestProcess p) throws Exception {
         Assert.isLess(out.ratio("unknown|break_compiled"), 0.002);
     }
 
-    @Test(mainClass = StringBuilderTest.class, debugNonSafepoints = true, arch = {Arch.ARM64, Arch.ARM32})
+    @Test(
+        mainClass = StringBuilderTest.class,
+        debugNonSafepoints = true,
+        arch = {Arch.ARM64, Arch.ARM32},
+        // C2 Inlining can cause some issues when aggressive inlining happens
```
fandreuz: If the problem is inlining, why can't we run this with `-XX:-Inline` on the versions affected by this behavior?

.github/workflows/test-and-publish-nightly.yml
```diff
@@ -148,6 +148,27 @@ jobs:
             java-version: 11
             java-distribution: corretto
             image: public.ecr.aws/async-profiler/asprof-builder-arm:latest
+          - runson:
+              display: linux-arm64
+              # There is no "latest" tag available (yet) as ARM runners are still in public preview
```
fandreuz: I think one of these comments is enough, especially because it does not seem to be a real problem for us.

test/test/comptask/ComptaskTests.java
Baraa-Hasheesh: After JDK-8367689 is released in next JDK25 LTS release we need to enable this test back to run on all JDKs

test/test/recovery/RecoveryTests.java
Baraa-Hasheesh: Let's not hide bugs here 
I can make the test "pass" in many ways
The problem should be ideally fixed then we can re-enable the tests

test/test/comptask/ComptaskTests.java
fandreuz: I see, the comment is quite unclear tho

test/test/recovery/RecoveryTests.java
fandreuz: > Let's not hide bugs here

I'm not proposing to hide it. I'm saying that instead of disabling the test altogether on modern JDK versions running on ARM, we could think of having a separate test which runs with `-XX:-Inline`, if that really solves the problem. It could still help us catching possible regressions, especially since there isn't a clear timeline towards a fix.

Baraa-Hasheesh: I like the idea, 

@apangin what are you're thoughts about adding a separate test with `-XX:-Inline` just to capture possible issues/regressions in Intrinsics on newer JDKs?

test/test/comptask/ComptaskTests.java
apangin: `TODO: remove version filter after the fix for JDK-8367689 is released`

test/test/recovery/RecoveryTests.java
apangin: The comment sounds hand-waving. How can you claim something about aggressive inlining if you haven't got to the root cause yet?
You can frame it like "C2 often loses PcDesc mapping from arraycopy intrinsic to the original bytecode."
Link JBS issue when you create it.

Baraa-Hasheesh: Just to note, I will rephrase the comment
As we've yet to find the exact root cause 
But we noticed that the issue is always accompanied with aggressive inlines
Disabling the inline does make the test "pass"

apangin: Testing with `-XX:-Inline` has no practical benefit.

krk:
Please wait for https://github.com/async-profiler/async-profiler/pull/1502 to be merged.

test/test/comptask/ComptaskTests.java
```diff
@@ -12,7 +12,9 @@ public class ComptaskTests {
         mainClass = Main.class,
         agentArgs = "start,features=comptask,collapsed,interval=1ms,file=%f",
         jvmArgs = "-Xcomp",
-        jvm = Jvm.HOTSPOT
+        jvm = Jvm.HOTSPOT,
+        // TODO: remove version filter after the fix for JDK-8367689 is released
+        jvmVer = {8, 24}
```
krk, line 17:
```suggestion
        jvmVer = {8, 25}
```

Baraa-Hasheesh:
```
    private static boolean applicable(Test test) {
        Os[] os = test.os();
        Arch[] arch = test.arch();
        Jvm[] jvm = test.jvm();
        int[] jvmVer = test.jvmVer();
        return (os.length == 0 || Arrays.asList(os).contains(currentOs)) &&
                (arch.length == 0 || Arrays.asList(arch).contains(currentArch)) &&
                (jvm.length == 0 || Arrays.asList(jvm).contains(currentJvm)) &&
                (jvmVer.length == 0 || (currentJvmVersion >= jvmVer[0] && currentJvmVersion <= jvmVer[jvmVer.length - 1]));
    }
```

values are inclusive, we don't want to run on 25 as it's currently broken until next release

PR #1511, opened 2025-09-25
Fix the pixel jump on hovering the flamegraph, on Firefox and Chrome

krk:
### Description
This fixes the one pixel jump on hovering frames in the flamegraph, which was foreshadowed in https://github.com/async-profiler/async-profiler/pull/1140#discussion_r2197332570.

Chrome needed an extra adjustment which is done by the `jankOffset` variable.

First 4 seconds from master, and fixed after that:

https://github.com/user-attachments/assets/150132a7-12ef-4b28-90eb-f8e5a78abb4b

### Motivation and context

When a frame is hovered, text should not be offset, only the highlighting rectangle should be drawn.

### How has this been tested?

Tested in Chrome and Firefox.

---

### DIFF
--- a/.assets/html/flamegraph.html
+++ b/.assets/html/flamegraph.html
@@ -215,9 +215,9 @@
 			if (f) {
 				if (f !== root) getSelection().removeAllRanges();
 				const border = 1;
-				hl.style.left = (Math.max(f.left - root.left, 0) * px + canvas.offsetLeft) + "px";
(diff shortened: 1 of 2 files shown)


apangin:
Unfortunately, this makes it even worse for me.
Now it's not only text that jumps, but also the highlighting rectangle does not fully cover frames.

krk:
Interesting, I tested on Linux Chrome/Firefox.

An alternative that will work everywhere is rendering the highlight rectangle on the canvas, rather than with an html element.

krk:
Closing this and may pick up the "render directly on canvas" approach at a later date.

Issue #1513, opened 2025-09-26
Test RecoveryTests.suppliers sometimes fail on alpine


krk:
### Describe the bug

```
 FAIL [105/119] RecoveryTests.suppliers took 8.949 s
java.lang.AssertionError: Expected 0.002008032128514056 < 0.002
	>  test/test/recovery/RecoveryTests.java:73
	>  Assert.isLess(out.ratio("unknown|break_compiled"), 0.002);
	at one.profiler.test.Assert.assertComparison(Assert.java:39)
	at one.profiler.test.Assert.isLess(Assert.java:76)
	at test.recovery.RecoveryTests.suppliers(RecoveryTests.java:73)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at one.profiler.test.Runner.run(Runner.java:124)
	at one.profiler.test.Runner.main(Runner.java:188)
```

e.g. at https://github.com/krk/async-profiler/actions/runs/18042779349/job/51345959848

### Async-profiler version

master

Baraa-Hasheesh:
This is likely due to MUSL unwinding issues 

```
[unknown];/lib/ld-musl-x86_64.so.1 1
[unknown];/lib/ld-musl-x86_64.so.1;entry_SYSCALL_64_after_hwframe_[k];do_syscall_64_[k];x64_sys_call_[k];__x64_sys_futex_[k];do_futex_[k];futex_wait_[k];__futex_wait_[k];futex_wait_queue_[k];schedule_[k];__schedule_[k];finish_task_switch.isra.0_[k] 2
[unknown];memcpy 1
```

I found 4 samples here that caused this failure

Baraa-Hasheesh:
@apangin can this be closed after the #1517 change

apangin:
Yes, I think so.

PR #1517, opened 2025-10-01
Use JavaFrameAnchor to find top Java frame with cstack=vm

apangin:
### Description

If a thread has `JavaFrameAnchor` set, prefer getting PC/SP/FP from the anchor, since the anchor's values are always reliable.

### Related issues

#1317, #1179, #1513

### Motivation and context

Async-profiler may fail to unwind native part of the stack when a library is compiled without frame pointers AND with no DWARF unwinding info in `.eh_frame` section (typically on Alpine Linux). In this case, stack walker may use incorrect values of SP/FP once it reaches the topmost Java frame or it can miss the top Java frame completely. This in turn results in a large number of truncated stack traces with `[unknown]` or `[break_interpreted]` frame.

The solution is to use JavaFrameAnchor saved in the VM's JavaThread structure to discover the top Java frame even when the native part of the stack cannot be traversed.

### How has this been tested?

Tested manually on Liberica JDK 11 with sample programs provided in #1179.
Also, updated `stackwalker` integration tests: previously they checked that stack traces were truncated, now they check that stack traces are complete.

---

### DIFF
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -267,7 +267,10 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
         if (CodeHeap::contains(pc)) {
             NMethod* nm = CodeHeap::findNMethod(pc);
             if (nm == NULL) {
-                fillFrame(frames[depth++], BCI_ERROR, "unknown_nmethod");
+                if (anchor == NULL) {
+                    // Add an error frame only if we cannot recover
+                    fillFrame(frames[depth++], BCI_ERROR, "unknown_nmethod");
+                }
                 break;
             }
 


src/stackWalker.cpp
```diff
@@ -445,6 +461,14 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
         }
     }
 
+    // If we did not meet Java frame but current thread has JavaFrameAnchor set,
+    // retry stack walking from the anchor
+    if (anchor != NULL && anchor->getFrame(pc, sp, fp)) {
+        anchor = NULL;
+        while (depth > 0 && frames[depth - 1].method_id == NULL) depth--;  // pop unknown frames
```
Baraa-Hasheesh, line 471: * shouldn't we also pop error frames as well -> top frame can be "unknown_nmethod" 
* I think we should add an error frame to indicate that we recovered from anchor (I.E. we unwinding failed in native layer before reaching the first proper java frame)

src/stackWalker.cpp
```diff
@@ -257,13 +257,32 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
         }
     }
 
-    // Walk until the bottom of the stack or until the first Java frame
+    JavaFrameAnchor* anchor = NULL;
+    if (detail > VM_BASIC && vm_thread != NULL && vm_thread->isJavaThread()) {
```
Baraa-Hasheesh, line 261: Not related to this change

Question: once VM becomes the default stack walker can we delete this VM_BASIC mode or do we still need it?

=> It's only used for alloc_tracer samples when VM structs are available by using the `JavaFrameAnchor`

src/stackWalker.cpp
apangin: OK, I no longer add `unknown_nmethod` error if anchor is available.
All other erros happen in Java context, where there is no anchor to recover from.

apangin: Recovering from anchor is a normal situation, we don't need to mark it with a synthetic error frame.
This typically happens at the border between a native Java method and its JNI implementation, like in your StackWalker tests:
```
    Java_test_stackwalker_StackGenerator_largeFrame
-->
    test/stackwalker/StackGenerator.largeFrame
    test/stackwalker/StackGenerator.main
```

src/stackWalker.cpp
apangin: `VM_BASIC` mode mimics what JVM TI GetStackTrace or AsyncGetCallTrace produce.
This can be probably removed, but only after the next release, when there will be a large cleanup of AGCT-related code.

src/stackWalker.cpp
Baraa-Hasheesh: My concern is that it's possible for the recovery to happen way before that boundary
For example:

```
NativeMethod3
NativeMethod2 --> Stack walker can fail here due to missing FPs & dwarf info
NativeMethod1
NativeJniMethod
JavaJniMethod
JavaMainMethod
```

The final collected stack will look like

```
NativeMethod3
NativeMethod2
                ------> Here we missed NativeMethod1 & NativeJniMethod
JavaJniMethod
JavaMainMethod
```

I understand there is no real way to know if we really missed any frames, so this suggested "synthetic error frame" could be wrong, but I think there's value to know it's possible to have missing frames as that does tell us there is potential to enhance our unwinding logic 

My point in short is I don't want to hide potential unwinding bugs in DWARF that we can fix/enhance

apangin: As a user, I'd prefer _not_ to have these error frames. They add noise and give a feeling that something works wrong, although this is exactly how the algorithm is designed to work.
Often, especially on arm64 where there is a dedicated register for frame pointers, native stack will be fully unwound.

After all, this behavior replicates how `cstack=dwarf` + `AsyncGetCallTrace` work right now, so there will be no surprises for users.

Issue #1519, opened 2025-10-02
Profile Java with a native agent


apangin:
### Describe the bug

[rust-agent](https://github.com/async-profiler/rust-agent) is a wrapper around async-profiler, it can be used to profile Rust applications. Sometimes Rust application cohabits with a JVM in the same process.

When `asprof_execute` is called for the first time, it detects a running JVM and attaches current thread to it. But if the next time `asprof_execute` is called from a different thread, it will crash the JVM, becasue `VM::loaded()` is `true`, but `VM::jni()` returns `NULL`.

There was a previous attempt to fix a similar bug: #1454, however, it only handled dump/stop commands, but the problem can also happen with start/resume commands. My proposal is to make `asprof_execute` take care of attachment and deattachment, as all other code paths are from Java threads.

### Async-profiler version

4.1

PR #1523, opened 2025-10-02
Fix crash when profiling java from a native agent

Baraa-Hasheesh:
### Description
In general it's possible for the user to execute various profiler commands via different threads.

If a native application has a JVM that is attached to the profiler, it becomes possible that the user will invoke the profiler from a non java thread.

this in turn may either result in: 
* bad sampling of the JVM -> User calling dump/stop
* outright crash -> user calls start which calls Profiler::startTimer which tries to the access the JVM

This fixes both issues by ensuring that the current thread is attached to the JVM if it exists at the start of any asprof command.

### Related issues
#1519

### Motivation and context
Improve experience for native applications & prevent crashes

### How has this been tested?
New tests have been added
These tests fail before the change & pass after it

---


src/asprof.cpp
```diff
@@ -24,6 +24,11 @@ DLLEXPORT const char* asprof_error_str(asprof_error_t err) {
 }
 
 DLLEXPORT asprof_error_t asprof_execute(const char* command, asprof_writer_t output_callback) {
+    SafeJvmContext safe_jvm_context;
+    if (!safe_jvm_context.safe()) { // JVM exists but we failed to attach
```
apangin: Naming is not good. It's very vague (for someone unfamiliar with the problem) what "jvm context" means here and how it can be "safe" or "unsafe".

src/asprof.cpp
```diff
@@ -24,6 +24,11 @@ DLLEXPORT const char* asprof_error_str(asprof_error_t err) {
...
+        return asprof_error("Could not attach to JVM");
```
apangin: `asprof_execute` is called from a native thread. I don't expect it to fail because of any JVM issues; after all, user may not be interested in JVM profiling.

src/vmEntry.h
```diff
@@ -187,4 +192,18 @@ class VM {
     static jvmtiError JNICALL RetransformClassesHook(jvmtiEnv* jvmti, jint class_count, const jclass* classes);
 };
 
+class SafeJvmContext {
```
apangin: While I generally like RAII idiom, I believe code will be cleaner here without it. There is only one place where it is used, and it has only one exit point.

src/asprof.cpp
```diff
@@ -24,6 +24,11 @@ DLLEXPORT const char* asprof_error_str(asprof_error_t err) {
...
+    }
```
apangin: Any attachment should happen _after_ parsing arguments and opening logs.
First, because there is no point in attaching if there are syntax errors in arguments;
second, because log file should be ready before we do something that may issue warnings.

Furthermore, I think we can skip attaching here if we know for sure that the command does not require any JNI access, e.g. `status`, `meminfo`, `list`, `version`.

src/asprof.cpp
```diff
@@ -24,6 +24,11 @@ DLLEXPORT const char* asprof_error_str(asprof_error_t err) {
...
+
```
apangin: The problem is likely less trivial than it seems. I forsee troubles with `loop` option where we restart profiling session without calling API explicitly. What if JVM is loaded in the middle of a loop?

I don't think this is a common use case, so I'd prefer to stay safe and simple. What about moving `VM::tryAttach()` from `Profiler::start` to here? This way, we'll avoid potential issues with `loop` and we'll also keep all dynamic attach logic in one place.

src/asprof.cpp
Baraa-Hasheesh: For native applications with no JVM instance everything will work fine 

This path will only be taken if & only of the native app has a JVM instance that the profiler is attached to

src/asprof.cpp
Baraa-Hasheesh: That should be fine 

I also thought about that but opted against moving the tryAttach here 

My reasoning being the following:
* The only time it makes sense to attach to a JVM is at the start of the profiling session (start command), doing that randomly mid profiling session will cause problems/complexities for the timer loop side specially for the auto flushing of chuncks for the JFR recorder (other complications might happen)

* I think this is more of a feature, if a JVM is started in the middle of a continuous profiling session the profiler will be smart enough to attach to it on the next stop/start that happens internally due to loop

src/asprof.cpp
Baraa-Hasheesh: We can't infer if the user cares about JVM capabilities or not

The only indications we have is that the user is using profiler in an application that has a JVM instance

src/asprof.cpp
Baraa-Hasheesh: --> Just to clarify what I mean by "randomly mid profiling session" is for example the user calling dump in the middle of an active profiling session

src/asprof.cpp
Baraa-Hasheesh: Logic updated https://github.com/async-profiler/async-profiler/pull/1523/commits/fcdf16105ea84f6588b8f2ad73008c4467997fcf

src/asprof.cpp
Baraa-Hasheesh: Logic updated https://github.com/async-profiler/async-profiler/commit/fcdf16105ea84f6588b8f2ad73008c4467997fcf

src/vmEntry.h
Baraa-Hasheesh: Logic updated https://github.com/async-profiler/async-profiler/commit/fcdf16105ea84f6588b8f2ad73008c4467997fcf

test/test/nonjava/non_java_app.cpp
```diff
@@ -5,7 +5,6 @@
 
 #include <dirent.h>
 #include <dlfcn.h>
-#include <iostream>
```
fandreuz, line 8: I'm curious, why do we need to make this change here and now? It's difficult to distinguish what is actually needed for #1519

Baraa-Hasheesh: This was the only native tests that was using "iostream", in a matter on fact it was the only place of the profiler using it

So I did some cleanups as I was touching the test, 
You can focus on the large multi line changes such as method5 & method6

src/asprof.cpp
apangin: This PR does not fully fix the original problem, but adds new problems.
1. Potential race between `asprof_execute` and `timerLoop`, where unattached thread may attempt to use JNI.
2. Inconsistency in detaching native threads from the JVM: the first thread to call `asprof_execute` will stay attached to the JVM forever, but subsequent threads will be detached.
3. Inability to attach to the JVM will prevent from running profiler even in "native" mode.

Baraa-Hasheesh: > 1. Potential race between `asprof_execute` and `timerLoop`, where unattached thread may attempt to use JNI.

I can't really picture that possibility with the current code 
* JVM is only loaded at start of "start" command, & if it's not loaded then everything runs in "native" mode
* It shouldn't be possible to trigger a JVM load while a different thread is in "start" command due to the [locks & state machine](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1066)

Am I missing something?

> 2\. Inconsistency in detaching native threads from the JVM: the first thread to call `asprof_execute` will stay attached to the JVM forever, but subsequent threads will be detached.

Sure I will check on how best to make the flow consistent 

> 3\. Inability to attach to the JVM will prevent from running profiler even in "native" mode.

I agree that this indeed an issue & I will look on how to best fix this.

> This PR does not fully fix the original problem

Just to confirm the current PR changed the behavior of a potential crash to the profiler refusing to start for safety reasons 

However the desired behavior is to allow the profiler (if possible) to run but lock/don't access the JVM capabilities to avoid crashing the application, am I reading this correctly?

Baraa-Hasheesh: > > 1. Potential race between `asprof_execute` and `timerLoop`, where unattached thread may attempt to use JNI.
> 
> I can't really picture that possibility with the current code
> 
> * JVM is only loaded at start of "start" command, & if it's not loaded then everything runs in "native" mode
> * It shouldn't be possible to trigger a JVM load while a different thread is in "start" command due to the [locks & state machine](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1066)
> 
> Am I missing something?

Actually never mind, I found the race condition but to confirm, it happens between the `loop` loading the VM on the next start & `asprof_execute` on other threads.

apangin: > it happens between the loop loading the VM on the next start & asprof_execute on other threads.

Yes. This kind of race is difficult to detect and to fix correctly, that's why I thought it would be simpler if attachment to the existing VM was handled uniformly in one place in a thread-safe manner (under a lock?)

Perhaps, putting it in `asprof_execute` was not the best idea, since it is not protected by the lock, whereas Profiler::start/stop/dump are already protected by the `_state_lock`.

Baraa-Hasheesh: > > it happens between the loop loading the VM on the next start & asprof_execute on other threads.
> 
> Yes. This kind of race is difficult to detect and to fix correctly, that's why I thought it would be simpler if attachment to the existing VM was handled uniformly in one place in a thread-safe manner (under a lock?)
> 
> Perhaps, putting it in `asprof_execute` was not the best idea, since it is not protected by the lock, whereas Profiler::start/stop/dump are already protected by the `_state_lock`.

Yes, I'm actually moving towards that implementation, but note there are some edge cases that will require some extra checks 

For example  if you attempt to stop a running profiler & attach fails

I moved this PR to draft as to try & find other locations of possible crashes/problems & add tests for them

src/flightRecorder.cpp
```diff
@@ -1259,7 +1259,7 @@ Error FlightRecorder::start(Arguments& args, bool reset) {
 
     char* filename_tmp = NULL;
     const char* master_recording_file = NULL;
-    if (args._jfr_sync != NULL) {
+    if (args._jfr_sync != NULL && VMCapabilities::available()) {
```
Baraa-Hasheesh: old crash on Master :)

Baraa-Hasheesh:
@apangin I refactored the logic to fix all of the crashes I found  & added a new test for JFR.
=> Should I also do a JFR sync test on a native app?

The new logic is using RAII on all commands & replaced all `VM::loaded()` checks with the new more safe `VMCapabilities::available()` check 

I chased the usages of JNI & JVMTI & I think we should be safe
=> I did manual testing &  if attach fails the profiler will either fail due to needing JVM access or it will start in pure native mode if possible

Baraa-Hasheesh:
moved to draft to consider a different impl

Issue #1525, opened 2025-10-03
Expose a dictionary of javaThreads from JfrReader


lbloder:
### Describe the feature

We would like to expose a Dictionary `javaThreadIds` that maps eventIds to javaThreadIds in `JfrReader`.

### Use Case

In order to implement profiling for the [sentry-java sdk](https://github.com/getsentry/sentry-java) we had to vendor-in the `JfrReader.java` in order to get the JavaThread Ids, which we need to map them to running transactions.

If this feature would be implemented we could depend on the jfr-converter as-is without vendoring and customizing.

We can also supply a PR for this, if you'd like.

### Proposed Solution

As part of the `readThread` method in `JfrReader` (https://github.com/async-profiler/async-profiler/blob/ea095462ca20581b93ac3d3d7360c9b6db457175/src/converter/one/jfr/JfrReader.java#L453-L464).

We would propose to track the `javaThreads` in a Dictionary just like the `threads` are tracked by their name, so we can map them from the event id.

Here's our version of the `JfrReader` for reference: https://github.com/getsentry/sentry-java/blob/main/sentry-async-profiler/src/main/java/io/sentry/asyncprofiler/vendor/asyncprofiler/jfr/JfrReader.java

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
Sounds good. Please go ahead and submit a PR. Thanks.

PR #1526, opened 2025-10-06
Expose dictionary that maps event id to javaThreadId

lbloder:
### Description
Adds a dictionary `javaThreads` that maps eventIds to javaThreadIds.

### Related issues
#1525 

### Motivation and context
For the [sentry-java sdk](https://github.com/getsentry/sentry-java?rgh-link-date=2025-10-03T13%3A38%3A47.000Z) we recently implemented continuous profiling with async-profiler. We had to vendor in parts of the converter in order to get to the `Java Thread Id` that we need to map between profiles and transactions. This PR exposes that mapping and would allow us to simply use the `jfr-converter` as a dependency. 
Hopefully this will be useful for others too.

### How has this been tested?
We implemented this change in sentry-java sdk and have tested it in that context.

---

### DIFF
--- a/src/converter/one/jfr/JfrReader.java
+++ b/src/converter/one/jfr/JfrReader.java
@@ -60,6 +60,7 @@ public class JfrReader implements Closeable {
     public final Dictionary<StackTrace> stackTraces = new Dictionary<>();
     public final Map<String, String> settings = new HashMap<>();
     public final Map<String, Map<Integer, String>> enums = new HashMap<>();
+    public final Dictionary<Long> javaThreads = new Dictionary<>();
 
     private final Dictionary<Constructor<? extends Event>> customEvents = new Dictionary<>();
 
@@ -460,6 +461,7 @@ public class JfrReader implements Closeable {
             long javaThreadId = getVarlong();
             readFields(fieldCount - 4);
             threads.put(id, javaName != null ? javaName : osName);
+            javaThreads.put(id, javaThreadId);
         }
     }
 


src/converter/one/jfr/JfrReader.java
```diff
@@ -460,6 +461,7 @@ private void readThreads(int fieldCount) {
             long javaThreadId = getVarlong();
             readFields(fieldCount - 4);
             threads.put(id, javaName != null ? javaName : osName);
+            javaThreads.put(id, javaThreadId);
```
apangin, line 466: Micro-optimization: `threads` dictionary is pre-sized to the required capacity in the beginning of this method, so can `javaThreads`.

lbloder: Thanks for the quick review, good catch.

I see two options here:

Option 1: less change to existing code
```
int count = threads.preallocate(getVarint());
javaThreads.preallocate(count);
```

Option 2: IMHO a bit more self-explanatory in terms of what `count` is and what value we use to size the dicts.
```
int count = getVarint();
threads.preallocate(count);
javaThreads.preallocate(count);
```

Any preference between these two?

apangin: Both are fine, but I agree Option 2 is a bit easier to understand.

lbloder: Went with option 2. Thanks again for the quick feedback.

apangin:

PR #1529, opened 2025-10-10
[POC] Packed frames in call trace storage

fandreuz:
### Description
`ASGCT_CallFrame` is arguably the object type which we allocate the most in Async-Profiler. We also keep many instances in memory for the whole lifetime of the profiler, until it is stopped or restarted.

This is the structure of `ASGCT_CallFrame`:
```
% pahole --class_name ASGCT_CallFrame ./build/lib/libasyncProfiler.so 
typedef struct {
        jint                       bci;                  /*     0     4 */
        jint                       padding;              /*     4     4 */
        jmethodID                  method_id;            /*     8     8 */

        /* size: 16, cachelines: 1, members: 3 */
        /* last cacheline: 16 bytes */
} ASGCT_CallFrame;
```
There's some padding between `bci` and `method_id` which takes 25% of the object.

In this PR I propose the following packed struct:
```
struct Packed_ASGCT_CallFrame {
        jint                       bci1;                 /*     0     4 */
        jint                       bci2;                 /*     4     4 */
        jmethodID                  method_id1;           /*     8     8 */
        jmethodID                  method_id2;           /*    16     8 */

        /* size: 24, cachelines: 1, members: 4 */
        /* last cacheline: 24 bytes */
};
```
It can store two frames in 24 bytes instead of 32. We can use it for frames stored in `CallTraceStorage`.

### Motivation and context
Using `Packed_ASGCT_CallFrame` could save around 25% of our memory usage in `CallTraceStorage`, which should result in:
- reduced memory footprint of the profiler;
- less frequent calls to `mmap` in `recordSample`. 

| Renaissance Benchmark                | Branch   | Call Trace Storage (KB) | Allocated Frames   |
|:--------------------------:|:---------:|:------------------------:|:--------------------:|
| akka-uct     | master  | 202 764                | 12 138 904                   |
| akka-uct     | packed-frames   | 169 996                |  —       |
| all          | master  | 3 119 128              | 204 653 358                  |
| all          | packed-frames   | 2 562 072              |     —    |

I used the following configuration to stress the profiler in seeing many different call traces:
```
start,jfr,all,event=cpu-clock,record-cpu,threads,sched
```
and default configuration for Renaissance.

### How has this been tested?
GHA, renaissance.

---


apangin:
Please, don't open this can of worms.
25% seems a good improvement in isolation, but what is an _actual_ impact on a typical server-side application?
IMO, even 25% does not justify a mess that this change introduces. Especially when there are techniques to achieve 2x or even 10x improvement by changing the data structure.
Now, what will you do when we need to attach some more information to each frame (read: add a new field)? I doubt that someone will roll back everything to normal layout.

fandreuz:
> Especially when there are techniques to achieve 2x or even 10x improvement by changing the data structure.

Do you mean some sort of compression?

apangin:
Let me correct myself. My main point was likely hidden behind emotions in the previous message.
The code in this PR is actually _not_ that terrible. It serves its purpose well. But.. I think the very concept of such optimization is broken. A stack trace is no longer stored as an array of frames, the key array properties are lost. It also introduces a new structure which pretends to look similar to `ASGCT_CallFrame` (and is even called similarly) but in fact isn't. For instance, what should one expect from bci2/method_id2 for the last unpaired frame of a stack trace? This all complicates the model of the CallTraceStorage and increases cognitive load required to deal with it. From this perspective, it would be better to achieve the same optimization outcome just by withdrawing alignment, i.e. let Frame structure be 12 byte in size, where method_id is not always 8-byte aligned. But this is not the problem to be solved right now, I think. We'll likely find a good use for these extra 4 bytes instead.
When we realize CallTraceStorage footprint is a bottleneck, we can revise the data structure completely rather than applying local optimizations. The biggest waste is duplication of frames in stack traces that share common prefix. That is HUGE. If we deduplicate those, we could achieve 10x savings. There are multiple ways of doing that, with prefix tree being one obvious example, although not the most efficient from computation perspective. Deduplication of stack chunks of 8 frames could be another option. We can discuss this later.

fandreuz:
Thanks for the clarification, that makes sense to me.

PR #1530, opened 2025-10-10
Timezone switcher between Local and UTC time in heatmaps

jerrinot:
### Description
Timestamps in heatmaps can be rendered in UTC. The state is stored in a local storage. 

### Related issues
I am not aware of an existing issue. 

### Motivation and context
Before the change, heatmaps would always use the local timezone from a browser. I find it confusing. I prefer UTC when analyzing logs. 

### How has this been tested?
On my local, in Firefox@Linux. 

---

[asprof_utc.webm](https://github.com/user-attachments/assets/23089240-61f1-478f-978b-4f2e4ec2ba31)

Disclosure: I used an AI assistant to develop this patch.

### DIFF
--- a/src/res/heatmap.html
+++ b/src/res/heatmap.html
@@ -3680,8 +3680,7 @@
 			let ms = origin + markIndex * oneMarkMs;
 			let markOffset = markIndex * markPixels - dx;
 
-			let options = currentTimezone === 'UTC' ? timeOptionsTinyWithTZ : timeOptionsTiny;
-			let title = formatTime(ms, options);
+			let title = formatTime(ms, currentTimezone === 'UTC' ? timeOptionsTinyWithTZ : timeOptionsTiny);
 			let x = markOffset + 4;
 			heatC.fillText(title, x, canvasTimeHeight - 1, markPixels);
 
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for your proposal.

I like the idea of a switch between UTC and local time, but I think it could be slightly better from the UI perspective.
Toggle buttons are always confusing: it's not obvious whether they show the current state or the state to switch to. It would be clearer if the current timezone is shown directly on the time axis. No need for a separate button: the time axis itself can serve as a switch, similarly to how the zoom label acts as a switch for changing zoom. That way, the interface stays minimal and intuitive - consistent with the rest of the tool.

Regarding GenAI assistants, I'd recommend to minimize usage of them in order to stay compliant with the project license and to simplify reviewers' work. For instance, I found that the clock icon in your PR literally replicates an MIT-licensed SVG image. Unfortunately, I don't have resources to verify origin of each contribution, so the general recommendation is to submit only your original work or third-party work explicitly marked as CC0 or equivalent.

jerrinot:
Thank you for your review, appreciate it. 

I agree toggles may be ambiguous; I recall there was some other SW where I was never sure about the same thing. 

I'll change to make the time axis clickable. I like the consistency with the Zoom feature. There is a slight issue with feature discoverability, but that's something to be possibly addressed later: for both zoom and timezones. 

Apologies for the situation with the icon. I used Claude Code as a starting point (I sucks at HTML) and then changed most of its code anyway since it went bananas with timezone conversions. but I failed to realize the generated icon could possibly be licensed.

apangin:
Thank you for working on the issue.
Upon review, I noticed that the text on the time axis is unreadable in some timezones:
timezone
Then, click on the timeline changes current range selection, that I didn't expect to happen.
It would be also better if cursor changed when hovering over the timeline, indicating that it is clickable.

jerrinot:
@apangin Thanks for the feedback! Changing the mouse cursor on hover and not cancelling selecting on the timezone switch was easy. I was unsure how to solve the issue with long time zones. I decided not to render a timezone at all when it's local (=default, the same behaviour as now) and only render it when it's UTC (=non-default, explicitly selected). Let me know what you think.

apangin:
> I decided not to render a timezone at all when it's local (=default, the same behaviour as now) and only render it when it's UTC (=non-default, explicitly selected).

That is smart. I like the idea :)

apangin:
Works like a charm now.

Issue #1330, opened 2025-06-06
Differential flamegraphs


franz1981:
### Describe the feature

One of the coolest features of heatmaps is the ability to produce with ease differential flamegraphs: it would be great to have a similar option in the converter - allowing to produce a diff flamegraph given two inputs jfr traces.
Clearly is not easy to compare if the duration of the two traces is different - but the same applies for heatmaps too, I think.

### Use Case

Running the same experiment with some performance fix in requires to compare before/after it: having a differential flamegraph is very handy for this.

### Proposed Solution

This is more a workaround for it: using folded output type and https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html can still get us there (I didn't tried, but I believe so), but it will miss lot of the nice visual of async profiler.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
Sorry it took longer; the feature has been finally merged.

[Usage](https://github.com/async-profiler/async-profiler/blob/master/docs/ConverterUsage.md#differential-flame-graph)

PR #1536, opened 2025-10-14
Add differential flamegraph feature to jfrconv

roy-soumadipta:
### Description
Add differential flamegraph feature to jfrconv similar to Brendan Gregg's differential flamegraph https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html.

Differential flamegraphs can be extended to heatmaps as well, however that is a much more complex task that needs more effort and separate handling. This PR addresses a quick win for customers to use and get benefit.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1330

### Motivation and context
Give customers the options to see regressions and improvements between two profiles generated.

### How has this been tested?
Tested with JFRs generated in integration test. Tested with JFR, HTML and Collapsed formats and also mix of formats.

---

[differential_test.html](https://github.com/user-attachments/files/22907721/differential_test.html)

### DIFF
--- a/docs/ConverterUsage.md
+++ b/docs/ConverterUsage.md
@@ -65,7 +65,7 @@ JFR options:
     --norm             Normalize names of hidden classes/lambdas, e.g. Original JFR transforms
                        lambda names to something like pkg.ClassName$$Lambda+0x00007f8177090218/543846639
                        which gets normalized to pkg.ClassName$$Lambda
-    --diff             Enable differential mode (auto-enables --norm). Outputs a differential flamegraph 
+    --diff             Enable differential mode (auto-enables --norm). Outputs a differential flamegraph
                        by acception two JFR/HTML/Collapsed files
     --dot              Dotted class names, e.g. java.lang.String instead of java/lang/String
     --from TIME        Start time in ms (absolute or relative)


src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,75 @@
...
+        Set<String> allStacks = new HashSet<>();
+        allStacks.addAll(folded1.keySet());
+        allStacks.addAll(folded2.keySet());
+
+        for (String stack : allStacks) {
+            long count1 = folded1.getOrDefault(stack, 0L);
+            long count2 = folded2.getOrDefault(stack, 0L);
+
+            // Always apply count normalization for meaningful differential analysis
+            if (total1 != total2 && total1 > 0) {
```
fandreuz: Should this be optional and configurable?

roy-soumadipta: I had this question in my mind, went with the current approach. I can make it optional. Let's see what Andrei has to say.

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static void processFiles(String file1, String file2, PrintStream output) throws IOException {
+        Map<String, Long> folded1 = readCollapsedFile(file1);
+        Map<String, Long> folded2 = readCollapsedFile(file2);
+
+        long total1 = folded1.values().stream().mapToLong(Long::longValue).sum();
+        long total2 = folded2.values().stream().mapToLong(Long::longValue).sum();
+
+        Set<String> allStacks = new HashSet<>();
+        allStacks.addAll(folded1.keySet());
```
fandreuz:
```suggestion
        Set<String> allStacks = new HashSet<>(folded1.keySet());
```

src/converter/Main.java
```diff
@@ -35,11 +35,19 @@ public static void main(String[] argv) throws Exception {
...
+
+            processDifferential(file1, file2, output, args);
+            return;
+        }
+
         for (int i = 0; i < fileCount; i++) {
             String input = args.files.get(i);
             String output = isDirectory ? new File(lastFile, replaceExt(input, args.output)).getPath() : lastFile;
 
-            System.out.print("Converting " + getFileName(input) + " -> " + getFileName(output) + " ");
```
fandreuz, line 42: Is this intended?

src/converter/one/convert/DiffFolded.java
roy-soumadipta: yes good catch

src/converter/Main.java
roy-soumadipta: No, must have been deleted with some other deletion locally

src/converter/Main.java
```diff
@@ -68,6 +76,54 @@ public static void convert(String input, String output, Arguments args) throws I
...
+        try {
+            String collapsed1 = getCollapsedFile(file1, args);
+            String collapsed2 = getCollapsedFile(file2, args);
+
+            String tempDiff = output + ".tmp.diff.collapsed";
+            DiffFolded.process(collapsed1, collapsed2, tempDiff);
+
+            convert(tempDiff, output, args);
+
+            if (!collapsed1.equals(file1)) new File(collapsed1).delete();
```
fandreuz: How about these instead?
- https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File-
- https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--

Baraa-Hasheesh:
Let's update the `ConverterUsage.md`

src/converter/Main.java
```diff
@@ -68,6 +76,54 @@ public static void convert(String input, String output, Arguments args) throws I
...
+        }
+
+        long endTime = System.nanoTime();
+        System.out.print("# " + (endTime - startTime) / 1000000 / 1000.0 + " s\n");
+    }
+
+    private static String getCollapsedFile(String inputFile, Arguments args) throws IOException {
+        if (inputFile.endsWith(".collapsed")) {
+            return inputFile;
+        } else if (isJfr(inputFile)) {
```
fandreuz: Is there any difference in JFR and HTML branches?

roy-soumadipta: No, and the entire condition can be simplified as well

src/converter/Main.java
```diff
@@ -35,11 +35,19 @@ public static void main(String[] argv) throws Exception {
             }
         }
 
+        if (args.differential && fileCount == 2 && !isDirectory) {
```
Baraa-Hasheesh: What if the user uses `args.differential` with a single file 
I think we should add validation here to return error if the user uses --differential on a single file

Also what if the user gives the same file twice?
Is it worth it to add a validation against that?
IMO that would save some sanity time when that mistake happens

src/converter/Main.java
```diff
@@ -35,11 +35,19 @@ public static void main(String[] argv) throws Exception {
...
+            String file1 = args.files.get(0);
+            String file2 = args.files.get(1);
+            String output = lastFile;
```
Baraa-Hasheesh: `lastFile = args.files.size() - 1` can be the same value as `file2 = args.files.get(1)`

This means you might just accidentally delete a user file

If user doesn't give a third file lets create an artificial differential file

src/converter/Main.java
```diff
@@ -68,6 +76,54 @@ public static void convert(String input, String output, Arguments args) throws I
...
+
+        long endTime = System.nanoTime();
+        System.out.print("# " + (endTime - startTime) / 1000000 / 1000.0 + " s\n");
+    }
+
+    private static String getCollapsedFile(String inputFile, Arguments args) throws IOException {
+        if (inputFile.endsWith(".collapsed")) {
+            return inputFile;
+        } else if (isJfr(inputFile)) {
+            String tempCollapsed = inputFile + ".tmp.collapsed";
```
Baraa-Hasheesh: rather than this let's create a proper temp file

src/converter/Main.java
```diff
@@ -68,6 +76,54 @@ public static void convert(String input, String output, Arguments args) throws I
...
+            Arguments tempArgs = new Arguments("--differential", "--output", "collapsed");
```
Baraa-Hasheesh: --differential? it's not needed 

Also What would happen if I give two jfr files & do `jfrconv f1.jfr f2.jfr output.html --differential --alloc` 

This will do a conversion always to Execution samples, what about other sample types?

src/converter/one/convert/Arguments.java
```diff
@@ -82,6 +83,11 @@ public Arguments(String... args) {
                 throw new IllegalArgumentException(arg);
             }
         }
+        
+        // Automatically enable normalization for differential flame graphs
+        if (differential) {
+            norm = true;
```
Baraa-Hasheesh, line 89: "--differential" is passed internally just to turn on this flag

The correct behaviour would be to pass "--norm" directly

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,75 @@
...
+        allStacks.addAll(folded2.keySet());
+
+        for (String stack : allStacks) {
+            long count1 = folded1.getOrDefault(stack, 0L);
+            long count2 = folded2.getOrDefault(stack, 0L);
+
+            // Always apply count normalization for meaningful differential analysis
+            if (total1 != total2 && total1 > 0) {
+                count1 = (long) (count1 * total2 / total1);
```
Baraa-Hasheesh, line 41: if `count1 = count2` there might be a difference in the profile due to a difference between `total1 & total2`

Is this intended, or should these two samples be recorded as no diff

src/res/flame.html
```diff
@@ -61,14 +61,42 @@ <h1>/*title:*/</h1>
...
 		const v = Math.random();
 		return '#' + (p[0] + ((p[1] * v) << 16 | (p[2] * v) << 8 | (p[3] * v))).toString(16);
 	}
 
+	function getDifferentialColor(delta) {
+		let r = 255, g = 255, b = 255;
+		
+		let maxDelta = 1;
+		if (typeof deltaMap !== 'undefined') {
+			for (const key in deltaMap) {
```
apangin: For every frame (there can be a million) you scan the entire deltaMap?
This can be precomputed, and not even in JS.

src/converter/one/convert/DiffFolded.java
apangin: No, it should not. At least, until there is a good use case for that. I don't know one from the top of my head.

src/converter/Main.java
```diff
@@ -68,6 +76,54 @@ public static void convert(String input, String output, Arguments args) throws I
...
+                + " -> " + getFileName(output) + " ");
+        System.out.flush();
+
+        long startTime = System.nanoTime();
+
+        try {
+            String collapsed1 = getCollapsedFile(file1, args);
+            String collapsed2 = getCollapsedFile(file2, args);
+
+            String tempDiff = output + ".tmp.diff.collapsed";
```
apangin: The tool does a series of intermediate conversions, writing temporary files to disk. This is a huge waste, both in terms of performance and extra code. There is no need to dump `diff.collapsed` file whose only purpose is to be parsed back to a flame graph. Similarly, there is no need to convert jfr to collapsed with the only purpose to be parsed again as `Map<String, Long>`. Sometimes this will not be even possible, e.g., in containers with small `/tmp` volume. 5MB flame graph can easily inflate to a 200MB collapsed file.

roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/res/flame.html
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/one/convert/DiffFolded.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/one/convert/Arguments.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/one/convert/DiffFolded.java
roy-soumadipta: [47bb491](https://github.com/async-profiler/async-profiler/pull/1536/commits/47bb49172d441ce4c2c20f5b2a8daa88571afa91)

src/converter/Main.java
```diff
@@ -35,6 +38,27 @@ public static void main(String[] argv) throws Exception {
...
+        if (args.differential) {
+            if (fileCount < 1 || fileCount > 2 || isDirectory) {
+                System.err.println("Error: --differential requires 2 input files and optionally 1 output file");
+                System.err.println("Usage: jfrconv --differential <file1> <file2> [output]");
+                System.exit(1);
+            }
+
+            String file1 = args.files.get(0);
+            String file2 = args.files.get(1);
+            String output = fileCount == 1 ? "differential-output.html" : lastFile;
```
Baraa-Hasheesh, line 50: **Style - Optional**
I think it's better to mash the file names together
{FILE1_NAME}_{FILE2_NAME}_diff.html

src/converter/Main.java
```diff
@@ -35,6 +38,27 @@ public static void main(String[] argv) throws Exception {
             }
         }
 
+        if (args.differential) {
```
Baraa-Hasheesh: --differential is too long for the user let's do --diff

src/converter/Main.java
```diff
@@ -68,10 +92,88 @@ public static void convert(String input, String output, Arguments args) throws I
...
+
+    private static void validateFileFormats(String file1, String file2) throws IOException {
+        String format1 = getFileFormat(file1);
+        String format2 = getFileFormat(file2);
+
+        if (!format1.equals(format2)) {
+            System.err.println("Error: Cannot create differential flame graph with different file formats");
+            System.err.println("File 1: " + getFileName(file1) + " (" + format1 + ")");
+            System.err.println("File 2: " + getFileName(file2) + " (" + format2 + ")");
+            System.err.println("Both files must be the same format (JFR, HTML, or collapsed)");
```
Baraa-Hasheesh, line 129: otlp not getting any love?

I can see that `otlp` to `collapsed` is not yet supported
Let's throw an exception at least?

src/converter/Main.java
```diff
@@ -68,10 +92,88 @@ public static void convert(String input, String output, Arguments args) throws I
...
+            System.exit(1);
+        }
+    }
+
+    private static String getFileFormat(String inputFile) throws IOException {
+        if (isJfr(inputFile))
+            return "JFR";
+        else if (inputFile.endsWith(".html"))
+            return "HTML";
+        else
```
Baraa-Hasheesh: This will use `otlp` files as `collapsed`

src/converter/Main.java
```diff
@@ -68,10 +92,88 @@ public static void convert(String input, String output, Arguments args) throws I
...
+            return "collapsed";
+    }
+
+    private static ProfileData getProfileData(String inputFile, Arguments args) throws IOException {
+        if (isJfr(inputFile)) {
+            return JfrToFlame.extractCollapsedDataWithDuration(inputFile, args);
+        } else if (inputFile.endsWith(".html")) {
+            Map<String, Long> collapsedData = FlameGraph.extractCollapsedDataFromHtml(inputFile, args);
+            return new ProfileData(collapsedData, null);
+        } else {
```
Baraa-Hasheesh, line 155: same as above

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+            long count2 = folded2.getOrDefault(stack, 0L);
+
+            // Apply normalization only when appropriate
+            if (shouldNormalize && total1 != total2 && total1 > 0) {
+                count1 = (long) (count1 * total2 / total1);
+            }
+
+            // Only output stacks that exist in profile2 (count2 > 0)
+            // This ensures the flame graph structure matches f2
+            if (count2 > 0) {
```
Baraa-Hasheesh, line 46: if `count2 == 0` that doesn't mean that there is no difference, the difference could be 100% (count1 > 0)

roy-soumadipta: In Brendan's implementation, this is optional with -negate flag. I wanted to avoid that in the initial implementation to keep things a little simpler. Also if something is missing in the second profile, it is definitely not a regression.
@apangin need your inputs, if you feel this is absolutely needed in this implementation I can add it

src/converter/Main.java
roy-soumadipta: I did do it first but changed it to a shorter name as the name can get pretty big depending on the filename

src/converter/one/convert/DiffFolded.java
apangin: Not needed

src/converter/Main.java
roy-soumadipta: [44c8dda](https://github.com/async-profiler/async-profiler/pull/1536/commits/44c8ddadeaae4952ca70f0a56479dac8a7abeca9)

src/converter/Main.java
roy-soumadipta: [44c8dda](https://github.com/async-profiler/async-profiler/pull/1536/commits/44c8ddadeaae4952ca70f0a56479dac8a7abeca9)

src/converter/Main.java
roy-soumadipta: [44c8dda](https://github.com/async-profiler/async-profiler/pull/1536/commits/44c8ddadeaae4952ca70f0a56479dac8a7abeca9)

src/converter/Main.java
roy-soumadipta: [44c8dda](https://github.com/async-profiler/async-profiler/pull/1536/commits/44c8ddadeaae4952ca70f0a56479dac8a7abeca9)

src/converter/one/convert/DifferentialResult.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.convert;
+
+import java.util.Map;
+
+public class DifferentialResult {
```
krk, line 10: nit: sometimes Diff sometimes Differential is used in naming. Let's use Diff everywhere?

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public final class DiffFolded {
```
krk, line 17: What does it mean to be folded here?

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+            maxDelta = Math.max(maxDelta, Math.abs(functionDelta));
+        }
+        
+        return new DifferentialResult(stackData, functionDeltas, maxDelta);
+    }
+
+    private static boolean shouldNormalizeCounts(ProfileData profile1, ProfileData profile2, long total1, long total2) {
+        if (profile1.durationSeconds == null || profile2.durationSeconds == null) return true;
+        
+        long roundedDuration1 = Math.round(profile1.durationSeconds);
```
krk, line 76: Why do we round to seconds to detect if normalization is required?

If one profile is exactly `1` second long and another `1.49` seconds, there will be ~50% bias when not normalized.

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+
+        Set<String> allStacks = new HashSet<>(folded1.keySet());
+        allStacks.addAll(folded2.keySet());
+
+        for (String stack : allStacks) {
+            long count1 = folded1.getOrDefault(stack, 0L);
+            long count2 = folded2.getOrDefault(stack, 0L);
+
+            // Apply normalization only when appropriate
+            if (shouldNormalize && total1 != total2 && total1 > 0) {
```
krk: `shouldNormalize` should contain the rest of the condition, to simplify this if to `if (shouldNormalize)`.

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+        Map<String, Long> functionDeltas = new HashMap<>();
+        long maxDelta = 1;
+        
+        Map<String, Long> folded1 = profile1.collapsedData;
+        Map<String, Long> folded2 = profile2.collapsedData;
+        
+        long total1 = folded1.values().stream().mapToLong(Long::longValue).sum();
+        long total2 = folded2.values().stream().mapToLong(Long::longValue).sum();
+
+        boolean shouldNormalize = shouldNormalizeCounts(profile1, profile2, total1, total2);
```
krk, line 30: Instead, we can get the normalization factor directly. 

```java
double factor = getNormalizationFactor(profile1, profile2)
```

and below `count1 = factor * count2;` instead of `if shouldNormalize` block.

src/converter/one/convert/FlameGraph.java
```diff
@@ -41,27 +49,88 @@ public void parseCollapsed(Reader in) throws IOException {
...
+            
+            String[] parts = counts.split(" ");
+            if (parts.length != 2) continue;
+            
+            long count2 = Long.parseLong(parts[1]);
+            buildStackAndAddSample(stackPart, count2, stack);
+        }
+    }
+
+    private void parseDifferentialLine(String line, CallStack stack) {
```
krk, line 79: I think we need to parse two `FlameGraph` instances and diff them, rather than working with text files.

src/converter/Main.java
krk: `out-diff.html`?

src/converter/Main.java
```diff
@@ -35,6 +38,27 @@ public static void main(String[] argv) throws Exception {
...
+                System.err.println("Usage: jfrconv --diff <file1> <file2> [output]");
+                System.exit(1);
+            }
+
+            String file1 = args.files.get(0);
+            String file2 = args.files.get(1);
+            String output = fileCount == 1 ? "differential-output.html" : lastFile;
+
+            if (file1.equals(file2)) {
+                System.err.println("Error: Cannot create differential flame graph with the same file twice");
```
krk, line 53: Why limit the user? Maybe they are testing that the result will be empty.

src/res/flame.html
```diff
@@ -61,14 +61,35 @@ <h1>/*title:*/</h1>
 		[0xcce880, 20, 20, 20],
```
krk, line 61: Please add a sample file to `.assets/html/flamegraph-diff.html` - and perhaps reference it from the docs.

krk (changes requested):

src/converter/one/convert/DiffFolded.java
roy-soumadipta: Inspired by Brendan Gregg's naming. As per resources like https://www.brendangregg.com/flamegraphs.html, the term comes from the flame graph community where stack traces are "folded" (collapsed) into a compact text format before being processed into visual flame graphs. This intermediate format makes it easy to:
* Store and transfer profiling data
* Perform differential analysis
* Generate flame graphs from the collapsed representation

So DiffFolded = Differential analysis of Folded stack traces

src/converter/one/convert/FlameGraph.java
roy-soumadipta: Like the difffolded naming reply, working with collapsed format looks like the general trend. For my knowledge why are you suggesting it, what are the pros and cons over this? Would like @apangin to weigh in on this as well.

src/converter/Main.java
roy-soumadipta: This doesn't seem a genuine use case to me. What is the need for such a scenario?

src/converter/one/convert/DiffFolded.java
Baraa-Hasheesh: `folded` is used in many projects such as `perf` & it's very similar to `collapsed` 
But for our usage I would prefer calling it diffCollapsed so we have one naming terminology rather than two

src/converter/one/convert/DiffFolded.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static boolean shouldNormalizeCounts(ProfileData profile1, ProfileData profile2, long total1, long total2) {
+        if (profile1.durationSeconds == null || profile2.durationSeconds == null) return true;
+        
+        long roundedDuration1 = Math.round(profile1.durationSeconds);
+        long roundedDuration2 = Math.round(profile2.durationSeconds);
+        
+        if (roundedDuration1 == roundedDuration2) return false;
+        
+        return true;
```
Baraa-Hasheesh, line 81: `return roundedDuration1 != roundedDuration2`

src/converter/one/convert/DiffFolded.java
roy-soumadipta: I do understand and had factored this in, but this is not a realistic scenario right? For realistic scenarios, the current impl suffices.

src/converter/one/convert/FlameGraph.java
krk: There is no good reason we would use temporary files and the collapsed text files with new format as intermediate representation for diffing. 

Everything you need to do the diff should already be in the FlameGraph object. If not, let's improve it instead of using temp text files as part of conversion/diffing.

src/converter/Main.java
krk: Why do you want to print an error message and exit, with valid inputs?

src/converter/one/convert/DiffFolded.java
krk: I don't understand. I find it very realistic to collect short duration profiles like 1-2 seconds and diff them.

src/converter/one/convert/FlameGraph.java
apangin: I agree with Kerem: FlameGraph is a good intermediate representation for building diffs, and a common denominator for various formats.

src/converter/Main.java
roy-soumadipta: Because the idea is to have different files for differential output

src/converter/one/convert/FlameGraph.java
apangin: I mean, not HTML format but the FlameGraph as a structure. `one.convert.Frame` object, to be precise.

src/converter/Main.java
Baraa-Hasheesh: I agree with @roy-soumadipta 

The reason I initially raised this is that I can a user doing this by mistake by pressing on the `tab` to auto complete the file name & incorrectly passing the same file twice 

& then a couple of minutes of analyzing how "this" happened & finally finding that the user simply ran the incorrect command

I don't see passing a --diff for the same file as a valid use case & this check will save users some time & sanity 
If the user wants to explicitly do it, the user can go through the extra effort of creating a copy of the file with a different name

olivergillespie:
Cool :)

It would be useful to have a note in the output specifying the two input files, ideally also mentioning which way the diff works.
```
Red: more in b.jfr
Blue: more in a.jfr
``` 
or the title of the flamegraph could be:

```
Diff (before: a.jfr, after: b.jfr)
```

or


```
Diff (baseline: a.jfr, comparison: b.jfr)
```

and I think it's probably intuitive that red means more in 'after' or 'comparison' (i.e. it got worse).

At the very least, documentation of the `--diff` argument should clarify the ordering.

olivergillespie:
Can you please share the two input files used to generate the example diff output? I'm a bit confused, for example `ParseGenerator::generate` shows `8 samples, 0.56%, +4.36%`. Shouldn't the frame be at least as wide as the delta?

olivergillespie:
I have my own diff implementation, and there I colour based on the delta relative to that frame's size, not the total. For example, if profiles a and b have 10,000 samples each and frame `foo::new` is only in profile b, 1 sample: the current logic says the delta is 0.01%, so the colour will be light red (depending on max delta). In my version, the relative delta is 100% so that frame is dark red. The width of the frame is already shown, which I think is naturally visually included in the weight of the colouring. I.e. a small dark red frame has similar visual weight to a large light red frame, so IMO they should represent the same 'amount' of diff samples. I think it works better in practice, personally. Another way of looking at it - imagine profile b is entirely distinct from profile a, no overlap. What colour should profile b be? In my view, entirely one shade of red, it's all equally brand new. Bigger new stuff is more important, as usual, and shown by the width.

My other concern is doing things relative to the max delta. There are obvious edge cases like if the profiles are identical except some big frame has one more sample, it will be bright red. My tool colours relative to 100%, not the max delta. That way two mostly similar profiles end up mostly white, which I think is the intuitive outcome.

olivergillespie:
In my tool I found it useful to add the two input profiles to set the shape of the diff profile, instead of just using the shape of the 'after' profile like you have here. That way, frames which get significantly smaller or entirely disappear are still shown in the diff, no need to run the diff in reverse. It's true that the new shape doesn't represent either state, but I haven't found that to be a problem. For me, better to have the full context in one view of both things that have increased and decreased.

apangin:
> I colour based on the delta relative to that frame's size, not the total

Yeah, I've already discussed that with @roy-soumadipta.
We'll reimplement this feature using a different approach.

roy-soumadipta:
Closing this PR in favor of an upcoming different implementation.

Issue #1533, opened 2025-10-13
Deprecate AsyncGetCallTrace


apangin:
### Describe the feature

In line with our strategy to elimiate all usages of `AsyncGetCallTrace` API in async-profiler, I propose to make `cstack=vm` the default stack walking mode and then deprecate `fp` and `dwarf` stack walking modes with the intention to remove them completely in future versions.

### Use Case

This is a follow up to #795 and #1073.

### Proposed Solution

`cstack=vm` already provides a superset of AsyncGetCallTrace capabilities, including all the tricks to recover from edge cases. At the same time, `cstack=vm` is safer and richer in functionality.

We will continue using vanilla `AsyncGetCallTrace` on Azul Zing only (without any additional stack recovery techniques) until we find an alternative.

### Acknowledgements

- [x] I may be able to implement this feature request
- [x] This feature might incur a breaking change

PR #1537, opened 2025-10-15
Support advanced stack walking features with cstack=vm

apangin:
### Description

This PR is a prerequisite to making `cstack=vm` the default stack walking mode.
It includes the following changes:

1. `vm` and `vmx` are no longer different stack walking modes. It's now a single `vm` mode with an optional `mixed` feature. There is already `-F` argument for choosing various stack walking features, and the ability to display mixed stack (where native frames interleave Java frames) fits this paradigm well. `cstack=vmx` remains for backward compatibility as a shorthand for `cstack=vm,features=mixed`.
2. `StackDetail` enum is gone. Now, I pass `features` argument instead.
3. VMStructs stack walker now supports `vtable` and `comptask` features.
4. For `nativemem` profiles, I chop everything above malloc hook frames, similarly to `fp` and `dwarf` modes.

I haven't changed logic around allocation and lock profiling. They still use JVMTI API for getting stack traces. This will keep familiar user experience after switching to `vm` by default.

### Related issues

#1533, #1370

### Motivation and context

VMStructs should support all stack walking features that `fp` and `dwarf` modes do.

### How has this been tested?

`--cstack vm -F comptask+vtable`

Integration tests will be amended in the subsequent PR that changes default mode.

---

### DIFF
--- a/test/test/vtable/VtableTests.java
+++ b/test/test/vtable/VtableTests.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test.vtable;
+
+import one.profiler.test.*;
+import test.recovery.Numbers;
+import test.recovery.Suppliers;
+
+public class VtableTests {
+
+    @Test(mainClass = Numbers.class, jvm = Jvm.HOTSPOT)
+    public void vtableStubs(TestProcess p) throws Exception {
+        Output out = p.profile("-d 3 -e cpu -i 1ms --cstack vm -F vtable -o collapsed");
+        assert out.contains("Numbers.avg;vtable stub;java.lang.Integer_\\[i]");
+        assert out.contains("Numbers.avg;vtable stub;java.lang.Long_\\[i]");
+    }
+
+    @Test(mainClass = Suppliers.class, jvm = Jvm.HOTSPOT)
+    public void itableStubs(TestProcess p) throws Exception {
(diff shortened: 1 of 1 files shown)


src/stackWalker.cpp
```diff
@@ -385,11 +396,20 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                 }
                 continue;
             } else {
+                if (features.vtable_target && nm->isVTableStub() && depth == 0) {
```
Baraa-Hasheesh, line 399: can we add a test for this?

Baraa-Hasheesh:
> I haven't changed logic around allocation

But wouldn't that keep the known issue of allocation profiling not collecting native samples for `alloc` samples when `vm` is selected?
For example the following samples 

image

image

When using VM stack walker they won't have these native frames,
This is because we're skipping native portion for `vm` on the `Profiler::getNativeTrace` function & only collect the java samples using the JVMTI stack walker 

You can test this change against the `AllocTests.startup` I think it will fail if you enable `vm`

src/stackWalker.cpp
apangin: Added tests for `vtable stub` and `itable stub`

apangin:
@Baraa-Hasheesh You are right: ideally, allocation profiling should use VMStructs stack walker too. But let's do it in a separate PR.

Baraa-Hasheesh:
> @Baraa-Hasheesh You are right: ideally, allocation profiling should use VMStructs stack walker too. But let's do it in a separate PR.

That's should be fine

Issue #1534, opened 2025-10-14
Stress testing JVM Instrumentation may cause a warning message during VM shutdown


fandreuz:
### Describe the bug

While instrumenting a significant number of classes, Async-Profiler may cause the following warnings:
```
[3.190s][warning][gc,alloc] DestroyJavaVM: Stall for VM-Shutdown timed out; allocation may fail with OOME
[ERROR] JVMTI_ERROR_OUT_OF_MEMORY occurred while calling RetransformClasses
```

The first comes from [stall_for_vm_shutdown](https://github.com/openjdk/jdk/blob/a3ee821f38d0d04f982465f5157170249c92cc49/src/hotspot/share/gc/shared/collectedHeap.cpp#L616), which is called when an allocation fails while the VM is shutting down. This logic was introduced in [JDK-8366865](https://bugs.openjdk.org/browse/JDK-8366865), which is why we never see this failure in our GHA.

The second comes from Async-Profiler: our call to `RetransformClasses` fails due to `JVMTI_ERROR_OUT_OF_MEMORY` (allocation failed). Thus, Async-Profiler dutifully reports the warning.

### Expected vs. actual behavior

Ideally `Instrument::stop` should never fail and never trigger a warning. This behavior is not a concern though.

The reason why the test fails is that we're particularly strict in the test verification, which is done to catch verification errors after bytecode rewriting: https://github.com/async-profiler/async-profiler/blob/fb673227c7fb311f872ce9566769b006b357ecbe/test/test/instrument/InstrumentTests.java#L239-L245

### Reproduction Steps

Non deterministic behavior on JDK > 25. Happens in tests like this:
https://github.com/async-profiler/async-profiler/blob/fb673227c7fb311f872ce9566769b006b357ecbe/test/test/instrument/InstrumentTests.java#L43-L57

### Additional Information/Context

```
#0  CollectedHeap::stall_for_vm_shutdown (this=0x7ffff22b3bc0) at /jdk/src/hotspot/share/gc/shared/collectedHeap.cpp:617
#1  0x00007ffff59bfb19 in CollectedHeap::satisfy_failed_metadata_allocation (this=0x7ffff22b3bc0, loader_data=0x7fffcfb416c0,
    word_size=14, mdtype=Metaspace::NonClassType) at /jdk/src/hotspot/share/gc/shared/collectedHeap.cpp:390
[...]

PR #1538, opened 2025-10-15
Skip `Instrument::stop` in `VMDeath` callback

fandreuz:
### Description
`Instrument::stop` can be skipped when it is called as part of a `VMDeath` JVMTI callback.

### Related issues
#1534

### Motivation and context
We noticed some test failures in the stress tests related to bytecode rewriting.

### How has this been tested?
n/a

---

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -32,7 +32,7 @@ int VM::_hotspot_version = 0;
 bool VM::_openj9 = false;
 bool VM::_zing = false;
 
-bool VM::_jvm_dying = false;
+bool VM::_terminating = false;
 
 GetCreatedJavaVMs VM::_getCreatedJavaVMs = NULL;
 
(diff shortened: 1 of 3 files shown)


src/vmEntry.cpp
```diff
@@ -419,6 +419,7 @@ void JNICALL VM::VMInit(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
 }
 
 void JNICALL VM::VMDeath(jvmtiEnv* jvmti, JNIEnv* jni) {
+    Profiler::instance()->set_jvm_dying();
```
apangin: Why in `Profiler`? You're in the `VM` class, and the flags makes sense for the VM.

fandreuz: Makes sense, fixed in e460841b0d83f9701012f078210be9807ead8791

PR #1539, opened 2025-10-15
Use VMStructs stack walking mode by default

apangin:
### Description

Make `cstack=vm` the default stack walking mode on HotSpot JVM.

### Related issues

#1533

### Motivation and context

We've been using `cstack=vm` in production for a while and it proved stable.
Recent changes (e.g. #1449, #1517, #1537) significantly improved quality of VMStructs stack traces, which is now superior to AsyncGetCallTrace. There are only advantages in using VMStructs, so it is a good time to change the defaults, thus making another step towards AsyncGetCallTrace deprecation.

### How has this been tested?

All existing tests pass with no changes, except for `recovery` tests that check soon-to-be-deprecated `safemode` option.

---


Baraa-Hasheesh:
This looks good to me 

The only comment I have is out of scope being it would properly be a good idea to prefer dwarf to FP when it's available on pure native apps

PR #1546, opened 2025-10-17
Fix Duplicated native stacks when mixed feature is used for FP & Dwarf stack walkers

Baraa-Hasheesh:
### Description
Bad usage from user may result in duplicated native stacks
This can be seen by doing `start,event=cpu,cstack=fp,features=mixed`

### Related issues
N/A

### Motivation and context
Prevent duplication of native stacks 

### How has this been tested?
manual testing 

---

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -77,7 +77,7 @@ const size_t EXTRA_BUF_SIZE = 512;
 //     interval=N              - sampling interval in ns (default: 10'000'000, i.e. 10 ms)
 //     jstackdepth=N           - maximum Java stack depth (default: 2048)
 //     signal=N                - use alternative signal for cpu or wall clock profiling
-//     features=LIST           - advanced stack trace features (vtable, comptask, pcaddr)"
+//     features=LIST           - advanced stack trace features (mixed, vtable, comptask, pcaddr)"
 //     safemode=BITS           - disable stack recovery techniques (default: 0, i.e. everything enabled)
 //     file=FILENAME           - output file name for dumping
 //     log=FILENAME            - log warnings and errors to the given dedicated stream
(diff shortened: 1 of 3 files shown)


apangin:
Good catch. But I think it's better to disallow this combination at all to avoid confusion.
Let's throw an error if `mixed` is specified with a mode other than CSTACK_VM.
Alternatively, we can hide `mixed` feature for internal use only, and let user choose between `vm` or `vmx` modes as before.

Baraa-Hasheesh:
> Alternatively, we can hide `mixed` feature for internal use only, and let user choose between `vm` or `vmx` modes as before.

I prefer this approach, this makes enabling the vmx possible via only a single path being `cstack=vmx` rather than 2 paths being 
cstack=vmx
cstack=vm,features=mixed

Baraa-Hasheesh:
@apangin mixed feature is now internal & user can't directly control it directly via feature arguments,
-> It will still be reported by the JFR in the features string as an indication for `vmx`

Baraa-Hasheesh:
Failure unrelated to change, it's the `JfrTests.parseMultiModeRecording` test being flakey again :(

apangin:
> I prefer this approach, this makes enabling the vmx possible via only a single path being cstack=vmx rather than 2 paths

Even though I proposed this idea, after rethinking I don't like it anymore.
It contradicts the previous intention declared in #1537 to unify `vm` and `vmx` modes. This is essentially the same mode with additional opt-in feature, and that matches how this mode is implemented. After `vm` stack walker became the default, other stack walking modes lost their significance: there are no advantages of selecting `fp` or `dwarf` over `vm`, and nobody actually used `lbr` since appearance of `dwarf`. Consequently, the whole concept of "stack walking modes" is not relevant anymore.

On the other hand, the concept of "stack walking features" expands, and users can now mix and match various featues (`vtable`, `comptask`, `pcaddr`) depending on the use case. "Mixed" stack traces (intermediate frames) fits this concept well. So I would argue that we'd better leave `-F mixed` instead of `--cstack vmx`.

We already introduced `-F mixed` in the version 4.2, so we already have two ways of doing the same thing. This is fine: `--cstack vmx` is left for backward compatibility during the transitional period, but it's the one that I'd prefer to eventually remove.

Baraa-Hasheesh:
My concern here is  I can see the following use case being problematic 

`start,event=cpu,features=mixed`

Now if the process is running on a hotspot JVM with vmstructs, then this will work fine, 
but if the JVM doesn't such as Azul Zing then this command will fail if we add the error checking

Should we add logic to ignore the `mixed` feature fully if default stack walker & no vm structs?

PR #1552, opened 2025-10-21
Parse FlameGraph title from HTML input

apangin:
### Description

Automatically extract FlameGraph title when parsing HTML source.
Without this change, original title is lost during convertion (it is always replaced with the default "Flame Graph", unless configured manually with the `--title` option).

### Motivation and context

Preserve original title when doing `jfrconv` manipulations, e.g., when upgrading an existing flame graph to a new template.

---

### DIFF
--- a/src/converter/one/convert/FlameGraph.java
+++ b/src/converter/one/convert/FlameGraph.java
@@ -74,8 +74,8 @@ public class FlameGraph implements Comparator<Frame> {
 
         try (BufferedReader br = new BufferedReader(in)) {
             for (String line; !(line = br.readLine()).startsWith("const cpool"); ) {
-                if (line.startsWith("<h1>")) {
-                    title = line.substring(4, line.lastIndexOf("</h1>"));
+                if (line.startsWith("<h1")) {
+                    title = line.substring(line.indexOf('>') + 1, line.lastIndexOf("</h1>"));
                 }
             }
             br.readLine();


src/converter/one/convert/FlameGraph.java
```diff
@@ -71,7 +73,11 @@ public void parseHtml(Reader in) throws IOException {
         boolean needRebuild = args.reverse || args.include != null || args.exclude != null;
 
         try (BufferedReader br = new BufferedReader(in)) {
-            while (!br.readLine().startsWith("const cpool")) ;
+            for (String line; !(line = br.readLine()).startsWith("const cpool"); ) {
+                if (line.startsWith("<h1>")) {
+                    title = line.substring(4, line.lastIndexOf("</h1>"));
```
Baraa-Hasheesh: how about adding a break here to terminate the loop early?

src/converter/one/convert/FlameGraph.java
```diff
@@ -71,7 +73,11 @@ public void parseHtml(Reader in) throws IOException {
         boolean needRebuild = args.reverse || args.include != null || args.exclude != null;
 
         try (BufferedReader br = new BufferedReader(in)) {
-            while (!br.readLine().startsWith("const cpool")) ;
+            for (String line; !(line = br.readLine()).startsWith("const cpool"); ) {
+                if (line.startsWith("<h1>")) {
```
Baraa-Hasheesh: How about checking for `<h1` 
My reasoning is that we might add html attributes to the header in the future
Checking `<h1>` exactly will make this flow never execute as expected if we add an attribute

src/converter/one/heatmap/Heatmap.java
```diff
@@ -91,7 +91,7 @@ public void dump(PrintStream stream) throws IOException {
         stream.print('E');
 
         tail = ResourceProcessor.printTill(stream, tail, "/*title:*/");
-        stream.print(args.title == null ? "Heatmap" : args.title);
+        stream.print(args.title != null ? args.title : "Heatmap");
```
Baraa-Hasheesh, line 94: Was there a reason for the switch of condition

on a different note I noticed this change (making args.title null by default) solves an old issue where the title in the heatmap is reported as "Flame Graph"

`<h1 id="title">Flame Graph</h1>`

src/converter/one/convert/FlameGraph.java
apangin: `break` will break the logic :)
This loop terminates (intentionally) only when it reaches `const cpool`, because the next loop is to parse constant pool.

Baraa-Hasheesh: ah I see, that makes sense

src/converter/one/heatmap/Heatmap.java
apangin: Nothing important, just wanted it to match the similar code in FlameGraph.java.

src/converter/one/convert/FlameGraph.java
apangin: OK, made it more flexible.

PR #1553, opened 2025-10-21
Differential Flame Graphs

apangin:
### Description

Add `--diff` option to produce differential Flame Graph from two input files: left ("baseline") and right ("new").
Each of the input files can be in `collapsed`, `html` or `jfr` format.

Remaining tasks:

- [x] Dedicated color for "new" frames (missing in the baseline profile).
- [x] Display diff% in the tooltip.
- [x] Add integration tests.
- [x] Update documentation.

### Related issues

#1330

### How has this been tested?

Added two integration tests that verify differential flame graphs in collapsed and html formats.

---

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -15,6 +15,7 @@ header:
       - 'src/jattach'
       - 'src/res'
       - '**/MANIFEST.MF'
+      - 'test/**/*.collapsed'
     license:
       content: |
         Copyright The async-profiler authors


src/converter/one/convert/Main.java
```diff
@@ -104,6 +133,7 @@ private static void usage() {
                 "\n" +
                 "Conversion options:\n" +
                 "  -o --output FORMAT    Output format: html, collapsed, pprof, pb.gz, heatmap, otlp\n" +
+                "     --diff             Create differential Flame Graph from two input files\n" +
```
Baraa-Hasheesh, line 138: Let's also update the documentation in the `ConverterUsage.md`

On a different note I can see it's missing some options like trace, nativelock

src/converter/one/convert/Main.java
```diff
@@ -35,6 +36,34 @@ public static void main(String[] argv) throws Exception {
...
+
+            String input1 = args.files.get(0);
+            String input2 = args.files.get(1);
+            String output = isDirectory ? new File(lastFile, replaceExt(input2, "diff." + args.output)).getPath() : lastFile;
+
+            System.out.print("Converting " + getFileName(input2) + " vs " + getFileName(input1) + " -> " + getFileName(output) + " ");
+            System.out.flush();
+
+            long startTime = System.nanoTime();
+            FlameGraph base = parseFlameGraph(input1, args);
```
Baraa-Hasheesh, line 57: If the input file is `otlp` this would end up reading it as collapsed

Let's add an exception for now, until otlp is conversion is supported

src/converter/one/convert/Main.java
```diff
@@ -35,6 +36,34 @@ public static void main(String[] argv) throws Exception {
             }
         }
 
+        if (args.diff) {
```
Baraa-Hasheesh, line 39: Is it possible to provide some output samples here

krk:
Diff result is missing items from one side. If this is intentional, can we consider the unified approach mentioned in https://github.com/async-profiler/async-profiler/pull/1536#issuecomment-3421970978?

left.txt:
```
A 30
A;B 20
A;B;C 10
D 40
```

right.txt:
```
A 30
A;B 19
A;B;C 11
E 50
```

```
$ASPROF_LIB/../bin/jfrconv -o html left.txt --diff right.txt out.html
Converting right.txt vs left.txt -> out.html # 0.022 s
```

(which file argument is the base is unclear to me).

Result does not have `D`:

image

About the colors, it is hard to read black text over these shades of blue and red.

apangin:
> Diff result is missing items from one side. If this is intentional

Yes, this is intentional. This implementation follows the spirit of [classic differential flame graphs](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) by Brendan Gregg. The size and shape of the flame graph is the same as a flame graph for the right ("new") profile. To see what frames have disappeared, build a reverse diff (swapping left and right arguments). In a combined flame graph, width of each frame does not carry meaningful semantics.

> About the colors, it is hard to read black text over these shades of blue and red.

This again was inspired by Brendan's differential flame graphs. I'm open to alternatives, though.

krk:
Thanks. For the colors, lighter shades would do. This can be checked from https://webaim.org/resources/contrastchecker/ (or visually).

e.g. `#8F8FFF` for blue and `#E35F5F` for red.

apangin:
Adjusted colors, added section in ConverterUsage.md doc, added integration tests.
Reviews are welcome.

olivergillespie:
Cool, glad to have this!

A couple of observations from trying it out.

* Colour is relative to max diff. That means two extremely similar profiles can still have bright red and blue, which to me is misleading.
* Diff works on self time for the frame, not total time.

In my own diff tool, I colour relative to the frame's own value. So a frame going from 1 to 2 samples has the same red as a frame going from 50 to 100. I think visually the width of the frame acts as a multiplier of the colour, so saturation x width = total change.
I also use frame total time, not self time. Combined with the above, I think it's still clear where the leaf 'source' of a change is, but it also gives a nice rollup effect for seeing the combined effect of multiple changes across children.

I also normalize for time range (and actually total sample count, though that's a bit more controversial, there is some benefit in seeing the whole profile red if traffic volume increased uniformly), not sure if that's a use-case you're interested in.

I already [mentioned](https://github.com/async-profiler/async-profiler/pull/1536#issuecomment-3421970978) my preference for a unified/summed profile shape, I understand it's not your preference, but if you only show one then it could be nice to be able to flip direction easily, not sure if the implementation is amenable to that. My own tool shows baseline on the left, comparison on the right and combined/unified diff in the middle with navigation synced between the 3, as I found that it's useful for new users to be able to reference the original profiles when understanding the diff.

PR #1558, opened 2025-10-22
Fix record-cpu bug in various stack walkers when kernel stacks are not available 

Baraa-Hasheesh:
### Description
Previously `record-cpu` would not have been reported for FP, DWARF & VM stack walkers if used explicitly & kernel stacks aren't available 

`if (_kernel_stack || _cstack == CSTACK_DEFAULT || _cstack == CSTACK_LBR)` 

if `_kernel_stack` was false then only default & LBR stack walkers could record CPU, for other cases the perf_event_mmap_page would not get created

### Related issues
N/A

### Motivation and context
Allow the stack walkers to record cpu even when kernel stacks are missing 

### How has this been tested?
manual testing

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -534,9 +534,9 @@ PerfEvent* PerfEvents::_events = NULL;
 PerfEventType* PerfEvents::_event_type = NULL;
 bool PerfEvents::_alluser;
 bool PerfEvents::_kernel_stack;
+bool PerfEvents::_use_perf_mmap;
 bool PerfEvents::_record_cpu;
 int PerfEvents::_target_cpu;
-bool PerfEvents::_use_perf_mmap;
 
 int PerfEvents::createForThread(int tid) {
     if (tid >= _max_events) {
--- a/src/perfEvents.h
+++ b/src/perfEvents.h
@@ -22,9 +22,9 @@ class PerfEvents : public CpuEngine {
     static PerfEventType* _event_type;
     static bool _alluser;
     static bool _kernel_stack;
+    static bool _use_perf_mmap;
     static bool _record_cpu;
     static int _target_cpu;
-    static bool _use_perf_mmap;
 
     static u64 readCounter(siginfo_t* siginfo, void* ucontext);
     static void signalHandler(int signo, siginfo_t* siginfo, void* ucontext);


src/perfEvents_linux.cpp
```diff
@@ -621,7 +621,7 @@ int PerfEvents::createForThread(int tid) {
     }
 
     void* page = NULL;
-    if (_kernel_stack || _cstack == CSTACK_DEFAULT || _cstack == CSTACK_LBR) {
+    if (_kernel_stack || _cstack != CSTACK_NO) {
```
apangin: This is not the right fix, unfortunately.
The idea of the condition here was to avoid creating locked memory mappings unless absolutely necessary, because this is a limited resource. See the discussion in #550.
The `--record-cpu` option added one more case when this mapping is "absolutely necessary", however the condition has never been updated to account for this option.

Baraa-Hasheesh: @apangin Thanks I wasn't aware of that discussion 
I will adjust the condition to account for `_record_cpu` as well

Baraa-Hasheesh: Updated here [4696592](https://github.com/async-profiler/async-profiler/pull/1558/commits/469659293ff940da3ed77b19067f4550c0ccc2e1)

src/perfEvents_linux.cpp
```diff
@@ -621,7 +621,7 @@ int PerfEvents::createForThread(int tid) {
     }
 
     void* page = NULL;
-    if (_kernel_stack || _cstack == CSTACK_DEFAULT || _cstack == CSTACK_LBR) {
+    if (_kernel_stack || _cstack == CSTACK_DEFAULT || _cstack == CSTACK_LBR || _record_cpu) {
```
apangin: Condition looks correct now, thanks, but it's getting long. To avoid recomputing it each time, I'd introduce a new variable (like `_use_perf_mmap`), which will be set only once in `PerfEvents::start`.

Baraa-Hasheesh: makes sense applied here [a2cc201](https://github.com/async-profiler/async-profiler/pull/1558/commits/a2cc20141e22c3dec6bc637f44b121e8fd2dc2df)

Baraa-Hasheesh:
> Can you add a new test for this fix? Or you can just amend existing `perfEventsRecordCpuEventsCount` test to run profiling one more time with `--all-user` option. (I noticed that adding `--all-user` also breaks this test, regardless of sysctl settings).

right, that will also cause the test to fail as you mentioned, I added the new profiling session as you suggested in the test here [a2cc201](https://github.com/async-profiler/async-profiler/pull/1558/commits/a2cc20141e22c3dec6bc637f44b121e8fd2dc2df)

Issue #1564, opened 2025-10-24
[GHA] Tests in `ApiTests` use the wrong Async-Profiler artifact


fandreuz:
### Describe the bug

Since corretto now ships with Async-Profiler included, it looks like some tests in our GHA are testing the wrong Async-Profiler version.

https://github.com/fandreuz/async-profiler/actions/runs/18781125495/job/53587389610#step:9:215

This happens because:
- The Async-Profiler JAR we use in GHA does not embed `libasyncProfiler.so`
- in `AsyncProfiler.java` we let `System.load` decide where to load Async-Profiler from
- `System.load` gives priority to libs coming with the JDK. 
 
Tests using `agentArgs` are unaffected because the artifact path is provided manually there.

### Expected vs. actual behavior

We should test `master`, but we're testing the Async-Profiler version released with corretto.

### Reproduction Steps

Run Async-Profiler test suite the latest corretto release for any JDK version: `make test-java TESTS=api`

### Async-profiler version

master

fandreuz:
I think chasing all usages of `AsyncProfiler.getInstance()` in our tests and setting the path explicitly would be cumbersome and not really a long-term solution.
- We could have an env variable `ASPROF_LIB` which `AsyncProfiler#getInstance` checks as part of its routine
- We could change our GHA to use a JAR which embeds `libasyncProfiler.so`

fandreuz:
Fixed in #1565.

PR #1565, opened 2025-10-24
Force test runs to use the correct Async-Profiler library

fandreuz:
### Description
In this PR I add the system property `one.profiler.embeddedLibraryPath`, which lets user of `AsyncProfiler` override the path to the embedded `libasyncProfiler.so`. We can use it to load `libasyncProfiler.so` from the classpath during tests, so it does not fall back to the binary coming from Corretto.

### Related issues
#1564 

### Motivation and context
Test runs should not use binaries coming from Corretto.

### How has this been tested?
GHA and new test.

---

### DIFF
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -141,7 +141,7 @@ public class TestProcess implements Closeable {
                 cmd.add("-XX:+DebugNonSafepoints");
             }
             cmd.add("-Done.profiler.libraryPath=" + System.getProperty("one.profiler.libraryPath", profilerLibPath()));
-            cmd.add("-Djava.library.path=" + System.getProperty("java.library.path", testLibPath()));
+            cmd.add("-Djava.library.path=" + testLibPath());
             cmd.add("-ea");
             addArgs(cmd, test.jvmArgs());
             if (!test.agentArgs().isEmpty()) {


test/test/api/ApiTests.java
```diff
@@ -11,25 +11,33 @@
...
         assert !out.contains("BusyLoops.method1");
         assert out.contains("BusyLoops.method2");
         assert !out.contains("BusyLoops.method3");
     }
+
+    // https://github.com/async-profiler/async-profiler/issues/1564
+    @Test(mainClass = PrintProfilerVersion.class, output = true)
+    public void version(TestProcess p) throws Exception {
+        String actual = p.waitForExit(TestProcess.STDOUT).toString().trim();
+        String expected = System.getenv("PROFILER_VERSION");
```
apangin: It's a bad test - at least, in its current shape.
A test must be self-sufficient, it must not depend on variables set outside. First, who guarantees that a the environment variable has the correct value? Second, if a test is run standalone, it will immediately crash with NPE.
We cannot require users/developers/vendors to run the test from Makefile only. Even in our company async-profiler tests are run by custom scripts bypassing Makefile.

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -57,9 +57,15 @@ public static synchronized AsyncProfiler getInstance(String libPath) {
     }
 
     private static File extractEmbeddedLib() {
-        String resourceName = "/" + getPlatformTag() + "/libasyncProfiler.so";
-        InputStream in = AsyncProfiler.class.getResourceAsStream(resourceName);
+        String embeddedLibraryPath = System.getProperty("one.profiler.embeddedLibraryPath");
```
apangin: Why "embedded"? Simply `libraryPath`.
Furthermore, if `one.profiler.libraryPath` is set and the specified file exists, we can just return this file as is without extraction. But if the file does not exist, I'd fallback to the default resource name instead of throwing an error.

test/test/api/ApiTests.java
fandreuz: Should we even have this test? It's not really testing anything other than the test infrastructure, and #1564 is probably a one-off thing?

apangin: > Should we even have this test?

That’s exactly what I was getting at :) I think we don't need it - for the reasons you've mentioned.

src/api/one/profiler/AsyncProfiler.java
fandreuz: > But if the file does not exist, I'd fallback to the default resource name instead of throwing an error.

If the property is specified, I expect it to be an error that the file cannot be found in the classpath. As a user I would want this to fail loudly.

fandreuz:
Test failure looks unrelated.

Baraa-Hasheesh:
I'm not sure what's the best location for this but I do believe the new property should be documented

fandreuz:
> I'm not sure what's the best location for this but I do believe the new property should be documented

I don't think we should make it public, have we ever seen the need for such a thing? Users typically use the JAR we distribute on maven, which embeds the `.so` in a known location. Alternatively, they can provide the lib path to the ctor of `AsyncProfiler`.

Baraa-Hasheesh:
> I don't think we should make it public, have we ever seen the need for such a thing? Users typically use the JAR we distribute on maven, which embeds the `.so` in a known location. Alternatively, they can provide the lib path to the ctor of `AsyncProfiler`.

While that is true, that's not really a guarantee, 
Let's say I'm a user & saw an interesting new feature that is available on the nightly release & downloaded the binaries & jar to try this new feature

I might do the following setup to launch the profiler

`java -cp ${PATH_TO_PROFILER_JAR} -java.library.path=${PATH_TO_PROFILER_SO}` 

This setup will have the same issue we have in our tests if it's using latest Corretto, no?

fandreuz:
> > I don't think we should make it public, have we ever seen the need for such a thing? Users typically use the JAR we distribute on maven, which embeds the `.so` in a known location. Alternatively, they can provide the lib path to the ctor of `AsyncProfiler`.
> 
> While that is true, that's not really a guarantee, Let's say I'm a user & saw an interesting new feature that is available on the nightly release & downloaded the binaries & jar to try this new feature
> 
> I might do the following setup to launch the profiler
> 
> `java -cp ${PATH_TO_PROFILER_JAR} -java.library.path=${PATH_TO_PROFILER_SO}`
> 
> This setup will have the same issue we have in our tests if it's using latest Corretto, no?

Yes, we should probably document that `libasyncProfiler.so` in `java.library.path` will be ignored on new releases of Corretto. But we don't need to make this system property public since there's already a workaround.

src/api/one/profiler/AsyncProfiler.java
apangin: Getting `IllegalArgumentException` when calling a method that has no arguments is... strange, to say the least.

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -57,17 +54,31 @@ public static synchronized AsyncProfiler getInstance(String libPath) {
...
+        if (!url.getPath().isEmpty()) {
+            return new File(url.getPath());
+        }
+
         try {
             String extractPath = System.getProperty("one.profiler.extractPath");
             File file = File.createTempFile("libasyncProfiler-", ".so",
                     extractPath == null || extractPath.isEmpty() ? null : new File(extractPath));
-            try (FileOutputStream out = new FileOutputStream(file)) {
+            file.deleteOnExit();
```
apangin: `deleteOnExit` does not work in many cases: if `System.halt` is called, if the process is terminated by the OOM killer, if JVM crashes. It's better to not leave garbage; deleting temporary .so after immediately loading it is a common practice.

src/api/one/profiler/AsyncProfiler.java
fandreuz: Replaced with `Runtime`: e69daf131fa10a22ad156ad87270b70243978f49

src/api/one/profiler/AsyncProfiler.java
fandreuz: 014d8ceeb1e797b2a51c94308fe4f7344cb449fb

test/one/profiler/test/TestProcess.java
```diff
@@ -111,8 +111,12 @@ public Os currentOs() {
         return this.currentOs;
     }
 
+    private String profilerLibFileName() {
+        return "libasyncProfiler." + currentOs.getLibExt();
+    }
+
```
apangin: This change is no longer needed?

src/api/one/profiler/AsyncProfiler.java
```diff
@@ -39,16 +39,22 @@ public static synchronized AsyncProfiler getInstance(String libPath) {
...
             } catch (UnsatisfiedLinkError e) {
-                File file = extractEmbeddedLib();
-                if (file != null) {
-                    try {
-                        System.load(file.getAbsolutePath());
-                    } finally {
-                        file.delete();
-                    }
+                String libraryPath = System.getProperty("one.profiler.libraryPath");
+                if (libraryPath != null) {
```
apangin: `&& !libraryPath.isEmpty()`

test/one/profiler/test/TestProcess.java
```diff
@@ -140,6 +144,7 @@ private List<String> buildCommandLine(Test test) {
                 cmd.add("-XX:+UnlockDiagnosticVMOptions");
                 cmd.add("-XX:+DebugNonSafepoints");
             }
+            cmd.add("-Done.profiler.libraryPath=" + System.getProperty("one.profiler.libraryPath"));
             cmd.add("-Djava.library.path=" + System.getProperty("java.library.path"));
```
apangin: Does it make sense to provide default paths here? E.g.
```
cmd.add("-Done.profiler.libraryPath=" + System.getProperty("one.profiler.libraryPath", profilerLibPath()));
```
This will eliminate need to pass the property manually when running test standalone, while keeping possibility to override the path with `-D` in the command line. This will also simplify Makefile command to run tests.

test/one/profiler/test/TestProcess.java
fandreuz: 0a81f37232c3d276fce7b01064752b6deabc1449

src/api/one/profiler/AsyncProfiler.java
fandreuz: 0a81f37232c3d276fce7b01064752b6deabc1449

test/one/profiler/test/TestProcess.java
fandreuz: Yeah that seems reasonable: 0a81f37232c3d276fce7b01064752b6deabc1449

test/one/profiler/test/TestProcess.java
```diff
@@ -140,6 +140,7 @@ private List<String> buildCommandLine(Test test) {
                 cmd.add("-XX:+UnlockDiagnosticVMOptions");
                 cmd.add("-XX:+DebugNonSafepoints");
             }
+            cmd.add("-Done.profiler.libraryPath=" + System.getProperty("one.profiler.libraryPath", profilerLibPath()));
```
apangin, line 143: Good, thanks!
Now, the same for `java.library.path` below? Use `testLibPath()` as the default.

fandreuz: 35b1c7475d16123a5b28d6f09c272a6d6f706085

fandreuz: Actually, the default value of `java.library.path` is `/usr/java/packages/lib:/usr/lib/x86_64-linux-gnu/jni:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/usr/lib/jni:/lib:/usr/lib`, not null.

This caused failures, see failed jobs in GHA. We could set it to empty manually, but I'm not sure that would be less ugly than what we do now.

apangin: Ah, you are right. Maybe then always set it to `testLibPath()` here instead of the Makefile?

fandreuz: yeah that makes sense

fandreuz: d7bbdb15dd121019dad12922c2da3d0126f22153

PR #1460, opened 2025-09-01
Expose profiler metrics via API

adinazugravescu:
### Description
Add `getMetrics()` API to expose internal metrics including memory usage for call trace storage, flight recording, dictionaries, code cache, and discarded samples count.

### Related issues
N/A

### Motivation and context
N/A

### How has this been tested?
N/A

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #1568, opened 2025-10-29
Expose Async-Profiler metrics

fandreuz:
### Description
In this PR I pick up the work we drafted in #1460. I made some small changes to the initial work proposed by @adinazugravescu.

The new API is a simple command (`metrics`) which returns Async-Profiler metrics in Prometheus format:
```
calltracestorage_bytes 10489856
flightrecorder_bytes 0
classmap_bytes 4104
threadfilter_bytes 65536
runtimestubs_bytes 35047
nativelibs_bytes 10140787
samples_total 0
sample_failures_total 0
calltracestorage_overflows_total 0
```

### Related issues
#1460 

I stumbled on #1564 while working on this.

### Motivation and context
Many products in the observability ecosystem provide some form of "self-observability": like the observed software, observability tools have some footprint on the machine resources. Having metrics on such footprint provides a full picture of the observed system.

### How has this been tested?
New GHA test.

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1460,7 +1460,7 @@ void Profiler::writeMetrics(Writer& out) {
 
     out << "samples_total " << _total_samples << '\n';
     out << "samples_skipped_total " << _failures[-ticks_skipped] << '\n';
-    out << "overflows_calltracestorage_total " << _call_trace_storage.overflow() << '\n';
+    out << "calltracestorage_overflows_total " << _call_trace_storage.overflow() << '\n';
 
     if (_total_stack_walk_time != 0) {
         out << "stackwalk_ns_total " << _total_stack_walk_time << '\n';
--- a/test/test/api/MetricsTest.java
+++ b/test/test/api/MetricsTest.java
@@ -24,7 +24,7 @@ public class MetricsTest {
             String[] pair = line.split(" ");
             assert pair.length == 2 : line;
             if (pair[1].startsWith("0")) {
-                assert "samples_skipped_total".equals(pair[0]) || "overflows_calltracestorage_total".equals(pair[0]) : line;
+                assert "samples_skipped_total".equals(pair[0]) || "calltracestorage_overflows_total".equals(pair[0]) : line;
             }
 
             if (pair[0].equals("samples_total")) {


src/arch.h
```diff
@@ -43,6 +43,10 @@ static inline u64 loadAcquire(u64& var) {
     return __atomic_load_n(&var, __ATOMIC_ACQUIRE);
```
fandreuz, line 43: I'll add some documentation on this new feature after we discuss the details in this PR.

src/arguments.cpp
```diff
@@ -163,6 +163,9 @@ Error Arguments::parse(const char* args) {
             CASE("version")
                 _action = ACTION_VERSION;
 
+            CASE("metrics")
+                _action = ACTION_METRICS;
```
fandreuz: Do we want to keep `meminfo`? `metrics` provide the same (and more) information in human readable format

apangin: Good point. We don't have to maintain two actions, it's ok to remove `meminfo`.

src/codeCache.h
```diff
@@ -232,9 +233,15 @@ class CodeCacheArray {
         return __atomic_load_n(&_count, __ATOMIC_ACQUIRE);
     }
 
+    size_t usedMemory() {
+        size_t out = count() * sizeof(CodeCache);
```
apangin: Why not including `sizeof(CodeCache)` directly in `CodeCache::usedMemory()`?

src/profiler.cpp
```diff
@@ -1475,6 +1469,23 @@ void Profiler::printUsedMemory(Writer& out) {
     out << buf;
 }
 
+void Profiler::writeMetrics(Writer& out) {
+    out << "calltracestorage_bytes " << (u64) _call_trace_storage.usedMemory() << '\n';
```
apangin: Maybe keep it in kb for readability?
Note that Linux typically reports memory stats in kB:
```
$ cat /proc/meminfo
MemTotal:       16204964 kB
MemFree:        15573588 kB
MemAvailable:   15567340 kB
Buffers:           13900 kB
Cached:           212264 kB
SwapCached:            0 kB
Active:            73628 kB
Inactive:         201364 kB
```
JVM does too:
```
$ jcmd `pgrep java` GC.heap_info
31427:
 garbage-first heap   total 256000K, used 2853K [0x0000000708a00000, 0x0000000800000000)
  region size 2048K, 1 young (2048K), 0 survivors (0K)
 Metaspace       used 83K, committed 320K, reserved 1114112K
  class space    used 5K, committed 128K, reserved 1048576K
```

src/profiler.cpp
```diff
@@ -1475,6 +1469,23 @@ void Profiler::printUsedMemory(Writer& out) {
...
+    out << "flightrecorder_bytes " << (u64) _jfr.usedMemory() << '\n';
+    out << "classmap_bytes " << (u64) _class_map.usedMemory() << '\n';
+    out << "threadfilter_bytes " << (u64) _thread_filter.usedMemory() << '\n';
+    out << "runtimestubs_bytes " << (u64) _runtime_stubs.usedMemory() << '\n';
+    out << "nativelibs_bytes " << (u64) _native_libs.usedMemory() << '\n';
```
apangin: Can you group metrics of the same category using a prefix? E.g.
```
mem_classmap_kb
mem_codecache_kb
```

src/profiler.cpp
```diff
@@ -1475,6 +1469,23 @@ void Profiler::printUsedMemory(Writer& out) {
...
+
+    out << "samples_total " << _total_samples << '\n';
+    out << "sample_failures_total " << _failures[-ticks_skipped] << '\n';
+    out << "calltracestorage_overflows_total " << _call_trace_storage.getOverflow() << '\n';
```
apangin: No need to add `_total` to everything that has no breakdown.

src/profiler.cpp
```diff
@@ -1475,6 +1469,23 @@ void Profiler::printUsedMemory(Writer& out) {
...
+
+    if (_total_stack_walk_time != 0) {
+        out << "stackwalk_ns_total " << _total_stack_walk_time << '\n';
```
apangin, line 1466: Average time is more useful than total.

src/callTraceStorage.h
```diff
@@ -60,6 +60,7 @@ class CallTraceStorage {
     void clear();
     u32 capacity();
     size_t usedMemory();
+    u64 getOverflow() { return loadRelaxed(_overflow); }
```
apangin: why not just `return _overflow`?

apangin: BTW, I prefer getters without prefix, like
```
u64 overflows() { return _overflows; }
```

src/codeCache.h
```diff
@@ -232,9 +233,15 @@ class CodeCacheArray {
...
+        return out + _used_memory;
+    }
+
     void add(CodeCache* lib) {
         int index = __atomic_load_n(&_count, __ATOMIC_ACQUIRE);
         _libs[index] = lib;
+        _used_memory += lib->usedMemory();
```
apangin, line 243: Yeah, it's a good idea to cache this 👍

src/callTraceStorage.h
fandreuz: 22820b89c60b5b14b469acd7d004fbda2830bfc0

fandreuz: > why not just return _overflow?

What's the difference between relaxed and a plain read? Do we risk tearing on some platforms?

src/profiler.cpp
fandreuz: Average is the most basic function provided in services like Prometheus. The usual pattern is to return both total and count, so one can compute the instantaneous value of the average for the time series.

I'd rather provide average in addition, rather than as a replacement. Would that work for you?

src/arguments.cpp
fandreuz: 82f88de3717bf12550e49d9e1e6b82658bc28240

src/lockTracer.cpp
```diff
@@ -53,18 +48,20 @@ Error LockTracer::start(Arguments& args) {
...
-    // Intercept Unsafe.park() for tracing contended ReentrantLocks
-    setUnsafeParkEntry(env, UnsafeParkHook);
+    _total_duration = 0;
 
     return Error::OK;
 }
 
-void LockTracer::stop() {
+void LockTracer::stop(bool restart) {
+    if (restart) {
```
Baraa-Hasheesh: @fandreuz I think you bundled multiple changes under the same PR by mistake

fandreuz: Ops, thanks: eac57a30604ab06d9c7624be307a3eb8e11cc7cb

src/callTraceStorage.h
fandreuz: Discussed offline

src/codeCache.h
fandreuz: Makes sense: 42d2c7a111931e0bb8d497a72f312cf271458287

src/profiler.cpp
fandreuz: d47aa6cd8a991c44aa5809935f7ec177e7fdfcab

src/profiler.cpp
fandreuz: d47aa6cd8a991c44aa5809935f7ec177e7fdfcab

src/profiler.cpp
fandreuz: It's a convention in [metric names](https://prometheus.io/docs/instrumenting/writing_exporters/#naming), if the value is a counter (as opposed to e.g. gauge) then the name is usually suffixed with `_total`. It's clearer to me, values without `_total` are usually gauges

src/profiler.cpp
fandreuz: c442e8a97192ac7190f6e1f699ca671b379502cf

src/profiler.cpp
```diff
@@ -1449,30 +1449,24 @@ Error Profiler::dump(Writer& out, Arguments& args) {
...
+    out << "mem_nativelibs_kb " << (u64) _native_libs.usedMemory() / KB << '\n';
 
-    size_t code_cache = _runtime_stubs.usedMemory();
-    int native_lib_count = _native_libs.count();
-    for (int i = 0; i < native_lib_count; i++) {
-        code_cache += _native_libs[i]->usedMemory();
-    }
-    code_cache += native_lib_count * sizeof(CodeCache);
+    out << "samples_total " << _total_samples << '\n';
+    out << "samples_failures_total " << _failures[-ticks_skipped] << '\n';
```
apangin: `samples_skipped_total`

src/profiler.cpp
```diff
@@ -1449,30 +1449,24 @@ Error Profiler::dump(Writer& out, Arguments& args) {
...
-             "        Code cache: %7zu KB\n"
-             "------------------------------\n"
-             "             Total: %7zu KB\n",
-             call_trace_storage / KB, flight_recording / KB, dictionaries / KB, code_cache / KB,
-             (call_trace_storage + flight_recording + dictionaries + code_cache) / KB);
-    out << buf;
+    if (_total_stack_walk_time != 0) {
+        out << "stackwalk_ns_total " << _total_stack_walk_time << '\n';
+        u64 stacks = _total_samples - _failures[-ticks_skipped];
+        out << "stackwalk_ns_average " << (_total_stack_walk_time / stacks) << '\n';
```
apangin: maybe `stackwalk_ns_avg`?

src/profiler.cpp
```diff
@@ -1987,9 +1981,8 @@ Error Profiler::runInternal(Arguments& args, Writer& out) {
             }
             break;
         }
-        case ACTION_MEMINFO: {
-            MutexLocker ml(_state_lock);
-            printUsedMemory(out);
+        case ACTION_METRICS: {
+            writeMetrics(out);
```
apangin, line 1986: Have you removed `MutexLocker` intentionally?
It can be dangerous to collect stats concurrently with profiler starting/stopping.

test/test/api/ApiTests.java
```diff
@@ -32,4 +32,10 @@ public void stopResume(TestProcess p) throws Exception {
         assert out.contains("BusyLoops.method2");
         assert !out.contains("BusyLoops.method3");
     }
+
+    @Test(mainClass = MetricsTest.class, jvmArgs = "-Djava.library.path=build/lib")
```
apangin: unneeded `java.library.path`?

src/profiler.cpp
```diff
@@ -1449,30 +1449,24 @@ Error Profiler::dump(Writer& out, Arguments& args) {
...
 
-    size_t code_cache = _runtime_stubs.usedMemory();
-    int native_lib_count = _native_libs.count();
-    for (int i = 0; i < native_lib_count; i++) {
-        code_cache += _native_libs[i]->usedMemory();
-    }
-    code_cache += native_lib_count * sizeof(CodeCache);
+    out << "samples_total " << _total_samples << '\n';
+    out << "samples_failures_total " << _failures[-ticks_skipped] << '\n';
+    out << "overflows_calltracestorage_total " << _call_trace_storage.overflow() << '\n';
```
apangin: nit: `calltracestorage_overflows_total` would be better

fandreuz: Mm, we already have `mem_calltracestorage_kb`, which is `what_where_unit`. Shall we swap also the `mem_*` metrics?

src/profiler.cpp
fandreuz: eead6d9b712705b1fdf4595dbc9baeea914d7f0c

src/profiler.cpp
fandreuz: eead6d9b712705b1fdf4595dbc9baeea914d7f0c

test/test/api/ApiTests.java
fandreuz: Right, thanks: f150ef8ceb7088b45c17b7b32fa4abb1c43b8b0c

src/profiler.cpp
fandreuz: Can we make them safe instead? I think we can keep a `_used_memory` field for `Dictionary` and `CallTraceStorage`, incremented every time we add a new row/table. Similarly to what I've done for `CodeCacheArray`.

apangin: Even if you make structures _safe_ in the sense `usedMemory()` does not access dynamically allocated objects, this won't eliminate need for synchronization, because querying structures is not atomic: e.g., CallTraceStorage memory consists of LongHashTable memory plus LinearAllocator memory.

src/profiler.cpp
apangin: I don't interpret it this way. I rather thought of it as `group_counter_unit`.

src/profiler.cpp
fandreuz: >  this won't eliminate need for synchronization, because querying structures is not atomic

But this is not something which we get just by owning `_state_lock`, is it?

apangin: The lock will at least guarantee you won't add up components from different profiling sessions.

fandreuz: Ok I see now, that makes sense. I restored the lock: 5e4e54cf5e927eb283eaa46a324a220338cbcd15

I still think there's room for improvement: the problem you pointed out can be addressed by asking the up-to-date size to the linear allocator when we know it may have updated its capacity (which happens in well known locations). But this can be a follow up PR.

src/profiler.cpp
fandreuz: a1e0e4e11f8354ae5326f3dca2d18f0c75aba665

PR #1571, opened 2025-10-29
Reduce mutex and rdlock iterations for nativelock tests

roy-soumadipta:
### Description
The commit includes reducing the mutex lock iteration count as in the worst case, by the time the test canAsprofTraceAllLockTypes ends in 3 seconds, we may not have profiled rd and wr locks. Also reducing rd lock iteration count for wr locks also to be sampled within the test completion duration.

### Related issues
N/A

### Motivation and context
NativelockTests.canAsprofTraceAllLockTypes fails intermittently on Jenkins.

### How has this been tested?
Was failing reliably on mac with `sudo purge`. Now it passes reliably. Tested on Linux as well.

---

### DIFF
--- a/test/test/nativelock/AllNativeLocks.java
+++ b/test/test/nativelock/AllNativeLocks.java
@@ -34,7 +34,7 @@ public class AllNativeLocks {
             threads.add(t);
             t.start();
         }
-         
+
         for (int i = 0; i < 2; i++) {
             Thread t = new Thread(() -> NativeLock.rdlockContentionThread());
             threads.add(t);


apangin:
Can you explain how the proposed fix helps?
I guess the main problem is that 12 threads run in parallel, that can be very undeterministic, especially in the environments with only a few vCPUs.

roy-soumadipta:
@apangin my intention is to reduce the overall mutex contention duration as the test which fails is defined with profiling duration as 3 secs. With the old iteration count and usleep configurations, it was highly probable that profiling stops even before rd and wr locks sampling takes place. But your comment makes sense, so should I limit the threads for each lock type to 2 along with the current change? I tested with 2 for each and tests pass as expected.

apangin:
> With the old iteration count and usleep configurations, it was highly probable that profiling stops even before rd and wr locks sampling takes place.

I don't really understand this part. As the code is currently written, it should start all test threads simultaneously, i.e. it does not wait mutex test to complete before running rdlock and wrlock tests. Did you originally plan to run them sequentially?

But leaving just 2 threads per each lock type makes sense anyway.

roy-soumadipta:
Oh yes, my deduction is completely wrong. It was never meant to be sequential. In that case, it was probably a thread starvation as you mentioned. In any case I would like to reduce the iteration count as well as limit threads to 2.

apangin:
OK, thanks. Let's see if this makes the test more reliable.

PR #1573, opened 2025-10-31
Allow looping with a duration

fandreuz:
### Description
In this PR I let `asprof collect` invocations with `--timeout` or `--loop` work with `--duration`.

### Related issues
n/a

### Motivation and context
The current behavior is to implicitly switch to `start`, which does not support a duration. I think we should either log an error, or take the fix from this PR. The current behavior is confusing (the command will just complete immediately, which is unexpected).

### How has this been tested?
Manual testing.

---

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -532,7 +532,6 @@ int main(int argc, const char** argv) {
 
         } else if (arg == "--timeout" || arg == "--loop") {
             params << "," << (arg.str() + 2) << "=" << args.next();
-            if (action == "collect") action = "start";
 
         } else if (arg == "--libpath") {
             libpath = args.next();


Baraa-Hasheesh:
In my personal opinion if a user uses loop, that would mean continuous profiling which doesn't make sense to be used with collect

As for `timeout` it wouldn't matter as the profiler will auto timeout either way, & using `start` saves on some unneeded work on the asprof process in terms of waiting to stop the profiler 
-> The profiler will auto stop it self

fandreuz:
> In my personal opinion if a user uses loop, that would mean continuous profiling which doesn't make sense to be used with collect

That sounds reasonable. I'd log an error in that case.

apangin:
It's not so simple.
What do you think is an expected behavior for each of the following commands? (note that `collect` command is not specified)

1. `asprof --timeout 120s`
2. `asprof -d 60 --timeout 30s`
3. `asprof --loop 30s`
4. `asprof -d 60 --loop 120s`

fandreuz:
When in doubt, leave it out :)

I propose we don't guess and we log an error in ambiguous cases, what do you think?

apangin:
I don't see how printing an error is better than the current behavior for the use cases like
`asprof --timeout 120s` and `asprof --loop 30s`.
Given that it's a breaking change with no obvious benefits, I'm inclined not to touch this.

Issue #1453, opened 2025-08-27
Codecache memory leak in lock profiling on stop/start


knbk:
### Describe the bug

The lock tracer installs a native hook to profile calls to `Unsafe.park`. The original method gets restored when the tracer is stopped.

HotSpot's JIT compiler will create a native adapter referring to the installed hook, which will get stored in the code cache. When the tracer is stopped or restarted, the method reference changes, and the native adapter is marked as not_entrant.

The issue we've seen is that the stale code cache entries for `Unsafe.park` are marked as not_entrant, but never cleared from the code cache. Combined with a short start/stop loop (e.g. using a short `--loop` option), this leads to a memory leak in the code cache. Eventually this can lead to a full code cache, disabled compilation and degraded performance.

### Expected vs. actual behavior

Expected: code cache memory usage remains stable.
Actual: code cache memory usage increases until the code cache is full and compilation is disabled.

### Reproduction Steps

See https://github.com/knbk/pyroscope-codecache-leak-reproduction for a reproduction project.

To reproduce the issue:
1. Create a continuous, tight loop calling `Unsafe.park(1_000_000)`. This encourages (re)compilation of the native method adapter.
2. Attach asprof with lock profiling and a short loop:
    ```bash
    asprof start -e lock --loop 1s <pid>
    ```
3. Watch as the number of entries for `Unsafe.park` in the codelist grows:
    ```bash
    jcmd <pid> Compiler.codelist | grep -c Unsafe.park
    ```

### Additional Information/Context

We encountered this issue while using Pyroscope's Java agent (https://github.com/grafana/pyroscope-java). Pyroscope creates a new profile for each upload interval (default 10 seconds), which it does by performing a stop+dump+start.

I think the core issue is with async-profiler, but let me know if I should create an issue for Pyroscope instead.

Grafana graph showing code cache usage over time. Pyroscope was enabled on 06/13. You can see the code cache for non-profiled nmethods hitting after about a week each time. We increased the code cache limit on 07/30, which delayed the issue but didn't prevent it.
<img width="3115" height="409" alt="Image" src="https://github.com/user-attachments/assets/e8f29593-2350-4a40-89ec-dd982bbb7980" />

Logs from `-Xlog:jit+compilation=debug` showing the recompilation of the `Unsafe.park` adapter:
[compilation.log](https://github.com/user-attachments/files/22003982/compilation.log)

### Async-profiler version

4.0, 4.1

### Environment details

[...]

PR #1575, opened 2025-11-03
Fix code cache memory leak in lock profiling while looping

fandreuz:
### Description
We decided to avoid resetting `Unsafe::park` in `LockTracer::stop` to avoid incurring in the problem reported in #1453. We can accept leaving around the hook since the alternative is a memory leak.

### Related issues
This is a workaround for #1453 and [JDK-8369219](https://bugs.openjdk.org/browse/JDK-8369219).

### Motivation and context
The [fix](https://github.com/openjdk/jdk/pull/27742) was merged in OpenJDK, but it will take some time to have it backported.

### How has this been tested?
The reproducer provided by the user is not reproducing the problem after the change:
```
$ ./script/run.sh --agent async-profiler --loop 1s
[monitor] Unsafe.park codelist occurrences: 0 | codecache: no-heap-lines
[monitor] Unsafe.park codelist occurrences: 1 | codecache: no-heap-lines
[monitor] Unsafe.park codelist occurrences: 1 | codecache: no-heap-lines
```

---

### DIFF
--- a/src/lockTracer.cpp
+++ b/src/lockTracer.cpp
@@ -70,6 +70,8 @@ void LockTracer::stop() {
     // Disable Java Monitor events
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTER, NULL);
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTERED, NULL);
+
+    // We don't reset the Unsafe::park hook due to JDK-8369219
 }
 
 Error LockTracer::initialize(jvmtiEnv* jvmti, JNIEnv* env) {


src/profiler.cpp
```diff
@@ -1314,7 +1314,7 @@ Error Profiler::stop(bool restart) {
     uninstallTraps();
 
     if (_event_mask & EM_WALL) wall_clock.stop();
-    if (_event_mask & EM_LOCK) lock_tracer.stop();
+    if (_event_mask & EM_LOCK && !restart) lock_tracer.stop();
```
Baraa-Hasheesh: I don't like this for 2 reasons
* Profiler is still collecting samples while it's technically off, this will be problamatic if the next start failed for any reason (One engine of the profiler remains in the background as a zombie) 
* This won't solve the issue in case the start/stop chain is handled externally via API mode agent

fandreuz: > if the next start failed for any reason

Next start is almost a no-op.

> is handled externally via API mode agent

I don't understand, can you clarify?

Baraa-Hasheesh: > Next start is almost a no-op.
> 
> >

That's not really true, you can have multiple point of failures, for example a perf event might fail to create


> I don't understand, can you clarify?

What I mean is if the user uses the Java/C APIs to start & stop the profiler & in between the user uploads the profile to an external location for example for post processing & then starts it again

src/lockTracer.cpp
```diff
@@ -74,9 +64,11 @@ void LockTracer::stop() {
 
     // Reset Unsafe.park() trap
     setUnsafeParkEntry(env, _orig_unsafe_park);
+
+    _initialized = false;
```
apangin: Initialization is supposed to be one-time thing. Calling it multiple times is not only expensive, but is wrong. E.g., you've just introduced another leak where a global JNI reference `_Unsafe` is allocated every time `start` is called.

src/profiler.cpp
apangin: I agree with @Baraa-Hasheesh that Engine::start/stop must be called anyway. They can be lightened for the continuous profiler case, but this logic should be in the engine itself, not outside.

src/lockTracer.cpp
fandreuz: I see, check the latest revision

src/lockTracer.h
```diff
@@ -28,7 +29,7 @@ class LockTracer : public Engine {
     static jfieldID _parkBlocker;
     static jmethodID _setEntry;
 
-    static Error initialize(jvmtiEnv* jvmti, JNIEnv* env);
+    static Error initialize(jvmtiEnv* jvmti, JNIEnv* env, const Arguments& args);
```
Baraa-Hasheesh: `args` is not used

src/lockTracer.h
```diff
@@ -60,7 +61,7 @@ class LockTracer : public Engine {
     }
 
     Error start(Arguments& args);
-    void stop();
+    void stop(bool restart);
```
Baraa-Hasheesh: we have 2 different stop functions being
* stop(bool) from LockTracer
* stop() from Engine 

how about we override `stop()` to call `stop(false)`?

src/lockTracer.cpp
```diff
@@ -36,35 +37,41 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
...
-    setUnsafeParkEntry(env, UnsafeParkHook);
+    _total_duration = 0;
+    _start_time = TSC::ticks();
 
     return Error::OK;
 }
 
-void LockTracer::stop() {
+void LockTracer::stop(bool restart) {
+    if (restart) return;
```
Baraa-Hasheesh: This doesn't really solve the issue reported in the ticket as I understand it 

> We encountered this issue while using Pyroscope's Java agent (https://github.com/grafana/pyroscope-java). Pyroscope creates a new profile for each upload interval (default 10 seconds), which it does by performing a **stop+dump+start**

As the stop & start are handled externally rather than via the `loop` option the memory leak will still happen 

https://github.com/grafana/pyroscope-java/blob/00f278fd21eb546bff5a781a38e0e33e8be3c06c/agent/src/main/java/io/pyroscope/javaagent/impl/SamplingProfilingScheduler.java#L80C18-L80C29

We can see that the flow is `start -> stop -> dump -> start -> ...`

I'm aware there is no clean solution for this, the most straight forward approach that I could think of would be never "unpatching" the `Unsafe.park` method but as discussed internally that would cause a slight performance hit

@apangin what do you think?

src/lockTracer.h
fandreuz: c092becc68b885c558ee63886236ac26d6a59f4b

src/lockTracer.h
fandreuz: Looks reasonable, I think eventually we'll just get rid of `stop()`.

08d2f01cb47229e6a8c430fd50a3eaa9a16f167d

src/lockTracer.cpp
fandreuz: My comment on this, as I said in person, is that I don't think we should fix 100% of the issue. Fixing `--loop` is enough because:
- it gives the user two possible solutions
     - upgrade to a newer jdk release, once the fix is backported
     - use `--loop` instead of `start/stop/start`
- this is part of a larger effort to make continuous profiling more lightweight, and we're not going to recommend `start/stop/start` for that

src/lockTracer.cpp
```diff
@@ -50,21 +46,36 @@ Error LockTracer::start(Arguments& args) {
...
+    _start_time = TSC::ticks();
 
     return Error::OK;
 }
 
 void LockTracer::stop() {
+    stop(false /* restart */);
+}
+
+void LockTracer::stop(bool restart) {
```
Baraa-Hasheesh: What if the `start` command failed in the next cycle, won't this leave this engine as fully on, even though the profiler is off?

For example it's possible for `ctimer` & `perfEvents` engine to fail in the restart process which will leave this engine as still on (Not likely but theoretically possible)

If `start` fails, we need to check which engines are in `light` stop & force stop them

src/lockTracer.cpp
```diff
@@ -21,6 +21,7 @@ INCLUDE_HELPER_CLASS(LOCK_TRACER_NAME, LOCK_TRACER_CLASS, "one/profiler/LockTrac
 
 
 bool LockTracer::_initialized = false;
+bool LockTracer::_started = false;
```
apangin: `_initialized`, `_enabled` and now `_started` - this becomes confusing.

src/lockTracer.cpp
```diff
@@ -36,11 +37,6 @@ UnsafeParkFunc LockTracer::_orig_unsafe_park = NULL;
 
 
 Error LockTracer::start(Arguments& args) {
-    // There is a JVM here, so TSC::frequency is calibrated from it
-    _ticks_to_nanos = 1e9 / TSC::frequency();
-    _interval = (u64)(args._lock * (TSC::frequency() / 1e9));
-    _total_duration = 0;
```
apangin, line 42: Many changes seem unrelated. To workaround the original issue, the only thing you need is to avoid calling `setUnsafeParkEntry`.

src/lockTracer.h
apangin: Two functions are confusing. Keep only one you need. Either call `stop(restart)` everywhere or assign the default value `stop(bool restart = false)` to minimize changes.

fandreuz: I think we need both, `stop()` is needed due to the polymorphic call to `Engine::stop()` in `Profiler::stop`.

src/lockTracer.cpp
fandreuz: We're trying to make continuous profiler lighter, right? I guess enabling/disabling JVMTI contended lock notifications must have some overhead, I guess there must be some synchronization.

src/lockTracer.cpp
fandreuz: Yes. Perhaps an enum of states would make it better? Probably not so much.

src/lockTracer.h
apangin: > I think we need both, stop() is needed due to the polymorphic call to Engine::stop() in Profiler::stop.

That's what I meant by saying two functions are confusing.
`lock_tracer.stop()` is no different from `any_other_engine.stop()`.

src/lockTracer.cpp
apangin: > We're trying to make continuous profiler lighter, right?

No. At least, not in this PR. Java locks will be the last thing to improve for confinuous profiling; by the time we get to it, design may have changed multiple times. There are other critical things to do first, like concurrent profile dumping.

Let's focus on a workaround for the memory leak here.

src/lockTracer.h
fandreuz: 357f418531b3c84a3019249aee15adda9b8de0d1

src/lockTracer.cpp
fandreuz: Latest revision should be fine

src/lockTracer.cpp
fandreuz: Got it, see the latest revision

test/test/lock/LockTests.java
```diff
@@ -13,13 +13,41 @@ public class LockTests {
...
+
+        p.profile("start -e lock --loop 1s");
+        Thread.sleep(2_000);
+        Output out = p.profile("stop -o collapsed");
+        assert out.contains("sun/nio/ch/DatagramChannelImpl.send");
+    }
+
+    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true, jvmVer = {11, Integer.MAX_VALUE})
+    public void datagramSocketLockStopStart(TestProcess p) throws Exception {
+        p.profile("start -e lock");
```
Baraa-Hasheesh: Why use `Thread.sleep` when you can do `p.profile("-e lock -d 2");`

test/test/lock/LockTests.java
```diff
@@ -13,13 +13,41 @@ public class LockTests {
...
+        checkCpu(p);
+
+        p.profile("start -e lock --loop 1s");
+        Thread.sleep(2_000);
+        Output out = p.profile("stop -o collapsed");
+        assert out.contains("sun/nio/ch/DatagramChannelImpl.send");
+    }
+
+    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true, jvmVer = {11, Integer.MAX_VALUE})
+    public void datagramSocketLockStopStart(TestProcess p) throws Exception {
```
Baraa-Hasheesh: Just wondering what does this test even do from a functional perspective?

What does it test, I don't really see why it's needed

test/test/lock/LockTests.java
```diff
@@ -13,13 +13,41 @@ public class LockTests {
...
+        Output out = p.profile("-e lock -d 3 -o collapsed");
+        assert out.contains("sun/nio/ch/DatagramChannelImpl.send");
+    }
+
+    @Test(mainClass = DatagramTest.class, debugNonSafepoints = true, jvmVer = {11, Integer.MAX_VALUE})
+    public void datagramSocketLockLoop(TestProcess p) throws Exception {
+        checkCpu(p);
+
+        p.profile("start -e lock --loop 1s");
+        Thread.sleep(2_000);
```
Baraa-Hasheesh: it's not really obvious what this test is even testing at all, how about adding a comment / linking the issue here

src/lockTracer.cpp
```diff
@@ -64,16 +64,18 @@ Error LockTracer::start(Arguments& args) {
...
     jvmtiEnv* jvmti = VM::jvmti();
     JNIEnv* env = VM::jni();
 
     // Disable Java Monitor events
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTER, NULL);
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTERED, NULL);
 
-    // Reset Unsafe.park() trap
-    setUnsafeParkEntry(env, _orig_unsafe_park);
+    if (!restart) {
```
Baraa-Hasheesh: Shouldn't you do something to prevent calling the same on the `start` side? or is that not necessary

fandreuz: We don't _need_ something in `start` thanks to [this](https://github.com/openjdk/jdk/blob/df414e0d19c1ed68f151d84dbb481a9dd6c65539/src/hotspot/share/oops/method.cpp#L1020) check. It would probably be good for performance, but we discussed not to focus on that aspect in this PR.

src/profiler.cpp
```diff
@@ -1314,13 +1314,13 @@ Error Profiler::stop(bool restart) {
     uninstallTraps();
 
     if (_event_mask & EM_WALL) wall_clock.stop();
-    if (_event_mask & EM_LOCK) lock_tracer.stop();
+    if (_event_mask & EM_LOCK) lock_tracer.stop(restart);
     if (_event_mask & EM_ALLOC) _alloc_engine->stop();
     if (_event_mask & EM_NATIVEMEM) malloc_tracer.stop();
     if (_event_mask & EM_NATIVELOCK) native_lock_tracer.stop();
     if (_event_mask & EM_METHOD_TRACE) instrument.stop();
```
apangin, line 1321: Since you've added `restart` argument to all engines, it is reasonable to pass `restart` to all engines, too.

src/lockTracer.cpp
```diff
@@ -64,16 +64,18 @@ Error LockTracer::start(Arguments& args) {
...
+        // Reset Unsafe.park() trap
+        setUnsafeParkEntry(env, _orig_unsafe_park);
+    }
```
apangin: After discussing this with @Baraa-Hasheesh, I start thinking this condition is not really correct/useful.
What if we don't reset the hook at all? In the worst case, we'll just have one extra call for each `Unsafe.park` invocation, which is a drop in the ocean comparing to all other stuff `Unsafe.park` is doing. This is a slow path anyway.

apangin: Or we may conditionally reset hook only on those JDK versions that have a fix.

fandreuz: >  I start thinking this condition is not really correct/useful.

Why?

apangin: Because this is only a partial workaround.
The overhead of resetting native entry is negligible and can be ignored for the sake of this conversation; the real problem is a memory leak. However, the workaround proposed here does not solve this leak in a common use case when continuous profiling is implemented via start/stop API. There is no available workaround outside profiler and JDK, so it's a responsibility of async-profiler to take care of this scenario too.

fandreuz: See the latest revision.

src/lockTracer.cpp
```diff
@@ -66,14 +66,10 @@ Error LockTracer::start(Arguments& args) {
...
 void LockTracer::stop() {
     jvmtiEnv* jvmti = VM::jvmti();
-    JNIEnv* env = VM::jni();
 
     // Disable Java Monitor events
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTER, NULL);
     jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_MONITOR_CONTENDED_ENTERED, NULL);
-
-    // Reset Unsafe.park() trap
-    setUnsafeParkEntry(env, _orig_unsafe_park);
```
Baraa-Hasheesh, line 76: since this is not longer reset, how about moving the `setUnsafeParkEntry(env, UnsafeParkHook);` to be done inside the `if (!_initialized)` check as it's only needed on the first call to start

fandreuz: It makes sense, my only argument against this change would be that the goal we discussed for this PR is to fix the memory leak. Reducing the overhead of continuous profiling is out of scope.

apangin: Can you add a comment here with a link to the bug why we don't reset the hook?

fandreuz: Sure: 2136dff91eab512c168c32a730a08c31ee9dcbc7

PR #1577, opened 2025-11-05
Remove redundant check declarations.

roy-soumadipta:
### Description
Remove redundant check declarations.

### Related issues
N/A

### Motivation and context
Check declaration left behind in header file

### How has this been tested?
Existing tests

---

### DIFF
--- a/src/allocTracer.h
+++ b/src/allocTracer.h
@@ -39,7 +39,6 @@ class AllocTracer : public Engine {
         return "bytes";
     }
 
-    Error check(Arguments& args);
     Error start(Arguments& args);
     void stop();
 
--- a/src/j9ObjectSampler.h
+++ b/src/j9ObjectSampler.h
@@ -15,7 +15,6 @@ class J9ObjectSampler : public ObjectSampler {
         return "j9_object_sampler";
     }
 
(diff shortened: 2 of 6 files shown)


src/allocTracer.h
```diff
@@ -39,7 +39,6 @@ class AllocTracer : public Engine {
         return "bytes";
     }
```
fandreuz, line 41: One more in `profiler.h`

roy-soumadipta: No that is needed. Profiler.cpp has the only check left with a deprecation warning.

src/j9ObjectSampler.h
```diff
@@ -15,7 +15,6 @@ class J9ObjectSampler : public ObjectSampler {
         return "j9_object_sampler";
     }
 
-    Error check(Arguments& args);
```
Baraa-Hasheesh, line 18: Why not remove if from `engine.h` & `engine.cpp` as well? or is it still needed?

roy-soumadipta: That was already done. Only these were left

apangin:
There are still a few left.
See ctimer.h and perfEvents.h

Issue #1562, opened 2025-10-23
`make test` should not recompile C sources every time


apangin:
### Describe the feature

`make test` unconditionally rebuilds all native test executables and native libraries, even when the corresponding C sources are unmodified. This unnecessarily extends the test feedback loop.

Additionally, I suggest not to run C++ unit tests when `TESTS` variable is non-empty. In these cases, developer is typically interested in particular Java integration tests.

### Use Case

Debug specific Java test failure.

### Proposed Solution

In order to keep the build system simple, it's OK not to track each C source individually. E.g., a valid implementation could recompile all C tests if at least one C file is changed.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
Handled in #1581

apangin:
Thanks! This one was particularly annoying me :)

PR #1581, opened 2025-11-05
Optimize make targets to not compile C/C++ files when not needed

Baraa-Hasheesh:
### Description
Currently make file does some unnecessary work as mentioned in ticket #1562
* Compiling unit tests for cases where user is testing integration `make test TESTS=cpu`
* Running unit tests for cases where user  is testing integration `make test TESTS=cpu`
* Compiling integration test binaries for unit tests 

After changes, 
unit tests won't compile for integration target  `make test TESTS=cpu`
unit tests won't run for integration target `make test TESTS=cpu`
integration binaries won't compile for unit test target `make test-cpp`

Note:
* I did consider splitting `build-test-libs` into 2 different targets for unit tests & integration tests (each has a unique group of libs) but opted against it for simplicity 
* I used if checks as they are the most straight forward approach in this case to check both  `MAKECMDGOALS` & `TESTS` to validate if unit tests are needed or not

### Related issues
#1562

### Motivation and context
Optimize make flow for faster feedback loop

### How has this been tested?
manual testing 

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ CPP_TEST_SOURCES := test/native/testRunner.cpp $(shell find test/native -name '*
 CPP_TEST_HEADER := test/native/testRunner.hpp
 CPP_TEST_INCLUDES := -Isrc -Itest/native
 TEST_LIB_SOURCES := $(wildcard test/native/libs/*)
-TEST_BIN_SOURCES := $(shell find test/test \( -name "*.c" -o -name "*.cpp" \))
+TEST_BIN_SOURCES := $(shell find test/test -name "*.c*")
 
 ifeq ($(JAVA_HOME),)
   JAVA_HOME:=$(shell java -cp . JavaHome)


Makefile
```diff
@@ -211,20 +227,22 @@ build/$(CONVERTER_JAR): $(CONVERTER_SOURCES) $(RESOURCES)
 
 build/test/cpptests: $(CPP_TEST_SOURCES) $(CPP_TEST_HEADER) $(SOURCES) $(HEADERS) $(RESOURCES) $(JAVA_HELPER_CLASSES)
 	mkdir -p build/test
+ifneq ($(RUN_CPP_TESTS),)
```
apangin: Too much conditional execution - not in the declarative spirit of Makefiles.
In fact, you can get without conditional execution at all: the only thing you need is one conditional target. Let `test` target depend on `$(TEST_CPP)` target, which is equal to `test-cpp` if `TESTS` variable is unset or an empty string otherwise.

apangin: E.g.:
```Makefile
ifeq ($(TESTS),)
    TEST_CPP=test-cpp
endif

test: $(TEST_CPP) test-java
```

Baraa-Hasheesh: Sounds good, applied here [1c639c9](https://github.com/async-profiler/async-profiler/pull/1581/commits/1c639c96b8dff1293022149a1f26b3cb5e64755d)

Makefile
```diff
@@ -75,6 +75,8 @@ TESTS ?=
 CPP_TEST_SOURCES := test/native/testRunner.cpp $(shell find test/native -name '*Test.cpp')
 CPP_TEST_HEADER := test/native/testRunner.hpp
 CPP_TEST_INCLUDES := -Isrc -Itest/native
+CPP_TEST_LIB_SOURCES := test/native/testRunner.cpp $(shell find test/native/libs \( -name "*.c" -o -name "*.cpp" \))
```
apangin: There are also other file types in that directory: `.s` and `.ld`. Why not just include all files? Since there are no subdirectories, you can just use built-in $(wildcard) instead of shell find.

apangin: Also, why does it depend on `testRunner.cpp`? This is only for unit tests.

Makefile
```diff
@@ -75,6 +75,8 @@ TESTS ?=
...
+CPP_TEST_BIN_SOURCES := test/native/testRunner.cpp $(shell find test/test \( -name "*.c" -o -name "*.cpp" \))
```
apangin: nit: `*.c*` to simplify the expression

Makefile
```diff
@@ -253,9 +255,10 @@ build-test-bins:
 	$(CC) -o $(TEST_BIN_DIR)/preload_malloc -Isrc test/test/nativemem/preload_malloc.c -ldl
 	$(CC) -o $(TEST_BIN_DIR)/nativemem_known_lib_crash -Isrc test/test/nativemem/nativemem_known_lib_crash.c -ldl
 	$(CXX) -o $(TEST_BIN_DIR)/non_java_app -std=c++11 $(INCLUDES) $(CPP_TEST_INCLUDES) test/test/nonjava/non_java_app.cpp $(LIBS)
+	@touch $@
 
 test-cpp: build-test-cpp
-	echo "Running cpp tests..."
+	@echo "Running cpp tests..."
```
apangin: Either change this everywhere (e.g. test-java), or nowhere at all.

Makefile
Baraa-Hasheesh: Updated [d37945e](https://github.com/async-profiler/async-profiler/pull/1581/commits/d37945efc87b680a3c2bd3fae4a5f74751604e02)

Makefile
Baraa-Hasheesh: `testRunner.cpp` was a copy paste error, `wildcard` used as suggested

Makefile
Baraa-Hasheesh: reverted to original

PR #1590, opened 2025-11-14
Support Compilation on modern JDKs & drop JDK7 support

Baraa-Hasheesh:
### Description
Currently building the profiler using a modern JDK, such as JDK21 & 25 isn't possible due to the fact that `target=7` is no longer supported 

This can be easily observed by changing Instrument.java file & running `make` using java 21/25

```
async-profiler % make
/Users/bhashesh/.sdkman/candidates/java/25.0.1-amzn/bin/javac -source 7 -target 7 -Xlint:-options -g:none src/helper/one/profiler/Instrument.java
error: Source option 7 is no longer supported. Use 8 or later.
error: Target option 7 is no longer supported. Use 8 or later.
make: *** [src/helper/one/profiler/Instrument.class] Error 2
```

This compile happens due to the following make target
`build/$(LIB_PROFILER): $(SOURCES) $(HEADERS) $(RESOURCES) $(JAVA_HELPER_CLASSES)`

which in turn is mapped to 

```
%.class: %.java
	$(JAVAC) -source 7 -target 7 -Xlint:-options -g:none $^
```

In this change I simply reuse the `JAVA_TARGET` variable to select source & target which is selected as `8` which is still supported on JDK25

I also updated the `.class` files by recompiling using  `-source 8 -target 8 -Xlint:-options -g:none`

This change will require dropping JDK7 support in the profiler, but that should be acceptable as JDK7 has been EOLed long time ago, & non of the available pipelines check for JDK7 

### Related issues
N/A

### Motivation and context

- Improve developer experience by using up-to-date build tools
- Reduce maintenance burden by giving up JDK 7 support

### How has this been tested?
`make` & `make test`

---


apangin:
The major consequence of this change not mentioned here is that certain profiler features will no longer work with JDK 7.
Yes, there are a few customers running profiler with JDK 7... at least, there were a year ago.

In fact, I think it's time to drop JDK 7 support in async-profiler, and therefore welcome this change, but this needs to be called out explicitly.

Baraa-Hasheesh:
@apangin is there a specific location of the documentation that need to be updated to reflect this?

apangin:
This PR.
Title, description and motivation.
In that way, this will get in the release notes (release notes are compiled from Issue and PR titles with the references to the corresponding issues).

Baraa-Hasheesh:
@apangin title & description updated

PR #1598, opened 2025-11-17
Add new target to generate a vscode workspace

krk:
### Description

Adds a new `vscode-project` make target that generates a VSCode workspace configuration with full IntelliSense support for the async-profiler C/C++ codebase - except for the cpp tests which need to support building with `MERGE=false` first. Support for the cpp tests can be added in a different PR.

### Motivation and context

IDEs like Visual Studio Code require a compilation database (compile_commands.json) to provide features like autocomplete and go-to-definition reliably, but generating this manually is tedious and error-prone. Without `compile_commands.json`, navigation in vscode is slow and "best effort", which fails to resolve e.g. jvm includes among other things.

This change adds a `vscode-project` target that automatically generates a complete VSCode workspace with proper IntelliSense configuration. It rebuilds the project while capturing compilation commands, uses `jq` for JSON generation with proper escaping, and detects the platform to configure the correct compiler mode (Clang for macOS, GCC for Linux).

### How has this been tested?

```
make vscode-project
code build/ide/async-profiler.code-workspace
```

---


Baraa-Hasheesh:
I can understand the idea behind this, but I don't personally like it

* This only does the C/Cpp setup, & leaves java unconfigured 
=> You can check on `Main.java` of the converter package, it has many places where IDE features aren't setup
image


* If the idea is to do C/C++, the default vscode setup does that well enough & you only need to do small changes to tell it which Java libs to use, having a small predefined template with some documentation would be simpler IMO

=> If I want to setup vscode for C/C++ on async-profiler it would take me a couple of minutes to do, as I just need to copy my `c_cpp_properties.json` file  

For example the following configuration should work well enough on every ARM mac

```
{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${workspaceFolder}/**",
                "${env:JAVA_HOME}/include",
                "${env:JAVA_HOME}/include/darwin"
            ],
            "defines": [],
            "compilerPath": "/usr/bin/clang",
            "cStandard": "c17",
            "cppStandard": "c++14",
            "intelliSenseMode": "macos-clang-arm64"
        }
    ],
    "version": 4
}
```

apangin:
I tend to agree with @Baraa-Hasheesh that the proposed implementation seems a bit over-engineered, especially the `collect-commands` trick. Why can't it be an almost static template?

krk:
The static template approach won't work correctly here. The Makefile uses dynamic defines, platform-specific include paths, and varying flags per OS/arch. A static `c_cpp_properties.json` with just `includePath` won't capture these, so IDEs will misevaluate the many `#ifdef __APPLE__` / `#ifdef __linux__` / `#ifdef __aarch64__` etc. blocks throughout the codebase.

The "collect-commands trick" is standard practice - it's what [bear](https://github.com/rizsotto/Bear) and [compiledb](https://github.com/nickdiego/compiledb) do. I just inlined a concise version to avoid a dependency. `compile_commands.json` is the https://clang.llvm.org/docs/JSONCompilationDatabase.html these tools are designed to consume.

Why should every new contributor have to figure out what these settings look like when they can just run `make vscode-project`? This approach also extends naturally to other IDEs in the future (CLion, etc.) since they all consume `compile_commands.json`.

Java support is orthogonal and explicitly out of scope per the PR description, having said that, there is nothing preventing a `make idea-project` or a similar command in the future.

apangin:
> The static template approach won't work correctly here. The Makefile uses dynamic defines

I think you exaggerate the problem. The only dynamic define is `PROFILER_VERSION` which can be any string and does not affect IDE capabilities.

> platform-specific include paths

Again, if it's just a matter of `$JAVA_HOME/include/linux` vs. `$JAVA_HOME/include/darwin`, simply add both.

>  so IDEs will misevaluate the many `#ifdef __APPLE__` / `#ifdef __linux__` / `#ifdef __aarch64__` etc. blocks throughout the codebase.

What do you mean by "misevaluate"? I don't want Linux code to be grayed out only because I run IDE on Mac.

> The "collect-commands trick" is standard practice - it's what [bear](https://github.com/rizsotto/Bear) and [compiledb](https://github.com/nickdiego/compiledb) do.

Well, if there are standard tools to do the thing, the value of replicating their function in async-profiler codebase becomes even less obvious to me. Maintaining extra code in async-profiler repo just to save a few developers from typing an extra command once in a year? I'm not sure.

Issue #1578, opened 2025-11-05
async-profiler perf events hang the whole machine


franz1981:
Hi!

I've recently upgraded my Ryzen 4 (7950X) workstation with Linux Fedora 43.
The (broken) kernel versions I've tested are:
```
kernel-0:6.17.1-300.fc43.x86_64
kernel-6.17.6-300.fc43.x86_64
```
The working kernel versions with Fedora 42 are:
```
kernel-6.15.3-200.fc42.x86_64 
kernel-6.15.4-200.fc42.x86_64
```
With the broken versions if I run async profiler to profile any JVM Hotspot application using perf events (cycles, cpu-clock whatever) once it attach to application, it freezes the whole machine.
It hang so bad that Linux `dmesg` and the `journal` are not populated, so IDK how to check what's going on.
I haven't yet debugged `async-profiler` with `gdb` yet and see step by step where it breaks.
If I run `perf record` with the same events (call chain collection/walk, cpu-clock and same frequency) it just work.
I've enabled paranoid and kernel ptr settings, as expected.

Both 4.0 and any other async-profiler version just fall similarly.

### Async-profiler version

4.0, 4.1, 4.2...till master

fandreuz:
I can reproduce this too:
```
AMD Ryzen 7 7840U
6.17.6-300.fc43.x86_64
```

franz1981:
I tried several times trying to capture any logs without success @fandreuz 
:"(

franz1981:
@fandreuz It looks like with HW perf events it just work fine.
Maybe the issue is due to
https://github.com/async-profiler/async-profiler/blob/fd269e6450431b26ea6b79f48bc5bf668be00fc9/src/perfEvents_linux.cpp#L567-L569
?

I will try change it.
For now using cycles is a decent temporary "workaround"

franz1981:
> I will try change it.

Changing it made the hang to be less immediate, but it happens the same.
I can confirm that Perf HW PMU events just work as expected i.e. no hang

fandreuz:
> > I will try change it.
> 
> Changing it made the hang to be less immediate, but it happens the same. I can confirm that Perf HW PMU events just work as expected i.e. no hang

Thanks @franz1981, I'll see if this can be reproduced without Async-Profiler.

franz1981:
@fandreuz I have found something funny...still to be verified (by you at this point because you have a similar env).

If you create this simple bash script
```bash
#!/usr/bin/env bash
[...]

PR #1599, opened 2025-11-18
Workaround for PERF_EVENT_IOC_REFRESH bug

apangin:
### Description

Implement workaround for the Linux kernel bug that causes the entire system to hang when running CPU profiler with perf_events engine.

### Related issues

#1578

### Motivation and context

`PERF_EVENT_IOC_REFRESH` control allows to disable perf_event automatically on counter overflow. Async-profiler uses this control to temporarily disable perf_event while it is processed inside the signal handler. Without this, event could be recursively triggered again and again during processing. Event is then re-enabled just before return from the signal handler.

In Linux 6.16.x and 6.17.x, there is a kernel bug that causes a deadlock when perf_event is automatically disabled inside hrtimer interrupt handler. To workaround this bug, we do not use `PERF_EVENT_IOC_REFRESH` feature, but instead disable and re-enable the event manually inside a signal handler. To avoid an extra syscall, the workaround is applied only when running CPU profiling on the affected kernel versions.

### How has this been tested?

Ran `cpu-clock` and `cycles` profiling on Linux versions 6.14, 6.16 and 6.17.

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -19,6 +19,7 @@
 #include <sys/resource.h>
 #include <sys/stat.h>
 #include <sys/syscall.h>
+#include <sys/utsname.h>
 #include <linux/perf_event.h>
 #include "arch.h"
 #include "fdtransferClient.h"


franz1981:
Many thanks again @apangin ❤️

Issue #1608, opened 2025-11-25
Limit reported stacks properly according to stackdepth argument


rkennke:
### Describe the bug

Currently, the jstackdepth argument only limits the Java part of the reported stacktraces. This can be confusing, because if a stacktrace contains native frames, then stacktraces can become much larger than what has been requested by jstackdepth. Restricting the number of reported stack-frames - Java and native combined - to jstackdepth is less confusing/more usable and also brings an opportunity to save memory both in the JFR recording and in-memory while constructing the stacktraces.

### Expected vs. actual behavior

I would have expected that jstackdepth=8 limits stacktraces to the top 8 frames. However, when native frames are involved, I often see much longer stacktraces.

### Reproduction Steps

Run with jstackdepth=8, observe number of reported stack-frames.

### Async-profiler version

4.21

krk:
It is correct that `jstackdepth` currently applies only to Java stack frames.

One possibility we could explore is introducing a new `stackdepth` argument to limit the total combined depth (Java + Native).

rkennke:
I made a first fix (see PR). But yes, it may be better to use a new option instead. This poses some interesting possible interactions (and restrictions) with the two options, though.

Baraa-Hasheesh:
Regarding this topic, currently the behaviour will depend on the stack walking mode being used

If the user selects the "vm" stack walker (default), `jstackdepth` will be fully respected.

The described issue will only happen for legacy stack walkers being `fp` & `dwarf`

rkennke:
> Regarding this topic, currently the behaviour will depend on the stack walking mode being used
> 
> If the user selects the "vm" stack walker (default), `jstackdepth` will be fully respected.
> 
> The described issue will only happen for legacy stack walkers being `fp` & `dwarf`

Right.
This seems to be an argument to *not* have a new option, but rather fix the code such that jstackdepth means total stack-depth regardless of the used stack walker.

Baraa-Hasheesh:
> This seems to be an argument to not have a new option, but rather fix the code such that jstackdepth means total stack-depth regardless of the used stack walker.

IMO introducing a new variable will just complicate the UX, especially when it comes to `vm` & `vmx`, but that can be visited if needed.

Regarding the suggested changes, it doesn't look the best, I would
* Refactor the java stack walking portion into a separate function similar to `getNativeTrace`
[...]

PR #1609, opened 2025-11-25
Limit reported stacks properly according to stackdepth argument (#1608)

rkennke:
### Description

Currently, the jstackdepth argument only limits the Java part of the reported stacktraces. This can be confusing, because if a stacktrace contains native frames, then stacktraces can become much larger than what has been requested by jstackdepth. Restricting the number of reported stack-frames - Java and native combined - to jstackdepth is less confusing/more usable and also brings an opportunity to save memory both in the JFR recording and in-memory while constructing the stacktraces.

### Motivation and context

It is confusing for the user when he specifies a maximum stack-depth, but then the reported stack-traces contain more frames than specified. The fact that the argument only applies to Java stack frames is not obvious and overall rather useless to the end user, especially in mixed stack traces. Even worse, the exact behaviour seems to depend on the stack-walker being used. It would improve the usefulness and usability of the stackdepth argument if it applies for the total number of frames (Java and native combined) and regardless of the stack-walker that is in use.

### How has this been tested?

I have ran the akka workload of the Renaissance benchmark with CPU profiling and the jstackdepth argument set to 8, and observed that many samples report more than 8 frames. With the fix, it consistently reports a maximum of 8 frames. It seems very difficult to make an automated test for this.

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -314,8 +314,8 @@ jmethodID Profiler::getCurrentCompileTask() {
 }
 
 int Profiler::getNativeTrace(void* ucontext, ASGCT_CallFrame* frames, EventType event_type, int tid, StackContext* java_ctx) {
+    const void* callchain[MAX_NATIVE_FRAMES]; // we can read one frame past when trying to figure out whether the result is truncated
     int max_depth = std::min(_max_stack_depth, MAX_NATIVE_FRAMES);
-    const void** callchain = new const void*[max_depth + 1]; // we can read one frame past when trying to figure out whether the result is truncated
     int native_frames;
 
     // Use PerfEvents stack walker for execution samples, or basic stack walker for other events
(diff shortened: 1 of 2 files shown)


Baraa-Hasheesh:
Generally speaking I would recommend filling the entire PR template in regards to motivation, context & how it was tested as that helps in release notes & understanding the change 

Other than that the change looks generally fine to me, most of my feedback would be style changes which is usually left to @apangin to comment on

rkennke:
> I left my high-level comment on this change [here](https://github.com/async-profiler/async-profiler/issues/1608#issuecomment-3587788948). I'm now back to work and can discuss the issue further. Let me know how you want to proceed with this PR.

I am sorry, I think I missed that last comment. The issue is not very important to us, and as you mentioned in the bug, would affect deprecated/discouraged code only anyway. I am fine with keeping the old code as it is.

apangin:
No worries, thank you for the clarification. I will close this PR then.
That said, I don't mind changing `jstackdepth` behavior in the future to make it consistent between HotSpot and OpenJ9.

Issue #1610, opened 2025-11-26
Add a filter for GraalVM on tests


Baraa-Hasheesh:
### Describe the feature

When running async-profiler tests against a GraalVM some failures will happen & are expected due to differences between GraalVM & regular JDKs 

Adding a filter to prevent tests from running against Graal should be a good filter to add as it makes it smoother to run async-profiler tests against graal 

### Use Case

Don't run some tests against Graal

### Proposed Solution

Add a flag/enum that checks for `bin/native-image` under `JAVA_HOME` & add an associate filter on `Test` interface

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
solved in #1611

PR #1611, opened 2025-11-26
Create a GraalVM filter for async-profiler tests

Baraa-Hasheesh:
### Description
Some tests in async-profiler should not be run with GraalVM due to differences with regular JDK

In this change I've added detection logic for the current JDK to check if is a GraalVM & filter tests out based on it

### Related issues
#1610

### Motivation and context
Remove noisy expected failures when running GraalVM

### How has this been tested?
manual testing by running `JAVA_HOME=GRAALVM_HOME make test`

---

### DIFF
--- a/test/one/profiler/test/Jvm.java
+++ b/test/one/profiler/test/Jvm.java
@@ -7,7 +7,7 @@ package one.profiler.test;
 
 public enum Jvm {
     HOTSPOT,
+    HOTSPOT_C2,
     ZING,
-    OPENJ9,
-    HOTSPOT_C2
+    OPENJ9
 }
--- a/test/one/profiler/test/Runner.java
+++ b/test/one/profiler/test/Runner.java
@@ -121,7 +121,7 @@ public class Runner {
         log.log(Level.INFO, "Running " + rt.testInfo() + "...");
 
         String testLogDir = logDir.isEmpty() ? null : logDir + '/' + rt.testName();
-        try (TestProcess p = new TestProcess(rt.test(), currentOs, testLogDir, currentJvm)) {
+        try (TestProcess p = new TestProcess(rt.test(), currentOs, currentJvm, testLogDir)) {
             Object holder = (rt.method().getModifiers() & Modifier.STATIC) == 0 ?
                     rt.method().getDeclaringClass().getDeclaredConstructor().newInstance() : null;
             rt.method().invoke(holder, p);
--- a/test/one/profiler/test/TestProcess.java
+++ b/test/one/profiler/test/TestProcess.java
@@ -60,18 +60,18 @@ public class TestProcess implements Closeable {
 
     private final Test test;
     private final Os currentOs;
-    private final String logDir;
     private final Jvm currentJvm;
+    private final String logDir;
     private final String[] inputs;
     private final Process p;
     private final Map<String, File> tmpFiles = new HashMap<>();
     private final int timeout = 30;
 
-    public TestProcess(Test test, Os currentOs, String logDir, Jvm currentJvm) throws Exception {
+    public TestProcess(Test test, Os currentOs, Jvm currentJvm, String logDir) throws Exception {
         this.test = test;
         this.currentOs = currentOs;
-        this.logDir = logDir;
         this.currentJvm = currentJvm;
+        this.logDir = logDir;
         this.inputs = test.inputs();
 
         List<String> cmd = buildCommandLine(test);


test/one/profiler/test/Runner.java
```diff
@@ -20,6 +20,7 @@ public class Runner {
     private static final Arch currentArch = detectArch();
     private static final Jvm currentJvm = detectJvm();
     private static final int currentJvmVersion = detectJvmVersion();
+    private static final boolean currentGraal = detectGraal();
```
fandreuz:
```suggestion
    private static final boolean isGraal = detectGraal();
```

test/test/vtable/VtableTests.java
```diff
@@ -11,14 +11,14 @@
 
 public class VtableTests {
 
-    @Test(mainClass = Numbers.class, jvm = Jvm.HOTSPOT)
+    @Test(mainClass = Numbers.class, jvm = Jvm.HOTSPOT, graal = false)
```
fandreuz: The name is a bit confusing. Just by looking at the test, I don't understand which of the following two is the actual selection logic:
```
testShouldRun = !isGraal || graal
testShouldRun = !isGraal ^ graal (XOR)
```

Baraa-Hasheesh: how do you feel about changing the variable to `runGraalVM` & it's`true` by default

test/one/profiler/test/Runner.java
Baraa-Hasheesh: Flag has been removed, Graal checks moved to be a `Jvm` check similar to HOTSPOT & Other types

test/test/vtable/VtableTests.java
Baraa-Hasheesh: Agreed, I moved Graal checks to be on `jvm`, so it's more clear now

test/one/profiler/test/Jvm.java
```diff
@@ -8,5 +8,6 @@
 public enum Jvm {
     HOTSPOT,
     ZING,
-    OPENJ9
+    OPENJ9,
+    GRAALVM
```
fandreuz: I'm not sure about this either. Doesn't GraalVM claim to be just like an Hotspot VM?

And we still don't have a way to prevent a test from running on GraalVM. How about a simple flag `noGraal`? `noGraal && graal` implies that the test won't run.

Baraa-Hasheesh: GraalVM isn't 100% same as Hotspot VM some differences exists 

for example try running a simple "HelloWorld" application in both with `-Xcomp`

```
$ time ${JAVA_HOME}/bin/java -Xcomp Main.java
Hello World!
java -Xcomp Main.java  7.71s user 0.25s system 99% cpu 7.977 total

$ time ${GRAALVM_HOME}/bin/java -Xcomp Main.java
Hello World!
java -Xcomp Main.java  73.29s user 2.05s system 99% cpu 1:15.42 total
```

> And we still don't have a way to prevent a test from running on GraalVM
 
Simply do `jvm = HOTSPOT` for HOTSPOT only 

or if you want all JDKs but not Graal you would do `jvm = {HOTSPOT, ZING, OPENJ9}`

fandreuz: > for example try running a simple "HelloWorld" application in both with -Xcomp

Time it takes to complete is not what I'd call a "public API". When we discriminate between different JVM flavors it's more about missing APIs.

Baraa-Hasheesh: > Time it takes to complete is not what I'd call a "public API"

Let's start by clarifying what you mean by a "public API"
* How is `-Xcomp` not a public API when it's a publicly available option for the JVM 
*  `vmstructs` are by no means a "public API" 

`-Xcomp` literally has different behavior (Different run time) between HOTSPOT & GRAALVM, how is that not a difference due to a "public API"

> When we discriminate between different JVM flavors it's more about missing APIs.

Again `vmstructs` are by no means a "public API" but they are significant enough to give HOTSPOT special testing treatment, 

GRAALVM JIT will behave differently from a HOTSPOT VM, do you consider that a "public API"?

fandreuz: > How is -Xcomp not a public API when it's a publicly available option for the JVM

Xcomp is public API.

> -Xcomp literally has different behavior (Different run time)

Time it takes to run a program is not public API. Running something on any modern computer will convince you.

apangin: GraalVM is an ambiguous term, it may mean different things. What we are interested in here is a HotSpot JVM with Graal Compiler. So yes, it is still a HotSpot JVM, but with an additional JIT compiler which may or may not replace C2 depending on the JVM flags.

I'd expect that whatever test runs on pure HotSpot should also run on Graal-enabled HotSpot by default. The only exception are tests that assume C2 JIT. For them, we can add a separate enum constant, e.g. `jvm = Jvm.HOTSPOT_C2`.

test/one/profiler/test/Runner.java
```diff
@@ -59,6 +59,10 @@ private static Jvm detectJvm() {
         // Example javaHome: /usr/lib/jvm/amazon-corretto-17.0.8.7.1-linux-x64
         File javaHome = new File(System.getProperty("java.home"));
 
+        if (new File(javaHome, "bin/native-image").exists()) {
```
apangin: This is a wrong way to detect Graal Compiler. As I mentioned above, GraalVM is ambiguous: it's an umbrella term that covers multiple technologies. Native Image is one of them, Graal Compiler is another. Absense of `native-image` tool tells nothing about availability of Graal Compiler. E.g. JVM included in GraalJS does not have `native-image` binary but still has Graal Compiler. A better way would be to detect `libjvmcicompiler` shared library.

test/one/profiler/test/Runner.java
```diff
@@ -18,8 +18,8 @@ public class Runner {
 
     private static final Os currentOs = detectOs();
     private static final Arch currentArch = detectArch();
-    private static final Jvm currentJvm = detectJvm();
     private static final int currentJvmVersion = detectJvmVersion();
```
Baraa-Hasheesh, line 21: Order of execution matters, I noticed that due to calling `detectJvm` first, the `currentJvmVersion` will be 0 during `detectJvm` which means the  `currentJvmVersion <= 8` is always true

test/one/profiler/test/Runner.java
```diff
@@ -84,7 +84,10 @@ private static Jvm detectJvm() {
             return Jvm.ZING;
         }
```
Baraa-Hasheesh, line 85: I was checking on other JDK checks here & I can see that we can simplify the `OpenJ9` check to a simple 

`new File(javaHome, "lib/J9TraceFormat.dat").exists()`

Also, I'm not sure about the Zing check 
* I couldn't understand what's the "Contents/Home" work around is for,
* I tried checking versus `Azul Zing` JDK, but I couldn't find such a file on it

test/one/profiler/test/Runner.java
```diff
@@ -101,9 +104,11 @@ private static boolean applicable(Test test) {
         Arch[] arch = test.arch();
         Jvm[] jvm = test.jvm();
         int[] jvmVer = test.jvmVer();
+        Jvm additionalJvm = currentJvm == Jvm.HOTSPOT_C2 ? Jvm.HOTSPOT : null;
+
         return (os.length == 0 || Arrays.asList(os).contains(currentOs)) &&
                 (arch.length == 0 || Arrays.asList(arch).contains(currentArch)) &&
-                (jvm.length == 0 || Arrays.asList(jvm).contains(currentJvm)) &&
+                (jvm.length == 0 || Arrays.asList(jvm).contains(currentJvm) || Arrays.asList(jvm).contains(additionalJvm)) &&
```
apangin: I'm not a fan of `additionalJvm` variable which can be null and thus leading to `List.contains(null)` check.
We can get rid of `additionalJvm` variable entirely by extending the condition:
```
(jvm.length == 0 || Arrays.asList(jvm).contains(currentJvm) ||
    (currentJvm == Jvm.HOTSPOT_C2 && Arrays.asList(jvm).contains(Jvm.HOTSPOT)))
```

test/one/profiler/test/TestProcess.java
```diff
@@ -61,15 +61,17 @@ private static MethodHandle getPidHandle() {
     private final Test test;
     private final Os currentOs;
     private final String logDir;
+    private final Jvm currentJVM;
```
apangin: `currentJvm` in one place and `currentJVM` in another.
Let's use `currentJvm` everywhere, similarly to `currentOs`.

test/one/profiler/test/Runner.java
Baraa-Hasheesh: sure, [3b672e7](https://github.com/async-profiler/async-profiler/pull/1611/commits/3b672e7452bdcd0387b7dfd0a170e75a846a0d9d)

test/one/profiler/test/TestProcess.java
Baraa-Hasheesh: my bad, when typing `JVM` my muscle memory just takes over & presses the shift key :) 

[3b672e7](https://github.com/async-profiler/async-profiler/pull/1611/commits/3b672e7452bdcd0387b7dfd0a170e75a846a0d9d)

PR #1614, opened 2025-12-03
Fix asprof incorrectly existing with non zero exit code

Baraa-Hasheesh:
### Description
In `collect` mode, the `asprof` will do the following:
* run `jattach` to start the profiler
* sleep/wait until the full profiling duration
* run `jattach` to stop the profiler 

During some testing, I noticed that it's possible for the Java process to exit during the call of the last `sleep` call that the `asprof` does to wait for session time to be exhausted 

If that happens the follow up `jattach` call will fail, as the process is already dead,

This change doesn't not fully solve the issue as the race condition still exists but will make much less likely to happen as it checks the process is still alive before calling `jattach` to stop the profiler

### Related issues
N/A

### Motivation and context
Fix exist code for the `asprof` in case Java process exiting before `stop` is called

### How has this been tested?
manual testing

---

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -588,25 +588,20 @@ int main(int argc, const char** argv) {
         signal(SIGTERM, sigint_handler);
 
         while (time_micros() < end_time) {
+            sleep(1);
+
             if (kill(pid, 0) != 0) {
                 fprintf(stderr, "Process exited\n");
                 if (use_tmp_file) print_file(file, STDOUT_FILENO);
                 return 0;
             }
-            sleep(1);
         }
 
         fprintf(stderr, end_time != 0 ? "Done\n" : "Interrupted\n");
         signal(SIGINT, SIG_DFL);
         // Do not reset SIGTERM handler to allow graceful shutdown
 
-        // It's possible for java process to die during the last `sleep` call in the waiting loop
-        // Running jattach in that case will cause the asprof to incorrectly conclude with non 0 exit code
-        if (kill(pid, 0) == 0) {
-            run_jattach(pid, String("stop,file=") << file << "," << output << format << ",log=" << logfile);
-        } else if (use_tmp_file) {
-            print_file(file, STDOUT_FILENO);
-        }
+        run_jattach(pid, String("stop,file=") << file << "," << output << format << ",log=" << logfile);
     } else {
         if (action == "start" || action == "resume") run_fdtransfer(pid, fdtransfer);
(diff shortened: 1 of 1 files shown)


src/main/main.cpp
```diff
@@ -600,7 +600,13 @@ int main(int argc, const char** argv) {
         signal(SIGINT, SIG_DFL);
         // Do not reset SIGTERM handler to allow graceful shutdown
 
-        run_jattach(pid, String("stop,file=") << file << "," << output << format << ",log=" << logfile);
+        // It's possible for java process to die during the last `sleep` call in the waiting loop
+        // Running jattach in that case will cause the asprof to incorrectly conclude with non 0 exit code
+        if (kill(pid, 0) == 0) {
+            run_jattach(pid, String("stop,file=") << file << "," << output << format << ",log=" << logfile);
```
fandreuz: Technically the race condition is still there. Can we do one of the following?
- Fix the problem in jattach
- Be resilient to this situation in Async-Profiler, without relying on the output of `kill(pid, 0)` at some arbitrary point in time.

Baraa-Hasheesh: I'm aware that race conditions still exists

> **This change doesn't not fully solve the issue as the race condition still exists** but will make much less likely to happen as it checks the process is still alive before calling jattach to stop the profiler

Now regarding the suggestions

> Fix the problem in jattach

That would need to be done on the correct package [here](https://github.com/jattach/jattach) which is out of scope for this change here 

And strictly speaking `jattach` is behaving correctly, so changes to avoid this won't exactly look the best & I would advice against it, but I would ask @apangin about his opinion 

> Be resilient to this situation in Async-Profiler, without relying on the output of kill(pid, 0) at some arbitrary point in time.

I can't really see how to do that without hard coding checks for acceptable error messages that may be reported back by the `jattach` command, which won't look the best & overriding the exit code for those cases

apangin: To keep the termination check logic in one place, will it be possible to re-shape the `while` loop above? E.g., what about just swapping `sleep(1)` with `if (kill(pid, 0) != 0)`?

Baraa-Hasheesh: @apangin that should achieve the same goal of making the race condition less likely, [35478e2](https://github.com/async-profiler/async-profiler/pull/1614/commits/35478e253c2577d7093055340da89225cadcabe7)

Baraa-Hasheesh: Also I created a PR [here](https://github.com/jattach/jattach/pull/86) for other race conditions on `jattach`

Issue #1617, opened 2025-12-09
JfrSync may deadlock the death of the Java process


Baraa-Hasheesh:
### Describe the bug

Thanks @apangin for the explanation about this deadlock 

While running some tests it was observed that the async-profiler may deadlock a java process from dying, this happens when using the `jfrsync` feature  

The deadlock happens between:
* the thread that is starting the profiler
* the thread that is calling the `VM::VMDeath` hook

This deadlock happens since `VM::VMDeath` will need to acquire the `_state_lock`, but that lock is currently being held by the thread that is currently starting the profiler 

The thread starting the profiler is waiting for the JVM to process the start of the jfr recording which won't happen until the `VM::VMDeath` is completed 

* Profiler is started with `jfrsync` option
* JVM starts it's death process after the profiler start is invoked but before the jfr recording is processed
* JVM won't process the `MSG_START` until the death hook is completed 

```
Thread 66 (Thread 0xffff913171f0 (LWP 24816) "java"):
#0  0x0000ffff929218d8 in __lll_lock_wait () from /usr/lib64/libpthread.so.0
#1  0x0000ffff92919ed4 in pthread_mutex_lock () from /usr/lib64/libpthread.so.0
#2  0x0000ffff910856e8 in Mutex::lock (this=0xffff8c031230) at src/mutex.cpp:17
#3  MutexLocker::MutexLocker (mutex=..., this=<synthetic pointer>) at src/mutex.h:41
#4  Profiler::shutdown (args=..., this=0xffff8c031230) at src/profiler.cpp:2027
#5  VM::VMDeath (jvmti=<optimized out>, jni=<optimized out>) at src/vmEntry.cpp:425
#6  0x0000ffff91dd17ec in JvmtiExport::post_vm_death() () from jdk-17/lib/server/libjvm.so
#7  0x0000ffff91bf2dac in before_exit(JavaThread*, bool) () from jdk-17/lib/server/libjvm.so
#8  0x0000ffff91caf45c in JVM_Halt () from jdk-17/lib/server/libjvm.so
#9  0x0000ffff7c36496c in ?? ()
#10 0x00000005003b7098 in ?? ()

Thread 52 (Thread 0xffff02e061f0 (LWP 24952) "Async-profiler "):
#0  0x0000ffff9291de38 in pthread_cond_wait@@GLIBC_2.17 () from /usr/lib64/libpthread.so.0
[...]

PR #1619, opened 2025-12-10
Prevent profiler from deadlocking the JVM death when using jfrsync 

Baraa-Hasheesh:
### Description
As described in #1617 it's possible to get a deadlock between the thread handling `VMDeath` thread & the thread handling the `Profiler::start`

This deadlock happens as the JFR recording is waiting for the JVM to process the start of the recording which won't happen as the Flight recorder monitor thread is dead, and at the same time the `VMDeath` thread is waiting for the start thread to relinquish the `_state_lock` lock, hence a deadlock happens between the two threads 

In my change I change the `Profiler::shutdown` to use `tryLock` to peek into the look in attempt to acquire it as to prevent the deadlock from happening 

Please note, I did consider using `pthread_mutex_timedlock` but opted against it as it's not available on macOS

### Related issues
#1617 

### Motivation and context
Prevent async-profiler from deadlocking the JVM death

### How has this been tested?
manual testing

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -46,8 +46,6 @@ const int MAX_STRING_LENGTH = 8191;
 const u64 MAX_JLONG = 0x7fffffffffffffffULL;
 const u64 MIN_JLONG = 0x8000000000000000ULL;
 
-static bool _in_jfr_sync = false;
-
 enum GCWhen {
     BEFORE_GC,
     AFTER_GC
@@ -60,6 +58,7 @@ static jclass _jfr_sync_class = NULL;
 static jmethodID _start_method;
 static jmethodID _stop_method;
 static jmethodID _box_method;
+static bool _jfr_starting = false;
 
 static const char* const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr", "vm"};
 
@@ -1403,9 +1402,9 @@ Error FlightRecorder::startMasterRecording(Arguments& args, const char* filename
     int event_mask = args.eventMask() |
                      ((args._jfr_options ^ JFR_SYNC_OPTS) << EVENT_MASK_SIZE);
 
-    __atomic_store_n(&_in_jfr_sync, true, __ATOMIC_RELEASE);
+    __atomic_store_n(&_jfr_starting, true, __ATOMIC_RELEASE);
     env->CallStaticVoidMethod(_jfr_sync_class, _start_method, jfilename, jsettings, event_mask);
-    __atomic_store_n(&_in_jfr_sync, false, __ATOMIC_RELEASE);
+    __atomic_store_n(&_jfr_starting, false, __ATOMIC_RELEASE);
 
     if (env->ExceptionCheck()) {
         env->ExceptionDescribe();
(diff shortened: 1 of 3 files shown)


src/flightRecorder.cpp
```diff
@@ -46,6 +46,8 @@ const int MAX_STRING_LENGTH = 8191;
 const u64 MAX_JLONG = 0x7fffffffffffffffULL;
 const u64 MIN_JLONG = 0x8000000000000000ULL;
 
+static volatile bool _in_jfr_sync = false;
```
apangin: `volatile` in C does not provide the same guarantees as in Java.
Strictly speaking, it's incorrect to rely on `volatile` for checking a flag, it's better to use `__atomic_load_n`/`__atomic_store_n`.

src/profiler.cpp
```diff
@@ -2024,7 +2023,17 @@ Error Profiler::restart(Arguments& args) {
 }
 
 void Profiler::shutdown(Arguments& args) {
-    MutexLocker ml(_state_lock);
+    // Potential deadlock may happen between current thread & profiling thread due to usage of jfrsync
+    // To avoid that we use `tryLock` rather than `lock`, we keep trying to acquire the lock until either:
+    //     - lock is acquired
+    //     - _in_jfr_sync is set which indicates a risky state & a potential deadlock JDK-8373439
```
apangin: Try to be concise. E.g.
```
Workaround for JDK-8373439: starting JFR duing VM shutdown may hang forever,
so avoid acquiring _state_lock in this case.
```

src/profiler.cpp
```diff
@@ -2024,7 +2023,17 @@ Error Profiler::restart(Arguments& args) {
...
+    while (!_state_lock.tryLock()) {
+        if (FlightRecorder::inJfrSync()) {
+            Log::warn("Unable to acquire lock for shutdown hook, hook is being skipped");
```
apangin: Warnings bother developers, but this message is not actionable (i.e. user cannot prevent it from happening), and the situation is actually not as bad that it requires careful attention. It certainly should not be a warning. I even think it is not important enough to be INFO. Can be probably DEBUG.

Baraa-Hasheesh: changed [b1d4854](https://github.com/async-profiler/async-profiler/pull/1619/commits/b1d4854722f19e69cfd628aad17278b52be7e4c2)

src/profiler.cpp
Baraa-Hasheesh: Looks good I applied it [b1d4854](https://github.com/async-profiler/async-profiler/pull/1619/commits/b1d4854722f19e69cfd628aad17278b52be7e4c2)

src/flightRecorder.cpp
Baraa-Hasheesh: [b1d4854](https://github.com/async-profiler/async-profiler/pull/1619/commits/b1d4854722f19e69cfd628aad17278b52be7e4c2)

apangin: `volatile` is now redundant: __atomic provides required guarantees.

Baraa-Hasheesh: makes sense, removed [c1a7b80](https://github.com/async-profiler/async-profiler/pull/1619/commits/c1a7b80b550b49eb16bd93056fde8978ee42debd)

Issue #1566, opened 2025-10-27
Filter cpu/wall profiles by latency


apangin:
### Describe the feature

Add `jfrconv` option to filter profile by the given latency, i.e. retain only those samples that occurred during execution of the longest transactions.

### Use Case

Method Tracing feature added in v4.2 allows to record duration of the selected Java methods. Developers are often interested in what exactly contribute to that duration.

With JFR output format, Method Tracing can run in parallel with CPU and/or Wall Clock profiling. When running tracing and sampling together, it is possible to determine time intervals (spans) of the longest transactions and then find the corresponding parts of the CPU/Wall profile by matching the event timestamps. However, doing it manually is difficult and time consuming. I therefore propose to automate this task in the JFR converter.

### Proposed Solution

Add a converter option that extracts parts of the profile responsible for long requests, e.g.
```
jfrconv --cpu --latency 10ms input.jfr output.html
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
Pass by comment 

When converting jfr to html, we loose the time information of when the sample was captured

On a long running JFR for example a 10 minute JFR, this would make it harder to associate the high latency samples with exact high latency times/durations 

I propose first that we first add a grouping mechanism passed on the sample times for example `duration-grouping` 
-> `jfrconv --cpu --duration-grouping 1m`

This would group each 1 minute of samples together using for example a synthetic frame,
-> This in my opinion is a better user experience that doing a bunch of `--from` & `--to`

Together with the above it would make the analysis experience much easier 
-> `jfrconv --cpu --latency 10ms --duration-grouping 100ms`

what do you think @apangin

apangin:
@Baraa-Hasheesh I did not say output would be flame graph only. `--latency` option will only serve as an intermediate filter, not imposing any limitations on the target format. If timestamps are necessary, JFR can be converted to a heatmap or otlp format.

yellowbuoy123:
This would be excellent. If I understand correctly, it would give a similar effect to the idea of "trace profiles" where one can identify slow traces (or spans) and then see the sampled CPU profile for the duration of that slow trace/span?

apangin:
@yellowbuoy123 Yes, that is correct.

Baraa-Hasheesh:
this was done in #1620

PR #1620, opened 2025-12-12
Filter JFR events by latency

fandreuz:
### Description
In this PR I introduce a new filtering options based on method tracing (#1497) for jfr-converter.

The new option (`--latency`) allows setting a minimum threshold (milliseconds): events which do not belong to traces exceeding this threshold will be excluded from the conversion.

### Related issues
#1566 

### Motivation and context
This will help users by reducing the noise of unrelated events.

### How has this been tested?
New test in `JfrconverterTests`.

---

### DIFF
--- a/docs/ConverterUsage.md
+++ b/docs/ConverterUsage.md
@@ -75,6 +75,7 @@ JFR options:
                        # an absolute time in hh:mm:ss or yyyy-MM-dd'T'hh:mm:ss format;
                        # a relative time from the beginning of recording;
                        # a relative time from the end of recording (a negative number).
+    --latency MS       Retain only samples within MethodTraces of at least MS milliseconds
 
 Flame Graph options:
     --title STRING     Convert to Flame Graph with provided title


fandreuz:
Needs some more testing, but feedback is welcome on the approach.

src/converter/one/convert/JfrConverter.java
```diff
@@ -33,14 +37,33 @@ public JfrConverter(JfrReader jfr, Arguments args) {
...
 
+        Predicate<Event> threadStatesPredicate = makeThreadStatesPredicate();
         while (jfr.hasMoreChunks()) {
             // Reset method dictionary, since new chunk may have different IDs
             methodNames = new Dictionary<>();
 
             collector.beforeChunk();
-            collectEvents();
+            collectEvents(event -> timePredicate.test(event) && threadStatesPredicate.test(event) &&
+                                   latencyPredicate.test(event));
```
apangin: This level of abstraction is abuse of OOP. If makes code harder to read, maintain, and likely kills performance.
Will a custom collector help?

src/converter/one/convert/JfrConverter.java
```diff
@@ -33,14 +37,33 @@ public JfrConverter(JfrReader jfr, Arguments args) {
...
     public void convert() throws IOException {
-        jfr.stopAtNewChunk = true;
+        Predicate<Event> timePredicate = makeTimePredicate();
+        Predicate<Event> latencyPredicate;
+        if (args.latency > -1) {
+            jfr.stopAtNewChunk = false;
+
+            TimeIntervals intervals = new TimeIntervals();
+            long minLatencyTicks = args.latency * jfr.ticksPerSec / 1000;
+            collectEvents(MethodTrace.class, timePredicate, methodTrace -> {
```
apangin: JFR chunks can be heterogeneous (e.g. in jfrsync mode the same file includes a JDK Flight Recorder chunk and an async-profiler chunk; another example is separate JFR files merged together). In particular, they can have different time scale. So, reading all MethodTrace events from all chunks first and then processing all samples does not work in general.

fandreuz: >  In particular, they can have different time scale.

So different value of `ticksPerSec`? If that's the problem, I'd just convert everything to UTC.

src/converter/one/convert/JfrConverter.java
fandreuz: Check a3d3a6fe1aac0f864078694066bd28ab53a318e4

src/converter/one/convert/JfrConverter.java
fandreuz: Check 9256c73587de8a0ff7e9a3fdeb60e1ad09a33329

Baraa-Hasheesh:
@fandreuz new test is failing, 

can you also add some flamegraph examples on this

fandreuz:
> @fandreuz new test is failing

I know. I'm trying to find the smallest sleep numbers that make CI happy.

> can you also add some flamegraph examples on this

What do you mean?

Baraa-Hasheesh:
> What do you mean?

I mean show a normal CPU flamegraph & a flamegraph with latency filtering in a comment to see the before & after

fandreuz:
> > @fandreuz new test is failing
> 
> I know. I'm trying to find the smallest sleep numbers that make CI happy.

Actually, I looked into the JFR file produced on alpaquita and I found this:
```
jdk.MethodTrace {
  startTime = 13:03:48.762
  duration = 489 ms
  method = N/A
  eventThread = "thread1" (javaThreadId = 20)
  stackTrace = [
    test.jfrconverter.Tracer.traceMethod(CountDownLatch) line: 17
    test.jfrconverter.Tracer.lambda$main$0(CountDownLatch) line: 36
    test.jfrconverter.Tracer$$Lambda.0x0000000800000a08.run() line: 0
    java.lang.Thread.runWith(Object, Runnable) line: 1596
    java.lang.Thread.run() line: 1583
    ...
  ]
}
```
which should not happen. I hope replacing `currentTimeMillis` with `nanoTime` makes it behave.

src/converter/one/convert/JfrConverter.java
apangin: I see you kept Consumer, but added one more level of indirection - FilteringCollector class that only serves as a holder of code that was previously a part of the loop. That's not what I meant.
Try to minimize changes to the existing design. Any refactoring made should reduce the amount of code, not increase it.

fandreuz: Should be better now: ec26d68de2286278d815d6849adc7971ec57d43d

src/converter/one/convert/JfrConverter.java
```diff
@@ -33,14 +33,21 @@ public JfrConverter(JfrReader jfr, Arguments args) {
     }
 
     public void convert() throws IOException {
-        jfr.stopAtNewChunk = true;
+        TimeIntervals timeIntervals = null;
+        if (args.latency > -1) {
+            timeIntervals = readLatencyTimeIntervals();
+            if (timeIntervals.isEmpty()) {
+                throw new RuntimeException("Found zero jdk.MethodTrace events");
```
Baraa-Hasheesh: I would say this isn't an exception but rather the converter should produce an empty Flamegraph/heatmap 

This IMO is not different from running `--nativemem` on a file with no malloc samples

fandreuz: As a user, I tend to favor explicit errors over empty outputs. But if other features in the converter behave as you say, I'll remove this check.

src/converter/one/convert/JfrConverter.java
apangin: Yes, it is. Thanks.

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: > As a user, I tend to favor explicit errors over empty outputs

Same here, but I would vote to keep a consistent behaviour, you can test locally by using the jfr generated in the `AllocTests.livenessJfrHasStacks` test

Baraa-Hasheesh: Also the error is not worded correctly, as you might have `jdk.MethodTrace` samples but non match the selected latency

src/converter/Main.java
```diff
@@ -128,6 +128,7 @@ private static void usage() {
                 "     --dot              Dotted class names\n" +
                 "     --from TIME        Start time in ms (absolute or relative)\n" +
                 "     --to TIME          End time in ms (absolute or relative)\n" +
+                "     --latency L        Retain only events recorded during a jdk.MethodTrace with duration at least L ms\n" +
```
apangin: A bit shorter description:
```suggestion
                "     --latency MS       Retain only samples within MethodTraces of at least MS milliseconds\n" +
```

src/converter/one/convert/JfrConverter.java
```diff
@@ -33,14 +33,21 @@ public JfrConverter(JfrReader jfr, Arguments args) {
     }
 
     public void convert() throws IOException {
-        jfr.stopAtNewChunk = true;
+        TimeIntervals timeIntervals = null;
+        if (args.latency > -1) {
```
apangin: nit: >= 0

apangin: I'd move all this logic in `readTimeIntervals` method and leave only
```
TimeIntervals timeIntervals = readTimeIntervals();
```
This will keep the main conversion method clean and also provide a convenient extension point for supplying custom time intervals (even though it is not a requirement).

src/converter/one/convert/JfrConverter.java
apangin: Good point from @Baraa-Hasheesh. I can agree that specifying `--latency` when a recording has no Method Traces at all is an error, but when Method Traces exist but `--latency` is too high - this is a valid situation and should not end up in an exception.

src/converter/one/jfr/JfrReader.java
```diff
@@ -21,11 +20,14 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
```
apangin: Not used

src/converter/one/jfr/JfrReader.java
```diff
@@ -718,6 +704,17 @@ private void seek(long pos) throws IOException {
         }
     }
 
+    public void reset() throws IOException {
```
apangin: rewind?

src/converter/one/jfr/JfrReader.java
```diff
@@ -718,6 +704,17 @@ private void seek(long pos) throws IOException {
...
+        seek(0);
+        state = STATE_NEW_CHUNK;
+
+        buf.flip();
+        ensureBytes(CHUNK_HEADER_SIZE);
+        if (!readChunk(0)) {
+            throw new IOException("Incomplete JFR file");
+        }
```
apangin: Is it all needed? `readChunk(0)` cannot fail if JfrReader was successfully constructed.
`hasMoreChunks()` or `readEvent` will handle beginning of a new chunk anyway.

src/converter/one/jfr/JfrReader.java
```diff
@@ -743,4 +740,9 @@ private boolean ensureBytes(int needed) throws IOException {
         buf.flip();
         return buf.limit() > 0;
     }
+
+    public long eventTimeToNanos(long time) {
+        long nanosSinceChunkStart = (time - chunkStartTicks) * SECONDS_TO_NANOS / ticksPerSec;
```
apangin: Multipliers can be too large even for `long` type. I suggest using floating point like in other similar places.
`1e9 / ticksPerSec` can be probably extracted to a separate variable.

src/converter/one/convert/TimeIntervals.java
```diff
@@ -0,0 +1,54 @@
...
+                timeIntervals.remove(floor.getKey());
+                startInstant = floor.getKey();
+                endInstant = Long.max(endInstant, floorEnd);
+            }
+        }
+
+        timeIntervals.put(startInstant, endInstant);
+    }
+
+    public boolean belongs(long instant) {
```
apangin: `intervals.contain(instant)`

src/converter/one/convert/TimeIntervals.java
```diff
@@ -0,0 +1,54 @@
...
+        Map.Entry<Long, Long> entry = timeIntervals.floorEntry(instant);
```
apangin: JfrConverter normally allocates only 1 object per event on average. This line adds two more allocations per event that I don't think JVM can optimize. Considering how hot this method is, it's worth rewriting it in a garbage-free way, e.g. by using binary search in a `long[]` array. Note: it's OK to use TreeMap during construction, but for lookup it's better to convert it to a different structure.

src/converter/one/convert/JfrConverter.java
fandreuz: e154be86f7b15f7de5399f127328dc3450526057

src/converter/one/convert/JfrConverter.java
fandreuz: e154be86f7b15f7de5399f127328dc3450526057

src/converter/one/jfr/JfrReader.java
fandreuz: 2c5aecb48b52c2d5303bfe633b4c5785dd25c00c

src/converter/one/jfr/JfrReader.java
fandreuz: c3f92606a49560abe968c3640b91b1b2729a6d7f

src/converter/one/jfr/JfrReader.java
fandreuz: c3f92606a49560abe968c3640b91b1b2729a6d7f

src/converter/one/convert/TimeIntervals.java
fandreuz: 2d10f32b0ba14478dfc470cc3024b30f79185d4a

src/converter/one/jfr/JfrReader.java
fandreuz: 60628d3ed5e17adf5dbec313e45eb945ec2d7e25

src/converter/one/convert/TimeIntervals.java
fandreuz: Sure: da6971fc81e92bf70f733fa1356c206d85dd3176

src/converter/one/convert/JfrConverter.java
```diff
@@ -10,6 +10,7 @@
 import one.jfr.JfrReader;
 import one.jfr.MethodRef;
 import one.jfr.event.*;
+import one.convert.TimeIntervals.TimeIntervalsList;
```
apangin: Bad naming.

src/converter/one/convert/JfrConverter.java
```diff
@@ -126,7 +155,7 @@ protected long toTicks(long millis) {
         } else if (millis < 1500000000000L) {
             nanos += jfr.startNanos;
         }
-        return (long) ((nanos - jfr.chunkStartNanos) * (jfr.ticksPerSec / 1e9)) + jfr.chunkStartTicks;
+        return (long) ((nanos - jfr.chunkStartNanos) / jfr.nanosecondsPerTick) + jfr.chunkStartTicks;
```
apangin: Keep previous version here. Since `nanosecondsPerTick` usually < 1, such replacement will worsen computation precision.

src/converter/one/convert/TimeIntervals.java
```diff
@@ -0,0 +1,79 @@
...
+        int index = 0;
+        for (Map.Entry<Long, Long> entry : timeIntervals.entrySet()) {
+            startIntervals[index] = entry.getKey();
+            endIntervals[index] = entry.getValue();
+            ++index;
+        }
+        return new TimeIntervalsList(startIntervals, endIntervals);
+    }
+
+    public static final class TimeIntervalsList {
```
apangin: Exposing a nested class with a repetitive name `TimeIntervals.TimeIntervalsList` looks weird.

A better idea would be to expose `TimeIntervals` for lookup and have `TimeIntervals.Builder` for construction. But I would just use the same `TimeIntervals` class for both purposes despite it's not how public API is usually designed. It's not a public API though.
Anyway, I'm OK with both approaches, as long as it's not `TimeIntervals.TimeIntervalsList`.

src/converter/one/jfr/JfrReader.java
```diff
@@ -84,20 +84,7 @@ public JfrReader(String fileName) throws IOException {
...
         this.fileSize = ch.size();
-
-        buf.flip();
-        ensureBytes(CHUNK_HEADER_SIZE);
-        if (!readChunk(0)) {
-            throw new IOException("Incomplete JFR file");
-        }
-    }
-
-    public JfrReader(ByteBuffer buf) throws IOException {
```
apangin, line 95: Did you remove constructor intentionally? It is used in other projects.

src/converter/one/jfr/JfrReader.java
```diff
@@ -718,6 +706,14 @@ private void seek(long pos) throws IOException {
         }
     }
 
+    public void rewind() throws IOException {
+        seek(0);
+        state = STATE_NEW_CHUNK;
```
apangin, line 725: These two lines are enough, no?
Of course, if constructor does not call `rewind` (which it should not, IMO)

src/converter/one/jfr/JfrReader.java
```diff
@@ -743,4 +739,9 @@ private boolean ensureBytes(int needed) throws IOException {
         buf.flip();
         return buf.limit() > 0;
     }
+
+    public long eventTimeToNanos(long time) {
+        long nanosSinceChunkStart = (long) ((time - chunkStartTicks) * nanosecondsPerTick);
+        return chunkStartNanos + nanosSinceChunkStart;
```
apangin: I see you like excessively long variable names and extra temporary variables )
What about
```
return chunkStartNanos + (long) ((time - chunkStartTicks) * nanosPerTick);
```

src/converter/one/convert/JfrConverter.java
fandreuz: 350b05393413f8a136c6bfc110885c829cd8da54

src/converter/one/jfr/JfrReader.java
fandreuz: No, `ensureBytes` is needed too due to this check in `readChunk`:
```
if (pos + CHUNK_HEADER_SIZE > buf.limit() || buf.getInt(pos) != CHUNK_SIGNATURE) {
   throw new IOException("Not a valid JFR file");
}
```

fandreuz: 6214086f0d3769411f1e372c69b4d82c30f21968

src/converter/one/jfr/JfrReader.java
fandreuz: Yeah that was intended, I didn't think about that. 6214086f0d3769411f1e372c69b4d82c30f21968

src/converter/one/jfr/JfrReader.java
fandreuz: My brain has a very small cache :)

e9cabff75b019c0d876bade3a405f44438316656

src/converter/one/convert/TimeIntervals.java
fandreuz: I like the builder: 4950e89b0f4b5bb0a7ceb41e260876e61fa2dc1f

apangin:
All good now.
Will you update docs/ConverterUsage.md?

fandreuz:
> All good now. Will you update docs/ConverterUsage.md?

Sure, I guess 5b4fc6aaad463e29a8e5ee3bbb9598817edce044 is enough, or do we need anything else in the docs?

Issue #1622, opened 2025-12-15
asprof collect mode is quite sensitive to race conditions related to VMDeath


Baraa-Hasheesh:
### Describe the bug

The `asprof` collect does the following:
* Start the profiler 
* Waits for the selected user duration 
* Stops the profiler 

This works generally fine, however during various tests of async-profiler we noticed that if the `asprof` starts at the end of a java process life it's possible that the `asprof` "stop" call will have various race conditions with the VMDeath hook calls 

These race conditions include but not limited to:
* Asprof finishes the wait while process is alive but the process dies before "stop" is called 
-> jattach will fail with process not found 

* Asprof finishes the wait while process is alive but the VM inside the process dies before the "stop" is called 
-> various possible errors will occur on jattach, from unable to attach to unexpected EOF errors 

* "stop" is called but a race condition occurs on what will execute first, the death hook or the stop call
-> Not there is also a race between the "dump" portion of the "stop" command & the VMDeath hook

These various race conditions will cause the "asprof" command to fail, which is wrong, as from a general perspective the profiles. were collected normally 

### Expected vs. actual behavior

`asprof` collect shouldn't fail due to VMDeath after the profiler was started successfully 

### Reproduction Steps

No solid reproducer, but several tests are hitting these states

### Async-profiler version

latest

Baraa-Hasheesh:
>"stop" is called but a race condition occurs on what will execute first, the death hook or the stop call

Fixed in #1623

Other races remain

PR #1623, opened 2025-12-15
Fix incorrect exit code for asprof when running in collect mode

Baraa-Hasheesh:
### Description
This change tackles one of the race conditions described in #1622

Being the race condition of what will execute first, the `stop` at the end of the collect command or the `VMDeath` hook

If `VMDeath` is executed first, then it will stop the profiler, dump the profile & proceed to terminate the profiler, 
Following that the `stop` command will fail as the profiler is terminated 

In this change, an internal flag is passed from `asprof` to the profiler, as to mark the current stop command as one coming from collect (Other stops commands should still fail) as to give it special handling 
 

### Related issues
#1622

### Motivation and context
asprof shouldn't have a non zero exit code from VMDeath 

### How has this been tested?
manual testing 

---

### DIFF
--- a/src/arguments.cpp
+++ b/src/arguments.cpp
@@ -518,10 +518,6 @@ Error Arguments::parse(const char* args) {
     return Error::OK;
 }
 
-const char* Arguments::rawFile() {
-    return _file;
-}
-
 const char* Arguments::file() {
     if (_file != NULL && strchr(_file, '%') != NULL) {
         return expandFilePattern(_file);
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -68,8 +68,6 @@ static Instrument instrument;
 
 static ProfilingWindow profiling_window;
 
-static const char* _start_file = NULL;
-
 struct MethodSample {
     u64 samples;
     u64 counter;
@@ -1274,8 +1272,6 @@ Error Profiler::start(Arguments& args, bool reset) {
         startTimer();
     }
 
-    _start_file = args.rawFile();
-
     return Error::OK;
 
 error7:
@@ -1372,7 +1368,7 @@ Error Profiler::flushJfr() {
 
 Error Profiler::dump(Writer& out, Arguments& args) {
     MutexLocker ml(_state_lock);
-    if (_state == TERMINATED && _start_file != NULL && args.rawFile() != NULL && strcmp(_start_file, args.rawFile()) == 0) {
+    if (_state == TERMINATED && _global_args._file != NULL && args._file != NULL && strcmp(_global_args._file, args._file) == 0) {
         return Error::OK;
     } else if (_state != IDLE && _state != RUNNING) {
         return Error("Profiler has not started");
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -296,8 +296,6 @@ class Arguments {
 
     Error parse(const char* args);
 
-    const char* rawFile();
-
     const char* file();
 
     bool hasTemporaryLog() const;


src/arguments.cpp
```diff
@@ -488,6 +488,9 @@ Error Arguments::parse(const char* args) {
             CASE("inverted")
                 _inverted = true;
 
+            CASE("_collect")
+                _collect = true;
```
apangin: No "secret" options please.

src/profiler.cpp
```diff
@@ -1368,7 +1368,9 @@ Error Profiler::flushJfr() {
 
 Error Profiler::dump(Writer& out, Arguments& args) {
     MutexLocker ml(_state_lock);
-    if (_state != IDLE && _state != RUNNING) {
+    if (args._collect && _state == TERMINATED) {
```
apangin: OK, stopping profiler in `TERMINATED` state is not an error. But why should behavior differ when running `asprof` standalone vs. from a script? And why is this check in `dump`, if asprof invokes `stop`?

Baraa-Hasheesh: It's mostly my personal interpretation on how the profiler should behave 

>  And why is this check in dump, if asprof invokes stop

The fall through from stop to dump makes having this check in stop redundant

The only edge case would be if user called stop with no output file, but then it will be reported as a failure with a clear error `Profiler is not active` 

> why should behavior differ when running asprof standalone vs. from a script

In asprof "collect" mode both start & end are called with output file, which means we're guaranteed for the profiler to dump the profile on disk via VMDeath + global args 

However for a regular start/stop commands, the dump isn't guaranteed which is arguably an error 

```
asprof start -e cpu  915
asprof stop -f out.html
```

src/arguments.cpp
Baraa-Hasheesh: The problem I kept hitting, is that profiler has different behavior depending on various condition, I wanted to keep the behaviors untouched as is, 

For example, the following won't generate a log

```
asprof start -cpu -f out.html 1234
asprof stop 1234
```

while, this will generate a log 

```
asprof start -cpu -f out.html 1234
VMDeath
asprof stop 1234
```

so a "secret" made sense to not effect existing behaviors

apangin: `asprof` is nothing but a binary script that loads profiler library in a target process and invokes commands on it. It could as well be a shell script or a custom program that does a similar thing. And I know people use custom scripts. A hidden option in asprof will not help them.

src/profiler.cpp
apangin: Again, this is not specific to asprof; asprof is just a wrapper.
It should be a decision of libasyncProfiler to return an error or not.

src/arguments.cpp
Baraa-Hasheesh: I've updated my approach to be determined on the usage, 

If the user starts the asprof with a file, then it's fine for the stop call to fail on termination & that failed stop can be considered a success (the VMDeath would have done the dump for the user)

So the user can address this race in any custom script by providing the output file in both start & stop as the asprof collect is currently doing [ee54ad2](https://github.com/async-profiler/async-profiler/pull/1623/commits/ee54ad2308da6f68054b67a20d288a41bf987ca9)

src/profiler.cpp
Baraa-Hasheesh: I've updated my approach to be determined on the usage, 

If the user starts the asprof with a file, then it's fine for the stop call to fail on termination & that failed stop can be considered a success (the VMDeath would have done the dump for the user)

So the user can address this race in any custom script by providing the output file in both start & stop as the asprof collect is currently doing [ee54ad2](https://github.com/async-profiler/async-profiler/pull/1623/commits/ee54ad2308da6f68054b67a20d288a41bf987ca9)

src/profiler.cpp
```diff
@@ -68,6 +68,8 @@ static Instrument instrument;
 
 static ProfilingWindow profiling_window;
 
+static const char* _start_file = NULL;
```
apangin: This is unnecessary. If `_start_file` comes from `_global_args`, then you can access it via `_global_args`. But if it comes from elsewhere, the pointer will be invalid.

src/profiler.cpp
```diff
@@ -1368,7 +1372,9 @@ Error Profiler::flushJfr() {
 
 Error Profiler::dump(Writer& out, Arguments& args) {
     MutexLocker ml(_state_lock);
-    if (_state != IDLE && _state != RUNNING) {
+    if (_state == TERMINATED && _start_file != NULL && args.rawFile() != NULL && strcmp(_start_file, args.rawFile()) == 0) {
```
apangin: This is a bit better than `_collect`, but is still a hack. I was thinking of something simpler, like always returning `Error::OK` for `stop` command if we are in `TERMINATED`. Although this is not fully accurate, I can't imagine a case when this is a real problem.

apangin: Alright, what about this hack:
```
if (_state == TERMINATED && args.hasTemporaryLog()) {
    return Error::OK;
}
```
The effect is the same as in your first revision, just with less code changes.

src/profiler.cpp
Baraa-Hasheesh: that make sense

src/profiler.cpp
Baraa-Hasheesh: > like always returning Error::OK for stop command if we are in TERMINATED

The real problematic part is the `dump` command not the `stop` & the `dump` will fail if profiler is terminated (Not safe to access JNI & JVMTI)

IMO if the output selected by the user isn't dumped, then that should be considered a failure, what are you thoughts on this @apangin 

From this perspective the only way for the profiler to dump it correclty, is for the user to provide the output file on the start command, hence my hack 

> Alright, what about this hack:
>
> if (_state == TERMINATED && args.hasTemporaryLog()) {
>    return Error::OK;
> }
> The effect is the same as in your first revision, just with less code changes.

This will also mark normal stops as a success even if the profiler didn't really write the profiles, which is wrong

```
asprof start -e cpu 1234
sleep 10
asprof stop -f out.html 1234 <--- this hit the race condition & unable to generate the out.html
```

apangin: OK, agreed - let's proceed with the `file` trick as I don't see a simple way to solve this cleanly without redesigning the entire protocol.
A good news is that it will be a tiny local change: all you need is to compare `args._file` to `_global_args._file`.

Baraa-Hasheesh: handled here [2c1c53d](https://github.com/async-profiler/async-profiler/pull/1623/commits/2c1c53d20da374185bf035ea686471208751da76)

PR #1624, opened 2025-12-17
Address breaking changes in OTLP 1.9.0

fandreuz:
### Description
[OTLP](https://github.com/open-telemetry/opentelemetry-proto/releases) `v1.8.0` and `v1.9.0` introduced some breaking changes. I'm addressing them in this PR:
- https://github.com/open-telemetry/opentelemetry-proto/pull/712
- https://github.com/open-telemetry/opentelemetry-proto/pull/672
- https://github.com/open-telemetry/opentelemetry-proto/pull/708
- https://github.com/open-telemetry/opentelemetry-proto/pull/724
- https://github.com/open-telemetry/opentelemetry-proto/pull/692
- https://github.com/open-telemetry/opentelemetry-proto/pull/649

### Related issues
n/a

### Motivation and context
We should pick up with the recent breaking changes in OTLP.

### How has this been tested?
Updated tests.

### Note for reviewers
This is a dense PR. I divided changes in a commit for each addressed PR to simplify reviews. Alternatively, I could open five different PRs.

---

### DIFF
--- a/src/otlp.cpp
+++ b/src/otlp.cpp
@@ -124,7 +124,7 @@ void Recorder::record(const std::vector<CallTraceSample*>& call_trace_samples, b
     protobuf_mark_t resource_profiles_mark = _otlp_buffer.startMessage(ProfilesData::resource_profiles);
     protobuf_mark_t scope_profiles_mark = _otlp_buffer.startMessage(ResourceProfiles::scope_profiles);
 
-    size_t unit_strindex = samples ? _strings.indexOf("count") : _engine_unit_strindex;
+    size_t unit_strindex = samples ? _count_strindex : _engine_unit_strindex;
     recordOtlpProfile(_engine_type_strindex, unit_strindex, samples);
 
     _otlp_buffer.commitMessage(scope_profiles_mark);
--- a/src/otlp.h
+++ b/src/otlp.h
@@ -107,6 +107,7 @@ class Recorder {
     const u64 _duration_nanos;
     const size_t _engine_type_strindex;
     const size_t _engine_unit_strindex;
+    const size_t _count_strindex;
 
     // Record a profile with a specified sample type
     void recordOtlpProfile(size_t type_strindex, size_t unit_strindex, bool samples);
@@ -116,7 +117,7 @@ class Recorder {
 
   public:
     Recorder(Engine* engine, FrameName& fn, u64 start_nanos, u64 duration_nanos) :
-        _otlp_buffer(ProtoBuffer(Otlp::OTLP_BUFFER_INITIAL_SIZE)),
+        _otlp_buffer(Otlp::OTLP_BUFFER_INITIAL_SIZE),
         _fn(fn),
         _thread_names(),
         _functions(),
@@ -124,6 +125,7 @@ class Recorder {
         _samples_info(),
         _engine_type_strindex(_strings.indexOf(engine->type())),
         _engine_unit_strindex(_strings.indexOf(engine->units())),
+        _count_strindex(_strings.indexOf("count")),
         _start_nanos(start_nanos),
         _duration_nanos(duration_nanos) {}
 


src/converter/one/convert/JfrToOtlp.java
```diff
@@ -28,6 +28,8 @@ public class JfrToOtlp extends JfrConverter {
     private final Index<String> functionPool = new Index<>(String.class, "");
     private final Index<Line> linePool = new Index<>(Line.class, Line.EMPTY);
     private final Index<KeyValue> attributesPool = new Index<>(KeyValue.class, KeyValue.EMPTY);
+    private final Index<int[]> stacksPool = new Index<>(int[].class, new int[0]);
```
apangin: Index, being a HashMap, does not work for arrays. You'll need a wrapper that implements equals/hashCode.

fandreuz: TIL, had no idea. Why is it done like that?

apangin: Well, it actually _works_, but not how you would expect.
Java arrays have default (i.e. identity) `equals` and `hashCode` inherited from java.lang.Object.

fandreuz: Yeah, I meant why doesn't array have a reasonable implementation for `equals` and `hashCode`. I'll go for a wrapper, thanks.

apangin: > why doesn't array have a reasonable implementation for `equals` and `hashCode`

If you dive deep in this question, you'll realize that the only reasonable implementation is identity-based :)

fandreuz: e9dc4b9dcdf7dbfc7db54b7e999f7e0905abc974

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -257,14 +264,36 @@ public boolean equals(Object o) {
...
+        private final int hash;
+
+        public IntArrayWrapper(int[] array) {
+            this.array = array;
+            this.hash = Arrays.hashCode(array);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof int[] && Arrays.equals(array, (int[]) o);
```
apangin: `o` is `IntArrayWrapper`, not `int[]`

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -257,14 +264,36 @@ public boolean equals(Object o) {
...
+            result = 31 * result + keyStrindex;
             return 31 * result + value.hashCode();
         }
     }
+
+    private static final class IntArrayWrapper {
+        public static final IntArrayWrapper EMPTY = new IntArrayWrapper(new int[0]);
+
+        private final int[] array;
+        private final int hash;
```
apangin: public vs. private modifiers seem to be placed randomly.
I suggest just to use default (package-private) visibility for all, similar to other nested classes in this file.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -144,55 +165,51 @@ public static void convert(String input, String output, Arguments args) throws I
...
-            proto.field(SAMPLE_attribute_indices, attributesPool.index(threadName));
-
-            long svMark = proto.startField(SAMPLE_value, MSG_SMALL);
-            proto.writeLong(samples);
-            proto.writeLong(factor == 1.0 ? value : (long) (value * factor));
-            proto.commitField(svMark);
-
-            proto.commitField(sMark);
+            SampleInfo si = new SampleInfo(timeNanos, attributesPool.index(threadNameKv), stackIndex, samples,
+                                            factor == 1.0 ? value : (long) (value * factor));
```
apangin: One extra space :)

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -52,38 +54,51 @@ public void convert() throws IOException {
...
-        proto.field(VALUE_TYPE_unit_strindex, stringPool.index(getSampleUnits()));
-        proto.field(VALUE_TYPE_aggregation_temporality, AGGREGATION_TEMPORARALITY_cumulative);
-        proto.commitField(stsMark);
-
-        long sttMark = proto.startField(PROFILE_sample_type, MSG_SMALL);
-        proto.field(VALUE_TYPE_type_strindex, stringPool.index(getValueType()));
-        proto.field(VALUE_TYPE_unit_strindex, stringPool.index(getTotalUnits()));
-        proto.field(VALUE_TYPE_aggregation_temporality, AGGREGATION_TEMPORARALITY_cumulative);
-        proto.commitField(sttMark);
+            writeSamples(samplesInfo, false /* samples */);
```
apangin: OMG. Is it now necessary to write a full copy of the profile every time you add a counter?

src/profiler.cpp
```diff
@@ -1641,91 +1641,104 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
-    std::vector<size_t> location_indices;
-    location_indices.reserve(call_trace_samples.size());
+std::vector<SampleInfo> Profiler::recordStacks(ProtoBuffer& otlp_buffer, FrameName& fn, Index& thread_names, Index& functions, const std::vector<CallTraceSample*>& call_trace_samples) {
+    using namespace Otlp;
 
-    FrameName fn(args, args._style & ~STYLE_ANNOTATE, _epoch, _thread_names_lock, _thread_names);
-    size_t frames_seen = 0;
+    std::vector<SampleInfo> samples_info;
     for (const auto& cts : call_trace_samples) {
+        samples_info.push_back(SampleInfo{0, 0, 0});
```
apangin: Why are we writing dummy samples?

src/profiler.h
```diff
@@ -128,6 +132,9 @@ class Profiler {
     Engine* activeEngine();
     Error checkJvmCapabilities();
 
+    std::vector<SampleInfo> recordStacks(ProtoBuffer& otlp_buffer, FrameName& fn, Index& thread_names, Index& functions, const std::vector<CallTraceSample*>& call_trace_samples);
+    void recordOtlpProfile(ProtoBuffer& otlp_buffer, const std::vector<SampleInfo>& samples_info, const SampleTypeStrings& st_strings, u64 duration_nanos, bool count);
+
```
apangin: Is it possible not to expose otlp-specific implementation details in the profiler header?

test/test/otlp/OtlpTests.java
```diff
@@ -140,4 +188,10 @@ private static Optional<AnyValue> getAttribute(Sample sample, ProfilesDictionary
     private static void classpathCheck() {
         ProfilesData.newBuilder();
     }
+
+    private static Path getTempFile() throws IOException {
```
apangin: I assume you can convert JFR to OTLP in memory without a temporary file.

src/converter/one/convert/JfrToOtlp.java
fandreuz: Ouch, thanks: 5dcf7d3b280e56e07b17bf9617e177123a7f55ba

src/converter/one/convert/JfrToOtlp.java
fandreuz: The idea was: `private` for stuff you shouldn't touch, `public` for stuff you can touch. But yeah it's not needed here I guess: 5dcf7d3b280e56e07b17bf9617e177123a7f55ba

src/converter/one/convert/JfrToOtlp.java
fandreuz: fdde3400bef4a53f55a0c504127462514d297dc1

test/test/otlp/OtlpTests.java
fandreuz: Right, thanks: 5a0dc2b1d2c976b9f392c48b29990627f78fbf6e

src/profiler.h
fandreuz: Sure: 12eb92859d993c2714d9e32eee4385c435a5eeb5, 4eed7e0672e0ee163e4b7b61e636dace366a2b26

src/profiler.cpp
fandreuz: Good point. This is an artifact of a previous implementation idea I had in mind: I wanted to have matching indexes between the vector of `SampleInfo` and `CallTraceSamples`. It's not needed anymore, though I missed this piece. I moved it down a couple of lines.

902f752a5d670104eeea6ca63d64afb2a1dfb15b

src/converter/one/convert/JfrToOtlp.java
fandreuz: Yes. There was some extensive discussion here and in the related issue: https://github.com/open-telemetry/opentelemetry-proto/pull/649

[This](https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-2772648117) is the content summarizing the whole discussion. In short, the confusion was about how to assign meaning to each `repeated` field in `Sample` timestamps, values, and value type (the latter coming from the enclosing `Profile`. The previous structure looked indeed a bit confusing to me.

src/profiler.cpp
```diff
@@ -1641,91 +1641,99 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
-    _call_trace_storage.collectSamples(call_trace_samples);
+struct SampleInfo {
+    u64 samples;
+    u64 counter;
+    size_t thread_name_index;
+};
 
-    std::vector<size_t> location_indices;
-    location_indices.reserve(call_trace_samples.size());
+std::vector<SampleInfo> recordStacks(ProtoBuffer& otlp_buffer, FrameName& fn, Index& thread_names_index, Index& functions,
```
apangin: `static`.

Alternatively, create a local class that holds otlp_buffer, indexes and other things that you otherwise pass as arguments.

src/profiler.cpp
```diff
@@ -1641,91 +1641,99 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
...
-    size_t frames_seen = 0;
+    std::vector<SampleInfo> samples_info;
     for (const auto& cts : call_trace_samples) {
         CallTrace* trace = cts->acquireTrace();
         if (trace == NULL || excludeTrace(&fn, trace) || cts->samples == 0) continue;
 
-        protobuf_mark_t sample_mark = otlp_buffer.startMessage(Profile::sample, 1);
-        otlp_buffer.field(Sample::locations_start_index, frames_seen);
-        otlp_buffer.field(Sample::locations_length, trace->num_frames);
+        samples_info.push_back(SampleInfo{cts->samples, cts->counter, 0 /* thread_name_index */});
```
apangin: nit: You can push SampleInfo after you get thread_name_index to avoid rewriting it in place.

src/profiler.cpp
```diff
@@ -1641,91 +1641,99 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
     }
 }
 
-static void recordSampleType(ProtoBuffer& otlp_buffer, Index& strings, const char* type, const char* units) {
+struct SampleTypeStrings {
+    size_t engine_type_strindex;
+    size_t engine_units_strindex;
+    size_t count_strindex;
```
apangin: Redundant structure. You use both of these variables just once - you can just pass the one you need directly to `recordOtlpProfile`

test/test/otlp/OtlpTests.java
```diff
@@ -5,70 +5,104 @@
...
+        // total
+        checkThreadNames(getProfile(profilesData, 1), profilesData.getDictionary());
+    }
+
+    @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,file=%f")
+    public void threadNameFromJfr(TestProcess p) throws Exception {
+        Output out = p.waitForExit("%f");
+        assert p.exitCode() == 0;
+
+        ProfilesData profilesData = profilesDataFromJfr(p.getFile("%f").toPath(), new Arguments("--cpu", "--output", "otlp"));
```
apangin: Why `toPath`, if you use the String representation anyway?

fandreuz: Right, thanks: f6b4ad6aa704a96c0beb0787ee9c2cef29cf41d5

src/profiler.cpp
fandreuz: b714ccbf0b59f62712478b99d7f75b881119d7dc

src/profiler.cpp
fandreuz: 4e31248d2e3e4050630329f5510d76f8f909957c

src/protobuf.cpp
```diff
@@ -53,6 +53,19 @@ void ProtoBuffer::field(protobuf_index_t index, u64 n) {
     putVarInt(n);
 }
 
+void ProtoBuffer::fieldFixed64(protobuf_index_t index, u64 n) {
+    tag(index, I64);
+    ensureCapacity(sizeof(u64));
+    _data[_offset++] = (char) n;
```
apangin: `_data` is `unsigned char` (or `u8`)

fandreuz: 1a7f3488c6545480eb8ca3fa30d47d4af624a959

src/profiler.cpp
fandreuz: 3b4495be3ee49d5ddac7d41024eaf8c1c818c527

src/converter/one/convert/JfrToOtlp.java
apangin: Unfortunately, this choice was likely based on wrong assumptions:
> this might cause memory profiles to become ~2x larger, but since CPU profiles are usually larger, we think this is not going to matter too much in practice

Anyhow, such a design implied that CPU profile will not have totals recorded, only the call counts.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -257,14 +264,36 @@ public boolean equals(Object o) {
...
         @Override
         public int hashCode() {
             int result = 17;
-            result = 31 * result + key.hashCode();
+            result = 31 * result + keyStrindex;
             return 31 * result + value.hashCode();
         }
     }
+
+    private static final class IntArrayWrapper {
```
apangin: `Wrapper` is a bad suffix: it makes class names sound enterprisish without adding much for the meaning.
(Why isn't `java.lang.Integer` called `java.lang.IntWrapper`?)

`IntArrayKey` would make more sense. Or just `IntArray`.

src/converter/one/convert/JfrToOtlp.java
fandreuz: Should we revisit our current decision to always encode both totals and counts in OTLP output?

src/converter/one/convert/JfrToOtlp.java
fandreuz: 06794e0f9acfa8a2172ec33f7ca1ec75f574df5d

src/converter/one/convert/JfrToOtlp.java
apangin: I intuitively think that totals and counts together don't make much sense for CPU profile, since the shape of the profile will be the same. It can be different for Allocation and Lock profiles though. I'd probably record only one (depending on the `--total` option) for CPU/perf_events profiles, and both for other types. I'm open to alternative opinions as it's not fully clear how this feature will be used.

fandreuz: I wouldn't do something special for an engine, I'd just let OTLP output behave like other formats.

fandreuz: fb66835555aa57c0ae16bc925c33b3c4f896a131

src/otlp.cpp
```diff
@@ -0,0 +1,134 @@
...
+    protobuf_mark_t sample_type_mark = _otlp_buffer.startMessage(Profile::sample_type, 1);
+    _otlp_buffer.field(ValueType::type_strindex, type_strindex);
+    _otlp_buffer.field(ValueType::unit_strindex, unit_strindex);
+    _otlp_buffer.commitMessage(sample_type_mark);
+}
+
+void Recorder::recordOtlpProfile(size_t type_strindex, size_t unit_strindex, bool samples) {
+    protobuf_mark_t profile_mark = _otlp_buffer.startMessage(ScopeProfiles::profiles);
+
+    _otlp_buffer.fieldFixed64(Profile::time_unix_nano, _start_nanos);
```
apangin, line 101: Is it the only place with fixed64 timestamps? I guess all unix_nano should be fixed64.
Btw, the same holds for Java converter.

src/protobuf.cpp
```diff
@@ -53,6 +53,19 @@ void ProtoBuffer::field(protobuf_index_t index, u64 n) {
...
+    tag(index, I64);
+    ensureCapacity(sizeof(u64));
+    _data[_offset++] = (unsigned char) n;
+    _data[_offset++] = (unsigned char) (n >> 8);
+    _data[_offset++] = (unsigned char) (n >> 16);
+    _data[_offset++] = (unsigned char) (n >> 24);
+    _data[_offset++] = (unsigned char) (n >> 32);
+    _data[_offset++] = (unsigned char) (n >> 40);
+    _data[_offset++] = (unsigned char) (n >> 48);
+    _data[_offset++] = (unsigned char) (n >> 56);
```
apangin: You can write it with a single operation: `*(u64*)(_data + _offset) = n;`
We only support little-endian architectures.

fandreuz: > We only support little-endian architectures.

Ah interesting, I didn't know.

ca2167afd13133174cc097a8b7ad50405660ba74

src/otlp.cpp
fandreuz: Thanks, I fixed the JFR writer and added a test to check that: 2256a6098647fad925e6b30086a6a3280d573bfd

Baraa-Hasheesh:
`make test` fails locally for me after this change 

```
/home/bhashesh/.sdkman/candidates/java/current/bin/javac -source 8 -target 8 -Xlint:-options -XDignore.symbol.file \
         -implicit:none \
         -cp "build/jar/*:test/deps/*:test/gen/*:test/stubs" \
         -d build/test/classes \
         test/one/profiler/test/RunnableTest.java test/one/profiler/test/TestResult.java test/one/profiler/test/Arch.java test/one/profiler/test/Os.java test/one/profiler/test/TestDeclaration.java test/one/profiler/test/Output.java test/one/profiler/test/Test.java test/one/profiler/test/TestProcess.java test/one/profiler/test/Runner.java test/one/profiler/test/SourceCode.java test/one/profiler/test/Jvm.java test/one/profiler/test/Tests.java test/one/profiler/test/TestStatus.java test/one/profiler/test/Assert.java test/test/nonjava/JavaClass.java test/test/nonjava/NonjavaTests.java test/test/api/DumpOtlp.java test/test/api/BusyLoops.java test/test/api/DumpCollapsed.java test/test/api/MetricsTest.java test/test/api/JavaAgent.java test/test/api/ApiTests.java test/test/api/StopResume.java test/test/jfrconverter/JfrconverterTests.java test/test/jfrconverter/Tracer.java test/test/kernel/KernelTests.java test/test/kernel/ListFiles.java test/test/instrument/MethodTracingStop.java test/test/instrument/Recursive.java test/test/instrument/CpuBurnerManyTargets.java test/test/instrument/InstrumentTests.java test/test/instrument/CpuBurner.java test/test/cstack/CstackTests.java test/test/cstack/LongInitializer.java test/test/vtable/VtableTests.java test/test/proc/IoIntensiveApp.java test/test/proc/MultiThreadApp.java test/test/proc/ProcTests.java test/test/proc/MemoryIntensiveApp.java test/test/proc/ManyProcessApp.java test/test/proc/BasicApp.java test/test/proc/ShortLivedApp.java test/test/proc/CpuIntensiveApp.java test/test/pmu/Dictionary.java test/test/pmu/PmuTests.java test/test/nativemem/CallsAllNoLeak.java test/test/nativemem/CallsMallocCalloc.java test/test/nativemem/NativememTests.java test/test/nativemem/CallsRealloc.java test/test/nativemem/Native.java test/test/otlp/OtlpProfileTimeTest.java test/test/otlp/OtlpTests.java test/test/otlp/CpuBurner.java test/test/cpu/CpuBurner.java test/test/cpu/CpuTests.java test/test/stackwalker/StackwalkerTests.java test/test/stackwalker/StackGenerator.java test/test/c/CTests.java test/test/alloc/Hello.java test/test/alloc/MapReader.java test/test/alloc/AllocTests.java test/test/alloc/RandomBlockRetainer.java test/test/alloc/MapReaderOpt.java test/test/lock/DatagramTest.java test/test/lock/LockTests.java test/test/nativelock/NativeLock.java test/test/nativelock/NativelockTests.java test/test/nativelock/AllNativeLocks.java test/test/jfr/Ttsp.java test/test/jfr/JfrMultiModeProfiling.java test/test/jfr/JfrCpuProfiling.java test/test/jfr/JfrTests.java test/test/jfr/CpuLoad.java test/test/smoke/Alloc.java test/test/smoke/SmokeTests.java test/test/smoke/Cpu.java test/test/smoke/LoadLibrary.java test/test/smoke/Threads.java test/test/vmstructs/VmstructsTests.java test/test/comptask/Main.java test/test/comptask/ComptaskTests.java test/test/wall/WallTests.java test/test/wall/BusyClient.java test/test/wall/SocketTest.java test/test/wall/IdleClient.java test/test/recovery/Numbers.java test/test/recovery/CodingIntrinsics.java test/test/recovery/StringBuilderTest.java test/test/recovery/RecoveryTests.java test/test/recovery/Suppliers.java test/test/recovery/TimeLoop.java
test/test/otlp/OtlpTests.java:171: error: cannot access ProtocolStringList
        for (; keyStrindex < dictionary.getStringTableList().size(); ++keyStrindex) {
                                                          ^
  class file for com.google.protobuf.ProtocolStringList not found
```

Issue #1621, opened 2025-12-13
Loop & Timeout are confusing together


Baraa-Hasheesh:
### Describe the bug

Currently in async-profiler the following descriptions exists for `loop` & `timeout` 

```
timeout: The profiling duration, in seconds. The profiler will run for the specified period of time and then automatically stop.

loop: Run profiler in a loop (continuous profiling).
```

From the provided description a user might assume it's possible to use both at the same time for example, the user might want to run the profiler in continuous profiling with 1 hour profiling session window with a max run duration of 24 hours 
-> `--loop 1h --timeout 86400`

However this won't work as expected, in the above case the profiler will work in `loop` mode & generate a profiling session every 24 hours 

```
            CASE("timeout")
                if (value == NULL || (_timeout = parseTimeout(value)) == -1) {
                    msg = "Invalid timeout";
                }

            CASE("loop")
                _loop = true;
                if (value == NULL || (_timeout = parseTimeout(value)) == -1) {
                    msg = "Invalid loop duration";
                }
```

### Expected vs. actual behavior

I should be able to use both loop & timeout at the same time 

### Reproduction Steps

run loop & timeout together 

### Async-profiler version

latest

Baraa-Hasheesh:
Fixed in #1625

PR #1625, opened 2025-12-17
Fix confusing timeout & loop combined usage

Baraa-Hasheesh:
### Description
Currently using both `timeout` & `loop` together is confusing,

This change makes so that `timeout` is used to correctly stop the profiler once reached, & `loop` is used to control the duration of the profiling session

### Related issues
#1621

### Motivation and context
Fix confusing `timeout` & `loop` behaviour & allow combined usage

### How has this been tested?
manual testing

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1863,7 +1863,7 @@ void Profiler::timerLoop(void* timer_id) {
         }
 
         if ((current_micros = OS::micros()) >= loop_limit) {
-            expire(_global_args, OS::micros() < _stop_time);
+            expire(_global_args, current_micros < _stop_time);
             return;
         }
 


src/profiler.cpp
```diff
@@ -1267,8 +1267,11 @@ Error Profiler::start(Arguments& args, bool reset) {
     _start_time = OS::micros();
     _epoch++;
 
-    if (args._timeout != 0 || args._output == OUTPUT_JFR) {
-        _stop_time = addTimeout(_start_time, args._timeout);
+    if (args._timeout != 0 || args._loop != 0 || args._output == OUTPUT_JFR) {
+        _loop_time = addTimeout(_start_time, args._loop);
+        if (args._file_num == 0) {
```
fandreuz, line 1272: Using `_file_num` as a proxy to check whether this is the first `start` is not the best way to do it IMO, how about we check if `_start_time == 0` before L1267?

src/profiler.cpp
```diff
@@ -1846,7 +1849,8 @@ void Profiler::stopTimer() {
 
 void Profiler::timerLoop(void* timer_id) {
     u64 current_micros = OS::micros();
-    u64 sleep_until = _jfr.active() ? current_micros + 1000000 : _stop_time;
+    u64 loop_limit = std::min(_stop_time, _loop_time);
+    u64 sleep_until = _jfr.active() ? current_micros + 1000000 : std::min(_stop_time, loop_limit);
```
```suggestion
    u64 sleep_until = _jfr.active() ? current_micros + 1000000 : loop_limit;
```

src/profiler.cpp
Baraa-Hasheesh: both approaches should work I have no preference, @apangin what do you think?

src/profiler.cpp
```diff
@@ -1998,8 +2002,9 @@ Error Profiler::run(Arguments& args) {
 
 Error Profiler::restart(Arguments& args) {
     MutexLocker ml(_state_lock);
+    bool restart = OS::micros() < _stop_time;
```
fandreuz: I'd call this `should_restart`

src/profiler.cpp
Baraa-Hasheesh: What if you do `--loop 79s --timeout 1000s`

The timeout wouldn't be respected, as you would end up running for 1027 seconds

fandreuz:
```
std::min(_stop_time, loop_limit) ==
std::min(_stop_time, std::min(_stop_time, _loop_time)) ==
std::min(_stop_time, _stop_time, _loop_time) ==
std::min(_stop_time, _loop_time) ==
loop_limit
```
unless I'm missing something.

Baraa-Hasheesh: loop limit is recalculated at every iteration  -> start time is updated at every loop 

So let's say we start at time 0 & timeout is 10 & loop is 3

* iteration 1 => _stop_time = 10, _loop_limit = 3, _start_time = 0
* iteration 2 => _stop_time = 10, _loop_limit = 6, _start_time = 3
* iteration 3 => _stop_time = 10, _loop_limit = 9, _start_time = 6
* iteration 4 => _stop_time = 10, _loop_limit = 12, _start_time = 9

Baraa-Hasheesh: my bad, thanks for raising it, I'm pretty sure I just had a brain fart here :)

src/profiler.cpp
Baraa-Hasheesh: @fandreuz actually I though about this & `_start_time == 0` won't work, 

```
asprof start --loop 10 --timeout 200 -e cpu 1234
sleep 210
asprof start --loop 10 --timeout 200 -e cpu 1234
```

In the second start the `_start_time` won't be zero as it is maintained & never reset to 0 after being set

src/profiler.cpp
Baraa-Hasheesh: https://github.com/async-profiler/async-profiler/pull/1625/commits/0ebfd5287b962f65c1c41df68f7bc5fb421d5273

src/profiler.cpp
Baraa-Hasheesh: https://github.com/async-profiler/async-profiler/pull/1625/commits/0ebfd5287b962f65c1c41df68f7bc5fb421d5273

src/profiler.cpp
```diff
@@ -1267,8 +1267,11 @@ Error Profiler::start(Arguments& args, bool reset) {
     _start_time = OS::micros();
     _epoch++;
 
-    if (args._timeout != 0 || args._output == OUTPUT_JFR) {
-        _stop_time = addTimeout(_start_time, args._timeout);
+    if (args._timeout != 0 || args._loop != 0 || args._output == OUTPUT_JFR) {
+        _loop_time = addTimeout(_start_time, args._loop);
+        if (args._file_num == 0) {
+            _stop_time = addTimeout(_start_time, args._timeout);
```
apangin, line 1273: What if `_timeout == 0`?

src/profiler.cpp
```diff
@@ -1998,8 +2002,9 @@ Error Profiler::run(Arguments& args) {
 
 Error Profiler::restart(Arguments& args) {
     MutexLocker ml(_state_lock);
+    bool should_restart = OS::micros() < _stop_time;
```
apangin: Isn't it better to pass the flag from outside? I mean, from `timerLoop` which should be responsible for time-related decisions.

src/profiler.cpp
Baraa-Hasheesh: `_stop_time` would be set to `0x7fffffffffffffffULL` which effectively mean profiler will never timeout

-> same as old behaviour

src/profiler.cpp
Baraa-Hasheesh: That makes sense

Baraa-Hasheesh: @apangin I don't really like the name of the function "restart"

How would you feel if I rename it to "rotate"

apangin: I agree "restart" isn't accurate. But neither is "rotate".
Maybe "expire"? Because it is called when timer is expired.

Baraa-Hasheesh: "expire" is good for me, [bd0d93b](https://github.com/async-profiler/async-profiler/pull/1625/commits/bd0d93bbe847c55d3f1791e685bc544f23f2dd39)

PR #1629, opened 2025-12-21
Harden crash protection in StackWalker

### LINKED ISSUE #1626: Enhance crash protection in async-profiler VM stack walker
### Describe the bug

While unwinding the stacks, the async-profiler will do multiple dereference operations, these operation may occasionally result in a Segfault

This is fine, as the async-profiler uses multiple mechanism in various stack walking modes to prevent this Segfault from crashing the process 

This protection is achieved via:
* `SafeAccess::load` in FP & Dwarf 
* `setjmp/longjmp` in VM 

The problem is that the `setjmp/longjmp` is only applied on threads that have the `VMThread` VM struct

This means if the crash happens on native thread that isn't related to the VM, the application may crash.

This may happen in native application that uses a VM instance, or in a java application that may spin a native thread.

Unwinding the stacks shouldn't crash the process, therefore crash protection should extend for native threads in VM stack walker mode

### Reproduction Steps

* Start a VM inside a native application & start profiler on it (you can start it in agent mode)
* Add deliberate segfaults in the VM stack walker  

_No response_

### Async-profiler version

latest

### Environment details

_No response_

apangin:
### Description

Rework signal-based crash protection in `StackWalker::walk` to make it more robust and support native threads.

I made the following changes:

1. Instead of using `ThreadShadow::_exception_file` field from VMStructs, store `jmp_buf` pointer in a one of 16 (CONCURRENCY_LEVEL) slots. `walkVM` is always called under one of 16 locks that protect `_calltrace_buffer`, so there is always a slot that a thread can exlusively use. This also works for non-Java threads.
2. In SIGSEGV handler, look for a slot that currently holds a pointer to `jmp_buf` of the current thread. This `jmp_buf` should be close to the current stack pointer - usually within 8 KB, but I increased the limit to 32 KB just in case. Since `walkVM` can be called recursively (e.g. when running `cpu` and `wall` profiling together), we need to scan through all slots and find the closest one.
3. Refactor `StackWalker::walkVM` to keep only one version of the function instead of three. This also helps to decrease the number of passed arguments: there were already 8 parameters and I wanted to add another one (lock_index).
4. Add an extra sanity check to validate SP on each unwinding step. This brakes `deepStack` test, but this is actually for a good reason. It's useful to have a hard limit of the stack walking depth (currently 1 MB) to ensure we don't go too far and accidentally jump beyond the current stack. Stacks deeper than 1 MB are very rare in practice.

### Related issues

Resolves #1626

### How has this been tested?

Injected random segfaults in stack walking code and verified that it does not crash JVM.
Tested with extreme profiling settings on x86 and ARM64.
```
-agentpath:libasyncProfiler.so=start,event=cpu,interval=100us,wall=1ms,alloc,file=out.jfr
```

---

### DIFF
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -492,8 +492,8 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
 
 void StackWalker::checkFault() {
     // Search for a crash protection context located on the current thread stack.
-    // Since one thread may use multiple contexts due to recursive walkVM calls,
-    // need to walk through all of them and find the nearest one.
+    // Since one thread may use multiple contexts because of simultaneous profiling engines,
+    // we need to walk through all of them and find the nearest one (i.e. the most recent).
     jmp_buf* nearest_ctx = NULL;
     uintptr_t stack_distance = 32768;  // maximum allowed stack distance
     const uintptr_t current_sp = (uintptr_t)&nearest_ctx;


src/profiler.cpp
```diff
@@ -663,9 +663,9 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
             num_frames += java_frames;
         }
     } else if (event_type >= ALLOC_SAMPLE && event_type <= ALLOC_OUTSIDE_TLAB && _alloc_engine == &alloc_tracer) {
```
Baraa-Hasheesh, line 665: Not now, but I think we should consider merging this branch with the one above it

src/stackWalker.cpp
```diff
@@ -206,75 +202,66 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
...
-        vm_thread->exception() = &crash_protection_ctx;
-        if (setjmp(crash_protection_ctx) != 0) {
-            vm_thread->exception() = saved_exception;
-            if (depth < max_depth) {
-                fillFrame(frames[depth++], BCI_ERROR, "break_not_walkable");
-            }
-            return depth;
+    if (setjmp(current_ctx) != 0) {
+        crash_protection_ctx[lock_index] = NULL;
+        if (depth < max_depth) {
```
Baraa-Hasheesh, line 234: Do we really need this if check?
Wouldn't it make sense to always add the error frame?

The only cases where I can see this condition not adding the error frame are comptask & vtable

src/stackWalker.cpp
```diff
@@ -507,19 +485,28 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
...
+    crash_protection_ctx[lock_index] = NULL;
 
     return depth;
 }
 
 void StackWalker::checkFault() {
-    if (VMThread::key() < 0) {
-        // JVM has not been loaded or VMStructs have not been initialized yet
-        return;
+    // Search for a crash protection context located on the current thread stack.
```
Baraa-Hasheesh, line 494: I don't like this comment, as while it's possible for a single thread to capture multiple locks & have multiple jmp buffers assigned to it, it doesn't happen due to recursive calls of walkVM, as there is no such recursion

It simply happens due to the profiler collecting profiles from multiple engines in parallel, 
-> if a thread is collecting a synchronized sample type such as `malloc` & during the execution of `walkVM` a `cpu`/`wall` engine hits the same thread to record an execution/wall sample, then it will hold 2 different locks & 2 different jump buffers 

can we reword the comment to make it clearer

test/test/stackwalker/StackwalkerTests.java
```diff
@@ -36,8 +36,8 @@ public void deepStack(TestProcess p) throws Exception {
         p.waitForExit();
         assert p.exitCode() == 0;
         Output output = Output.convertJfrToCollapsed(p.getFilePath("%f"));
-        assert output.contains("test/stackwalker/StackGenerator.main[^;]*;" +
-                "test/stackwalker/StackGenerator.deepFrame[^;]*;" +
+        // Cannot reach stack bottom because of the MAX_WALK_SIZE hard limit
```
Baraa-Hasheesh, line 39: Do we need to apply workarounds for this or is this acceptable limitation of the stack walker?

-> for example we can adjust the bottom value when we recover from anchor to simulate a fresh start of the stack walking logic

src/profiler.cpp
apangin: Yes, they became more similar, but there are still important nuances that make it non-trivial to merge.

src/stackWalker.cpp
apangin: I think it's better to leave this check to help readers. All other frames are filled inside `while (depth < max_depth)` but this one is not. Even if we can't reach here from places other than loop, it's not immediately obvious by looking at the code.

test/test/stackwalker/StackwalkerTests.java
apangin: In your original version, this test checked that a stack trace cannot be fully recovered because of a too deep stack. I changed the test in [this commit](https://github.com/async-profiler/async-profiler/commit/ea095462ca20581b93ac3d3d7360c9b6db457175#diff-ef10a3b1ac8c872574e879135594d309f9598189ac73a7ab9e6cbf41eb4e1a67L50), but that was probably a bad idea as it was contrary to the original intention.

I believe a hard limit is necessary as an additional safety measure to make sure we don't accidentally jump too far even when using an anchor. And there should be a test verifying that the hard limit actually works; `deepStack` is exactly such a test.

src/stackWalker.cpp
apangin: By "recursive" I actually meant "reentrant".
Thanks for noticing - I rephrased the comment.

Baraa-Hasheesh: looks much better now, thanks

src/stackWalker.cpp
```diff
@@ -206,75 +202,66 @@ int StackWalker::walkDwarf(void* ucontext, const void** callchain, int max_depth
...
     JavaFrameAnchor* anchor = NULL;
-    if (details && vm_thread != NULL && vm_thread->isJavaThread()) {
-        anchor = vm_thread->anchor();
+    VMThread* vm_thread = VMThread::current();
+    if (vm_thread != NULL && vm_thread->isJavaThread()) {
+        // For simple stack traces (e.g. for allocation profiling)
+        // jump directly to the first Java frame
+        if (details) {
+            anchor = vm_thread->anchor();
+        } else if (!vm_thread->anchor()->restoreFrame(pc, sp, fp)) {
```
Baraa-Hasheesh, line 250: This chain `vm_thread->anchor()->restoreFrame(pc, sp, fp)` looks risky to me 

let's add `vm_thread->anchor() != null`

apangin: `VMThread::anchor()` does not dereference anything on stack, it just adds a constant offset to `VMThread*` pointer.
So if `vm_thread` is non-null, `vm_thread->anchor()` is guaranteed to be non-null.

apangin:
Thank you for the review.

PR #1630, opened 2025-12-23
Fix issue where the jfrconv uses native lock samples for leak detection

Baraa-Hasheesh:
### Description
`jfrconv` might be used to find memory leaks by using `--leak --nativemem`

However currently if the user selects `--leak --nativemem --nativelock` the `jfrconv` will end up using the leak aggregator on NativeLock events which is wrong 

This change gives priority for `nativemem` when selected to fix that, also for better UX `--leak` will now auto enable `nativemem`, so the user can just do `jfrconv file.jfr --leak` 

### Related issues
N/A

### Motivation and context
Fix cast exceptions when use user selects both `--leak` & `--nativelock` at same time 

### How has this been tested?
manual testing 

---

### DIFF
--- a/src/converter/Main.java
+++ b/src/converter/Main.java
@@ -35,10 +35,6 @@ public class Main {
             }
         }
 
-        if (args.leak) {
-            args.nativemem = true;
-        }
-
         for (int i = 0; i < fileCount; i++) {
             String input = args.files.get(i);
             String output = isDirectory ? new File(lastFile, replaceExt(input, args.output)).getPath() : lastFile;
--- a/src/converter/one/convert/JfrConverter.java
+++ b/src/converter/one/convert/JfrConverter.java
@@ -29,7 +29,7 @@ public abstract class JfrConverter extends Classifier {
         this.args = args;
 
         EventCollector collector = createCollector(args);
-        this.collector = args.leak ? new MallocLeakAggregator(collector, args.tail) : collector;
+        this.collector = args.nativemem && args.leak ? new MallocLeakAggregator(collector, args.tail) : collector;
     }
 
     public void convert() throws IOException {


src/converter/one/convert/Arguments.java
```diff
@@ -84,6 +84,10 @@ public Arguments(String... args) {
                 throw new IllegalArgumentException(arg);
             }
         }
+
+        if (leak) {
+            nativemem = true;
+        }
```
apangin: `Arguments` class is just a holder for user-specified options. It should not have any business logic.

Baraa-Hasheesh: Moved to the main of the converter [8d11605](https://github.com/async-profiler/async-profiler/pull/1630/commits/8d116052d2498d4d5f6b1a74c327b797f958ce9f)

apangin:
> also for better UX `--leak` will now auto enable `nativemem`

Whether it's a better UX is debatable. `leak` is an ambiguous word on its own: it's not obvious if it relates to Java heap leaks or native memory leaks. Should we choose to make `leak` a separate option, we would need to update documentation. Also, we would need to treat it separately in the code instead of implicitly modifying `Arguments`. IMO, it's not worth doing it now, especially in the context of this PR intended to fix a different issue.

Baraa-Hasheesh:
@apangin agreed, removed from leak from this change [4dabdd8](https://github.com/async-profiler/async-profiler/pull/1630/commits/4dabdd8ff4dc2d8118d8c01f9659d559e5ea110a) 

Should I create a separate issue to make using leak easier for user?

apangin:
> Should I create a separate issue to make using leak easier for user?

I'd say no. This does not seem a problem at all.

Issue #1585, opened 2025-11-10
perf stat and -e cycles --total disagree


franz1981:
### Describe the bug

Running perf stat with async profiler -e cycles --total against the same application produce a different number of total cycles.

### Expected vs. actual behavior

perf stat and async profiler -e cycles --total shouldn't disagree

### Reproduction Steps

```bash
# git clone https://github.com/franz1981/quarkus-profiling-workshop

# cd quarkus-profiling-workshop

# mvn clean package -DskipTests

# cd scripts

# ./benchmark.sh -e cycles -p -g -u time/cached
```
It should produce an output like 
```
----- Install ap-loader -----
[jbang] A script with name 'ap-loader' already exists, use '--force' to install anyway.
----- Install Hyperfoil -----
[jbang] A script with name 'wrk2' already exists, use '--force' to install anyway.
[jbang] A script with name 'wrk' already exists, use '--force' to install anyway.
----- Benchmarking endpoint http://localhost:8080/time/cached
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
 -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2025-11-10 05:28:44,678 INFO  [io.quarkus] (main) hello-world-app 1.0.0-SNAPSHOT on JVM (powered by Quarkus 3.29.0) started in 0.392s. Listening on: http://0.0.0.0:8080
2025-11-10 05:28:44,679 INFO  [io.quarkus] (main) Profile prod activated. 
2025-11-10 05:28:44,679 INFO  [io.quarkus] (main) Installed features: [cdi, rest, rest-jackson, scheduler, smallrye-context-propagation, vertx]
----- Quarkus running at pid 215978 using 2 I/O threads
----- Start all-out test and profiling
----- Waiting 16 seconds before profiling for 20 seconds
Running 40s test @ http://localhost:8080/time/cached
  2 threads and 10 connections
----- Starting async-profiler on quarkus application (215978)
----- Collecting perf stat on 215978
----- Showing stats for 16 seconds
Linux 6.17.6-300.fc43.x86_64 (fedora.fritz.box)         11/10/2025      _x86_64_        (32 CPU)

05:29:02 AM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
[...]

PR #1631, opened 2025-12-23
Scale counter in case of multiplexing (#1585)

vish-chan:
### Description
If there are more events than counters, the kernel uses time multiplexing to give each event (only PMU events) a chance to access monitoring hardware. With multiplexing, an event is not measured all the time. We need to scale the counter value based on total time enabled vs time running to estimate what the count would have been, had the event been measured during the entire run.

### Related issues
* https://github.com/async-profiler/async-profiler/issues/1585

### Motivation and context
* Concurrent profilers interfere with each other's PMU access, causing incorrect measurements.  While concurrent profiler usage isn't typical in most applications, it's a legitimate use case. This change improves the correctness of async-profiler's measurements by properly handling multiplexed profiling scenarios, ensuring accurate metrics regardless of other active profilers.

### How has this been tested?
* Local tests executed via `make test`:
```
Total test duration: 298.367 s
Results Summary:
PASS: 132 (100.0%)
FAIL: 0
SKIP (disabled): 0
SKIP (config mismatch): 4
SKIP (missing JAR): 5
TOTAL: 141
```
* Verified the change on the reproducer mentioned in [1585](https://github.com/async-profiler/async-profiler/issues/1585). 
```
# jbang VolatileSpin.java 2
``` 

```
perf stat -d -p 61571 sleep 10 & build/bin/asprof --total -e cycles -d 10 -t 61571 & wait | less
```

Results of perf stat: Scaling ratio = ~1.49 (100/66.90) 
```
 Performance counter stats for process id '61571':
    48,625,772,042      cycles                    #    2.447 GHz                      (66.90%)
 10.006516992 seconds time elapsed
```

Results of async-profiler: Scaling ratio = ~1.54 (41455/26785)
```
       total  percent  samples  top
  ----------  -------  -------  ---
 41455746368   99.84%    26785  VolatileSpin.lambda$main$0
```

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -73,7 +73,6 @@ struct MultiplexState {
 static const unsigned int MAX_MULTIPLEXED_FD = 65536;
 
 static MultiplexState multiplex_state[MAX_MULTIPLEXED_FD];
-
 static bool multiplex_state_dirty = false;
 
 static int fetchInt(const char* file_name) {
@@ -605,6 +604,9 @@ int PerfEvents::createForThread(int tid) {
     attr.disabled = 1;
     attr.wakeup_events = 1;
 
+    // flags for multiplexing support
+    attr.read_format = PERF_FORMAT_TOTAL_TIME_ENABLED | PERF_FORMAT_TOTAL_TIME_RUNNING;
+
     if (_alluser) {
         attr.exclude_kernel = 1;
     }
@@ -630,9 +632,6 @@ int PerfEvents::createForThread(int tid) {
     if (_record_cpu) {
         attr.sample_type |= PERF_SAMPLE_CPU;
     }
-
-    // flags for multiplexing support
-    attr.read_format = PERF_FORMAT_TOTAL_TIME_ENABLED | PERF_FORMAT_TOTAL_TIME_RUNNING;
 
     int fd;
     if (FdTransferClient::hasPeer()) {
@@ -728,8 +727,7 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
         default: {
             // Read counter with multiplexing metadata for accurate scaling
             struct PerfCounter counter;
-            size_t num_of_bytes = read(siginfo->si_fd, &counter, sizeof(counter));
-            if (num_of_bytes == sizeof(counter)) {
+            if (read(siginfo->si_fd, &counter, sizeof(counter)) == sizeof(counter)) {
                 u64 current_val = counter.value;
                 if (counter.time_enabled > counter.time_running) {
                     int fd = siginfo->si_fd;
@@ -746,11 +744,11 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
 
                         if (delta_running > 0 && delta_enabled > delta_running) {
                             // scaled counter = (counter) * (delta_enabled / delta_running)
-                            double ratio = (double) delta_enabled / delta_running;
+                            double ratio = (double)delta_enabled / delta_running;
                             return (u64)(current_val * ratio);
                         }
                     } else if (counter.time_running > 0) {
-                        double ratio = (double) counter.time_enabled / counter.time_running;
+                        double ratio = (double)counter.time_enabled / counter.time_running;
                         return (u64)(current_val * ratio);
                     }
                 }


src/perfEvents_linux.cpp
```diff
@@ -700,8 +709,19 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
...
+            // Read counter with multiplexing metadata for accurate scaling
+            struct read_result result;
+            size_t num_of_bytes = read(siginfo->si_fd, &result, sizeof(result));
+            if(num_of_bytes == sizeof(result)) {
+                // Calculate scaling ratio to compensate for multiplexing
+                // ratio > 1.0 indicates the counter was time-sliced
+                double ratio = (result.time_running > 0 && result.time_enabled > result.time_running) ? (double)result.time_enabled / result.time_running : 1.0;
+                double scaled_counter = result.counter * ratio;
+                return (u64) scaled_counter;
+            } else if (num_of_bytes == sizeof(u64)) {
```
apangin: When is this branch taken?

src/perfEvents_linux.cpp
```diff
@@ -700,8 +709,19 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
...
-            u64 counter;
-            return read(siginfo->si_fd, &counter, sizeof(counter)) == sizeof(counter) ? counter : 1;
+            // Read counter with multiplexing metadata for accurate scaling
+            struct read_result result;
+            size_t num_of_bytes = read(siginfo->si_fd, &result, sizeof(result));
+            if(num_of_bytes == sizeof(result)) {
+                // Calculate scaling ratio to compensate for multiplexing
+                // ratio > 1.0 indicates the counter was time-sliced
+                double ratio = (result.time_running > 0 && result.time_enabled > result.time_running) ? (double)result.time_enabled / result.time_running : 1.0;
+                double scaled_counter = result.counter * ratio;
```
apangin: This does not look right.
`counter` is reset to `0` on each sample by `ioctl(siginfo->si_fd, PERF_EVENT_IOC_RESET, 0);`
whereas `time_enabled` and `time_running` keep growing. These are cumulative times from the creation of perf_event, but `counter` is delta from the previous sample.

vish-chan: You're right - I missed our earlier discussion on this. I'll update the implementation to address your feedback. 

Adding the snippet from perf man for other's reference:
```
 PERF_EVENT_IOC_RESET
              Reset the event count specified by the file descriptor
              argument to zero.  This resets only the counts; there is no
              way to reset the multiplexing time_enabled or time_running
              values.

```

src/perfEvents_linux.cpp
```diff
@@ -700,8 +709,19 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
         case 3: return StackFrame(ucontext).arg2();
         case 4: return StackFrame(ucontext).arg3();
         default: {
-            u64 counter;
-            return read(siginfo->si_fd, &counter, sizeof(counter)) == sizeof(counter) ? counter : 1;
+            // Read counter with multiplexing metadata for accurate scaling
+            struct read_result result;
+            size_t num_of_bytes = read(siginfo->si_fd, &result, sizeof(result));
+            if(num_of_bytes == sizeof(result)) {
```
fandreuz: Nit: space after `if`

src/perfEvents_linux.cpp
vish-chan: This was just defensive coding, not sure if this branch will ever be taken in practice.

src/perfEvents_linux.cpp
```diff
@@ -58,6 +58,24 @@ enum {
     HW_BREAKPOINT_X  = 4
 };
 
+struct PerfSample {
```
apangin: "Sample" is a momentary thing, it does not have time_running.
`PerfCounter` would be a better name (and its `counter` field could be `value`).

src/perfEvents_linux.cpp
```diff
@@ -58,6 +58,24 @@ enum {
...
+    u64 _counter;
```
apangin: `struct` fields do not have an underscore prefix.

src/perfEvents_linux.cpp
```diff
@@ -700,8 +725,37 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
...
         case 4: return StackFrame(ucontext).arg3();
         default: {
-            u64 counter;
-            return read(siginfo->si_fd, &counter, sizeof(counter)) == sizeof(counter) ? counter : 1;
+            // Read counter with multiplexing metadata for accurate scaling
+            struct PerfSample result;
+            size_t num_of_bytes = read(siginfo->si_fd, &result, sizeof(result));
+            if (num_of_bytes == sizeof(result)) {
+                u64 current_val = result._counter;
+                if (result._time_running > 0 && result._time_enabled > result._time_running) {
```
apangin: `_time_running > 0` - what for?

src/perfEvents_linux.cpp
```diff
@@ -810,6 +864,11 @@ Error PerfEvents::start(Arguments& args) {
 
     adjustFDLimit();
 
+    if (multiplex_data_dirty) {
+        memset(multiplex_state_data, 0, sizeof(multiplex_state_data));
```
apangin: This will touch (and thus page-in) the entire array even when a few elements are accessed.
Perhaps it's better not to reset `multiplex_data_dirty` flag and clear individual counters in `PerfEvents::createForThread` as you do now.

src/perfEvents_linux.cpp
```diff
@@ -660,6 +681,10 @@ int PerfEvents::createForThread(int tid) {
         err = errno;
         Log::warn("perf_event ioctl failed: %s", strerror(err));
     } else {
+        if (multiplex_data_dirty && fd < MAX_MULTIPLEXED_FD) {
+            multiplex_state_data[fd]._time_enabled = 0;
+            multiplex_state_data[fd]._time_running = 0;
```
apangin: You need to reset it _before_ enabling the counter.

src/perfEvents_linux.cpp
vish-chan: Added this as defensive coding to prevent potential division by zero. But I don't have concrete evidence it could be zero in practice. 
Happy to remove if you feel it's unnecessary.

src/perfEvents_linux.cpp
vish-chan: Yes makes sense. We'll end up allocating 1MB memory even if only one counter was touched. I'll get rid of this code.

src/perfEvents_linux.cpp
vish-chan: Yes, I'll move this before the ioctl

src/perfEvents_linux.cpp
apangin: You don't even divide by `_time_running` except for the fallback branch (fd >= MAX_MULTIPLEXED_FD), so you can move the check there.

vish-chan: True, but I was thinking of the if statement as a `Can I scale check`. If `time_running == 0`, then we just return the counter value. So having `time_running  > 0` in the if statement makes it semantically more clear imo.

apangin: In theory, `time_running` can be 0 only because of an overflow. This is a valid (although unlikely) situation. This is not a problem as long as you compute diff using unsigned arithmetic. So I'd either move the check down or remove it altogether.

vish-chan: I'll move it below for completeness.

src/perfEvents_linux.cpp
```diff
@@ -58,6 +58,24 @@ enum {
...
+
+// Per-FD struct for storing perf-event multiplexing data
+struct MultiplexState {
+    u64 time_enabled; /* stores previous time_enabled */
+    u64 time_running; /* stores previous time_running */
+};
+
+static const unsigned int MAX_MULTIPLEXED_FD = 65536;
+
+static MultiplexState multiplex_state_data[MAX_MULTIPLEXED_FD];
```
apangin: nit: `data` is a bad word for a variable: it only makes a name longer without adding anything for the meaning. Every variable holds some sort of data :) `multiplex_state` would be just fine.
Similarly, `multiplex_data_dirty` -> `multiplex_state_dirty`

src/perfEvents_linux.cpp
```diff
@@ -700,8 +726,37 @@ u64 PerfEvents::readCounter(siginfo_t* siginfo, void* ucontext) {
         case 3: return StackFrame(ucontext).arg2();
         case 4: return StackFrame(ucontext).arg3();
         default: {
-            u64 counter;
-            return read(siginfo->si_fd, &counter, sizeof(counter)) == sizeof(counter) ? counter : 1;
+            // Read counter with multiplexing metadata for accurate scaling
+            struct PerfCounter result;
```
apangin: nit: result -> counter

src/perfEvents_linux.cpp
vish-chan: `Every variable holds some sort of data` - can't agree more.

Issue #1582, opened 2025-11-05
[jfrconv] Support --include/--exclude for heatmap and other formats


apangin:
### Describe the feature

`--include`/`--exclude` options allow filtering stack traces during JFR-to-FlameGraph conversion.
It would be useful to support such filtering for other output formats.

### Use Case

- Produce a heatmap or otlp for the specific threads only.
- Compose filter and search.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
> Produce a heatmap or otlp for the specific threads only.

wouldn't it be better to create a new argument `--filer` or `--thread-filter` for this?

> Compose filter and search.

Currently the API isn't the most user friendly for matching multiple patterns due to the need to maintain complex regexes (`pattern1|pattern2|pattern3`), how do you feel about supporting multiple --include & --exclude?

apangin:
> wouldn't it be better to create a new argument `--filer` or `--thread-filter` for this?

If the goal can be easily achieved with existing arguments, I see no reason to introduce new.

> how do you feel about supporting multiple --include & --exclude?

Again, if the problem can be easily solved today by existing means, there is little value in such changes. 
We can revise CLI if needed, but not in the scope of this issue.

apangin:
Thanks to @visheshruparelia, `jfrconv` now supports `--include`/`--exclude` options for JFR-to-heatmap, JFR-to-pprof and JFR-to-OTLP conversions.

PR #1633, opened 2025-12-24
support include/exclude flag for jfr to heatmap conversion

visheshruparelia:
### Description
Add support for `--include / --exclude` flag in JFR Converter when converting JFR to Heatmap.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1582 (Implementation for other output modes will be done in separate PR)

#### Functional testing (manual)
Command: `build/bin/jfrconv ./profile.jfr -o heatmap out.html`
Total samples (in heatmap view): 49,682,638
Total samples in the heatmap which match the regex `.*compressed_graph.*` : 49,195,508 (computed using the "regexp" search  feature in the heatmap).
##### `--include`
Command: `build/bin/jfrconv ./profile.jfr -o heatmap out.html --include ".*compressed_graph.*" `
Expected samples: 49,195,508 (as computed above)
Actual samples: 49,195,508

##### `--exclude`
Command: `build/bin/jfrconv ./profile.jfr -o heatmap out.html --exclude ".*compressed_graph.*" `
Expected samples: 49,682,638 - 49,195,508 = 487,130
Actual samples: 487,130

#### Performance testing
JFR size: ~760 MB
Command 1: `build/bin/jfrconv ./profile.jfr -o heatmap out.html`
Time taken: 5.335 s

Command 2: `build/bin/jfrconv ./profile.jfr -o heatmap out.html --include ".*compressed_graph.*" `
Time taken:  5.263 s

Command 3: `build/bin/jfrconv ./profile.jfr -o heatmap out.html --exclude ".*compressed_graph.*" `
Time taken: 1.09s

---

### DIFF
--- a/src/converter/Main.java
+++ b/src/converter/Main.java
@@ -104,8 +104,8 @@ public class Main {
                 "\n" +
                 "Conversion options:\n" +
                 "  -o --output FORMAT    Output format: html, collapsed, pprof, pb.gz, heatmap, otlp\n" +
-                "  -I --include REGEX    Include only stacks with the specified frames (only for flamegraph and heatmap)\n" +
-                "  -X --exclude REGEX    Exclude stacks with the specified frames (only for flamegraph and heatmap)\n" +
+                "  -I --include REGEX    Include only stacks with the specified frames\n" +
+                "  -X --exclude REGEX    Exclude stacks with the specified frames\n" +
                 "\n" +
                 "JFR options:\n" +
                 "     --cpu              CPU profile (ExecutionSample)\n" +
--- a/src/converter/one/heatmap/Heatmap.java
+++ b/src/converter/one/heatmap/Heatmap.java
@@ -419,19 +419,19 @@ public class Heatmap {
             if (method.methodName == 0) {
                 return symbolTable.getKey(method.className);
             }
-            return symbolTable.getKey(method.className) + "." + symbolTable.getKey(method.methodName);
+            return symbolTable.getKey(method.className) + '.' + symbolTable.getKey(method.methodName);
         }
 
         private boolean includeStack(int prototypeId) {
             if (args.include == null && args.exclude == null) {
                 return true;
             }
-            return includeCache.computeIfAbsent(prototypeId, k -> applyIncludeExcludeFilter(prototypeId));
+            return includeCache.computeIfAbsent(prototypeId, stackId -> applyIncludeExcludeFilter(stackId));
         }
 
         // Returns true if the stack should be included
-        private boolean applyIncludeExcludeFilter(int prototypeId) {
-            int[] stack = stackTracesRemap.get(prototypeId);
+        private boolean applyIncludeExcludeFilter(int stackId) {
+            int[] stack = stackTracesRemap.get(stackId);
             Pattern include = args.include;
             Pattern exclude = args.exclude;
             for (int i = 0; i < stack.length; i++) {
@@ -479,9 +479,9 @@ public class Heatmap {
                 cachedStackTrace[stackSize - 1] = getMethodIndex(key);
             }
 
-            int newPrototypeId = stackTracesRemap.index(cachedStackTrace, stackSize);
-            if (includeStack(newPrototypeId)) {
-                sampleList.add(newPrototypeId, timeMs);
+            int newStackId = stackTracesRemap.index(cachedStackTrace, stackSize);
+            if (includeStack(newStackId)) {
+                sampleList.add(newStackId, timeMs);
             }
         }
 


src/converter/Main.java
```diff
@@ -140,6 +140,10 @@ private static void usage() {
                 "                        and for default stacktraces from flamegraph to icicle\n" +
                 "  -I --include REGEX    Include only stacks with the specified frames\n" +
                 "  -X --exclude REGEX    Exclude stacks with the specified frames\n" +
-                "     --highlight REGEX  Highlight frames matching the given pattern\n");
+                "     --highlight REGEX  Highlight frames matching the given pattern\n" +
+                "\n" +
+                "Heatmap options:\n" +
+                "  -I --include REGEX    Include only stacks with the specified frames\n" +
```
Baraa-Hasheesh: It might be better to move these options to the common section rather than adding a new Heatmap section, what do you think?

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,18 +410,62 @@ private static class State {
             this.sampleList = new SampleList(blockDurationMs);
         }
 
+        private String resolveFrameName(Method method) {
```
Baraa-Hasheesh, line 415: IMO, doing this in the `JfrConverter` class would be easier, less code & will automatically propagate to all "JFR -> X" conversions for free

Now doing that in the `JfrConverter` might have some side effects that some might consider undesired, for example, it might filter out `free` events which might result in some `malloc` samples being incorrectly marked  as leak when using the `--leak` option 

However for that we can add a special check to always add the `free` event for LeakAggregator bypassing user inputs for both latency & include/exclude
-> Note, doing that special check will fix a current bug where `latency` filtering will remove the `free` event

@apangin what are you thoughts on this

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,18 +410,62 @@ private static class State {
...
+            }
+
+            Boolean cached = excludeCache.get(prototypeId);
+            if (cached != null) {
+                return cached;
+            }
+
+            boolean result = applyIncludeExcludeFilter(stack, stackSize);
+            excludeCache.put(prototypeId, result);
+            return result;
```
Baraa-Hasheesh: Function can can be simplified to. 

`return excludeCache.computeIfAbsent(prototypeId, prototypeId -> applyIncludeExcludeFilter(stack, stackSize));`

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,18 +410,62 @@ private static class State {
...
-            if (sampleList.getRecordsCount() >= LIMIT) {
+            if (sampleList.getRecordsCount() >= LIMIT || stackTraceId == 0) {
                 return;
             }
 
             int prototypeId = stackTracesCache.get(stackTraceId);
+            int[] prototype = stackTracesRemap.get(prototypeId);
             if (classId == 0 && !args.threads) {
-                sampleList.add(prototypeId, timeMs);
+                if (!excludeStack(prototypeId, prototype, prototype.length)) {
```
Baraa-Hasheesh: why not change the function to return true if the stack should be included rather than negating it's return value, that would make it more readable IMO

src/converter/Main.java
apangin: include/exclude options should be listed only once.

src/converter/one/heatmap/Heatmap.java
visheshruparelia: Yeah I don't mind. I did it this way to stay consistent with other excludeStack method defined [here](https://github.com/async-profiler/async-profiler/blob/2a4f329cbae8f849642d4cd74b766ced79ed3557/src/converter/one/convert/FlameGraph.java#L311).

src/converter/one/heatmap/Heatmap.java
apangin: That was one of the @visheshruparelia's ideas from the beginning, but we rejected it, since it's too early to filter events at collection time, because JfrConverter by design does not know how collector is going to consume events.

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.convert;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class BiDirectionalIndex<T> extends Index<T> {
+    private final Map<Integer, T> reverseIndex;
```
apangin: Map is not the best structure here.
Growable array would be more efficient, both in time and space.

src/converter/one/heatmap/Heatmap.java
visheshruparelia: [e2d12d9](https://github.com/async-profiler/async-profiler/pull/1633/commits/e2d12d9ef96bb2c07b593de3f4fbac8ecb2730ca)

src/converter/one/heatmap/Heatmap.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1633/commits/b6e66c20191a7d50e1fe7954e89d1eacb4fd00bb

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1633/commits/fe263e6678a8d7a9a206041bc4acc1176bf7ec61 

We could use `ArrayList<>` too but in other places our codebase maintains our own logic of regrowing arrays so went with that approach to stay consistent. I don't like type casting `(T[]) new Object[]`, is there a better way to do it?

src/converter/Main.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1633/commits/e090ef33baf544510a92d3fa2f7b232e3368af01 
Does this work?

src/converter/one/convert/BiDirectionalIndex.java
fandreuz: > I don't like type casting (T[]) new Object[], is there a better way to do it?

`ArrayList` would solve that for you

visheshruparelia: > We could use ArrayList<> too but in other places our codebase maintains our own logic of regrowing arrays so went with that approach to stay consistent

Yes I know ^

fandreuz: I wonder if there's a motivation to keep the manually-resized array? If there isn't any, I'd vote for `ArrayList` as it would simplify the code.

visheshruparelia: Something to consider - If we use `ArrayList<>`, we will have to do `list.add(idx, key)`. On paper, this is `O(n)` since it will shift all the elements starting from `idx` by one position to the right. However, `idx` is monotonically increasing based on the implementation of `super.index()` so amortized complexity will be `O(1)`. 
Will wait for @apangin's thoughts.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -394,8 +395,11 @@ private static class State {
         // Maps stack trace ID to prototype ID in stackTracesRemap
         final DictionaryInt stackTracesCache = new DictionaryInt();
         final Map<MethodKey, Integer> methodCache = new HashMap<>();
-        final Index<Method> methods = new Index<>(Method.class, Method.EMPTY);
-        final Index<String> symbolTable = new Index<>(String.class, "");
+        final BiDirectionalIndex<Method> methods = new BiDirectionalIndex<>(Method.class, Method.EMPTY);
+        final BiDirectionalIndex<String> symbolTable = new BiDirectionalIndex<>(String.class, "");
+
+        // Cache for exclude/include filter results per prototype ID
+        final Map<Integer, Boolean> includeCache = new HashMap<>();
```
Baraa-Hasheesh: Do you need to clear these data structures between chucks?

visheshruparelia: `Heatmap#State` class has it's own `StackStorage` where we remap JFR's stack trace ID to Heatmap's own ID i.e `PrototypeID`. When `addStack()` is invoked, `StackStorage#index()` is used to compute the `PrototypeID `. `StackStorage#index()` computes the hash based on the contents of the stack. This means, we have following conditions:
 1. We get two `stackTraceId`s with same ID but differing Method indices across chunks.
 2. We get two `stackTraceId`s with same ID and same Method indices across chunks but the human readable name of these indices are different than what they were in the other chunk.

For `1.`, we will get a fresh Prototype ID so we are good.
For `2.`, we will get the same Prototype ID. It sounds like an edge case but it indeed is possible. 

I think we should clear the `includeCache`. Thoughts?

src/converter/one/convert/BiDirectionalIndex.java
apangin: OK, let's analyze this from the beginning. What we need in this problem is a data structure that provides an efficient mapping from Key to Index and back from Index to Key. This is essentially a BiMap. Since Java does not have BiMap in the standard library, the usual pattern is to implement it with two HashMaps - that's what Vishesh initially did with `Map<Integer, T> reverseIndex`. This is simple and clear. Often, it would be totally fine to stop here. However, what I typically propose to senior engineers is to take one extra step and for the price of a smallish complication make the structure more efficient. What's important here is that the complication is really small, but the benefit is tangible - this is what distinguishes ahead-of-time optimization from premature optimization.

So we need an efficient map from small integers to objects. A map with two operations: `get(index)` and ...

Here we get to the main point. If we were to implement an arbitrary BiMap, we would need `put(index, value)`. `get(index)`+`put(index, value)` is the basic functionality of arrays. Note: we don't use ArrayList capabilities in this case: no insertion, no removal, no appending, no iteration. The only ArrayList "feature" we can reuse is auto-growth, but this is more like a side-effect of appending rather than a feature. After all, array growth is a matter of one-liner, not a strong reason for reuse.

However, in our case indices grow sequentially and we never rewrite object at a given index. So instead of `put(index, value)` we actually need `append(value)`. With this in mind, `ArrayList` starts making more sense.

Why manual array growing in other places then? Because of primitive values. There is no such thing as `ArrayList<long>` or `ArrayList<int>` in Java, so we either stay with boxed versions where performance is not important or implement our own lists (growable arrays) for primitives.

visheshruparelia: 9f82876

src/converter/one/heatmap/Heatmap.java
visheshruparelia: 9f82876

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+    private final ArrayList<T> reverseIndex;
+
+    public BiDirectionalIndex(Class<T> cls, T empty) {
+        this(cls, empty, 256);
+    }
+
+    public BiDirectionalIndex(Class<T> cls, T empty, int initialCapacity) {
+        super(cls, empty, initialCapacity);
+        // Super already checks if initial capacity is < 0
+        initialCapacity = initialCapacity == 0 ? 1 : initialCapacity;
```
apangin: Not needed for ArrayList

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package one.convert;
+
+import java.util.ArrayList;
+
+public class BiDirectionalIndex<T> extends Index<T> {
+    private final ArrayList<T> reverseIndex;
+
+    public BiDirectionalIndex(Class<T> cls, T empty) {
```
apangin: `Bidirectional` or just `Bidi`

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+        this.reverseIndex = new ArrayList<>(initialCapacity);
+        this.reverseIndex.add(empty);
+    }
+
+    @Override
+    public int index(T key) {
+        assert super.size() == reverseIndex.size();
+        int idx = super.index(key);
+        if (idx < reverseIndex.size()) {
+        // This means the key already exists in the index and calling reverseIndex.add(idx, key) will corrupt it.
```
apangin: Indentation.
The second part of the comment is meaningless. "Key already exists" should be enough.

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+            return idx;
+        }
+        if (idx > reverseIndex.size()) {
+            throw new IllegalStateException("Reverse index is out of sync");
+        }
```
apangin: Taking this branch would mean a flaw in the algorithm. This is an assertion failure.
`assert idx == reverseIndex.size();`

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+        reverseIndex.add(key);
+        return idx;
+    }
+
+    public T getKey(int idx) {
+        if (idx < 0 || idx >= reverseIndex.size()) {
+            throw new IllegalArgumentException("Invalid index: " + idx);
+        }
```
apangin: No need to repeat checks that ArrayList already does.

src/converter/one/convert/BiDirectionalIndex.java
```diff
@@ -0,0 +1,50 @@
...
+        T key = reverseIndex.get(idx);
+        if (key == null) {
+            throw new IllegalArgumentException("No key exists at index: " + idx);
+        }
```
apangin: When can this happen?

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,18 +412,55 @@ private static class State {
...
+        }
+
         public void addEvent(int stackTraceId, int threadId, int classId, byte type, long timeMs) {
-            if (sampleList.getRecordsCount() >= LIMIT) {
+            if (sampleList.getRecordsCount() >= LIMIT || stackTraceId == 0) {
                 return;
             }
 
             int prototypeId = stackTracesCache.get(stackTraceId);
+            int[] prototype = stackTracesRemap.get(prototypeId);
```
apangin: If there are no include/exclude filters (common case), this will be a redundant operation.

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: 2e71cc7

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: 2e71cc7

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: 2e71cc7

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: 2e71cc7

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: 2e71cc7

src/converter/one/convert/BiDirectionalIndex.java
visheshruparelia: Hmm, it won't. 
2e71cc7

src/converter/one/heatmap/Heatmap.java
visheshruparelia: Did some refactor, lmk if that works.
I still dont like the fact that we will do args.include and args.exclude non null check twice in case we invoke the overloaded `includeStack(int prototypeID)`.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,14 +412,59 @@ private static class State {
...
+            if (args.include == null && args.exclude == null) {
+                return true;
+            }
+            return includeCache.computeIfAbsent(prototypeId, k -> applyIncludeExcludeFilter(stack, stackSize));
+        }
+
+        private boolean includeStack(int prototypeId) {
+            if (args.include == null && args.exclude == null) {
+                return true;
+            }
```
Baraa-Hasheesh, line 428: This check is not needed

Baraa-Hasheesh: On this note, is the `private boolean includeStack(int prototypeId, int[] stack, int stackSize) {` function really needed?

```
private boolean includeStack(int prototypeId) {
    if (args.include == null && args.exclude == null) {
        return true;
    }

    return includeCache.computeIfAbsent(prototypeId, k -> applyIncludeExcludeFilter(prototype, prototype.length));
```

visheshruparelia: Yes it is needed, ref: https://github.com/async-profiler/async-profiler/blob/2a4f329cbae8f849642d4cd74b766ced79ed3557/src/converter/one/heatmap/Heatmap.java#L423

visheshruparelia: > This check is not needed

How so?

Baraa-Hasheesh: > > This check is not needed
> 
> How so?

This will call `includeStack(prototypeId, prototype, prototype.length);` 

That function also has the check making the check here redundant

Baraa-Hasheesh: > Yes it is needed, ref:

But at the point the `one.heatmap.StackStorage#index` is already called, which means at this point the stacks & it's length is already saved inside `stackTracesRemap`

So you don't really need to depend on the `cachedStackTrace` value & read any needed value directly from `stackTracesRemap`, 

please correct me if I'm reading the flow incorrectly?

visheshruparelia: >That function also has the check making the check here redundant

The whole point of creating an overloaded method is to avoid doing `stackTracesRemap.get(prototypeId);` when args are null. See Andrei's comment here: https://github.com/async-profiler/async-profiler/pull/1633#discussion_r2656695550

visheshruparelia: > > Yes it is needed, ref:
> 
> But at the point the `one.heatmap.StackStorage#index` is already called, which means at this point the stacks & it's length is already saved inside `stackTracesRemap`
> 
> So you don't really need to depend on the `cachedStackTrace` value & read any needed value directly from `stackTracesRemap`,
> 
> please correct me if I'm reading the flow incorrectly?

It is needed because we iterate the array here: https://github.com/async-profiler/async-profiler/pull/1633/files/e090ef33baf544510a92d3fa2f7b232e3368af01#diff-bbadbb048180426b18993c9676ea5b057a6a946f24bdb574b83e5bed32326cd9R434
If I don't know the upper bound of the index I will start reading garbage values resulting in incorrect application of filters.

Baraa-Hasheesh: I still don't understand why you can't do the following

```
private boolean includeStack(int prototypeId) {
    if (args.include == null && args.exclude == null) {
        return true;
    }
    return includeCache.computeIfAbsent(prototypeId, k -> applyIncludeExcludeFilter(prototype, prototype.length));
}
```

Again when you call `one.heatmap.StackStorage#index`, the stack is saved with it's exact size [here](https://github.com/async-profiler/async-profiler/blob/master/src/converter/one/heatmap/StackStorage.java#L49) 

In current implementation you call the `one.heatmap.StackStorage#index` before calling the `includeStack`, which again means by this point you access the stack information from the `stackTracesRemap`

visheshruparelia: Yeah, I see your point now. Will update.

apangin: Hold on. In the discussion of cosmetic changes, one important point was lost. Calling `stackTracesRemap.index` adds a new stack trace in the `StackStorage` (or ensures it already exists). `StackStorage` contents is then dumped in the result heatmap. This means, if `stackTracesRemap.index` is called before filtering, all stack traces will get recorded in the heatmap even when a filter includes only 1% of them. This results in a huge waste in terms of the heatmap size. Is it possible to retain only filtered stack traces in the output file?

Baraa-Hasheesh: We will need to have 2 different filtering logics at 2 different locations but that should be possible 

First we apply the filtering on the addStack side
* Prevent stack from being added to storage

Then in the addEvent we only check for thread & classId elements before calling the index method

apangin: Filtering at `addStack` won't work - it may reject valid traces too early.
OK, preventing traces from getting into StackStorage can be challenging; it will be likely easier to filter them at dump time. In fact, we can probably skip this optimization for now.

src/converter/one/heatmap/Heatmap.java
```diff
@@ -406,14 +412,52 @@ private static class State {
...
+            return symbolTable.getKey(method.className) + "." + symbolTable.getKey(method.methodName);
+        }
+
+        private boolean includeStack(int prototypeId) {
+            if (args.include == null && args.exclude == null) {
+                return true;
+            }
+            int[] prototype = stackTracesRemap.get(prototypeId);
+            return includeCache.computeIfAbsent(prototypeId, k -> applyIncludeExcludeFilter(prototype));
```
apangin: I'd move `stackTracesRemap.get(prototypeId)` inside lambda - so that it's not executed in the case of cache hit.

visheshruparelia: [fa64cd6](https://github.com/async-profiler/async-profiler/pull/1633/commits/fa64cd6da1a9f2bd81133b07e68c0fb964564969)

src/converter/one/heatmap/Heatmap.java
visheshruparelia: > it may reject valid traces too early.

Curious - In which scenario will we reject traces early?

Baraa-Hasheesh: `--threads --include {pattern}`

You can't reject the trace if it doesn't match the pattern in the `addStack` as the matching frame could be the thread name 

But in the common use case where the `--threads` isn't not specified & user generates a heatmap for execution samples, then it will work just fine

PR #1635, opened 2025-12-29
support include/exclude flag for jfr to otlp conversion

visheshruparelia:
### Description
Add support for `--include / --exclude` flag in JFR Converter when converting JFR to OTLP.

### Related issues
https://github.com/async-profiler/async-profiler/issues/1582

#### Functional testing (manual)
Command: `build/bin/jfrconv ./profile.jfr -o otlp out.proto`
Total samples: 3996 (This is computed by using [this](https://github.com/fandreuz/otel-profile-cli) and with command: `otel-profile-cli show out.proto | grep "sample: {" | wc -l`)
##### `--include`
Command: `build/bin/jfrconv ./profile.jfr -o otlp out_include.proto --include ".*compressed_graph.*"`
Count of samples: 351 (`otel-profile-cli show out_include.proto | grep "sample: {" | wc -l`)

Command: `otel-profile-cli show out_include.proto | grep "compressed" | wc -l`
Output: `1`

##### `--exclude`
Command: `build/bin/jfrconv ./profile.jfr -o otlp out_exclude.proto --exclude ".*compressed_graph.*" `
Expected samples: 3996 - 351 = 3645
Actual: 3645 (`otel-profile-cli show out_exclude.proto | grep "sample: {" | wc -l`)
Command: `otel-profile-cli show out_exclude.proto | grep "compressed" | wc -l`
Output: `0`

#### Performance impact analysis
JFR size: ~760 MB
Command 1: `build/bin/jfrconv ./profile.jfr -o otlp out.proto`
Time taken: 0.804 s

Command 2: `build/bin/jfrconv ./profile.jfr -o otlp out_include.proto --include ".*compressed_graph.*"`
Time taken: 0.867 s

Command 3: `build/bin/jfrconv ./profile.jfr -o otlp out_exclude.proto --exclude ".*compressed_graph.*"`
Time taken: 0.92s

---

### DIFF
--- a/src/converter/one/convert/JfrToOtlp.java
+++ b/src/converter/one/convert/JfrToOtlp.java
@@ -177,17 +177,13 @@ public class JfrToOtlp extends JfrConverter {
             this.samplesInfo = samplesInfo;
         }
 
-        private boolean excludeEvent(Event event) {
-            if (args.include == null && args.exclude == null) {
-                return false;
-            }
-            return excludeStack(event.stackTraceId, event.tid);
-        }
-
         private boolean excludeStack(int stackId, int threadId) {
-            StackTrace stackTrace = jfr.stackTraces.get(stackId);
             Pattern include = args.include;
             Pattern exclude = args.exclude;
+            if (include == null && exclude == null) {
+                return false;
+            }
+
             if (args.threads) {
                 String threadName = getThreadName(threadId);
                 if (exclude != null && exclude.matcher(threadName).matches()) {
@@ -198,6 +194,8 @@ public class JfrToOtlp extends JfrConverter {
                     include = null;
                 }
             }
+
+            StackTrace stackTrace = jfr.stackTraces.get(stackId);
             for (int i = 0; i < stackTrace.methods.length; i++) {
                 String name = getMethodName(stackTrace.methods[i], stackTrace.types[i]);
                 if (exclude != null && exclude.matcher(name).matches()) {
@@ -213,9 +211,10 @@ public class JfrToOtlp extends JfrConverter {
 
         @Override
         public void visit(Event event, long samples, long value) {
-            if (excludeEvent(event)) {
+            if (excludeStack(event.stackTraceId, event.tid)) {
                 return;
             }
+
             String threadName = getThreadName(event.tid);
             KeyValue threadNameKv = new KeyValue(threadNameIndex, threadName);
             int stackIndex = stacksIndexCache.computeIfAbsent(event.stackTraceId, key -> stacksPool.index(makeStack(key)));


visheshruparelia:
IMO the manual testing approach mentioned in the PR Description should be sufficient. Since I am no expert, welcome to any suggestions if there are better ways to verify the behavior.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -156,10 +158,44 @@ private final class OtlpEventToSampleVisitor implements EventCollector.Visitor {
...
+                    if (exclude == null) return true;
+                    include = null;
+                }
+            }
+            return include == null;
         }
 
         @Override
         public void visit(Event event, long samples, long value) {
+            if (!includeEvent(event)) {
```
Baraa-Hasheesh: **Style**

I would personally flip this function to be `exludeEvent`

My rule of thump is if the function output is always negated, it would be more readable to simply negate the function definition rather than always negating the output

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -156,10 +158,44 @@ private final class OtlpEventToSampleVisitor implements EventCollector.Visitor {
...
+        }
+
+        private boolean includeStack(StackTrace stackTrace, int threadId) {
+            Pattern include = args.include;
+            Pattern exclude = args.exclude;
+            for (int i = 0; i < stackTrace.methods.length + 1; i++) {
+                String name;
+                if (i == stackTrace.methods.length) {
+                    name = getThreadName(threadId);
+                } else {
```
Baraa-Hasheesh: This is quite hacky IMO & will always be applied even when user doesn't use `--threads`
-> Let's at least only apply this when `--threads` is selected by the user to be similar to other conversion flows

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -156,10 +158,44 @@ private final class OtlpEventToSampleVisitor implements EventCollector.Visitor {
...
+            this.includeCache = new HashMap<>();
+        }
+
+        private boolean includeEvent(Event event) {
+            if (args.include == null && args.exclude == null) {
+                return true;
+            }
+            StackTrace st = jfr.stackTraces.get(event.stackTraceId);
+            int tid = event.tid;
+            return includeCache.computeIfAbsent(event.stackTraceId + ":" + event.tid, k -> includeStack(st, tid));
```
Baraa-Hasheesh: Wouldn't it be better to have two maps?
* map to store if stackTraceId is included
* map to store if threadId is included 

you only include the event if both return true

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -147,6 +148,7 @@ public static void convert(String input, String output, Arguments args) throws I
     private final class OtlpEventToSampleVisitor implements EventCollector.Visitor {
         private final List<Integer> locationIndices;
         private final double factor = counterFactor();
+        private final Map<String, Boolean> includeCache;
```
Baraa-Hasheesh: I believe this should be cleared between chuncks

src/converter/one/convert/JfrToOtlp.java
fandreuz: +1

src/converter/one/convert/JfrToOtlp.java
fandreuz: Can you elaborate? Switching chunk won't change the result of `includeStack` for a given key

Baraa-Hasheesh: Multiple chunks can be a result of merging multiple JFR files 

Same thread Id can exist across multiple chunks but refers to a different thread in each chunk

In chunk 1, thread 1 can refer to MyThread1
While in chunk 10, that same thread ID can refer to MyThread10


Same for StackTraceID, it can theoretically be repeated across chunks but refers to different stacks in each chunk

src/converter/one/convert/JfrToOtlp.java
fandreuz: I would also extract it from the loop. If duplication is really a problem you can extract L183 to L189 to a separate function, but iterating +1 times and checking if we're at the sentinel iteration every time is both confusing and potentially inefficient.

src/converter/one/convert/JfrToOtlp.java
fandreuz: I got confused by the fact that the key is `String` so I overlooked that we're still storing IDs rather than human readable values. Yeah we definitely need to clear then.

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: I kinda expected controversy when I was raised the PR :) 

I see why it is hacky, how is it inefficient?

Baraa-Hasheesh: You perform the index check check at every loop

If you extract to outside the loop you will only check one time by doing `if (args.threads)`

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: >Multiple chunks can be a result of merging multiple JFR files

I had no idea, thank you for pointing out.

>I overlooked that we're still storing IDs rather than human readable values

Is there a way to store human readable "version" of a stack trace ID? Or am I missing something @fandreuz ?

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: That was my exact original implementation. I landed up in finding a bug while testing that implementation. Take this example: `--include "someRegex"`. Now, "someRegex" does not match thread name so `includeThread()` returns `false` but it matches a frame in the stack trace so `includeStack()` returns `true`. Since we do an `AND`, we end up _**excluding**_ which we don't want.

I am not saying it is not possible to extract thread cache and stack cache into separate maps, it just requires some more creativity. Let me think about it, but let me ask you -- is the current way of doing it really a big problem?

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: Hmm, okay point taken. Are index checks that big of a deal?

>If you extract to outside the loop you will only check one time by doing if (args.threads)

I will do this once we agree on an approach here: https://github.com/async-profiler/async-profiler/pull/1635#discussion_r2651913644

Baraa-Hasheesh: > Hmm, okay point taken. Are index checks that big of a deal?

Free performance & makes code more readable, I would say win/win

src/converter/one/convert/JfrToOtlp.java
Baraa-Hasheesh: I see, technically speaking you have 3 states

Matches include
Matches exclude 
No match

I'm okay with it as is, but if you want you can hack it using  NULL as no match (confirm with @apangin if acceptable)

So basically you would have the following truth table (you will need to start using `containsKey`)

True & Null  = true
False & Null = false
Null & Null = false

Baraa-Hasheesh: -> Again I'm fine with keeping it as is,

src/converter/one/convert/JfrToOtlp.java
fandreuz: > Is there a way to store human readable "version" of a stack trace ID?

Nope, I think we'll just have to clear that cache as we do in other places.

src/converter/one/convert/JfrToOtlp.java
fandreuz: > Hmm, okay point taken. Are index checks that big of a deal?

JIT is probably smart enough to make this a counted loop and extract the last iteration, my main point here is about the code being a bit confusing.

src/converter/one/convert/JfrToOtlp.java
apangin: Making a composite key with string concatenation is almost always a bad idea. Especially here where you can merge two `int`s in one `long` almost for free.

src/converter/one/convert/JfrToOtlp.java
apangin: +1 to the above comments: move thread name check out of the loop.

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1635/commits/4f55b70b48c6ea7d233586e1cc710f647aab2cb4

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1635/commits/4f55b70b48c6ea7d233586e1cc710f647aab2cb4

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1635/commits/4f55b70b48c6ea7d233586e1cc710f647aab2cb4

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1635/commits/4f55b70b48c6ea7d233586e1cc710f647aab2cb4

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -171,13 +174,53 @@ private final class OtlpEventToSampleVisitor implements EventCollector.Visitor {
...
+            if (args.threads) {
+                String threadName = getThreadName(threadId);
+                if (exclude != null && exclude.matcher(threadName).matches()) {
+                    return true;
+                }
+                if (include != null && include.matcher(threadName).matches()) {
+                    return false;
+                }
+            }
+            return include != null;
```
Baraa-Hasheesh, line 209: wouldn't it be better to check for `args.threads` first as it can lead to fast path if a match is found against the thread name?

```
private boolean excludeStack(int stackId, int threadId) {
    StackTrace stackTrace = jfr.stackTraces.get(stackId);
    Pattern include = args.include;
    Pattern exclude = args.exclude;
    if (args.threads) {
        String threadName = getThreadName(threadId);
         if (exclude != null && exclude.matcher(threadName).matches()) {
             return true;
         }
         if (include != null && include.matcher(threadName).matches()) {
             if (exclude == null) return false;
             include = null;
         }
    }

    for (int i = 0; i < stackTrace.methods.length; i++) {
         ....
    }
    return include != null;
}
```

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -73,6 +75,7 @@ protected void convertChunk() {
         writeSamples(samplesInfo, !args.total /* samples */);
 
         proto.commitField(pMark);
+        otlpEventToSampleVisitor.excludeCache.clear();
```
Baraa-Hasheesh: This should be in the `beforeChunk`

-> (Keep it consistent with other converters like heatmap)

Baraa-Hasheesh:
Question, do we need to check for "classId" here?

OTLP drops the "classId" element when doing the conversion:
* Heatmap, flamegraph, collapsed & pprof will have that information & include it in the filtering 
* OTLP will not apply any filtering on "classId"

src/converter/one/convert/JfrToOtlp.java
apangin: `excludeCache.clear()` makes no sense: neither here nor anywhere else.
`otlpEventToSampleVisitor` becomes garbage immediately after this line and the next chunk will create a new `OtlpEventToSampleVisitor` with a new `excludeCache`.

apangin: Furthermore, `excludeCache` itself is useless. `OtlpEventToSampleVisitor` runs on an instance of `EventAggregator` which aggregates events by stack traces. This means all visited stack traces are unique, and the cache will be never hit.

visheshruparelia:
> Question, do we need to check for "classId" here?
> 
> OTLP drops the "classId" element when doing the conversion:
> 
> * Heatmap, flamegraph, collapsed & pprof will have that information & include it in the filtering
> * OTLP will not apply any filtering on "classId"

@Baraa-Hasheesh I believe the OTLP spec is still evolving and thus the converter also has limited support as pointed out to me by @fandreuz. Since the converter drops `classId` by default, I don't see how adding a filter on it gives us any value.

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: [b431ca4](https://github.com/async-profiler/async-profiler/pull/1635/commits/b431ca4158713095ad05d7b3534eb0231a0826b0)

src/converter/one/convert/JfrToOtlp.java
visheshruparelia: https://github.com/async-profiler/async-profiler/pull/1635/commits/b431ca4158713095ad05d7b3534eb0231a0826b0

Issue #1112, opened 2025-01-20
Record failure counts in more places


apangin:
### Discussed in https://github.com/async-profiler/async-profiler/discussions/1111

Emit a JFR event whenever profiler fails to acquire a buffer lock.

PR #1637, opened 2026-01-02
Record JFR event to report buffer lock acquisition failures

fandreuz:
### Description
In this PR I introduce a custom JFR event to report currently unreported profiler failures, as discussed in the linked issues:
```
profiler.Failures {
  startTime = 10:45:28.729
  ticksSkipped = 2
}
```
The event is reported once per chunk, at the end of the chunk.

### Related issues
#1111, #1112 

### Motivation and context
The counter is already included in our profiler metrics (#1568), and having it as a JFR event doesn't cost much and could provide insights to our users about the profiler health.

### How has this been tested?
Manual testing. Hard to write a deterministic test since we don't have a knob to reduce `CONCURRENCY_LEVEL`.

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -258,6 +258,8 @@ class Recording {
     RecordingBuffer _proc_buf;
     ProcessSampler _process_sampler;
 
+    u64 _last_ticks_skipped;
+
     static float ratio(float value) {
         return value < 0 ? 0 : value > 1 ? 1 : value;
     }
@@ -272,6 +274,7 @@ class Recording {
         _bytes_written = 0;
         _memfd = -1;
         _in_memory = false;
+        _last_ticks_skipped = Profiler::instance()->_failures[-ticks_skipped];
 
         _chunk_size = args._chunk_size <= 0 ? MAX_JLONG : (args._chunk_size < 262144 ? 262144 : args._chunk_size);
         _chunk_time = args._chunk_time <= 0 ? MAX_JLONG : (args._chunk_time < 5 ? 5 : args._chunk_time) * 1000000ULL;
@@ -330,11 +333,16 @@ class Recording {
         close(_fd);
     }
 
+    void writePerChunkEvents() {
+        writeProfilerFailures(_buf);
+        writeNativeLibraries(_buf);
+    }
+
     off_t finishChunk() {
         flush(&_monitor_buf);
         flush(&_proc_buf);
 
-        writeNativeLibraries(_buf);
+        writePerChunkEvents();
 
         for (int i = 0; i < CONCURRENCY_LEVEL; i++) {
             flush(&_buf[i]);
@@ -803,6 +811,19 @@ class Recording {
         }
 
         jvmti->Deallocate((unsigned char*)keys);
+    }
+
+    void writeProfilerFailures(Buffer* buf) {
+        u64 new_ticks_skipped = Profiler::instance()->_failures[-ticks_skipped];
+        if (new_ticks_skipped == _last_ticks_skipped) return;
+
+        int start = buf->skip(1);
+        buf->put8(T_PROFILER_FAILURES);
+        buf->putVar64(TSC::ticks());
+        buf->putVar64(new_ticks_skipped - _last_ticks_skipped);
+        buf->put8(start, buf->offset() - start);
+
+        _last_ticks_skipped = new_ticks_skipped;
     }
 
     void writeNativeLibraries(Buffer* buf) {
--- a/src/jfrMetadata.cpp
+++ b/src/jfrMetadata.cpp
@@ -296,6 +296,11 @@ JfrMetadata::JfrMetadata() : Element("root") {
                 << field("ioRead", T_LONG, "I/O Read Bytes", F_BYTES)
                 << field("ioWrite", T_LONG, "I/O Write Bytes", F_BYTES))
 
+            << (type("profiler.Failures", T_PROFILER_FAILURES, "Profiler failures")
+                << category("Profiler")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
+                << field("ticksSkipped", T_LONG, "Profiler ticks skipped"))
+
             << (type("jdk.jfr.Label", T_LABEL, NULL)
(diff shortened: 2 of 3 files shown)


src/jfrMetadata.cpp
```diff
@@ -296,6 +296,11 @@ JfrMetadata::JfrMetadata() : Element("root") {
                 << field("ioRead", T_LONG, "I/O Read Bytes", F_BYTES)
                 << field("ioWrite", T_LONG, "I/O Write Bytes", F_BYTES))
 
+            << (type("profiler.Failures", T_PROFILER_FAILURES, "Profiler failures")
+                << category("Profiler")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
+                << field("ticksSkipped", T_LONG, "Profiler ticks skipped"))
```
Baraa-Hasheesh, line 302: This name is confusing to me as you're recording the number of skipped samples 

It doesn't make sense to me to call it ticks skipped 

https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1425

src/flightRecorder.cpp
```diff
@@ -272,6 +274,7 @@ class Recording {
         _bytes_written = 0;
         _memfd = -1;
         _in_memory = false;
+        _last_ticks_skipped = Profiler::instance()->_failures[-ticks_skipped];
```
Baraa-Hasheesh, line 277: Is this really needed?

Failures are set to 0 when a new start command is issued 

https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1119

Baraa-Hasheesh:
General suggestion, rather than calling the item "failures", how about we call it stats & report other stats in it?

Like total walk time, total samples, etc..

fandreuz:
> General suggestion, rather than calling the item "failures", how about we call it stats & report other stats in it?

Yeah that sounds reasonable. I mainly called it this way because of how the ticket describes the intended usage, but something like stats or metrics would be better. Perhaps its content could report everything we have in `Profiler::writeMetrics`.

src/flightRecorder.cpp
fandreuz: This increases the hidden coupling between `Recording` and `Profiler`, I'd rather use `Profiler`s interface than relying on hidden behavioral dependencies.

Baraa-Hasheesh: I do see where you are coming from, but I don't fully agree, as this is basically a "X - 0" operation,

Also if the user uses `resume` command, this would result in only reporting the difference between profiling sessions, which doesn't seem right to me

apangin:
Thank you for looking into this issue.
In fact, this is not a minor feature; it may have long-term consequences and not only to async-profiler. Imagine that other tools integrating async-profiler (e.g. IntelliJ IDEA) will need to support this new event and maintain it forever even if we later decide to report skipped samples differently (or if we modify the algorithm to record all samples without skipping). So I suggest to think about the problem carefully and come up with a solid design proposal first.

The first question to answer is - can we rework locking scheme in a way that we never miss samples? If we manage to do so, that would be a preferrable solution.

Then, if it is not feasible to avoid lost samples, can we record them in a way that requires minimum changes (ideally none) on the reader/visualizer side? E.g. something like a preallocated synthetic frame where we only increment counter. For JFR, just as an idea, we can emit a synthetic WallClock event every second in `FlightRecorder::timerTick`. Happy to hear more thoughts.

Finally, if we choose to stick with a new JFR event, would it make sense to reuse `jdk.CPUTimeSamplesLost` since it serves a similar purpose?

Baraa-Hasheesh:
> would it make sense to reuse jdk.CPUTimeSamplesLost since it serves a similar purpose?

I would vote to not use the mentioned event as it would be confusing, as in async-profiler the lost sample can be of any sample type (CPU, WALL, ALLOC, LOCK, etc...)

fandreuz:
The lock guards the following resources:
| Resource                          | Current Solution                          | Possible Alternative                     |
|--------------------------------------|-------------------------------------------|------------------------------------------|
| Call trace buffer for `ASGCT_CallFrame`/`jvmtiFrameInfo` | `CallTraceBuffer* _calltrace_buffer[CONCURRENCY_LEVEL]`, allocated in `Profiler::start` | Spin-lock guarded buffer |
| `jmp_buf` for stack walking          | `static jmp_buf* crash_protection_ctx[CONCURRENCY_LEVEL]`  | `static thread_local jmp_buf* crash_protection_ctx[ENGINES_COUNT]` |
| JFR buffer                           | `RecordingBuffer _buf[CONCURRENCY_LEVEL]`         | ? |

For the first row, stack allocation would solve the problem nicely, but I guess that's not feasible (2048 + 138 * 16byte = ~35kb).

And even if we had some simple lock-free buffer, we would still have problems handling re-entrancy: if `recordSample` is interrupted half-way through, we can't reuse the same buffer as that would corrupt the stopped recording. So, to guarantee zero dropped samples, we need:
- at least `ENGINES_COUNT` buffers (or we deadlock);
- to accept that a thread may be stuck in `recordSample`, spinning until a buffer becomes available.

And something similar goes for JFR buffers. OpenJDK's JFR implementation seems to be using thread local buffers.

fandreuz:
> Then, if it is not feasible to avoid lost samples, can we record them in a way that requires minimum changes (ideally none) on the reader/visualizer side?

I'm just throwing the idea here, perhaps this kind of profiler telemetry is something we could propose upstream first, and implement in Async-Profiler later? I checked in [metadata.xml](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/jfr/metadata/metadata.xml) and couldn't find anything providing extensive information about the health of the profiler (be it JFR or Async-Profiler).

krk:
`CPUTimeSamplesLost.lostSamples` is recently added that to report lost sample counts because of buffer overruns (and other possible reasons): https://github.com/openjdk/jdk/blob/49f7265894652ea243f3a531cf3f9d0b06e53565/src/hotspot/share/jfr/metadata/metadata.xml#L1003. This currently refers to the loss of `CPUTimeSample` events.

apangin:
Closing for now, as we will likely implement this feature in a different way later.

Issue #1636, opened 2025-12-31
The CPU flame graph doesn't include Java code stack traces.   Invalid Recrod use V4.2.1


thinkhejie:
### Describe the bug

The CPU flame graph doesn't include Java code stack traces.  

<img width="3002" height="588" alt="Image" src="https://github.com/user-attachments/assets/27db6a87-bff4-4d62-a39c-723077ab5a71" />

### Async-profiler version

4.2.1

thinkhejie:
cmd    

./asprof -d 30 -f /tmp/cpu_profile.0924.html --all-user 70483

thinkhejie:
if use async-profiler v4.0 for recording  ; then the recrod html file include Java code stack traces 

<img width="3018" height="616" alt="Image" src="https://github.com/user-attachments/assets/3906b826-b6ca-43ff-8ff1-dc3cc847b827" />

Baraa-Hasheesh:
Is it possible to provide a reproducer for this and more details regarding it
* Java version being used
* System details: kernel version, arm/x86

apangin:
It seems that you profile an idle application that does not consume CPU resources, therefore CPU profile is empty. This is an expected behavior. If you run profiler in the wall-clock mode (`-e wall`) you'll likely see all Java threads, including idle.

Also, why did you add `--all-user` option? It has no effect on macOS.

thinkhejie:
I use macOS 

25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:40 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6000 arm64

JDK is  version 21

thinkhejie:
I use JDK 25 ，the problem is also。

apangin:
Thank you for the update.
Given the information provided, I see no issues with your profile.
As mentioned [above](https://github.com/async-profiler/async-profiler/issues/1636#issuecomment-3703849436), this is an expected outcome of CPU-profiling an idle process. If you tell more about the original problem you want to solve with async-profiler, I'll try to recommend some options that may help.

thinkhejie:
[cpu_profile.0106.1.html](https://github.com/user-attachments/files/24426228/cpu_profile.0106.1.html)
[cpu_profile.0106.2.html](https://github.com/user-attachments/files/24426229/cpu_profile.0106.2.html)
[Cpu1.java](https://github.com/user-attachments/files/24426231/Cpu1.java)

[...]

Issue #819, opened 2023-09-19
Some unwinding problems with the new vm cstack mode on AArch 64


yanglong1010:
Hi Andrei

Recently I was trying out the new stack unwinding implementation (--cstack vm) and found that there may be two problems here on AArch64.

reproducer

```
Linux iZbp1gsbpsgn4675kr8vfeZ 4.18.0-348.20.1.el7.aarch64 #1 SMP Wed Apr 13 20:57:50 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux`

openjdk version "1.8.0_352"
OpenJDK Runtime Environment (build 1.8.0_352-b08)
OpenJDK 64-Bit Server VM (build 25.352-b08, mixed mode)
```
```java
public class Test {
    public static void main(String[] args) {
        long time = System.currentTimeMillis();
        while (true) {
            long n = wrapper();
            if (n - time > 1000) {
                System.out.println("tick");
                time = n;
            }
        }
    }

    private static long wrapper() {
        return test();
    }

    private static long test() {
        return System.currentTimeMillis();
    }
}
```
```
java -Xcomp Test

asprof -e cpu -f /root/profile.html -d 5 --cstack vm jps
```

The first problem is that the call_stub and native methods below `main` are `unknown`.

![image](https://github.com/async-profiler/async-profiler/assets/24515208/ca5ae3b3-bed8-41f6-8253-01677513489a)

After discussing  with [D-D-H](https://github.com/D-D-H), he thought it might be because vm leaf call would reserve [2 words](https://github.com/openjdk/jdk/pull/6597) on the stack.

```
# {method} {0x0000ffffaf1f0348} 'main' '([Ljava/lang/String;)V' in 'Test'
...
  0x0000ffffb051727c: mov	x8, #0xdec0                	// #57024
                                                ;   {runtime_call}
  0x0000ffffb0517280: movk	x8, #0xbdb6, lsl #16
  0x0000ffffb0517284: movk	x8, #0xffff, lsl #32
  0x0000ffffb0517288: stp	xzr, x9, [sp, #-16]! // reserve 2 words   <----
  0x0000ffffb051728c: blr	x8
  0x0000ffffb0517290: add	sp, sp, #0x10   ;*invokestatic currentTimeMillis
                                                ; - Test::test@0 (line 18)
                                                ; - Test::wrapper@0 (line 14)
                                                ; - Test::main@4 (line 5)
```

I slightly modified stackWalker.cpp. 
[...]

PR #1641, opened 2026-01-05
Unwind vDSO correctly on Linux-ARM64

apangin:
### Description

This PR fixes two issues related to [vDSO](https://man7.org/linux/man-pages/man7/vdso.7.html):

1. vDSO symbols (e.g. `__kernel_gettimeofday`) are not resolved if vDSO object has no `DT_RELAENT` program header. In this case, `parseDynamicSection` returns early despite that `loadSymbolTable` does not need the value of `DT_RELAENT`.
2. If `vDSO` does not have DWARF unwinding info (`PT_GNU_EH_FRAME` program header), assume an empty frame layout instead of the default layout. There was a partial fix for this problem in #819, but that fix was incomplete: it did not handle the case when `PT_GNU_EH_FRAME` was absent.

### Related issues

#819, #1636

### Motivation and context

When testing an example program from #1636 on linux-arm64, I found that stack walking of `System.currentTimeMillis` was broken because of the two issues mentioned in the description.

**Before the fix:**

before

**After the fix:**

after

### How has this been tested?

Ubuntu 20.04 AArch64
Amazon Corretto 11
```
java -agentpath:build/lib/libasyncProfiler.so=start,event=cpu,file=out.html Cpu1
```

\+ Added new test.

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -416,7 +416,7 @@ void ElfParser::parseDynamicSection() {
             }
         }
 
-        if (symtab == NULL || strtab == NULL || syment == 0 || relent == 0) {
+        if (symtab == NULL || strtab == NULL || syment == 0) {
             return;
         }
 
@@ -425,7 +425,7 @@ void ElfParser::parseDynamicSection() {
         }
 
         const char* base = this->base();
-        if (jmprel != NULL && pltrelsz != 0) {
+        if (jmprel != NULL && pltrelsz != 0 && relent != 0) {
             // Parse .rela.plt table
             for (size_t offs = 0; offs < pltrelsz; offs += relent) {
                 ElfRelocation* r = (ElfRelocation*)(jmprel + offs);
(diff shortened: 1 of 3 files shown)


src/symbols_linux.cpp
```diff
@@ -425,7 +425,7 @@ void ElfParser::parseDynamicSection() {
         }
 
         const char* base = this->base();
```
Baraa-Hasheesh, line 427: nit: wouldn't it be better to return immediately after loading the symbol table rather than checking the value twice?

```
if (relent == 0) return;
```

apangin: I thought about this option, but decided that this condition alone would be less clear than having it directly in the context where `relent` is used.

test/test/recovery/RecoveryTests.java
```diff
@@ -73,4 +73,13 @@ public void intrinsics(TestProcess p) throws Exception {
         Assert.isLess(out.ratio("^\\[unknown"), 0.01, "No more than 1% of unknown frames");
         Assert.isLess(out.ratio("^[^ ;]+(;[^ ;]+)? "), 0.01, "No more than 1% of short stacks");
     }
+
+    // Verify that System.currentTimeMillis() intrinsic is unwound correctly
+    @Test(mainClass = TimeLoop.class, jvm = Jvm.HOTSPOT, jvmVer = {11, Integer.MAX_VALUE}, debugNonSafepoints = true)
```
fandreuz, line 78: Why not JDK 8?

apangin: There are `[unknown]` frames on Liberica 8-arm64 for a different reason.
Since this is an exotic configuration, I don't really want to waste time fixing this right now.

PR #1645, opened 2026-01-08
[POC] Support multiple engine types in OTLP JFR converter

fandreuz:
### Description
In this POC PR I propose a possible implementation to support multiple engine types in our OTLP JFR converter.

### Related issues
n/a

### Motivation and context
The current implementation targets only the event types selected by the argument combination, the other events are discarded. The same result could be achieved by multiple calls to the converter with different arguments; however, this may be cumbersome for our users, wasteful in terms of resources (JFR is scanned multiple times), and increases disk usage (stacktraces are not deduplicated across different engine types).

### How has this been tested?
POC, will add better testing after discussion with the team.

---

### DIFF
--- a/test/test/otlp/OtlpTests.java
+++ b/test/test/otlp/OtlpTests.java
@@ -104,6 +104,7 @@ public class OtlpTests {
     }
 
     private static void checkSamples(Profile profile, ProfilesDictionary dictionary) {
+        assert profile != null;
         Output collapsed = toCollapsed(profile, dictionary);
         assert collapsed.containsExact("test/otlp/CpuBurner.lambda$main$0;test/otlp/CpuBurner.burn") : collapsed;
     }


src/converter/one/convert/JfrConverter.java
```diff
@@ -310,10 +312,57 @@ public String getTotalUnits() {
         return "nanoseconds";
     }
 
+    public String getValueType(Class<? extends Event> eventClass) {
+        if (eventClass == AllocationSample.class) {
+            return "live-objects";
```
Baraa-Hasheesh, line 320: This isn't correct, `LiveObject` isn't the same as `AllocationSample`

src/converter/one/convert/JfrConverter.java
```diff
@@ -310,10 +312,57 @@ public String getTotalUnits() {
...
+            return "methods";
+        }
+        if (eventClass == ExecutionSample.class) {
+            return "cpu";
+        }
+        // TODO: wall
+        return null;
+    }
+
+    public String getSampleUnits(Class<? extends Event> eventClass) {
```
Baraa-Hasheesh, line 341: can you add a small comment on what's the plan here, as I don't think this function is needed

src/converter/one/convert/JfrConverter.java
```diff
@@ -310,10 +312,57 @@ public String getTotalUnits() {
...
+        if (eventClass == MallocEvent.class) {
+            return "malloc";
+        }
+        if (eventClass == MethodTrace.class) {
+            return "methods";
+        }
+        if (eventClass == ExecutionSample.class) {
+            return "cpu";
+        }
+        // TODO: wall
```
Baraa-Hasheesh, line 337: wall is technically `ExecutionSample.class`, you will need to actually check into the event instance to decide if it's a CPU, CPUTimeSample or Wall sample

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -139,6 +148,13 @@ private void writeProfileDictionary() {
         proto.commitField(profilesDictionaryMark);
     }
 
+    private boolean isSupported(Event event) {
+        if (getValueType(event.getClass()) == null) return false;
+        if (!(event instanceof MallocEvent)) return true;
+        MallocEvent mallocEvent = (MallocEvent) event;
+        return mallocEvent.size > 0; // discard 'free' event
```
Baraa-Hasheesh, line 155: is it not possible to keep the "Free" events? -> these events are useful when analyzing memory leaks

src/converter/one/convert/JfrConverter.java
```diff
@@ -310,10 +312,57 @@ public String getTotalUnits() {
...
+            eventClass == NativeLockEvent.class ||
+            eventClass == MethodTrace.class ||
+            eventClass == ExecutionSample.class) {
+            return "nanoseconds";
+        }
+        return null;
+    }
+
     public double counterFactor() {
         return (args.lock || args.nativelock) ? jfr.nanosPerTick : 1.0;
```
Baraa-Hasheesh, line 361: Old bug but if you do `--nativemem --lock` you would have a `counterFactor = jfr.nanosPerTick` which is wrong

src/converter/one/convert/JfrConverter.java
```diff
@@ -89,6 +89,7 @@ protected void collectEvents(TimeIntervals timeIntervals) throws IOException {
                 : args.alloc ? AllocationSample.class
                 : args.lock ? ContendedLock.class
                 : args.trace ? MethodTrace.class
+                : args.all ? null
```
Baraa-Hasheesh, line 92: I don't personally like this as it's all-in, can't we make our algorithm smart to allow user to select specific events 

For example `--nativemem --alloc` -> I'm currently analyzing memory usage/allocations 

-> we can keep the `all` option as a short hand for all samples

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: Also note, there are multiple different allocation samples, do you want to give each sample a unique type?

InNewTLAB vs OutsideTLAB vs jvmti engine sample

fandreuz: Open to discussion here, ideally there should be a conversion map from JFR event to OTLP value type. I'll ask around after the holidays.

> This isn't correct, LiveObject isn't the same as AllocationSample

I missed a branch: 5b716af2ce47d9be09f2d0ba47970d1a2a08ca6f

src/converter/one/convert/JfrConverter.java
fandreuz: Let's discuss the big things first, it's a POC PR.

src/converter/one/convert/JfrConverter.java
fandreuz: I know. If this is approved, I'll probably add a new event subclass for wall events.

src/converter/one/convert/JfrToOtlp.java
fandreuz: I'd argue `free` events are much less useful after the address is discarded, but open to discussion here.

src/converter/one/convert/JfrConverter.java
fandreuz: New GH issue then?

src/converter/one/convert/JfrConverter.java
fandreuz: > For example --nativemem --alloc -> I'm currently analyzing memory usage/allocations

Is this supported in other converters?

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: > new event subclass for wall events

This will break one important feature of the `jfrconverter` which is that it's able to automatically dump `wall` samples if no `cpu` samples exists in jfr 

This in turn means you will need to adjust the converter & reader classes to be smart enough to know if no `cpu` sample exist then use `wall` samples

src/converter/one/convert/JfrToOtlp.java
Baraa-Hasheesh: > after the address is discarded 

I didn't realize that in OTLP we drop the address when converting, if that's the case it does make sense to drop the free events 

but I can see this logic break if you do `--all --nativemem --leak`
-> you will hit casting exceptions 

You will need to modify the `MallocLeakAggregator` to be smart enough to handle this new flow
==> currently using `--nativemem --leak` will only save leaky allocation to OTLP

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: feel free to create the ticket, I commented on it to highlight it

src/converter/one/convert/JfrToOtlp.java
fandreuz: I ran the code and didn't spot any casting exception.

src/converter/one/convert/JfrConverter.java
Baraa-Hasheesh: Open for discussion, as other converters don't support multiple sample types as there is no way to distinguish between samples

src/converter/one/convert/JfrToOtlp.java
Baraa-Hasheesh: you're right that's because if I do `--all --nativemem --leak` it would only collect nativemem samples

apangin:
As discussed with @fandreuz, the use case for this feature is not very clear; there is no real-world problem that it solves.
I'm therefore closing this PR for now. Can be reopened once we have additional input.

Issue #1516, opened 2025-10-01
Enable TLAB-related allocation events for modern Hotspot versions


franz1981:
### Describe the feature

Recentlish, async profiler is automatically using sampled allocation events instead of TLAB ones, but there's still a value in the TLAB-related ones, as a developer.
I would like a new `--alloc-tlab` event which enable to capture the old allocation type of event in new OpenJDK versions too.

### Use Case

To be able to answer to the question: "I should really spend time optimizing to lower the allocation rate?"
Having slow path TLAB allocation events vs TLAB ones (the ratio) can help to understand if makes sense looking at the more fine grain information of the "bigger allocators" via the existing `--alloc` event.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
For this case I would recommend using a different feature of async-profiler that achieves what's requested being the `jfrsync` feature 

You can profile using the following command `asprof -e cpu -d 10s --jfrsync +jdk.ObjectAllocationInNewTLAB+jdk.ObjectAllocationOutsideTLAB`

This will enable the async-profiler for cpu profiling & also use the JFR events for allocation profiling which do report each event type separately as desired 

If you want to add custom sampling thresholds per event you can define a proper config file for the `jfrsync` feature & provide it as your input.

franz1981:
Thanks @Baraa-Hasheesh I am aware of jfrsync and it is indeed a viable option (sorry if I didn't mentioned in the issue description), but it means using the JFR engine instead of just async-profiler only - including losing the ability of observing the whole call chain when the allocation is triggered by some JNI code calling a Java one.

Baraa-Hasheesh:
> but it means using the JFR engine instead of just async-profiler only

That should be generally fine, as JFR does offer some value in some cases

> including losing the ability of observing the whole call chain when the allocation is triggered by some JNI code calling a Java one

This makes sense, but to confirm, do you mean that you want to see the full stack for cases where a native code calls java code

`NativeFrame1;NativeFrame2;{multipleJniFrames};JavaFrame1;JavaFrame2`

If my understanding is incorrect please feel free to correct me

apangin:
As discussed with @Baraa-Hasheesh, `--tlab` flag would be better than a new event.
So, you can now run `asprof -e alloc --tlab` for the old style allocation profiling on JDK 11+.

[...]

Issue #1424, opened 2025-07-30
Lock contention profiling triggers seg fault in j9 jvm runtime


obeton:
### Describe the bug

For any given java process (I've attached the test process I used), within seconds of attaching the async-profiler to it for contention lock profiling (e.g., `./build/bin/asprof -d 30 -e lock -i 10ms <pid>`), a seg fault will occur within the jvm. There appears to be an assert that is triggered due to a "race condition" like situation.

Console output of process at segfault:
`18:50:23.122 0x1e7f00    j9vm.225    *   ** ASSERTION FAILED ** at /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-openj9/workspace/build/src/openj9/runtime/vm/VMAccess.cpp:133: (!(vmThread->inNative))`

Stack trace of the two overstepping threads:
`#0  0x00007fc607e8bedc in __pthread_kill_implementation () from /lib64/libc.so.6`
`#1  0x00007fc607e3eb46 in raise () from /lib64/libc.so.6`
`#2  0x00007fc607e28833 in abort () from /lib64/libc.so.6`
`#3  0x00007fc607e2875b in __assert_fail_base.cold () from /lib64/libc.so.6`
`#4  0x00007fc607e37886 in __assert_fail () from /lib64/libc.so.6`
`#5  0x00007fc6061a407e in MM_JNICriticalRegion::releaseAccess(J9VMThread*, unsigned long*) () from /usr/lib/jvm/ibm-semeru-open-21-jdk/lib/default/libj9gc29.so`
`#6  0x00007fc6061144d2 in MM_EnvironmentBase::acquireExclusiveVMAccessForGC(MM_Collector*, bool) () from /usr/lib/jvm/ibm-semeru-open-21-jdk/lib/default/libj9gc29.so`
`#7  0x00007fc6061e53ed in MM_MemorySubSpaceSemiSpace::allocationRequestFailed(MM_EnvironmentBase*, MM_AllocateDescription*, MM_MemorySubSpace::AllocationType, MM_ObjectAllocationInterface*, MM_MemorySubSpace*, MM_MemorySubSpace*)
    () from /usr/lib/jvm/ibm-semeru-open-21-jdk/lib/default/libj9gc29.so`
[...]

PR #1647, opened 2026-01-09
Disallow incorrect usage of jfrsync

Baraa-Hasheesh:
### Description
`jfrsync` is only allowed with jfr output format & only on a process that has a JVM instance

Currently it's possible for the user to pass `jfrsync` on a native process which will cause the application to crash 

Following checks will fail early & report back to user with clear warning messages 

### Related issues
N/A

### Motivation and context
prevent bad user input on async-profiler from crashing the application 

### How has this been tested?
manual testing

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1095,6 +1095,10 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("Profiling event is not supported with non-Java processes");
     }
 
+    if (args._jfr_sync && !VM::loaded()) {
+        return Error("jfrsync is not supported with non-Java processes");
+    }
+
     if (args._fdtransfer) {
         if (!FdTransferClient::connectToServer(args._fdtransfer_path)) {
             return Error("Failed to initialize FdTransferClient");
@@ -1198,14 +1202,6 @@ Error Profiler::start(Arguments& args, bool reset) {
 
     if (_cstack != CSTACK_VM && _features.mixed) {
         return Error("mixed feature is only allowed with VMStructs stack walking");
-    }
-
-    if (args._jfr_sync && args._output != OUTPUT_JFR) {
-        return Error("jfrsync is only supported with jfr output");
-    }
-
-    if (args._jfr_sync && !VM::loaded()) {
-        return Error("jfrsync is not supported with non-Java processes");
     }
 
     // Kernel symbols are useful only for perf_events without --all-user


src/profiler.cpp
```diff
@@ -1200,6 +1200,14 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("mixed feature is only allowed with VMStructs stack walking");
     }
 
+    if (args._jfr_sync && args._output != OUTPUT_JFR) {
+        return Error("jfrsync is only supported with jfr output");
+    }
+
+    if (args._jfr_sync && !VM::loaded()) {
```
```
Program received signal SIGSEGV, Segmentation fault.
FlightRecorder::startMasterRecording (this=0x55555558b9f0, filename=0x55555570d993 "profile.jfr", args=...) at src/flightRecorder.cpp:1360
1360            if (env->FindClass("jdk/jfr/FlightRecorderListener") == NULL) {
(gdb) p env
$1 = (JNIEnv *) 0x0
```

How about we add a check after `JNIEnv* env = VM::jni();` and return an error instead? It seems better to me to handle the error where it happens.

Baraa-Hasheesh: I do see the argument for that, but checking `JNIEnv* env = VM::jni()` isn't fully correct 

As it's possible for a JVM instance to exist but for the current thread not being attached to it 

-> #1519 

This bug was discovered previously while experimenting on that ticket, 
I was going through my open PRs & decided it makes sense to add this small check separately as it does a basic check that a JVM was loaded/exist at all  (handles a different bug)


For exact checks on `JNIEnv`, I would vote to keep in the relevant ticket

src/profiler.cpp
```diff
@@ -1200,6 +1200,14 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("mixed feature is only allowed with VMStructs stack walking");
     }
 
+    if (args._jfr_sync && args._output != OUTPUT_JFR) {
```
apangin: This should be harmless, right?
`jfrsync` automatically selects JFR output, so this condition may happen only when user explicitly specifies output format after `jfrsync`. It's fine to ignore `jfrsync` in this case. E.g. I can play with output formats without modifying the rest of the command line.

src/profiler.cpp
apangin: It's a cheap check - you can move this up before engine selection, resetting buffers, etc.

src/profiler.cpp
Baraa-Hasheesh: My reading on it is that the jfrsync internally may cause multiple sample types to be collected (multiple engines),
for example you can use the jfrsync to collect lock & alloc samples are collected via the flight recorder while async-profiler is used to collect CPU samples 

-> This can be used as workarounds for #1424 (until fixed) & #1516 (until implemented) 

So it makes sense to hard check this to make it consistent with using multiple sampling engines which is only allowed on jfr

But I'm fine with not checking it, what are your thoughts?

src/profiler.cpp
Baraa-Hasheesh: that makes sense

src/profiler.cpp
apangin: async-profiler has 70 different arguments and works with a variety of JVMs. There are countless combinations that are not fully supported; it's impossible and pointless to check all of them.

My philosophy in async-profiler development is to focus on the intended use of the tool and keep error checking to a reasonable minimum. The reasonable minimum includes common user mistakes and unexpected cases that may cause bad consequences like crashes. The reason is that making a tool fool-proof requires much efforts and expands the codebase for a little-to-none value. After all, async-profiler is a tool for power users.

I'm not particularly against this specific check, I just want to share my vision to save you from adding excessive checks in the future as well as to save myself from reviewing that.

src/profiler.cpp
Baraa-Hasheesh: [6278129](https://github.com/async-profiler/async-profiler/pull/1647/commits/6278129d133d86d1c1a5abeeff8bb52e5ee5f6c7)

src/profiler.cpp
Baraa-Hasheesh: Thanks for sharing this vision, agreed, check removed [6278129](https://github.com/async-profiler/async-profiler/pull/1647/commits/6278129d133d86d1c1a5abeeff8bb52e5ee5f6c7)

PR #1648, opened 2026-01-10
Fix stop sequence in `Profiler::start`

fandreuz:
### Description
I spotted a small mistake in the stop sequence in `Profiler::start`. As a result, when an error occurs during any of `MallocTracer::start`, `NativeLockTracer::start`, and `Instrument::start`, `WallClock::stop` won't be called.

### Related issues
n/a

### Motivation and context
`error5` is stopping `malloc_tracer`, but it should stop `wall_clock`.

### How has this been tested?
n/a

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1257,13 +1257,13 @@ Error Profiler::start(Arguments& args, bool reset) {
     return Error::OK;
 
 error7:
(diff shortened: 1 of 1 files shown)


apangin:
Good catch. Thanks.

PR #1651, opened 2026-01-12
Do not record cpu frame on non perf samples

Baraa-Hasheesh:
### Description
While analyzing some async-profiler JFRs on some custom tests, I noticed that it's possible for the profiler to incorrectly record an invalid CPU frame when no such capabilities exist for the sample type

This happens when user runs `event=cpu-clock,record-cpu,wall,file=profile.jfr`

When a wall sample is collected, the profiler will incorrectly assume it has a valid CPU number but that is wrong, as such a check was added to only add the CPU frame for perf samples

This is relevant for all non perf profiling engines when run in parallel with perf

### Related issues
N/A

### Motivation and context
Do not record invalid CPU information in the collected stacks of non perf samples

### How has this been tested?
A new test was added 

---

### DIFF
--- a/test/test/jfr/JfrTests.java
+++ b/test/test/jfr/JfrTests.java
@@ -20,7 +20,7 @@ import java.util.*;
 
 public class JfrTests {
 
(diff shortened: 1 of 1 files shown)


apangin:
Makes sense. Thanks!

PR #1652, opened 2026-01-12
Prevent fdtransfer connection from not closing in case of start failure

Baraa-Hasheesh:
### Description
Noticed while fixing #1651, in many locations after initial connection is created it's possible for `start` to return error without closing the connection

Moving the if condition to just before starting the CPU engine should work well enough

### Related issues
N/A

### Motivation and context
Prevent fdtransfer connection from not closing in case of start error

### How has this been tested?
`make test` & manual testing

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1198,7 +1198,7 @@ Error Profiler::start(Arguments& args, bool reset) {
     if (args._fdtransfer) {
         if (!FdTransferClient::connectToServer(args._fdtransfer_path)) {
             error = Error("Failed to initialize FdTransferClient");
-            if (error) goto error1;
+            goto error1;
         }
     }
 
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1059,6 +1059,21 @@ Error Profiler::start(Arguments& args, bool reset) {
         VM::tryAttach();
     }
 
+    if (args._fdtransfer) {
+        if (!FdTransferClient::connectToServer(args._fdtransfer_path)) {
+            return Error("Failed to initialize FdTransferClient");
+        }
+    }
+
+    Error error = startInternal(args, reset);
+    if (error) {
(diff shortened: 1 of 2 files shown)


Baraa-Hasheesh:
@apangin how would you feel about adding a check to not start the fdtransfer connection in case it's not perf engine 

-> any other engine we can skip it even if specified by user

apangin:
> how would you feel about adding a check to not start the fdtransfer connection in case it's not perf engine

This refers back to my [previous comment](https://github.com/async-profiler/async-profiler/pull/1647#discussion_r2683507331) about making async-profiler fool-proof.
If user specifies `-e alloc --fdtransfer` for some reason, I don't want async-profiler to guess what was actually meant by that, was it a mistake or not. Instead, let it just follow user's instructions literally.

`--fdtransfer` option does two things: it starts a separate server process and connects to it from the JVM process. Server will wait until the JVM connects. If you want to skip connection, you need to do it on both sides. This becomes complicated since `asprof` launcher does not know what engine will be used. So what's the point of adding these complications if user won't even notice the difference?

src/profiler.cpp
```diff
@@ -1201,6 +1195,13 @@ Error Profiler::start(Arguments& args, bool reset) {
         }
     }
 
+    if (args._fdtransfer) {
+        if (!FdTransferClient::connectToServer(args._fdtransfer_path)) {
+            error = Error("Failed to initialize FdTransferClient");
+            goto error1;
+        }
+    }
```
apangin: `updateSymbols` call above relies on fdtransfer connection for reading kernel symbols.

Also, it is better to establish connection earlier since the server can timeout if it takes too long for a client to start.

Baraa-Hasheesh:
Additional area where this happens noticed while reviewing #1791 

If the `out.is_open` fails during the `expire` call during a `loop` session then the fdtransfer won't be closed

PR #1654, opened 2026-01-13
Prefer perf-events engine when record-cpu or target-cpu are selected

Baraa-Hasheesh:
### Description
While running recordCpuMultiEngine test in virtualized environments, `event=cpu` resulted in using ctimer which in turn caused the test to fail due to selecting `record-cpu` with a non perf event engine

However other tests that already use `cpu-clock` passed just fine, as perf events will automatically skip `_kernel_stack` if not supported 

If a user selects `event=cpu` with `record-cpu/target-cpu`, the user intentions are clear, perf events are desired as such additional checks where added to prefer perf events if such cases are observed on the user inputs 

Note: in worst case we will be replacing one error with another 
* Old error will be that the `record-cpu/target-cpu` are only supported on perf events
* New error if perf events are not supported will be that perf events aren't supported/creating perf events failed

### Related issues
N/A

### Motivation and context
Better following of user intentions when using `event=cpu`

### How has this been tested?
make test

---

### DIFF
--- a/test/test/jfr/JfrTests.java
+++ b/test/test/jfr/JfrTests.java
@@ -42,7 +42,7 @@ public class JfrTests {
         assert out.contains(normalLoadPattern);
     }
 
-    @Test(mainClass = CpuLoad.class, agentArgs = "start,event=cpu-clock,wall,record-cpu,file=%profile.jfr", os = Os.LINUX)
+    @Test(mainClass = CpuLoad.class, agentArgs = "start,event=cpu,wall,record-cpu,file=%profile.jfr", os = Os.LINUX)
     public void recordCpuMultiEngine(TestProcess p) throws Exception {
         p.waitForExit();
         assert p.exitCode() == 0;


test/test/jfr/JfrTests.java
```diff
@@ -42,7 +42,7 @@ public void cpuLoad(TestProcess p) throws Exception {
         assert out.contains(normalLoadPattern);
     }
 
-    @Test(mainClass = CpuLoad.class, agentArgs = "start,event=cpu,wall,record-cpu,file=%profile.jfr", os = Os.LINUX)
+    @Test(mainClass = CpuLoad.class, agentArgs = "start,event=cpu-clock,wall,record-cpu,file=%profile.jfr", os = Os.LINUX)
```
apangin: Does it make sense to choose `perf_events` in `selectEngine` if `event=cpu` and either of `record-cpu` or `target-cpu` options is specified?

Baraa-Hasheesh: I mean it's either perf engine is not selected & you get an error or perf event isn't supported & you also get an error 

So it does make sense to go with option 2 as it's less likely to error out as if kernel symbols aren't supported it will run in user mode, should I update this PR to add that change?

Baraa-Hasheesh: I added the conditions for that [5878526](https://github.com/async-profiler/async-profiler/pull/1654/commits/58785263750dcb9015f98a3f5c0e2065f944a316) 

As a side effect of the change if the user selects these options with `cpu` on macOS it will fail due to unsupported perf events, which should be fine as it would have previously failed due not selecting perf events

apangin: Good, thanks. Shall we revert the test back to `event=cpu` to have this code path executed?
Please also update PR title/description to reflect new fix.

Baraa-Hasheesh: Done

Issue #377, opened 2020-12-04
JVM Crash with 1.8.2 + OpenJDK 8u275


SylvainJuge:
This issue has been initially reported in [Elastic forum](https://discuss.elastic.co/t/jvm-corrupted-with-java-agent-1-19-0-centos/257599).

Context:
- elastic-apm-agent version `1.19.0` that ships with async-profiler `1.8.2` (md5 of `.so` is `7b0005bb0afc7e225ffe8372878fd01a`)
- JVM : OpenJDK 8 upate 275

We don't have the regular `hs_err_pid*` crash report, but the following error has been reported:

```
*** Error in `/opt/java/1.8.0_275/bin/java': double free or corruption (!prev): 0x00007f0dc47550b0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81299)[0x7f0ecfbce299]
/opt/java/1.8.0_275/jre/lib/amd64/server/libjvm.so(+0x64edab)[0x7f0ecf151dab]
/opt/java/1.8.0_275/jre/lib/amd64/server/libjvm.so(+0x7aa513)[0x7f0ecf2ad513]
/opt/java/1.8.0_275/jre/lib/amd64/server/libjvm.so(+0x79ebc4)[0x7f0ecf2a1bc4]
/home/temp/tomcat_temp/libasyncProfiler-linux-x64-7b0005bb0afc7e225ffe8372878fd01a.so(_ZN8Profiler17getJavaTraceJvmtiEP15_jvmtiFrameInfoP15ASGCT_CallFramei+0x64)[0x7f0dde2b58f4]
/home/temp/tomcat_temp/libasyncProfiler-linux-x64-7b0005bb0afc7e225ffe8372878fd01a.so(_ZN8Profiler17getJavaTraceAsyncEPvP15ASGCT_CallFramei+0x15b)[0x7f0dde2b5c9b]
/home/temp/tomcat_temp/libasyncProfiler-linux-x64-7b0005bb0afc7e225ffe8372878fd01a.so(_ZN8Profiler12recordSampleEPvyiP10_jmethodID11ThreadState+0x20c)[0x7f0dde2b632c]
/lib64/libpthread.so.0(+0xf630)[0x7f0ed0346630]
/lib64/libpthread.so.0(pthread_cond_timedwait+0x132)[0x7f0ed0342de2]
[...]

PR #1656, opened 2026-01-14
#1528: Add a hard-coded limit on the maximum number of jmethodIDs

yanglong1010:
Hi,
Could I have a review of this patch ?

### Description
Add a hard-coded limit on the maximum number of jmethodIDs

### Related issues
https://github.com/async-profiler/async-profiler/discussions/1528

### Motivation and context
Every jmethodid occupies at least 8 bytes in c heap, [see JNIMethodBlockNode](https://github.com/openjdk/jdk21u-dev/blob/6201b183cae8be105659568251f1aa53ffc196e4/src/hotspot/share/oops/method.cpp#L2033), and this type of memory is never freed. Some Java apps create classes on-the-fly and continuously. Though the classes are unloaded, the RSS keep increasing. In the worst case, this can lead to Java process be killed by system OOM killer.

As the comments in the discussion #1528 , before jmethodid is no longer used, add a limit is acceptable.

The current implementation has a limitation:
- Class retransformation/redefinition can add/delete methods to already loaded classes. From JDK 13 and later, need to add -XX:+AllowRedefinitionToAddDeleteMethods. In VM::RedefineClassesHook/VM::RetransformClassesHook, because the operation is in progress, it is unsafe to retrieve the method count of the old (I tried, it leads to crash). So the current implementation just assumes no method is added/deleted.

### How has this been tested?
- JNI: one.profiler.AsyncProfiler.getInstance(libPath), limit works
- Agent: -agentpath:..., limit works
- Attach: asprof ..., limit works

If this patch is acceptable, I will add unit tests.

---

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -368,12 +368,12 @@ void VM::applyPatch(char* func, const char* patch, const char* end_patch) {
     }
 }
 
-int VM::loadMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni, jclass klass, bool update_count) {
+void VM::loadMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni, jclass klass, bool update_count) {
     if (loadAcquire(_jmethod_id_count) > JMETHOD_ID_LIMIT) {
         if (__sync_bool_compare_and_swap(&_jmethod_id_count_warned, false, true)) {
             Log::warn("Total number of generated jmethod-ids exceeds %d, stop generating more", JMETHOD_ID_LIMIT);
         }
-        return -1;
+        return;
     }
     if (VMStructs::hasClassLoaderData()) {
         VMKlass* vmklass = VMKlass::fromJavaClass(jni, klass);
@@ -397,9 +397,7 @@ int VM::loadMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni, jclass klass, bool update_co
             atomicInc(_jmethod_id_count, method_count);
         }
         jvmti->Deallocate((unsigned char*)methods);
-        return method_count;
     }
-    return -1;
 }
 
 void VM::loadAllMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni) {
--- a/src/vmEntry.h
+++ b/src/vmEntry.h
@@ -113,7 +113,7 @@ class VM {
 
     static void ready();
     static void applyPatch(char* func, const char* patch, const char* end_patch);
-    static int loadMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni, jclass klass, bool update_count = true);
+    static void loadMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni, jclass klass, bool update_count = true);
     static void loadAllMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni);
     static bool hasJvmThreads();
 


src/vmEntry.cpp
```diff
@@ -392,14 +413,15 @@ void VM::loadAllMethodIDs(jvmtiEnv* jvmti, JNIEnv* jni) {
...
-            loadMethodIDs(jvmti, jni, classes[i]);
+            loadMethodIDs(jvmti, jni, classes[i], true);
         }
         jvmti->Deallocate((unsigned char*)classes);
     }
 }
 
 void JNICALL VM::VMInit(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
     ready();
+    jmethod_id_limit = _global_args._jmethod_id_limit;
```
Baraa-Hasheesh: This is not correct, by this point the `_global_args` are not always set

For example, the following flow,
* attach is called & the init is called from [here](https://github.com/async-profiler/async-profiler/blob/master/src/vmEntry.cpp#L494)
* Profiler is started after the VM details are initialized [here](https://github.com/async-profiler/async-profiler/blob/master/src/vmEntry.cpp#L499)
* global args are updated [here](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L1096) 

You will actually need to update global args before calling the VM::init in the  for this to work which can be risky as a different profiling session could already be running so you will need to check for those details

src/main/main.cpp
```diff
@@ -95,6 +95,8 @@ static const char USAGE_STRING[] =
     "  --fdtransfer        run separate fdtransfer process to serve perf requests\n"
     "                      from the non-privileged target\n"
     "  --target-cpu cpu    sample threads on a specific CPU (perf_events only, default: -1)\n"
+    "  --jmid-limit limit  the number of methods to generate jmethod id (default 0, i.e. no limit).\n"
```
Baraa-Hasheesh: nit, I would put -1 as no limit & 0 as don't save any id

src/vmEntry.cpp
yanglong1010: Thanks for you review.

The flow you described is attach mode. In this mode, VM::VMInit is not invoked, refer to below.

In my code, I used different execute paths to set `jmethod_id_limit`:
- Attach mode, `jmethod_id_limit` is set to the value of `args._jmethod_id_limit` which is parsed from attach options [code](https://github.com/yanglong1010/async-profiler/blob/512e39cfa67c9961d118342e1b8ece03cedb9feb/src/vmEntry.cpp#L517)
- Agent mode, `jmethod_id_limit` is set to the value of `_global_args._jmethod_id_limit` which is parsed from agent options [code](https://github.com/yanglong1010/async-profiler/blob/512e39cfa67c9961d118342e1b8ece03cedb9feb/src/vmEntry.cpp#L424)
- Java code mode, `jmethod_id_limit` is set to the value of `args._jmethod_id_limit` which is parsed from the argument of AsyncProfiler.execute [code](https://github.com/yanglong1010/async-profiler/blob/512e39cfa67c9961d118342e1b8ece03cedb9feb/src/profiler.cpp#L1057)

Are there any other execution paths beyond the modes I described above ?

apangin:
Thank you for working on async-profiler.
As I wrote in the [comment](https://github.com/async-profiler/async-profiler/discussions/1528#discussioncomment-14668215), I'm looking for a fix that eliminates jmethodIDs completely. This should be possible with VMStructs, although laborious. In the meantime, I'm fine with adding a hard-coded limit. What I don't really like is a new public facing argument that is to shift responsibility of dealing with a bug to end-users.

src/vmEntry.cpp
yanglong1010: Because the limit is changed to hard-coded, so this line of code is deleted. The limit will work correctly in any mode.

src/main/main.cpp
yanglong1010: Changed to hard-coded limit, this line of code is deleted.

yanglong1010:
I have changed the limit to hard-coded, please review again ~

src/vmEntry.cpp
```diff
@@ -25,6 +25,10 @@
 const int ARGUMENTS_ERROR = 100;
 const int COMMAND_ERROR = 200;
 
+static u32 _jmethod_id_limit = 1024 * 1024 * 500 / 8; // 500 MiB memory, about 65 million methods
```
apangin: `const` or `constexpr`

src/vmEntry.cpp
```diff
@@ -449,7 +462,7 @@ jvmtiError VM::RetransformClassesHook(jvmtiEnv* jvmti, jint class_count, const j
         JNIEnv* env = jni();
         for (int i = 0; i < class_count; i++) {
             if (classes[i] != NULL) {
-                loadMethodIDs(jvmti, env, classes[i]);
+                loadMethodIDs(jvmti, env, classes[i], false);
```
apangin, line 465: Doesn't class redefinition create new jmethodIDs? It should.. at least for redefined methods.

src/vmEntry.cpp
yanglong1010: Pushed

src/vmEntry.cpp
yanglong1010: For existing methods, redefinition/retransformation replaces the contents of jmethods with the new Methods, no new jmethod id is created.

see
https://github.com/openjdk/jdk21u-dev/blob/master/src/hotspot/share/prims/jvmtiRedefineClasses.cpp#L3820C1-L3830C6

As a supporting evidence, I commented out the code in `RetransformClassesHook/RedefineClassesHook` that generates jmethod ids. For existing methods, after class redefinition/retransformation, the profiling result is correct.

For added methods, we need to generate new jmethod ids for them. I added code to get the method counts before and after transformation/redefinition, and add the diff to `_jmethod_id_count`.

apangin: The reason why I added RedefineClasses hook was definitely not about newly added methods.
I remembered - they were not jmethodIDs that were re-created after class redifinition, but jmethodID cache in `InstanceKlass`. That caused JVM crash #377. Since there is no more code that previously crashed, the hack with intercepting RedefineClasses is no longer needed. I'd prefer to get rid of it completely. The downside of removing it is that methods added by RedefineClasses won't be visible in a profile, but this is a very rare case and it violates the spec anyway. WDYT?

yanglong1010: Regarding #377, based on my understanding, if no method is added, the jmethod id cache in InstanceKlass is always up to date, the stack trace of #377 should not be encountered. Unless methods are added, then, multiple threads extend the jmethod id cache and release the original cache more than 1 time. I can't reproduce the crash, so I don't know whether methods were actually added.

- If methods were added in #377 , then it should be ok to remove the hooks.
- If no method was added, it means your assumption (redefinition invalidates jmethod_id cache) is true, remove the hooks may lead to invalid stack trace for existing methods.

If we cannot figure out which situation it is, keep the hooks seems more safe.

yanglong1010: In the jdk code, it says `Leave arrays of jmethodIDs and itable index cache unchanged`, but you said `redefinition invalidates jmethod_id cache`, this makes me confused.

https://github.com/openjdk/jdk21u-dev/blob/2ce809a93eccb9a425bee8aeaaaa195a93597209/src/hotspot/share/prims/jvmtiRedefineClasses.cpp#L4387

apangin: OK, you may be right - I don't know exactly how classes were redefined in #377.
Seems that implications of RedefineClasses need more research, not in the scope of this PR though.
For now, I think, it is fine to leave RedefineClasses hook as it is without incrementing counter.

yanglong1010: Pushed.

PR #1659, opened 2026-01-16
Provide non-aggregated samples in OTLP converter

fandreuz:
### Description
The current implementation of the `JfrToOtlp` converter aggregates events belonging to the same group while going through the JFR (i.e. same stacktrace ID, same thread ID, same event type). This happens because the converter is implicitly using the `EventAggregator`.

### Related issues
n/a

### Motivation and context
The current behavior is not technically incorrect, except for the fact that the resulting `Sample` gets a random timestamp. There are two solutions:
1. Keep the aggregation, but drop the timestamp ([timestamp](https://github.com/open-telemetry/opentelemetry-proto/blob/d6dc40fe54b8d441fd7a920af29d96c3ba6ed36a/opentelemetry/proto/profiles/v1development/profiles.proto#L413) is not obligatory according to the protocol)
2. Drop the aggregation, and let each event be mapped to its own `Sample` (this PR).

I'd prefer (1) because (2) is a lossy conversion. However, (1) would produce a closer output to `Profiler::dumpOtlp`, and would have a much simpler implementation (we'd just need to drop [two lines](https://github.com/async-profiler/async-profiler/blob/872be632202bb1963c66fbb153ced191363baa14/src/converter/one/convert/JfrToOtlp.java#L188-L189) from `JfrToOtlp`). Happy to hear more thoughts on this.

### How has this been tested?
New test.

---

### DIFF
--- a/test/test/otlp/OtlpTests.java
+++ b/test/test/otlp/OtlpTests.java
@@ -71,7 +71,8 @@ public class OtlpTests {
     @Test(mainClass = CpuBurner.class, agentArgs = "start,otlp,file=%f.pb")
     public void samples(TestProcess p) throws Exception {
         ProfilesData profilesData = waitAndGetProfilesData(p);
-        checkSamples(getProfile(profilesData, 0), profilesData.getDictionary());
+        Output collapsed = toCollapsed(getProfile(profilesData, 0), profilesData.getDictionary());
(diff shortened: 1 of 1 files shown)


fandreuz:
Found a small mistake, will reopen later.

PR #1660, opened 2026-01-16
Provide non-aggregated samples in OTLP converter

fandreuz:
### Description
The current implementation of the `JfrToOtlp` converter aggregates events belonging to the same group while going through the JFR (i.e. same stacktrace ID, same thread ID, same event type). This happens because the converter is implicitly using `EventAggregator`. Consequently, resulting `Sample` are not 1-1 mappings from JFR events.

The problem addressed by this PR is that the time information contained in each event could be preserved (`Sample::timestamp_unix_nano` is [repeated](https://github.com/open-telemetry/opentelemetry-proto/blob/d6dc40fe54b8d441fd7a920af29d96c3ba6ed36a/opentelemetry/proto/profiles/v1development/profiles.proto#L413)), but is discarded in the current implementation. Instead of providing aggregated samples, e.g.
```
Sample {
   ...
   timestamps_unix_nano: T
   values: X
}
```
where `X = a + b + c` and `T = random_choice(t_a, t_b, t_c)`, we'll record:
```
Sample {
   ...
   timestamps_unix_nano: [t_a, t_b, t_c]
   values: [a, b, c]
}
```

### Related issues
n/a

### Motivation and context
The current behavior is incorrect because each `Sample` gets a random timestamp.

Another solution we could consider: keep the aggregation, and remove the timestamp from all `Sample` messages.

### How has this been tested?
New test.

---

### DIFF
--- a/src/converter/one/convert/JfrToOtlp.java
+++ b/src/converter/one/convert/JfrToOtlp.java
@@ -142,7 +142,7 @@ public class JfrToOtlp extends JfrConverter {
     }
 
     private void writeSample(int stackTraceId, int tid, AggregatedEvent ae) {
-        // 34 is the sum of:
+        // 24 is the sum of:
         // 4 tags: 1 byte
         // 5 * 2: max size of thread name and stack idx
         // 5 * 2: max size of timestamps/values arrays


src/converter/one/jfr/event/EventAggregatorWithTime.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright The async-profiler authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package one.jfr.event;
+
+import java.util.Arrays;
+
+public class EventAggregatorWithTime implements EventCollector {
```
apangin: Looks overly complex + there's much copy-pasted code.
Can it be a method-local EventCollector that creates a `HashMap<SampleKey, TLongList>`?

fandreuz: Sure, check e1ea43cd2ecc470c7b88b778ee043e2c356dd339

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -25,6 +25,7 @@ public class JfrToOtlp extends JfrConverter {
     // Size in bytes to be allocated in the buffer to hold the varint containing the length of the message
     private static final int MSG_LARGE = 5;
     private static final int MSG_SMALL = 1;
+    private static final int VALUES_PER_EVENT = 2; /* time, samples/value */
```
apangin: No longer needed

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -70,20 +117,57 @@ protected void convertChunk() {
         proto.fieldFixed64(PROFILE_time_unix_nano, jfr.chunkStartNanos);
         proto.field(PROFILE_duration_nanos, jfr.chunkDurationNanos());
 
-        writeSamples(samplesInfo, !args.total /* samples */);
+        aggregatedEvents.forEach(new Dictionary.Visitor<AggregatedEvent>() {
```
apangin: Why not lambda?

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -285,4 +310,21 @@ public int hashCode() {
             return hash;
         }
     }
+
+    private static final class AggregatedEvent {
+        private long[] timestamps = new long[1];
+        private long[] values = new long[1];
+        private int eventsCount = 0;
```
apangin: nit: I tend to make such fields package-private: first, to indicate that they are accessed from the outer class; second, to get rid of a synthetic accessor methods generated by javac prior to JDK 11.

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -70,20 +117,57 @@ protected void convertChunk() {
...
+
+    private Line makeLine(StackTrace stackTrace, int i) {
+        String methodName = getMethodName(stackTrace.methods[i], stackTrace.types[i]);
+        int lineNumber = stackTrace.locations[i] >>> 16;
+        int functionIdx = functionPool.index(methodName);
+        return new Line(functionIdx, lineNumber);
+    }
+
+    private void writeSample(int stackTraceId, int tid, AggregatedEvent ae) {
+        long sMark = proto.startField(PROFILE_samples, MSG_LARGE);
```
apangin: Does it make sense to estimate the size of a sample message? There is usually a long tail of unique stack traces with only one sample.

fandreuz: Yeah that would make sense. I'll add a new commit later.

src/converter/one/convert/JfrToOtlp.java
fandreuz: TIL

14a1af5c218dd8542c3d9d71681d6c26b553eab2

src/converter/one/convert/JfrToOtlp.java
fandreuz: 14a1af5c218dd8542c3d9d71681d6c26b553eab2

src/converter/one/convert/JfrToOtlp.java
fandreuz: 14a1af5c218dd8542c3d9d71681d6c26b553eab2

src/converter/one/convert/JfrToOtlp.java
```diff
@@ -70,20 +116,62 @@ protected void convertChunk() {
...
+        int lineNumber = stackTrace.locations[i] >>> 16;
+        int functionIdx = functionPool.index(methodName);
+        return new Line(functionIdx, lineNumber);
+    }
+
+    private void writeSample(int stackTraceId, int tid, AggregatedEvent ae) {
+        // 44 is the sum of:
+        // 4 tags: 1 byte
+        // 10 * 2: max size of thread name and stack idx
+        // 10 * 2: max size of timestamps/values arrays
```
apangin: Why 10? All indices and array lengths are `int`, that is 5 bytes maximum.

fandreuz: Right, thanks: 87e4c42b4180a1ef3cacf2863c381d4a704a25c2

apangin: Array length too (timestamps/values).

fandreuz: 6c3b3145754f55d09abbe040d999d73708f823a4

Issue #1044, opened 2024-11-01
User-space part of a trace is not reported if Kernel symbols are unavailable


AntonIvanov87:
### Describe the bug

Say there is a trace that leads into the kernel:
```
  ...
  [13] do_syscall_64_[k]
  [14] entry_SYSCALL_64_after_hwframe_[k]
  [15] send
  [16] io.netty.channel.unix.Socket.sendAddress
```
and consumes much CPU cycles.

If Kernel symbols are not available:
```
[WARN] Kernel symbols are unavailable due to restrictions. Try
  sysctl kernel.perf_event_paranoid=1
  sysctl kernel.kptr_restrict=0
```
such traces are not reported at all.

This creates an illusion that CPU is spend elsewhere.

### Expected vs. actual behavior

Actual behaviour: such traces are ignored.

Expected behaviour: at least user-space part of such traces should be reported, e.g.:
```
  [0] send
  [1] io.netty.channel.unix.Socket.sendAddress
```
Or even:
```
  [0] Kernel symbols are unavailable
  [1] send
  [2] io.netty.channel.unix.Socket.sendAddress
```

### Async-profiler version

3.0

### Environment details

OS: CentOS 9, Linux 5.15.160
JDK: 11
CPU: x86_64

apangin:
Previously, when `/proc/sys/kernel/perf_event_paranoid` settings allowed only user-space measurements (default since Linux 4.6), async-profiler automatically enabled `alluser` option, corresponding to `perf_event_attr.exclude_kernel` bit. It was a big oversight, because `exclude_kernel` option silently discarded all execution samples occurred in the kernel-space, leading to the issue you've described.

If kernel-space measurements are not allowed by system configuration, it's better to switch to `ctimer` mode based on `timer_create` API. I've modified the logic of CPU profiling as follows:
 - perf_events availability is now automatically checked by trying to create a dummy perf_event.
 - If kernel-space profiling using perf_events is not available (including when restricted by `perf_event_paranoid` setting or by seccomp), async-profiler transparently falls back to `ctimer` mode.
 - If perf_events are available, but kernel symbols are hidden (e.g., by `kptr_resitrct` setting), async-profiler continues to use perf_events, emits a warning and does not show kernel stack traces.
 - To force using perf_events for user-space only profiling, specify `-e cpu-clock --all-user` instead of `-e cpu`.
 - `allkernel` option has been removed.
[...]

PR #1663, opened 2026-01-20
Event=cpu not switching alluser flag when kernel symbols are not available

Baraa-Hasheesh:
### Description
Missed when doing #1654, I noticed that `strcmp(args._event, EVENT_CPU)` basically prevented the `event=cpu` from switching the `alluser` flag 

This check no longer makes sense as one of the following is correct:
* event=cpu & `PerfEvents::supported` is true 
* event=cpu-clock
* event=cpu, `PerfEvents::supported` is false & `record_cpu/target_cpu` is true

For the last case async-profiler would have previously selected ctimer but after change in #1654, async-profiler will select perf events

So the previous change was incomplete, as this branch was in need to be updated to allow the third case to switch the alluser flag if needed

### Related issues
N/A

### Motivation and context
enhance UX by automatically switching the alluser flag in case event=cpu & kernel symbols are not available 

### How has this been tested?
manual, `make test`

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -796,8 +796,8 @@ Error PerfEvents::start(Arguments& args) {
                   "  sysctl kernel.perf_event_paranoid=1\n"
                   "  sysctl kernel.kptr_restrict=0");
         _kernel_stack = false;
-        // Automatically switch on alluser for non-CPU events, if kernel profiling is unavailable
-        _alluser = strcmp(args._event, EVENT_CPU) != 0 && !supported();
+        // Automatically switch on alluser if kernel profiling is unavailable
+        _alluser = !supported();
     }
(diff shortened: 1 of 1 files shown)


apangin:
`alluser` must not be enabled automatically for CPU profiling since it has serious implications, see #1044.

Baraa-Hasheesh:
@apangin 

even with check enabled, if the user selects "event=cpu-clock" the `alluser` will be enabled as we only check for `event=cpu`, 

so technically speaking `cpu-clock` will work while `cpu` won't, & I don't think that is desired,

Should we extend the check to also check for `cpu-clock`?

apangin:
That was a deliberate choice when fixing #1044 - so that users (and tests) relying on previous behavior could easily switch from `cpu` to `cpu-clock`. I suppose certain tests will break in restricted environments with the proposed change. So I don't want to revise previous decisions unless there are known customer issues.

Baraa-Hasheesh:
noted, I will close this PR

PR #1671, opened 2026-02-04
Allow user to select the TLAB based alloc tracer engine in newer JDKs (11+)

Baraa-Hasheesh:
### Description
Currently when the async-profiler is run with allocation sampling enabled it will try to use the best engine it can, as such
* If ObjectSampler JVMTI based engine is available async-profiler will use it
* If that engine is not available (Locked by a different profiler or we're using JDK8) then async-profiler will use the AllocTracer engine which instruments JFR methods to preform the sampling

It's not possible for the user to manually override this behaviour, so this change is being introduced to allow for manual override by the user to force the selection of the TLAB allocation engine 

Notes:
* I did consider adding this as a feature, but declaring it as a full flag looked to be better & more inline with other flags such as `record-cpu`

### Related issues
#1516

### Motivation and context
Allow user to use TLAB based allocation profiling in newer JDKs

### How has this been tested?
A new test was created 

---


apangin:
> I did find an old bug in allocTracer where the "0" value was not being translated to the DEFAULT_ALLOC_INTERVAL

It's not a bug - it's by design.
Unlike JVM TI ObjectSampler, there is no way to adjust threshold when JVM triggers AllocTracer events. These events are driven by TLABs, whose size depends on many factors and changes dynamically. The default and intended behavior of AllocTracer is to record all such events.

src/arguments.cpp
```diff
@@ -230,6 +230,9 @@ Error Arguments::parse(const char* args) {
                     if (_lock < 0) _lock = DEFAULT_LOCK_INTERVAL;
                 } else if (strcmp(value, EVENT_NATIVELOCK) == 0) {
                     if (_nativelock < 0) _nativelock = DEFAULT_LOCK_INTERVAL;
+                } else if (strcmp(value, EVENT_ALLOC_TLAB) == 0) {
```
apangin: IMO, accepting it as a new event type does not make sense - it only makes things confusing.
I like the idea of having it as a flag that modifies `alloc` behavior - let's keep it purely a flag then.
I'd also call it `--tlab` just for the sake of brevity (`--alloc --alloc-tlab` looks redundant and difficult to type).

Baraa-Hasheesh:
> > I did find an old bug in allocTracer where the "0" value was not being translated to the DEFAULT_ALLOC_INTERVAL
> 
> It's not a bug - it's by design. Unlike JVM TI ObjectSampler, there is no way to adjust threshold when JVM triggers AllocTracer events. These events are driven by TLABs, whose size depends on many factors and changes dynamically. The default and intended behavior of AllocTracer is to record all such events.

I reverted it here https://github.com/async-profiler/async-profiler/pull/1671/commits/5f7f91d88f468d55fa7edf4fe6feaa69d5e399d3

src/arguments.cpp
Baraa-Hasheesh: Applied here https://github.com/async-profiler/async-profiler/pull/1671/commits/5f7f91d88f468d55fa7edf4fe6feaa69d5e399d3

test/test/alloc/AllocTests.java
```diff
@@ -61,6 +63,25 @@ public void humongous(TestProcess p) throws Exception {
...
+    @Test(mainClass = MapReaderOpt.class)
+    public void tlabAllocSampler(TestProcess p) throws Exception {
+        p.profile("-e alloc --tlab -d 3 -f %profile.jfr");
+        boolean tlabEvent = false;
+
+        try (RecordingFile recordingFile = new RecordingFile(p.getFile("%profile").toPath())) {
+            while (recordingFile.hasMoreEvents()) {
+                RecordedEvent event = recordingFile.readEvent();
+                String eventName = event.getEventType().getName();
+                    if (eventName != null && eventName.equals("jdk.ObjectAllocationOutsideTLAB")) {
```
apangin: Indentation

src/arguments.cpp
```diff
@@ -249,6 +249,9 @@ Error Arguments::parse(const char* args) {
             CASE("alloc")
                 _alloc = value == NULL ? 0 : parseUnits(value, BYTES);
 
+            CASE("tlab")
```
apangin, line 253: There is another place at the top of this file where all agent arguments are listed.
(In the future, I think it makes sense to eliminate redundancy and leave ProfilerOptions.md the single source of truth).

src/main/main.cpp
```diff
@@ -69,6 +69,7 @@ static const char USAGE_STRING[] =
     "\n"
     "  --loop time         run profiler in a loop\n"
     "  --alloc bytes       allocation profiling interval in bytes\n"
+    "  --tlab              In allocation profiling use TLAB events\n"
```
apangin: Wording: "use TLAB events for allocation profiling"

Baraa-Hasheesh: updated wording in all locations here [8dc1a56](https://github.com/async-profiler/async-profiler/pull/1671/commits/8dc1a563c836f2ce721df62c3c88c3e55982acd0)

src/arguments.cpp
Baraa-Hasheesh: Agreed I do think it doesn't make sense to keep this information in 3 different locations

I added it in the remaining location here. [8dc1a56](https://github.com/async-profiler/async-profiler/pull/1671/commits/8dc1a563c836f2ce721df62c3c88c3e55982acd0)

test/test/alloc/AllocTests.java
Baraa-Hasheesh: fixed thanks [8dc1a56](https://github.com/async-profiler/async-profiler/pull/1671/commits/8dc1a563c836f2ce721df62c3c88c3e55982acd0)

Issue #1675, opened 2026-02-06
Cleanup all AGCT recovery tricks and remove `safemode` option


apangin:
### Describe the feature

There is a large chunk of intricate code in `getJavaTraceAsync` to recover from cases when `AsyncGetCallTrace` fails. AGCT often returns one of `ticks_unknown_Java`, `ticks_not_walkable_Java`, `ticks_unknown_not_Java`, or `ticks_not_walkable_not_Java` error codes. When async-profiler encounters these errors, it attempts to adjust SP/FP or otherwise "fix" stack layout and retry AGCT call. Such adjustments are not reliable, they caused several issues in the past. `safemode` option was introduced to selectively disable recovery techniques for troubleshooting purposes.

After switching to `vm` stack walker by default, maintaining this fragile code is no longer reasonable.

### Use Case

`AsyncGetCallTrace` remains only for Zing which does not have VMStructs.
However, AGCT recovery tricks are not used on Zing anyway, so it's safe to remove them.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1683, opened 2026-02-09
More AGCT cleanup after removal of recovery tricks 

Baraa-Hasheesh:
### Description
This PR continues work done in #1680

StackContext in practice is now only used to record the value of the CPU number when pref events are selected with "record-cpu", so most of it's related usages can be safely removed 

### Related issues
https://github.com/async-profiler/async-profiler/issues/1675

### Motivation and context
Cleanup of the profiler code

### How has this been tested?
`make test`

---

### DIFF
--- a/src/perfEvents_linux.cpp
+++ b/src/perfEvents_linux.cpp
@@ -784,8 +784,8 @@ void PerfEvents::signalHandlerJ9(int signo, siginfo_t* siginfo, void* ucontext) 
     if (_enabled) {
         u64 counter = readCounter(siginfo, ucontext);
         J9StackTraceNotification notif;
-        StackContext java_ctx;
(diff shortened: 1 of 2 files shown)


apangin:
Good catch: StackContext is no longer needed in `getJavaTraceAsync`, indeed.
But removal of `unwindAtomicStub` was premature: it breaks unwinding of non-Java threads, e.g. a stack of a GC thread may have an atomic stub on the top with the rest of the stack composed of the JVM native frames.
Also, `StackContext` structure itself is now redundant.

Baraa-Hasheesh:
> But removal of unwindAtomicStub was premature

I've reverted this part of the change

> StackContext structure itself is now redundant.

I did consider deleting it completely but wasn't 100% sure if it would ever be needed again, however I will delete it & if it's ever needed it can be re-introduced

apangin:
All good now, thanks.

PR #1684, opened 2026-02-10
Unwind AArch64 generated stubs on JDK 26+

apangin:
### Description

1. Strip `Stub Generator` prefix introduced with JDK-8336658.
2. Change `strcmp` to `strncmp` to account for stub suffices added in JDK 26 and JDK 27.
3. Remove dead code related to AsyncGetCallTrace.

### Related issues

`RecoveryTests.intrinsics` fails on JDK 26.

### Motivation and context

Fix unwinding of runtime stubs on JDK 26 / ARM64.

### How has this been tested?

Integration tests pass on JDK 26 and 27-ea.
Manually profiled spring-petclinic and searched for stubs.

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -117,6 +117,10 @@ void Profiler::addJavaMethod(const void* address, int length, jmethodID method) 
 }
 
 void Profiler::addRuntimeStub(const void* address, int length, const char* name) {
+    if (name[0] == 'S' && strncmp(name, "Stub Generator ", 15) == 0) {
+        name += 15;  // useless prefix introduced with JDK-8336658
+    }
+
     _stubs_lock.lock();
     _runtime_stubs.add(address, length, name, true);
     _stubs_lock.unlock();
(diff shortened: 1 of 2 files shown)


Baraa-Hasheesh:
> Strip Stub Generator prefix introduced with JDK-8336658.
Change strcmp to strncmp to account for stub suffices added in JDK 26 and JDK 27.

Since we're already removing the "Stub Generator" prefix from the name before we add it to the `_runtime_stubs` wouldn't it make sense to also remove the "_stub" suffix before we add it?

With the current change `base64_encodeBlock` stub will be reported in the stack as `base64_encodeBlock_stub` which doesn't look nice

apangin:
@Baraa-Hasheesh
> wouldn't it make sense to also remove the "_stub" suffix

This was my first intention, but things appeared to be more complicated.
JDK 27 changed stub names once again, unifying them across all blob types.
E.g. `generic_arraycopy` is now `generic_arraycopy_stub (stub gen)`,
and `C2 Runtime new_instance` became `new_instance_blob (C2 runtime)`.

You may ask what if we remove everything in parentheses, but that's not ideal too, as it won't be possible to distinguish `new_instance_blob (C1 runtime)` from `new_instance_blob (C2 runtime)`. So I decided to leave the original JDK names - after all, that's what JDK prints in its own stack traces.

Issue #1258, opened 2025-04-24
Consider running Java integration tests in parallel


fandreuz:
We should consider running Java integration tests in parallel to speed up our GHA.

rudraroop:
@fandreuz can you assign this to me?

apangin:
Resolved by #1688

PR #1688, opened 2026-02-17
Adding a multi-threaded test executor

lutkerd:
### Description
 Use thread pools for executing tests with a default of 8 test threads. Also creates a single thread executor to run tests that didn't initially work in multi-threaded execution. Fixing tests to run in multi-threaded mode will be left to another PR.

On my host, execution time went from ~330s to ~60s.

### Motivation and context

https://github.com/async-profiler/async-profiler/issues/1258

### How has this been tested?
Ran locally and all tests are passing, need to run through GHAs to test more platforms.

---

### DIFF
--- a/test/one/profiler/test/Runner.java
+++ b/test/one/profiler/test/Runner.java
@@ -175,11 +175,9 @@ public class Runner {
         }
 
         int pass = statusCounts.get(TestStatus.PASS.ordinal());
-        String totalDuration = String.format("%.3f s", totalTestDuration / 1e9);
-        String actualExecutionDuration = String.format("%.3f s", executionDuration / 1e9);
 
-        System.out.println("\nTotal test duration: " + totalDuration);
-        System.out.println("\nActual execution duration: " + actualExecutionDuration);
+        System.out.printf("\nTotal test duration: %.3f s\n", totalTestDuration / 1e9);
+        System.out.printf("Actual execution duration: %.3f s\n", executionDuration / 1e9);
         System.out.println("Results Summary:");
         System.out.printf("PASS: %d (%.1f%%)\n", pass, 100.0 * pass / (pass + fail));
         System.out.println("FAIL: " + fail);
@@ -249,7 +247,7 @@ public class Runner {
                     failedTests.add(rt.testInfo());
                 }
 
-                System.out.printf("tid[%d] %s [%d/%d] %s took %.3f s\n", Thread.currentThread().getId(), result.status(), i.getAndIncrement(), testCount, rt.testInfo(), durationNs / 1e9);
+                System.out.printf("%s [%d/%d] tid[%d] %s took %.3f s\n", result.status(), i.getAndIncrement(), testCount, Thread.currentThread().getId(), rt.testInfo(), durationNs / 1e9);
                 if (result.throwable() != null) {
                     result.throwable().printStackTrace(System.out);
                 }


test/test/cpu/CpuTests.java
```diff
@@ -47,7 +47,7 @@ public void itimerTotal(TestProcess p) throws Exception {
         assertCloseTo(out.total(), 2_000_000_000, "itimer total should match profiling duration");
     }
 
-    @Test(mainClass = CpuBurner.class, os = Os.LINUX)
```
Baraa-Hasheesh, line 50: singleThreaded is not the best name IMO, maybe something like "isolated"?

Also what's the reasoning for having these tests run one by one?

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +185,87 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+        List<String> failedTests = Collections.synchronizedList(new ArrayList<>());
+        Map<TestStatus, Integer> statusCounts = Collections.synchronizedMap(new EnumMap<>(TestStatus.class));
+        BlockingQueue<Runnable> multiThreadWorkQueue = new ArrayBlockingQueue<>(testCount);
 
-            totalTestDuration += durationNs;
-            statusCounts.put(result.status(), statusCounts.getOrDefault(result.status(), 0) + 1);
-            if (result.status() == TestStatus.FAIL) {
-                failedTests.add(rt.testInfo());
-            }
+        final ThreadPoolExecutor executor = new ThreadPoolExecutor(threadCount, threadCount, 60L, TimeUnit.SECONDS, multiThreadWorkQueue);
```
Baraa-Hasheesh: historically speaking we have had multiple flakey failures in tests due to low number of cores/threads on the Github runners side, can you re-run the tests multiple times to make sure that this issue doesn't become worse with running tests in parallel

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +185,87 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+        }
+    }
     public static void main(String[] args) throws Exception {
         configureLogging();
 
         TestDeclaration decl = TestDeclaration.parse(args);
         List<RunnableTest> allTests = decl.getRunnableTests();
         final int testCount = allTests.size();
         final int retryCount = Integer.parseInt(System.getProperty("retryCount", "0"));
+        final int threadCount = Integer.parseInt(System.getProperty("threadCount", "8"));
```
Baraa-Hasheesh: maybe "parallelism" would be a better name

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +185,87 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
         System.out.println("TOTAL: " + testCount);
     }
 
+    private static void waitForExecutorTermination(ThreadPoolExecutor executor) {
+        executor.shutdown(); // Initiate orderly shutdown
+        try {
+            // Wait indefinitely (or a specific time) for all tasks to finish
+            boolean terminated = executor.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
```
Baraa-Hasheesh: I believe 3600s would be more than sufficient here rather than an infinite wait

In a matter of fact I think the sink time here can be much lower than 3600s 

-> Let's avoid any possible infinite wait

test/test/cpu/CpuTests.java
lutkerd: Changed to runIsolated.

These 2 tests failed locally when running multithreaded. I think they expected only a certain amount of change in a perf counter and running multiple tests broke that assumption.

test/one/profiler/test/Runner.java
lutkerd: Yes, that was the plan to run the actions a few times to make sure tests that needed isolation would be detected across all platforms.

test/one/profiler/test/Runner.java
lutkerd: changed to 5 minutes for now and I will adjust again based on results.

lutkerd:
I've run this 5 times with 8 threads and about the same with 2 threads on the GHA's and there haven't been any failures. It seems stable enough and it's easy to change the default in the Makefile or override in the integ.yml for just GHA.

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +186,99 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
         System.out.println("TOTAL: " + testCount);
     }
 
+    private static void waitForExecutorTermination(ThreadPoolExecutor executor) {
+        executor.shutdown(); // Initiate orderly shutdown
+        try {
+            // Wait indefinitely (or a specific time) for all tasks to finish
+            boolean terminated = executor.awaitTermination(5 * 60L, TimeUnit.SECONDS);
```
apangin: 5 minutes is too short - it may not be enough to run even normal tests single-threaded.
Furthermore, we wanted to add extended tests once tiered strategy is implemented.
Each individual test already has its own timeout. I don't think another timeout is necessary. There is an overall GHA timeout anyway. (I saw Bara's comment, but it does not seem to be relevant in this particular case).

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +186,99 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
-                result = run(rt, decl);
-                attempt++;
-            }
+        log.log(Level.INFO, "Running with " + threadCount + " test threads.");
 
-            long durationNs = System.nanoTime() - start;
+        AtomicLong i = new AtomicLong(1);
+        AtomicLong totalTestDuration = new AtomicLong();
+        Set<String> failedTests = Collections.synchronizedSet(new HashSet<>());
+        Map<TestStatus, Integer> statusCounts = Collections.synchronizedMap(new EnumMap<>(TestStatus.class));
```
apangin: `ConcurrentHashMap` / `ConcurrentHashMap.newKeySet()` are almost always better than `Collections.synchronizedMap` / `Collections.synchronizedSet`.

In this case, however, it's not just about performance but more about correctness.
`statusCounts.put(result.status(), statusCounts.getOrDefault(result.status(), 0) + 1);` is not safe with either implementation, but with `ConcurrentHashMap`, it can be rewritten in a correct way using `compute` method.

What we need here is a per-status counter - I would probably not use a map for it at all. Simple `AtomicIntegerArray` is a good alternative for concurrent `EnumMap<Enum, Integer>`.

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +186,99 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+                    totalTestDuration.addAndGet(durationNs);
+                    statusCounts.put(result.status(), statusCounts.getOrDefault(result.status(), 0) + 1);
+                    if (result.status() == TestStatus.FAIL) {
+                        failedTests.add(rt.testInfo());
+                        if (!rerunTests.contains(rt)) {
+                            log.log(Level.INFO, "Adding " + rt.testInfo() + " to rerun list.");
+                            // Track if the test was rerun due to a failure so we don't rerun in a loop.
+                            rerunTests.add(rt);
+                            singleThreadWorkQueue.add(this);
+                        }
```
apangin: Mmm, I don't really like this logic. A test should be either tolerant to concurrent execution or it should always run isolated. It would be ideal for all tests to support concurrent execution, but that's a separate problem. For now, I would like not to hide potential issues by rerunning failed tests single-threaded.

test/one/profiler/test/Runner.java
```diff
@@ -176,45 +186,99 @@ private static void printSummary(EnumMap<TestStatus, Integer> statusCounts, List
...
+                    }
+
+                    System.out.printf("tid[%d] %s [%d/%d] %s took %.3f s\n", Thread.currentThread().getId(), result.status(), i.getAndIncrement(), testCount, rt.testInfo(), durationNs / 1e9);
+                    if (result.throwable() != null) {
+                        result.throwable().printStackTrace(System.out);
+                    }
+                }
+            };
+            if (rt.test().runIsolated()) {
+                singleThreadWorkQueue.add(task);
```
apangin: This is a rather creative way to submit a list of tasks to an executor )
The canonical approach is to call `threadPool.invokeAll` or just `list.forEach(threadPool::execute)`

Makefile
```diff
@@ -62,7 +62,8 @@ LOG_DIR=build/test/logs
 LOG_LEVEL=
 SKIP=
 RETRY_COUNT=0
-TEST_FLAGS=-DlogDir=$(LOG_DIR) -DlogLevel=$(LOG_LEVEL) -Dskip='$(subst $(COMMA), ,$(SKIP))' -DretryCount=$(RETRY_COUNT)
+TEST_THREADS ?= 8
```
Baraa-Hasheesh, line 65: Rather than having this as a static value of 8 which might break other CI/CD systems that run async-profiler, it would be better for this to be smart enough to avoid running on more CPU cores than available 

For example on macOS you should be able to get number of cores using `sysctl -n hw.ncpu` 
on linux `nproc --all` should do the job (alternatively you can read the information from `/proc/cpuinfo`

test/one/profiler/test/Runner.java
lutkerd: Sure, I can put this behind the retry flag, I found it useful for figuring out if a failure was because of multi-threaded execution or because I broke something.

test/one/profiler/test/Runner.java
lutkerd: will fix in next rev

test/one/profiler/test/Runner.java
lutkerd: I was trying to avoid creating another Collection in addition to the work queues to hold the tasks while filtering to multi-threaded and single-threaded. The invokeAll() may be better since it would block until all tasks complete given the comments about the timeouts on waiting for termination.

Makefile
lutkerd: I would like to revisit this after we are able to make all of the tests run reliably multithreaded mode, adding more concurrency causes more instability and less reproducibility since the results across machines will had differing thread counts. In some local testing, on a host with 48 cpus, caused 1-3 more failures and only saved an additional 20 seconds, ~85s -> ~65s. If/when we do this, we could easily get the available processors in java in a cross platform way and remove the default in the make file.

Here are some timing results from my host with various thread counts, after ~20 threads there is no big improvement.

```
log_1:Actual execution duration: 335.226 s
log_3:Actual execution duration: 145.142 s
log_5:Actual execution duration: 105.667 s
log_7:Actual execution duration: 90.695 s
log_8:Actual execution duration: 86.140 s
log_10:Actual execution duration: 78.135 s
log_12:Actual execution duration: 74.423 s
log_14:Actual execution duration: 71.345 s
log_16:Actual execution duration: 69.494 s
log_18:Actual execution duration: 67.282 s
log_20:Actual execution duration: 65.531 s
log_22:Actual execution duration: 64.881 s
log_24:Actual execution duration: 63.773 s
log_26:Actual execution duration: 62.299 s
log_28:Actual execution duration: 63.074 s
log_30:Actual execution duration: 62.115 s
log_32:Actual execution duration: 61.648 s
log_34:Actual execution duration: 60.969 s
log_36:Actual execution duration: 61.143 s
log_38:Actual execution duration: 60.755 s
log_40:Actual execution duration: 60.447 s
```

test/one/profiler/test/Runner.java
```diff
@@ -156,65 +167,125 @@ private static void configureLogging() {
...
+        final ThreadPoolExecutor executor = new ThreadPoolExecutor(threadCount, threadCount, 60L, TimeUnit.SECONDS, new ArrayBlockingQueue<>(testCount));
+        final ThreadPoolExecutor singleExecutor = new ThreadPoolExecutor(1, 1, 60L, TimeUnit.SECONDS, new ArrayBlockingQueue<>(testCount));
 
-            System.out.printf("%s [%d/%d] %s took %.3f s\n", result.status(), i, testCount, rt.testInfo(), durationNs / 1e9);
-            if (result.throwable() != null) {
-                result.throwable().printStackTrace(System.out);
+        final ArrayList<Callable<TestResult>> multithreadedTests = new ArrayList<>();
+        final ArrayList<Callable<TestResult>> singlethreadedTests = new ArrayList<>();
+
+        HashSet<RunnableTest> rerunTests = new HashSet<>();
```
apangin: Not thread-safe

test/one/profiler/test/Runner.java
```diff
@@ -156,65 +167,125 @@ private static void configureLogging() {
...
+
+        long startTime = System.nanoTime();
+        for (RunnableTest rt : allTests) {
+            Callable task = new Callable<TestResult>() {
```
apangin: Replace with a lambda expression?

test/one/profiler/test/Runner.java
apangin: Test reliability is more important than execution speed. However, what this code says is "it's not a problem if a test fails in multithreaded mode". Such failures may remain unnoticed as long as `retry` parameter is set (which is always set on macOS, for example).

If this retry logic is removed, how will it affect test pass ratio?

test/one/profiler/test/Runner.java
lutkerd: With a lambda I can't access `this` so you can't requeue the task for single threaded execution.

test/one/profiler/test/Runner.java
lutkerd: I will remove this then. Most of the problematic tests are already marked to runIsolated, I don't expect this to really change anything for the vast majority of executions.

test/one/profiler/test/Runner.java
```diff
@@ -156,65 +167,110 @@ private static void configureLogging() {
...
+        int fail = statusCounts.get(TestStatus.FAIL.ordinal());
         if (fail > 0) {
             System.out.println("\nFailed tests:");
             failedTests.forEach(System.out::println);
         }
 
-        int pass = statusCounts.getOrDefault(TestStatus.PASS, 0);
+        int pass = statusCounts.get(TestStatus.PASS.ordinal());
         String totalDuration = String.format("%.3f s", totalTestDuration / 1e9);
+        String actualExecutionDuration = String.format("%.3f s", executionDuration / 1e9);
```
apangin: You can just inline this in `printf` below.

test/one/profiler/test/Runner.java
```diff
@@ -156,65 +167,110 @@ private static void configureLogging() {
...
 
         System.out.println("\nTotal test duration: " + totalDuration);
+        System.out.println("\nActual execution duration: " + actualExecutionDuration);
```
apangin: nit: `\n` is redundant - at least, on the second line. It would look better without an empty line in between.

test/one/profiler/test/Runner.java
```diff
@@ -156,65 +167,110 @@ private static void configureLogging() {
...
+                statusCounts.incrementAndGet(result.status().ordinal());
+
+                if (result.status() == TestStatus.FAIL) {
+                    failedTests.add(rt.testInfo());
+                }
 
-            System.out.printf("%s [%d/%d] %s took %.3f s\n", result.status(), i, testCount, rt.testInfo(), durationNs / 1e9);
-            if (result.throwable() != null) {
-                result.throwable().printStackTrace(System.out);
+                System.out.printf("tid[%d] %s [%d/%d] %s took %.3f s\n", Thread.currentThread().getId(), result.status(), i.getAndIncrement(), testCount, rt.testInfo(), durationNs / 1e9);
```
apangin: super-nit: it's easier to visually track progress when the status line starts with `PASS [131/150]` rather than `tid[20]`.

Issue #1676, opened 2026-02-06
Unify `dwarf` and `vm` stack walking modes


apangin:
### Describe the feature

Merge `StackWalker::walkDwarf` with `StackWalker::walkVM`.
Make `cstack=dwarf` an alias of `cstack=vm`.

### Use Case

Simplify stack walking code.
Reduce the amount of duplication.
Leverage crash protection in `StackWalker::walkVM` for native applications.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1689, opened 2026-02-18
Migrate existing users from dwarf & fp stack walkers to vm if available

Baraa-Hasheesh:
### Description
While working on #1676 many challenges were observed, these are mainly related to differences between hotspot & openj9 JDKs in terms of how the stack is collected 

Merging the code is doable but will require many changes that will make the code highly unmergable without spreading the change across multiple PRs

Even after the change we will end up with effectively a larger codebase as the savings acquired by cleaning up DWARF function will be largely overshadowed by the changes needed to address J9 side of the stack walking 

As such as discussed internally it's a better idea to simply move existing users of the async-profiler to use VM when available & only expose legacy stack walkers via a new feature flag that is off by default

As part of this change the `no` stackwalker was replaced to use VM if available & fallback to agct if not

### Related issues
N/A

### Motivation and context
Move users to the better VM stack walker if available

### How has this been tested?
`make test`

---

### DIFF
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -112,7 +112,8 @@ struct StackWalkFeatures {
     unsigned short vtable_target : 1;   // show receiver classes of vtable/itable stubs
     unsigned short comp_task     : 1;   // display current compilation task for JIT threads
     unsigned short pc_addr       : 1;   // record exact PC address for each sample
-    unsigned short _padding      : 9;   // pad structure to 16 bits
+    unsigned short no_native     : 1;   // do not collect native stacks
+    unsigned short _padding      : 8;   // pad structure to 16 bits
 };
 
 
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -107,11 +107,12 @@ constexpr int EVENT_MASK_SIZE = 7;
 struct StackWalkFeatures {
     unsigned short stats         : 1;   // collect stack walking duration statistics
     unsigned short jnienv        : 1;   // verify JNIEnv* obtained using VMStructs
+    unsigned short agct          : 1;   // force usage of legacy stack walkers when selected rather than silently using vm
     unsigned short mixed         : 1;   // mixed stack traces with Java and native frames interleaved
     unsigned short vtable_target : 1;   // show receiver classes of vtable/itable stubs
     unsigned short comp_task     : 1;   // display current compilation task for JIT threads
     unsigned short pc_addr       : 1;   // record exact PC address for each sample
-    unsigned short _padding      : 10;  // pad structure to 16 bits
+    unsigned short _padding      : 9;   // pad structure to 16 bits
 };
 
 
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -60,7 +60,6 @@ enum SHORT_ENUM CStack {
     CSTACK_FP,       // walk stack using Frame Pointer links
     CSTACK_DWARF,    // use DWARF unwinding info from .eh_frame section
     CSTACK_VM,       // unwind using HotSpot VMStructs
-    CSTACK_INVALID   // added to stop compilation warning related to assert statement in flightRecorder
 };
 
 enum SHORT_ENUM Clock {
@@ -112,7 +111,7 @@ struct StackWalkFeatures {
     unsigned short vtable_target : 1;   // show receiver classes of vtable/itable stubs
     unsigned short comp_task     : 1;   // display current compilation task for JIT threads
     unsigned short pc_addr       : 1;   // record exact PC address for each sample
-    unsigned short no_native     : 1;   // do not collect native stacks
+    unsigned short java_only     : 1;   // only collect java stacks
     unsigned short _padding      : 8;   // pad structure to 16 bits
 };
 
--- a/src/arguments.h
+++ b/src/arguments.h
@@ -57,10 +57,10 @@ enum Style {
 // Whenever enum changes, update SETTING_CSTACK in FlightRecorder
 enum SHORT_ENUM CStack {
     CSTACK_DEFAULT,  // use perf_event_open stack if available or Frame Pointer links otherwise
-    CSTACK_NO,       // do not collect native frames
     CSTACK_FP,       // walk stack using Frame Pointer links
     CSTACK_DWARF,    // use DWARF unwinding info from .eh_frame section
-    CSTACK_VM        // unwind using HotSpot VMStructs
+    CSTACK_VM,       // unwind using HotSpot VMStructs
+    CSTACK_INVALID   // added to stop compilation warning related to assert statement in flightRecorder
 };
 
 enum SHORT_ENUM Clock {
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -59,7 +59,7 @@ static const char USAGE_STRING[] =
     "  -I include          output only stack traces containing the specified pattern\n"
     "  -X exclude          exclude stack traces with the specified pattern\n"
     "  -L level            log level: debug|info|warn|error|none\n"
-    "  -F features         advanced stack trace features: mixed, vtable, comptask, pcaddr\n"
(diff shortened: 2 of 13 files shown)


apangin:
> we might even want to delete the `no` walker as it has some buggy behaviour related to `record-cpu` currently

What behavior do you mean? This is totally fine if `record-cpu` does not work with `cstack=no`, but `cstack=no` itself is an important option that I know customers use. It must not be silently ignored, and it should not degrade. Ideally, `cstack=no` should also use `StackWalker::walkVM`.

src/profiler.cpp
```diff
@@ -937,14 +937,11 @@ Error Profiler::start(Arguments& args, bool reset) {
...
 
-    if (_cstack == CSTACK_DEFAULT) {
-        if (VMStructs::hasStackStructs()) {
-            // Use VMStructs by default when possible
-            _cstack = args._cstack = CSTACK_VM;
-        } else if (VM::isOpenJ9() && DWARF_SUPPORTED) {
-            // OpenJ9 libs are compiled with frame pointers omitted
-            _cstack = args._cstack = CSTACK_DWARF;
-        }
+    if (VMStructs::hasStackStructs() && !_features.agct) {
```
apangin, line 942: The main argument for merging `dwarf` and `vm` (or "migrating" from one to another) was not that much about code deduplication (that would be nice, but as we discussed, not quite simple) - it was about bringing benefits of `vm` (like crash protection) to users of `dwarf` automatically. This also applies to native applications. I wonder how easy it will be to switch from `dwarf` to `vm` for native apps.

Baraa-Hasheesh:
> > we might even want to delete the `no` walker as it has some buggy behaviour related to `record-cpu` currently
> 
> What behavior do you mean? This is totally fine if `record-cpu` does not work with `cstack=no`, but `cstack=no` itself is an important option that I know customers use. It must not be silently ignored, and it should not degrade. Ideally, `cstack=no` should also use `StackWalker::walkVM`.

`record-cpu` will always record CPU as 0 if `no` is selected as we never read that information from the perf event

src/profiler.cpp
Baraa-Hasheesh: > I wonder how easy it will be to switch from dwarf to vm for native apps

That should be already be possible, as if you check [here](https://github.com/async-profiler/async-profiler/blob/master/src/profiler.cpp#L936) a pure native application is already allowed to use VM walker due to the fact that the VM is not loaded

This was something introduced to allow a native that later launches a JVM to easily transition from pure native stack walking to mixed Java & native stack walking when running in `loop` mode (quite useful for LD_PRELOAD)

Baraa-Hasheesh: > native applications

For this I intentionally didn't touch the path even though it's possible as it will require some extra checks for the possibility of a non hotspot instance being launched & the check will be quite confusing

Baraa-Hasheesh:
I've updated the PR with an option not to collect native stacks for VM & stopped recording the CPU in case user selects `no` walker

src/main/main.cpp
```diff
@@ -59,7 +59,7 @@ static const char USAGE_STRING[] =
     "  -I include          output only stack traces containing the specified pattern\n"
     "  -X exclude          exclude stack traces with the specified pattern\n"
     "  -L level            log level: debug|info|warn|error|none\n"
-    "  -F features         advanced stack trace features: mixed, vtable, comptask, pcaddr\n"
+    "  -F features         advanced stack trace features: nonative, mixed, vtable, comptask, pcaddr\n"
```
apangin: Showing native stacks is a feature; not showing is an anti-feature.
In the code, `no_native` is mostly used with negation (`!no_native`) that makes it look intricate. Can we flip the feature to the positive meaning?

src/arguments.h
```diff
@@ -57,10 +57,10 @@ enum Style {
 // Whenever enum changes, update SETTING_CSTACK in FlightRecorder
 enum SHORT_ENUM CStack {
     CSTACK_DEFAULT,  // use perf_event_open stack if available or Frame Pointer links otherwise
-    CSTACK_NO,       // do not collect native frames
     CSTACK_FP,       // walk stack using Frame Pointer links
     CSTACK_DWARF,    // use DWARF unwinding info from .eh_frame section
-    CSTACK_VM        // unwind using HotSpot VMStructs
+    CSTACK_VM,       // unwind using HotSpot VMStructs
+    CSTACK_INVALID   // added to stop compilation warning related to assert statement in flightRecorder
```
apangin: What warning are you referring to?

src/main/main.cpp
Baraa-Hasheesh: what do you think about `java_only`?

src/arguments.h
Baraa-Hasheesh: on macOS after removing the "no" walker, the compiler started showing a warning on 
`assert(args._cstack < sizeof(SETTING_CSTACK) / sizeof(char*));` 

=> condition is always true

Baraa-Hasheesh:
```
In file included from <stdin>:15:
./src/flightRecorder.cpp:631:29: warning: result of comparison of constant 4 with expression of type 'CStack' is always true [-Wtautological-constant-out-of-range-compare]
  631 |         assert(args._cstack < sizeof(SETTING_CSTACK) / sizeof(char*));
      |                ~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/assert.h:72:25: note: expanded from macro 'assert'
   72 |     (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __ASSERT_FILE_NAME, __LINE__, #e) : (void)0)
```

apangin: But that is exactly what this assertion is about - it is to ensure that `SETTING_CSTACK` array is large enough to match any value of `_cstack`.

Baraa-Hasheesh: That is true, but apparently before removing the "no" string the compiler wasn't smart enough to figure out that it's always true, currently after the change it found out that the condition is indeed always true which caused the warning to start appearing only after the change 

I'm fine with removing this extra enum value but the warning message will stay

apangin: This should not be a choice of one over another. Can we have a clean code with a safety check and no warnings?

Baraa-Hasheesh: I used pointer arithmetics to obfuscate the check from the compiler [9882634](https://github.com/async-profiler/async-profiler/pull/1689/commits/98826344c85bd52c220c2717c5883da683663488)

src/main/main.cpp
Baraa-Hasheesh: changed to javaonly [97b5556](https://github.com/async-profiler/async-profiler/pull/1689/commits/97b5556c01c3efe3bde5b9a0e3378da78f3cb426)

apangin:
Support for `--cstack no` in VMStructs stack walker is important, but it makes the implementation a bit messy. To minimize changes for the next release, I propose to split this PR into two parts.
As the first step, we'll just silently convert `dwarf`->`vm`, unless `agct` feature is specified.

https://github.com/async-profiler/async-profiler/compare/master...dwarf-to-vm

This is similar to what you've proposed in the first revision, except for two changes:
- `fp` remains unaffected;
- if `agct` feature is specified, it is mentioned in the JFR ActiveSetting event.

Baraa-Hasheesh:
@apangin sounds reasonable, I checked the changes they look good to me

Just a general question about FP, is the plan to eventually add a special feature to always use FP on the native stacks in the VM walker?

apangin:
Thanks. I don't think there will be such a feature, but I see benefits in 1) making `walkVM` smarter so it can automatically adjust strategy if DWARF unwinding info is not available; and 2) probably an option to use perf_events' built-in stack walker (which is FP-based).

apangin:
Modified version of this PR was merged in #1726.
A change for `cstack=no` will follow in a separate PR.

PR #1696, opened 2026-03-10
flamegraph: support Shift+Click for remove-stack

### LINKED ISSUE #1672: Alt+Click keybinding doesn't work on Linux
### Describe the bug

The keybinding introduced in #896 doesn't work on some desktop environments as it clashes with the default DE keybinding for moving windows. e.g Cinnamon

Maybe we could switch to `Shift+Click` instead? This would also better align with `N` and `Shift+N`

### Reproduction Steps

_No response_

_No response_

### Async-profiler version

v4.3

### Environment details

Linux Mint 22.3 with Cinnamon

SergioChan:
### Description
- Allow `Shift+Click` as an alternative to `Alt+Click` for "Remove stack" in flame graph view.
- Keep existing `Alt+Click` behavior for backwards compatibility.
- Update the in-page hotkey legend to document both modifiers.

### Related issues
Fixes #1672

### Motivation and context
Some desktop environments reserve `Alt+Click` for window management, so the existing shortcut can be intercepted before it reaches the browser. Adding `Shift+Click` provides a reliable cross-desktop fallback while preserving current behavior.

### How has this been tested?
- Manual code-path verification in `src/res/flame.html` to ensure remove-stack now triggers when either `event.shiftKey` or `event.altKey` is pressed.
- Attempted to run `make test-java`, but local environment is missing JVMTI headers (`jvmti.h`), so native build/test could not complete in this environment.

---

### DIFF
--- a/.assets/html/flamegraph.html
+++ b/.assets/html/flamegraph.html
@@ -57,7 +57,7 @@
 </dl>
 <dl class='hotkeys'>
 	<dt>Click frame</dt><dd>Zoom into frame</dd>
-	<dt>Alt+Click</dt><dd>Remove stack</dd>
+	<dt>Ctrl+Click / Alt+Click</dt><dd>Remove stack</dd>
 	<dt>0</dt><dd>Reset zoom</dd>
 	<dt>I</dt><dd>Invert graph</dd>
 	<dt>Ctrl+F</dt><dd>Search</dd>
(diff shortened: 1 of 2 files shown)


apangin:
Thank you for the contribution, I appreciate it.
However, as mentioned in the original bug report, Shift+Click is not an option: Shift+Click is typically used for opening a link in a new window or for extending text selection. In general, Shift key is a "light" modifier - it's uncommon to associate it with a destructive secondary action.

On another note, if you modify a flamegraph template, please modify the demo flamegraph correspondingly.

SergioChan:
Thanks for the quick review and for clarifying.

Agreed — I won’t push the Shift+Click approach.

Would you prefer `Ctrl+Click` as the fallback modifier here (while keeping `Alt+Click` for existing behavior), or a different gesture entirely? Once you confirm the preferred interaction, I’ll update both the flamegraph template and the demo flamegraph together.

SergioChan:
Thanks for the quick review �� addressed in a4fd304.

Changes made:
- Replaced the Shift fallback with Ctrl, so remove-stack is now `Ctrl+Click` or `Alt+Click`.
- Kept existing `Alt+Click` behavior for backward compatibility.
- Updated both files to stay in sync:
  - `src/res/flame.html` (template)
  - `.assets/html/flamegraph.html` (demo asset)

I also kept the click handler explicit (`function(event)`) while updating the modifier check.

apangin:
An alternative fix has been already pushed.

PR #1697, opened 2026-03-12
Ensure remaining buffer is sufficient for event data in JfrReader

alevymyers:
### Description

`JfrReader.readEvent()` only calls `ensureBytes(CHUNK_HEADER_SIZE)` (68 bytes) at the top of its read loop. It does not ensure the buffer contains enough data for the full event before dispatching to event readers. If an event exceeds the remaining buffer capacity, the read methods (`getString()` in my example) throw `BufferUnderflowException`. 

Any event larger than 68 bytes is technically susceptible, including, I believe, the new `ProcessSample`. Although an event close to 68 bytes should rare hit this in practice.

#### Reproducer

See my [branch](https://github.com/alevymyers/async-profiler/commit/1c084fc48f8049a4f8949bafa2e9c22cb27726da) with a naive reproducer that can create a JFR with a ThreadDump event exceeding the default buffer size. Along with that, the required changes to cause the error in `JfrReader` and `JfrConverter`. 

### Motivation and context

Will crash under certain conditions if the code is extended, in a way that is supported.

### How has this been tested?

Verified using the reproducer branch and confirmed the `BufferUnderflowException` before the fix and successfully reading the JFR after.

---

### DIFF
--- a/src/converter/one/jfr/JfrReader.java
+++ b/src/converter/one/jfr/JfrReader.java
@@ -181,8 +181,6 @@ public class JfrReader implements Closeable {
                 }
                 return null;
             }
-            long eventEnd = filePosition + pos + size;
-            ensureBytes(size - (buf.position() - pos));
 
             if (type == executionSample || type == nativeMethodSample) {
                 if (cls == null || cls == ExecutionSample.class) return (E) readExecutionSample(false);
@@ -213,6 +211,8 @@ public class JfrReader implements Closeable {
             } else {
                 Constructor<? extends Event> customEvent = customEvents.get(type);
                 if (customEvent != null && (cls == null || cls == customEvent.getDeclaringClass())) {
+                    long eventEnd = filePosition + pos + size;
+                    ensureBytes(size - (buf.position() - pos));
                     try {
                         return (E) customEvent.newInstance(this);
                     } catch (ReflectiveOperationException e) {
@@ -223,7 +223,7 @@ public class JfrReader implements Closeable {
                 }
             }
 
-            seek(eventEnd);
+            seek(filePosition + pos + size);
         }
 
         state = STATE_EOF;


src/converter/one/jfr/JfrReader.java
```diff
@@ -181,6 +181,8 @@ public <E extends Event> E readEvent(Class<E> cls) throws IOException {
                 }
                 return null;
             }
+            long eventEnd = filePosition + pos + size;
```
apangin: Thank you for the report and the bug fix. This makes sense, but please avoid touching code unrelated to the fix, unless there is a strong reason for it. Thanks.

alevymyers: Thank you for the quick response, in this case it is required by the fix. `ensureBytes()` can advance `filePosition` when it compacts or reallocates the buffer. The existing `seek(filePosition + pos + size)` calls in those cases would then use the modified `filePosition` with the old `pos`. I am capturing it while they are consistent with eachother.

apangin: I see, thank you for the explanation.
This, however, raises another concern: `JfrReader` will now allocate a buffer for large events and unconditionally read all of them from disk, even when no one is interested in these events. Like in your reproducer, a good example is a `ThreadDump` event that can easily exceed 10MB for a normal Java service.
Since all standard events that are hard-coded in the parser (with an exception of `ActiveSetting`) are smaller than `CHUNK_HEADER_SIZE`, will it make sense to move `ensureBytes` logic into a branch for custom events only?

alevymyers: That makes sense, I've moved `ensureBytes` logic into the custom event branch (d8146c7a0734453b78bafcd44755d2231b9fac48). It's worth noting that this does add an implicit assumption that all hard-coded events fit within `CHUNK_HEADER_SIZE`, but as you pointed out that is true outside `ActiveSetting`, which in practice, I do not think will trigger the exception.

apangin:
Merged. Thank you for the contribution!

PR #1701, opened 2026-03-18
Optimize NMethod type check

apangin:
### Description

Currently, `NMethod::isInterpreter`, `isNMethod`, `isStub` and `isVTableStub` are implemented by comparing the name using `strcmp`/`strncmp`. We can make the check faster by inlining string comparison.

Also, for interpreted frames, instead of checking NMethod name we could check NMethod pointer itself, because Interpreter is a singleton.

On some benchmarks, improvement in average stack walking speed can reach 10% as measured by `features=stats`:

| Benchmark | Before | After | % |
| --- | --- | --- | --- |
| spring-petclinic | 13042 | 11656 | -10.6% |
| renaissance dotty | 18490 | 16322 | -11.7% |

### Related issues

https://github.com/async-profiler/async-profiler/pull/1662

### Motivation and context

This change was inspired by @fandreuz's PR: https://github.com/async-profiler/async-profiler/pull/1662
However, what I'm not happy about in that implementation is that it makes `isType` methods stateful. I found that the actual performance issue was not in the string comparison itself, but rather in an indirect call to a standard library function. I went further and replaced `isInterpreter` with a single pointer comparison, since HotSpot has exactly one Code Blob with the name `Interpreter`.

---

### DIFF
--- a/src/vmStructs.h
+++ b/src/vmStructs.h
@@ -510,7 +510,7 @@ class NMethod : VMStructs {
 
     bool isNMethod() {
         const char* n = name();
-        return n != NULL && n[0] == 'n' && (startsWith(n, "nmethod\0") || startsWith(n, "native nmethod\0"));
+        return n != NULL && (startsWith(n, "nmethod\0") || startsWith(n, "native nmethod\0"));
     }
 
     bool isStub() {


src/stackWalker.cpp
```diff
@@ -280,44 +280,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                 anchor = NULL;
             }
 
-            if (nm->isNMethod()) {
```
apangin, line 283: No actual changes; just swapped `nm->isInterpreter()` and `nm->isNMethod()` branches.

PR #1702, opened 2026-03-19
Fix null vaddr_diff in ELF program header address resolution

jbachorik:
### Description

Fix undefined behavior in `ElfParser::at(ElfProgramHeader*)` when `_vaddr_diff` is NULL.

For position-independent ELF binaries whose first `PT_LOAD` segment has `p_vaddr == 0` and the library is loaded at base address 0, `_vaddr_diff` computes as `NULL` (i.e. `0 - 0`). The existing code then performs `NULL + pheader->p_vaddr`, which is undefined behavior per the C++ standard (pointer arithmetic on null pointers).

The fix handles the `_vaddr_diff == NULL` case by falling back to treating `p_vaddr` as an absolute virtual address, matching the `ET_EXEC` code path.

### Related issues

None.

### Motivation and context

Discovered during AddressSanitizer testing in [DataDog/java-profiler](https://github.com/DataDog/java-profiler). While the UB is currently benign on most platforms (adding 0 to a null pointer), it is flagged by sanitizers and could theoretically be optimized away by a sufficiently aggressive compiler.

### How has this been tested?

- Builds successfully with `make` on macOS (arm64) and Linux (x86_64)
- Verified under AddressSanitizer — the UB warning on `_vaddr_diff + pheader->p_vaddr` is eliminated
- Existing async-profiler test suite passes

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -243,7 +243,10 @@ class ElfParser {
     }
 
     const char* at(ElfProgramHeader* pheader) {
-        return _header->e_type == ET_EXEC ? (const char*)pheader->p_vaddr : _vaddr_diff + pheader->p_vaddr;
+        if (_header->e_type == ET_EXEC) {
+            return (const char*)pheader->p_vaddr;
+        }
+        return _vaddr_diff == NULL ? (const char*)pheader->p_vaddr : _vaddr_diff + pheader->p_vaddr;
     }
 
     const char* base() {


apangin:
Hi Jaroslav, thank you for raising the issue.

As a matter of fact, `base + offset` is UB in C++ for any `base` that does not belong to an array object, not just nullptr. To be conformant, async-profiler needs to deal with `uintptr_t` everywhere, but that would be a mess.

It turns out that async-profiler uses `base + offset` pattern (where `base` can be nullptr) in many more places. If AddressSanitizer detects it in one place but not in other - this is more like a problem with a tool, not with async-profiler. I don't mind fixing this particular place, but what I don't want is to get a series of follow-up PRs solely for making analyzers happy.

apangin:
Let me close the issue due no activity in last 6 months.
Feel free to reopen once if you have updates.

PR #1704, opened 2026-03-20
Unify uses of gcc atomics

apangin:
### Description

- Make `atomicInc`/`atomicDec`/`loadAcquire`/`storeRelease` template functions so that they can be used with various types.
- Replace inline uses of `__atomic_load_n`/`__atomic_store_n` with the above templates.

### Motivation and context

Make code a little bit cleaner; provide atomic operations for a broader range of integral types without `std::atomic` boilerplate.

---

### DIFF
--- a/src/cpuEngine.cpp
+++ b/src/cpuEngine.cpp
@@ -85,7 +85,7 @@ void CpuEngine::enableThreadHook() {
 
 void CpuEngine::disableThreadHook() {
     *_pthread_entry = (void*)pthread_setspecific;
-    storeRelease(_current, NULL);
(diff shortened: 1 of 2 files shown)


src/lookup.cpp
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <string.h>
```
Baraa-Hasheesh, line 6: was this import added intentionally?

apangin: Yes, one of my previous commits broke compilation with `MERGE=false` on Linux. I found this when validating this PR.

PR #1705, opened 2026-03-22
An option to limit size of the call trace storage

apangin:
### Description

Add `--memlimit` option to limit the amount of memory that CallTraceStorage can use.

### Motivation and context

Async-profiler's internal storage for stack traces (CallTraceStorage) consists of two structures:
- open-address hash tables for mapping stack trace fingerprints to stack trace contents;
- chunked array for storing stack frames.

Both structures can grow indefinitely. In extreme cases (e.g. deep recursion with high variation), this may cause rapid growth of async-profiler's memory consumption. This PR introduces a new option that limits such memory growth. Once this limit is reached, no new stack traces will be recorded. A synthetic stack trace with a single `storage_overflow` frame will be used instead. Samples that share one of the previously collected stacks will remain valid.

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -643,6 +643,7 @@ class Recording {
         writeIntSetting(buf, T_ACTIVE_RECORDING, "jfropts", args._jfr_options);
         writeIntSetting(buf, T_ACTIVE_RECORDING, "chunksize", args._chunk_size);
         writeIntSetting(buf, T_ACTIVE_RECORDING, "chunktime", args._chunk_time);
+        writeIntSetting(buf, T_ACTIVE_RECORDING, "memlimit", args._mem_limit);
 
         char str[256];
         writeStringSetting(buf, T_ACTIVE_RECORDING, "features", getFeaturesString(str, sizeof(str), args._features));


src/callTraceStorage.cpp
```diff
@@ -241,6 +243,12 @@ u32 CallTraceStorage::put(int num_frames, ASGCT_CallFrame* frames, u64 counter)
 
     while (keys[slot] != hash) {
         if (keys[slot] == 0) {
+            if (usedMemory() > _mem_limit) {
```
Baraa-Hasheesh, line 246: nit, after this point there is no value in unwinding the stacks for any sample, what would you think about adding a check on the `recordSample` side to skip stack unwinding after this happens one time?

Baraa-Hasheesh, line 246: This doesn't look right to me

Currently we allocate memory in chunks of around 8MB

Let's say we have a limit of 10MB, the new chunk will be allocated once we use 4MB of data 

8MB -> 16MB

So technically speaking the profiler will stop processing new stacks after using 4MB of storage

```
void* LinearAllocator::alloc(size_t size) {
    Chunk* chunk = _tail;

    do {
        // Fast path: bump a pointer with CAS
        for (size_t offs = chunk->offs; offs + size <= _chunk_size; offs = chunk->offs) {
            if (__sync_bool_compare_and_swap(&chunk->offs, offs, offs + size)) {
                if (_chunk_size / 2 - offs < size) {
                    // Stepped over a middle of the chunk - it's time to prepare a new one
                    reserveChunk(chunk);
                }
                return (char*)chunk + offs;
            }
        }
    } while ((chunk = getNextChunk(chunk)) != NULL);

    return NULL;
}
```

src/arguments.cpp
```diff
@@ -166,6 +166,9 @@ Error Arguments::parse(const char* args) {
                     msg = "Invalid loop duration";
                 }
 
+            CASE("memlimit")
+                _mem_limit = value == NULL ? 0 : parseUnits(value, BYTES);
```
Baraa-Hasheesh, line 170: shouldn't this be 10MB by default & we should add checks against this ?

-> `The lowest possible limit is 10 MB`

apangin: `CallTraceStorage` is unlimited by default. `0` here means no limit.
Values lower than 10M are not useful in practice, but can be set to confirm that the mechanism works, i.e., no stack traces are collected and memory usage does not increase.

src/callTraceStorage.cpp
apangin: I can't agree there is no value. Stack traces are not unique: after memory limit is reached, no new stack traces will be recorded, but new samples may still refer to previously collected stacks. For example, a deep recursion may build up only 10% of the overall profile, while the remaining 90% are comprised of the repeated stacks.
My first intention was to stop profiler after reaching the memory limit, but I realized we could miss lots of useful data in this case: for instance, MethodTrace events that often share a narrow set of repetitive stack traces.

apangin: Yes, memory estimation is not accurate here, but it does not need to be. The main purpose of the new option is to prevent infinite growth of the CallTraceStorage. Our typical limit will be around 100 MB. Whether it will stop growing at 96 MB or at 104 MB is not that important in practice.
Making estimation more accurate is possible, but I'd like to keep the code simple and fast.

apangin: OK, I made it a bit more accurate by incrementing memory counter after _switching_ to a new chunk rather than at allocation time. This allows using the whole 8 MB chunk before triggering the limit.
Sub-chunk granularity is out of consideration.

src/arguments.cpp
Baraa-Hasheesh: I think it's best to add a default value here rather than 0 (unlimited)

For example if I run profiler with agent/API mode with `start,event=cpu, memlimit` I would expect the profiler to not allow the `CallTraceStorage` to grow infinitely & use some default value to limit it

what do you think?

apangin: There is no obvious default value that fits all, it's better to set the limit explicitly.
`memlimit` without a value has no special meaning; I just made it a synonym for `memlimit=0` similarly to other options.

PR #1707, opened 2026-03-25
Workaround for JFR shutdown race

apangin:
### Description

When async-profiler tries to stop `jfrsync` recording, it should account for cases when the same recording is being stopped concurrently (e.g. by the JFR shutdown hook). Otherwise async-profiler may end up overwriting a not-yet-finalized JFR chunk.

### Related issues

This seems to be a OpenJDK issue that [dumping](https://github.com/openjdk/jdk/blob/cf424480f42ac220adee7034e0319cee0e9039db/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecording.java#L173) happens outside the state lock when the recording has already switched to the `STOPPED` state.

### Motivation and context

We've seen `jfrsync` recordings in production where an async-profiler chunk overlaps a JFR chunk.
All these cases are observed during JVM shutdown. There is apparently a race between async-profiler and JFR related to JVM shutdown, which I'm addressing here.

### How has this been tested?

Added an artificial delay at [this line](https://github.com/openjdk/jdk/blob/cf424480f42ac220adee7034e0319cee0e9039db/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecording.java#L167) in OpenJDK code to exercise the race.

---


src/helper/one/profiler/JfrSync.java
```diff
@@ -77,13 +78,24 @@ public static void start(String fileName, String settings, int eventMask) throws
...
             // Disable state change notification before stopping
             masterRecording = null;
-            recording.stop();
+            try {
+                recording.stop();
+            } catch (IllegalStateException e) {
+                // Workaround the JDK issue: JFR shutdown hook may stop the recording concurrently
+                // then populate the target file outside the state lock.
+                // Once the file is completely written, the recording state is changed to CLOSED.
+                for (int pause = 10; recording.getState() != RecordingState.CLOSED && pause < 1000; pause *= 2) {
```
Baraa-Hasheesh, line 92: nit, this in worst case (most likely will not happen) might cause the JVM shutdown to be delayed by a couple of seconds

src/flightRecorder.cpp
```diff
@@ -1411,7 +1411,9 @@ Error FlightRecorder::startMasterRecording(Arguments& args, const char* filename
 
 void FlightRecorder::stopMasterRecording() {
     JNIEnv* env = VM::jni();
-    env->CallStaticVoidMethod(_jfr_sync_class, _stop_method);
+    if (env->CallStaticBooleanMethod(_jfr_sync_class, _stop_method) == JNI_FALSE) {
+        Log::warn("Failed to stop JFR recording");
```
Baraa-Hasheesh, line 1415: In this case wouldn't it be best not to `appendRecording` as the JFR status isn't fully deterministic in this case

src/helper/one/profiler/JfrSync.java
apangin: I don't think this is the case. If we've got here, this means another thread is in the process of stopping/closing the same recording.

src/flightRecorder.cpp
apangin: Yeah, I considered that option too, but decided to stay with a simpler approach for two reasons:
1. Just because it's less changes. That aligns with the overall project philosophy not to over-focus on rare edge cases.
2. The more information, the better. If we don't append async-profiler chunk, consistency of JFR part is still undefined: we can end up with an empty file, a partially written file, or a complete JFR recording. If, however, we append an async-profiler chunk, there is a high chance that we'll get at least async-profiler's part right. Yes, the whole file may not be still readable with standard tools, but it should be possible to extract async-profiler chunk manually - that's exactly how I inspected broken recordings in our system.

Issue #1653, opened 2026-01-12
RecoveryTests.currentTimeMillis sometimes fails on linux-arm64-corretto11


apangin:
### Describe the bug

`RecoveryTests.currentTimeMillis` test is unstable because vDSO symbols are not resolved correctly.

E.g. https://github.com/async-profiler/async-profiler/actions/runs/20931738800/job/60144232325

### Expected vs. actual behavior

**Expected:**

There should be less than 1% of `[unknown]`, `[break_*]` and `[vdso]` frames.

**Actual:**
```
[break_stack_range];[unknown];__kernel_gettimeofday 2
__kernel_gettimeofday;[vdso] 147
```

### Reproduction Steps

GHA integration tests

### Async-profiler version

master

### Environment details

Linux ARM64 / Corretto 11

apangin:
Pushed a temporary workaround.
Still need to investigate the root cause.

Baraa-Hasheesh:
I noticed a failure on macOS https://github.com/Baraa-Hasheesh/async-profiler/actions/runs/20950626106/job/60202984247 

```
test/recovery/TimeLoop.main 3
thread_start;_pthread_start;thread_native_entry;Thread::call_run;ConcurrentGCThread::run;G1YoungRemSetSamplingThread::run_service;G1YoungRemSetSamplingThread::sleep_before_next_cycle;Monitor::wait;Monitor::IWait;os::PlatformEvent::park;__psynch_cvwait 1
test/recovery/TimeLoop.main;gettimeofday 2
test/recovery/TimeLoop.main;os::javaTimeMillis;gettimeofday;mach_absolute_time 32
test/recovery/TimeLoop.main;os::javaTimeMillis 2
test/recovery/TimeLoop.main;os::javaTimeMillis;gettimeofday 1
os::javaTimeMillis 1
thread_start;_pthread_start;thread_native_entry;Thread::call_run;WatcherThread::run;WatcherThread::sleep;Monitor::wait;Monitor::IWait;os::PlatformEvent::park;__psynch_cvwait 2
thread_start;_pthread_start;thread_native_entry;Thread::call_run;WatcherThread::run;WatcherThread::sleep;Monitor::wait;Thread::muxAcquire 1
[...]

PR #1708, opened 2026-03-25
Reattempt stack walking using Framepointer in case link register fails to unwind the stack  

Baraa-Hasheesh:
### Description
In some edge cases on ARM (vdso for example) it's possible for the link register to fail in unwinding the stack, this happen in 2 cases 

* Link register results in `pc == prev_pc`
* multiple frames that have to use link register, i.e. `f->pc_off == DW_LINK_REGISTER && depth > 1`

In this change we instruct the stack walker to re-attempt stack walking using FP as a final attempt to recover to a sane state

### Related issues
#1653

### Motivation and context
Enhance stack walking for ARM

### How has this been tested?
make test

---

### DIFF
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -424,6 +424,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
         }
 
         uintptr_t prev_fp = fp;
+        const void* prev_pc = pc;
         FrameDesc* f = native_lib != NULL ? native_lib->findFrameDesc(pc) : &FrameDesc::default_frame;
 
         unwind_frame:
@@ -449,7 +450,6 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
             break;
         }
 
-        const void* prev_pc = pc;
         if (f->fp_off & DW_PC_OFFSET) {
             pc = (const char*)pc + (f->fp_off >> 1);
         } else {
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -423,8 +423,10 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
             fillFrame(frames[depth++], BCI_NATIVE_FRAME, method_name);
         }
 
+        uintptr_t prev_fp = fp;
         FrameDesc* f = native_lib != NULL ? native_lib->findFrameDesc(pc) : &FrameDesc::default_frame;
 
+        unwind_frame:
         u8 cfa_reg = (u8)f->cfa;
         int cfa_off = f->cfa >> 8;
         if (cfa_reg == DW_REG_SP) {
@@ -457,10 +459,15 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
 
             if (EMPTY_FRAME_SIZE > 0 || f->pc_off != DW_LINK_REGISTER) {
                 pc = stripPointer(*(void**)(sp + f->pc_off));
-            } else if (depth == 1) {
-                pc = (const void*)frame.link();
             } else {
-                break;
+                if (depth == 1) pc = (const void*)frame.link();
+
+                if (pc == prev_pc) {
+                    f = &FrameDesc::default_frame;
+                    sp = prev_sp;
+                    fp = prev_fp;
+                    goto unwind_frame;
+                }
             }
 
             if (EMPTY_FRAME_SIZE == 0 && cfa_off == 0 && f->fp_off != DW_SAME_FP) {


src/stackWalker.cpp
```diff
@@ -457,10 +459,15 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
 
             if (EMPTY_FRAME_SIZE > 0 || f->pc_off != DW_LINK_REGISTER) {
                 pc = stripPointer(*(void**)(sp + f->pc_off));
-            } else if (depth == 1) {
-                pc = (const void*)frame.link();
             } else {
-                break;
+                if (depth == 1) pc = (const void*)frame.link();
+
+                if (pc == prev_pc) {
```
apangin, line 465: Can you provide a bit more details when this happens? I.e., what is a frame layout and what instruction is being executed? I'm trying to understand the pattern here.

Baraa-Hasheesh: There are two cases here both are from VDSO 

First normal case of having multiple frames inside the VDSO such as the following trace 

```
<Signal Handler>
#4  0x0000fffff7ffc534 in ?? ()
#5  0x0000fffff7ffc7a0 in __kernel_clock_gettime () <--- stack walking will fail here 
#6  0x0000fffff7f0099c in __GI___clock_gettime (clock_id=0, tp=0xfffff6797ea0) at ../sysdeps/unix/clock_gettime.c:115
#7  0x0000fffff74e56a4 in os::javaTimeMillis() () from /local/home/bhashesh/.sdkman/candidates/java/25-amzn/lib/server/libjvm.so
#8  0x0000ffffe015e950 in ?? ()
#9  0x00000000fcf36590 in ?? ()
```

We can only use link register on top frame at depth == 1 & fail other wise, so the unwinding breaks on the depth == 2

------

Second case which is quite sharp but can still happen  is the following 

```
<Signal Handler>
#4  0x0000fffff7ffc7a0 in __kernel_clock_gettime ()
#5  0x0000fffff7f0099c in __GI___clock_gettime (clock_id=0, tp=0xfffff6797ea0) at ../sysdeps/unix/clock_gettime.c:115
#6  0x0000fffff74e56a4 in os::javaTimeMillis() () from /local/home/bhashesh/.sdkman/candidates/java/25-amzn/lib/server/libjvm.so
#7  0x0000ffffe015eb50 in ?? ()


(gdb) disas 0x0000fffff7ffc7a0
Dump of assembler code for function __kernel_clock_gettime:
   0x0000fffff7ffc794 <+0>:     stp     x29, x30, [sp, #-16]!
   0x0000fffff7ffc798 <+4>:     mov     x29, sp
   0x0000fffff7ffc79c <+8>:     bl      0xfffff7ffc500
   0x0000fffff7ffc7a0 <+12>:    ldp     x29, x30, [sp], #16 <- Sample happens here 
   0x0000fffff7ffc7a4 <+16>:    ret
```

In the second case, the sample happens exactly after returning from the function defined at address `0xfffff7ffc500` & at this point `r30` still points to the return address @ __kernel_clock_gettime 

Recovery from Link register fails as link register has outdated data (`prev_sp == sp && pc == prev_pc`)

--------- 

I did check the structure of functions defined inside VDSO & noticed that they are mixed, some have Framepointer setup while others don't

For example __kernel_clock_gettime has framepointer setup, while function at 0xfffff7ffc500 doesn't 

```
(gdb) disassemble (void*)0xfffff7ffc500, (void*)0xfffff7ffc5ff
Dump of assembler code from 0xfffff7ffc500 to 0xfffff7ffc5ff:
   0x0000fffff7ffc500:  cmp     w0, #0xf
   0x0000fffff7ffc504:  b.hi    0xfffff7ffc5c0  // b.pmore
   0x0000fffff7ffc508:  mov     w2, #0x1                        // #1
   0x0000fffff7ffc50c:  mov     w3, #0x883                      // #2179
   0x0000fffff7ffc510:  lsl     w2, w2, w0
   0x0000fffff7ffc514:  adr     x4, 0xfffff7ffa000
   0x0000fffff7ffc518:  ands    w3, w2, w3
   0x0000fffff7ffc51c:  b.eq    0xfffff7ffc5e0  // b.none
```

Baraa-Hasheesh: Note here the true problem is that in VDSO we always use emptyFrame which is incorrect 

Generally using FP unwinding for arm should work better than what is happening currently but that's not guaranteed 

The only perfect solution would be to create a custom unwinding table for VDSO but it will be quite complex for the benefit it provides 

In general here it's my opinion that using FP always on ARM for VDSO is a good choice with the caveat that sometimes we will skip some top frames

This change is a different way to reattempt stack walking when link register fails (pc == prev_pc) to give a best effort of not skipping some of the top frames

apangin:
The idea to retrying unwinding with the deault_frame when empty_frame fails is nice.
I took your suggestion but slight rewrote it:
- I only retry a truly empty frame; in other cases when unwinding using link register fails, I abort stack walking as before, just to stay safe.
- Reduced the scope of `goto` and eliminated storing/restoring of `prev_fp` and `prev_sp`.

Here is an updated version: https://github.com/async-profiler/async-profiler/compare/master...vdso
WDYT?

Baraa-Hasheesh:
> The idea to retrying unwinding with the deault_frame when empty_frame fails is nice. I took your suggestion but slight rewrote it:
> 
> * I only retry a truly empty frame; in other cases when unwinding using link register fails, I abort stack walking as before, just to stay safe.
> * Reduced the scope of `goto` and eliminated storing/restoring of `prev_fp` and `prev_sp`.
> 
> Here is an updated version: [master...vdso](https://github.com/async-profiler/async-profiler/compare/master...vdso) WDYT?

I have some concerns

When we calculate the next frame pointer value we use the following formula `fp = *(uintptr_t*)(sp + f->fp_off);` 
When unwinding using default frame this value will be effectively the following 

`fp = *(uintptr_t*)(fp + f->fp_off);`  -> `fp = *(uintptr_t*)(fp);` 

With the suggested change it will be something like this 

`fp = *(uintptr_t*)(prev_sp + (pref_f->cfa >> 8));` 

Also `sp` is never really assigned the updated value from `fp` => `sp = fp + cfa_off;` , this in turn will effect PC caluclations `pc = stripPointer(*(void**)(sp + f->pc_off));` 

-----------

Second concern is something I noticed while testing the performance of the above change, is that the retry will happen in cases where it obviously shouldn't, in particular at the bottom of the stack 

I would suggest running a small app & observe when this retry happen you will observe that it does happen on the root frame of `thread_start` (Tested on Linux arm)

apangin:
Good catch, thanks. I changed the condition to retry only if `fp == sp`. When the condition fails - it's not the default gcc frame, so retrying does not make sense.
I also moved the label to your original location. It does not make difference in this particular case, but is less error-prone.
https://github.com/async-profiler/async-profiler/compare/master...vdso

Baraa-Hasheesh:
@apangin The new changes look good to me

apangin:
Thank you for the review. I merged the change co-authored with you.

Issue #1272, opened 2025-04-29
Heatmap usability improvements


NN---:
### Describe the feature

First, thanks for this great feature !

The heatmap is great, but the default view is not showing all the time range.
In addition there is no simple button to show the whole time range.

Please consider adding these features.

### Use Case

Working with heatmap.

### Proposed Solution

Default show all time range, button to show all time range.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1709, opened 2026-03-26
Heatmap usability - show full time range by default and add Show All button

rlaope:
### Description

Improve heatmap usability by showing the full time range on initial load and adding a toolbar button to reset the
view.

> If you find this feature useful, I'd be happy to see it merged.

Changes to `src/res/heatmap.html` only (+14 lines):
- Add 4th zoom level (`1 min` per square, `1 hr` per column) to `zoomToGroupSize` — the matching `zoomToText`
entry already existed but had no corresponding zoom level
- Add "All" button in the right-side toolbar that resets zoom to fit the entire time range, clears selection, and
renders the full flame graph
- With the additional zoom level, `calculateInitialZoom()` naturally selects a zoom that fits all data on initial
load

### Related issues

#1272

### Motivation and context

The heatmap default view does not show the full time range for longer profiling sessions (several minutes or
more), requiring horizontal scrolling. There is also no button to reset the view to show the entire time range
after zooming in.

**Before** — 5-minute profiling session requires horizontal scrolling:

image

**After** — full time range visible by default, with "All" button in toolbar:

image
image

**New 4th zoom level** — `1 hr : 1 min` for long profiling sessions:

image

### How has this been tested?

- Built with `make` on macOS (arm64, JDK 17)
- Generated heatmaps from 30-second and 5-minute JFR recordings
- Verified default view fits entire time range without scrolling
- Verified sidebar click cycles through 4 zoom levels (20ms → 1s → 5s → 1min)
- Verified "All" button resets zoom, clears selection, renders full flame graph
- Verified existing interactions (click, Shift+drag, Ctrl+click diff, Ctrl+F search) work normally

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0
license].


apangin:
Adding more buttons does not improve usability; the toolbar already has too many buttons.
4th zoom level is a controversial change; to me, it's an anti-feature, as recordings for analyzing on a single page almost never exceed 24h span. In any case, this change is out of scope of the original issue.

PR #1710, opened 2026-03-26
Fix flush pending WALL_BATCH samples when thread is removed from filter

### LINKED ISSUE #1249: Inaccurate Sample Count in WALL_BATCH Mode (v4.0)
### Describe the bug

First, thank you for your great work on async-profiler!

We've encountered an issue with sample count accuracy in WALL_BATCH mode and would like to report it. 

When using WALL_BATCH mode, the number of samples collected varies significantly and appears inconsistent. The sample counts fluctuate between extremely high and extremely low values for similar operations.

 The sample counts vary dramatically between runs:
- Maximum samples observed: 400+
- Minimum samples observed: 2
- Expected samples for 2s duration at 10ms interval: ~200 (with some expected variance)

### Reproduction Steps

1. Sampling configuration:
   - Sampling period: 10ms
   - Duration: 1 minute continuous sampling
   - Mode: WALL_BATCH

2. Test operation:
    - Call `addThread` to add the thread before execution
   - Execute a code segment that sleeps for 2 seconds
   - Call `removeThread` to remove the thread after execution

- This behavior is consistently reproducible in our environment
- The significant variance occurs even for identical operations
- We've verified that the thread is properly added/removed each cycle

### Async-profiler version

4.0

### Environment details

os:  macos sonoma
jdk: 1.8

rlaope:
Thank you for maintaining such a great project. I reviewed this issue and implemented the fix following the direction you suggested. I've verified the results with a reproducer test. If this looks like the right approach, please consider reviewing and merging.

### Description

Flush pending WALL_BATCH idle samples when a thread is removed from the thread filter via `removeThread()`.

In `wallClock.cpp` `timerLoop()`, when a thread is rejected by the thread filter (line 208), the loop skips it
with `continue`. However, if the thread had accumulated idle samples in `tss.counter` (WALL_BATCH optimization),
those samples were never flushed — they remained in `thread_sleep_state` until `stop()` was called.

This fix checks for pending batch samples at the filter rejection point and flushes them before skipping.

**Change:** `src/wallClock.cpp` only (+7 lines)

```cpp
if (thread_filter_enabled && !thread_filter->accept(thread_id)) {
    // NEW: flush pending batch for threads removed from filter
    if (mode == WALL_BATCH) {
        ThreadSleepMap::iterator it = thread_sleep_state.find(thread_id);
        if (it != thread_sleep_state.end() && it->second.counter != 0) {
            recordWallClock(it->second, THREAD_SLEEPING, thread_id);
            // thread_sleep_state.erase(it); # revise
            it->second.counter = 0;
        }
    }
    continue;
}
```

### Related issues

Fixes #1249

### Motivation and context

In WALL_BATCH mode, sleeping threads are not signaled every tick. Instead, a counter (`tss.counter`) accumulates
and is flushed when the thread wakes up or the counter reaches `MAX_IDLE_BATCH` (1000). This is a performance
optimization.

The problem: when `removeThread()` is called, the thread is removed from the filter bitmap. On subsequent timer
loop iterations, the thread is skipped at the filter check — but its pending counter is never flushed. If `dump`
is called (rather than `stop`), these samples are permanently lost.

### How has this been tested?

Built on macOS arm64 with JDK 17. Tested with the following reproducer:

```java
import one.profiler.AsyncProfiler;
import java.util.concurrent.*;

public class WallBatchBugTest {

    static int countSamples(AsyncProfiler profiler) throws Exception {
        String collapsed = profiler.execute("collapsed");
        int total = 0;
        for (String line : collapsed.split("\n")) {
            line = line.trim();
            if (line.isEmpty()) continue;
            int lastSpace = line.lastIndexOf(' ');
            if (lastSpace > 0) {
                try {
                    total += Integer.parseInt(line.substring(lastSpace + 1));
                } catch (NumberFormatException ignored) {}
            }
        }
        return total;
    }

        profiler.execute("start,wall=10ms,filter,threads");

        ExecutorService executor = Executors.newFixedThreadPool(4);
        CountDownLatch allDone = new CountDownLatch(10);

            executor.submit(() -> {
                profiler.addThread(null);
                try { Thread.sleep(200); } catch (InterruptedException ignored) {}
                profiler.removeThread(null);
                allDone.countDown();
            });
        }

        allDone.await();
        Thread.sleep(300);

        int dumpSamples = countSamples(profiler);
        System.out.println("Samples at dump time: " + dumpSamples);
        System.out.println("Expected: ~200 (10 tasks * 200ms / 10ms)");
        System.out.println(dumpSamples >= 180 ? "PASS" : "FAIL (samples lost!)");

        profiler.execute("stop");
        executor.shutdown();
    }
}
```

**Results (10 tasks × 200ms sleep, 10ms interval, expected ~200 samples):**

| | Samples at dump time | Result |
|---|---|---|
| **Before fix** | **125** | FAIL — ~75 samples lost |
| **After fix** | **204** | PASS |

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0
license].

### DIFF
--- a/src/wallClock.cpp
+++ b/src/wallClock.cpp
@@ -206,6 +206,13 @@ void WallClock::timerLoop() {
                 continue;
             }
             if (thread_filter_enabled && !thread_filter->accept(thread_id)) {
+                if (mode == WALL_BATCH) {
+                    ThreadSleepMap::iterator it = thread_sleep_state.find(thread_id);
+                    if (it != thread_sleep_state.end() && it->second.counter != 0) {
+                        recordWallClock(it->second, THREAD_SLEEPING, thread_id);
+                        it->second.counter = 0;
+                    }
+                }
                 continue;
             }
 


apangin:
Unfortunately, this is a wrong solution to a wrong problem. I would like to see a more systematic approach that addresses the common problem as discussed in the original ticket, not a particular edge case.

apangin:
Sorry, I should have clarified what I meant. The proposed PR does not solve the root cause of the issue, which has nothing to do with filtering. E.g. if you run `dump` without stopping profiler (even when no filter is set), there will be still the same problem.

There can be multiple approaches to solve the problem: for example, to flush pending WallClock events periodically or to do that at fixed points like before dumping and at the end of every JFR chunk. To choose the right one, I'd like to see analysis with use cases, limitations, and performance impact of each approach. Simplicity/efficiency of the implementation is also an important factor.

rlaope:
Ah... I see. That was my mistake. Thank you for explaining.

PR #1713, opened 2026-03-27
Fix mmap failure check and munmap size in OS::copyFile on macOS

### LINKED ISSUE #1712: Incorrect mmap failure check in OS::copyFile on macOS
### Describe the bug

`OS::copyFile` in `os_macos.cpp` checks the `mmap` return value against `NULL`, but `mmap` returns `MAP_FAILED` on failure. The NULL check never catches the failure, so `write()` is called with an invalid pointer, silently corrupting the JFR recording.

The same file correctly uses `MAP_FAILED` in `OS::safeAlloc`.

Additionally, `munmap` is called with the mutated `offset` instead of the original mapped size.

When mmap fails, `copyFile` should return early. Instead, it proceeds with an invalid pointer, and the resulting JFR file is corrupted.

### Reproduction Steps

Use `--jfrsync` on macOS under conditions where mmap fails.

### Async-profiler version

4.3 (master, commit `7e92b5c`)

macOS only — `OS::copyFile` in `os_macos.cpp` is macOS-specific.

rlaope:
Fix two bugs in `OS::copyFile` (`src/os_macos.cpp`):

1. `mmap` returns `MAP_FAILED` on failure, not `NULL`. The `NULL` check never catches mmap failure, so `write()` is called with an invalid pointer, silently corrupting the JFR recording.

2. `munmap` uses the mutated `offset` instead of the original mapped size, causing a partial unmap.

Fixes #1712

### DIFF
--- a/src/os_macos.cpp
+++ b/src/os_macos.cpp
@@ -374,8 +374,9 @@ int OS::createMemoryFile(const char* name) {
 }
 
 void OS::copyFile(int src_fd, int dst_fd, off_t offset, size_t size) {
-    char* buf = (char*)mmap(NULL, size + offset, PROT_READ, MAP_PRIVATE, src_fd, 0);
-    if (buf == NULL) {
+    size_t map_size = size + offset;
+    char* buf = (char*)mmap(NULL, map_size, PROT_READ, MAP_PRIVATE, src_fd, 0);
+    if (buf == MAP_FAILED) {
         return;
     }
 
@@ -388,7 +389,7 @@ void OS::copyFile(int src_fd, int dst_fd, off_t offset, size_t size) {
         size -= (size_t)bytes;
     }
 
-    munmap(buf, offset);
+    munmap(buf, map_size);
 }
 
 void OS::freePageCache(int fd, off_t start_offset) {


apangin:
Thank you for contributing to async-profiler. There is indeed a bug, and the fix makes total sense.

But I want to ask you: please **avoid pasting AI-generated rubbish** in issues, PR descriptions, comments, and of course, in the code.
- Such content is overly verbose and hard to read.
- GenAI tends to explain trivial things (like that -1 is 0xFFFFFFFFFFFFFFFF and not NULL) but misses the most important part, e.g. why a recording gets corrupted in one case but not in another, even though the very first `write` fails.
- It hallucinates. In particular, this fix is macOS-only, but the Motivation section talks about cgroup and vm.max_map_count, which are Linux-specific concepts.

Please keep communication concise, accurate, and to the point.
I will de-prioritize reviewing any AI-generated content.
Thank you for your understanding.

rlaope:
I see. thank you

apangin:
> misses the most important part, e.g. why a recording gets corrupted in one case but not in another, even though the very first write fails.

OK, it's getting more interesting. I was really puzzled by the above effect so I did some research to realize that you've found a bug not just in async-profiler, but in the _macOS kernel!_

If I attempt to write a knowingly invalid buffer such as
```
write(fd, (void*)-1, 65536);
```
to a file, the call will expectedly fail with `errno = EFAULT`, but the file will be _actually updated_ with 65536 null bytes.
There is no such a bug on Linux.

rlaope:
I've cleaned up the issue and PR descriptions — removed the verbose content and Linux-specific references.

That's a fascinating find — I didn't investigate why the corrupted bytes were written despite `EFAULT`. Should I add a note about this macOS kernel behavior to the issue?

PR #1722, opened 2026-04-08
Size comparison

benty-amzn:
### Description
Adds script and workflows to post a comment to pull requests with the size comparison for Linux and macos binaries included

### How has this been tested?

Ran the workflow on a test branch in my fork: https://github.com/benty-amzn/async-profiler/actions/runs/25021267551/job/73282029524

The table is reproduced here for convenience:

### linux-x64

| File | Base (KiB) | Treatment (KiB) | Delta (KiB) | % Change |
| --- | ---: | ---: | ---: | ---: |
| `bin/asprof` | 81.72 | 81.72 | +0.00 | +0.00% |
| `bin/jfrconv` | 133.73 | 133.73 | +0.00 | +0.00% |
| `lib/libasyncProfiler.so` | 590.01 | 590.01 | +0.00 | +0.00% |
### linux-arm64

| File | Base (KiB) | Treatment (KiB) | Delta (KiB) | % Change |
| --- | ---: | ---: | ---: | ---: |
| `bin/asprof` | 89.59 | 89.59 | +0.00 | +0.00% |
| `bin/jfrconv` | 133.73 | 133.73 | +0.00 | +0.00% |
| `lib/libasyncProfiler.so` | 609.29 | 609.29 | +0.00 | +0.00% |
### macos

| File | Base (KiB) | Treatment (KiB) | Delta (KiB) | % Change |
| --- | ---: | ---: | ---: | ---: |
| `bin/asprof` | 116.74 | 116.74 | +0.00 | +0.00% |
| `bin/jfrconv` | 133.73 | 133.73 | +0.00 | +0.00% |
| `lib/libasyncProfiler.dylib` | 991.34 | 991.34 | +0.00 | +0.00% |

---

### DIFF
--- a/scripts/compare-binary-sizes.sh
+++ b/scripts/compare-binary-sizes.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright The async-profiler authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+function list-archive() {
+    case "${1}" in
+        *.tar.gz) tar -tzvf "${1}" --wildcards '*/bin/*' '*/lib/*' | awk '$1 ~ /^-/ { print $NF, $3 }' ;;
+        *.zip)    zipinfo -l "${1}"           '*/bin/*' '*/lib/*'  | awk '$1 ~ /^-/ { print $NF, $4 }' ;;
+    esac | cut -d/ -f2- | sort
+}
+
+function main() {
+    if [[ ${#} -ne 2 ]]; then
+        printf 'Usage: compare-binary-sizes.sh <base-archive> <treatment-archive>\n' >&2
+        exit 1
+    fi
+
+    [[ ${1} =~ (linux-x64|linux-arm64|macos) ]]
+    printf '### %s\n\n| File | Base (KiB) | Treatment (KiB) | Delta (KiB) | %% Change |\n| --- | ---: | ---: | ---: | ---: |\n' "${BASH_REMATCH[1]}"
+    join -a1 -a2 -e0 -o '0,1.2,2.2' \
+         <(list-archive "${1}") \
+         <(list-archive "${2}") \
+        | awk '{
+            b = $2; t = $3; d = t - b; pct = b > 0 ? d * 100 / b : 0
+            printf "| `%s` | %.2f | %.2f | %+.2f | %+.2f%% |\n", $1, b/1024, t/1024, d/1024, pct
+        }'
+}
+
+[[ "${BASH_SOURCE[0]}" == "${0}" ]] && main "$@"
--- a/.github/workflows/compare-binary-sizes.yaml
+++ b/.github/workflows/compare-binary-sizes.yaml
@@ -1,0 +1,28 @@
+name: compare-binary-sizes
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+permissions:
+  contents: read
+  pull-requests: write
+  actions: read
+env:
+  GH_TOKEN: "${{ github.token }}"
+  REPO: "${{ github.repository }}"
+  PR_SHA: "${{ github.event.workflow_run.head_sha }}"
+jobs:
+  compare:
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          gh api "repos/${REPO}/contents/scripts/compare-binary-sizes.sh" -H 'Accept: application/vnd.github.raw' > compare.sh
+          read pr base_sha < <(gh api "repos/${REPO}/commits/${PR_SHA}/pulls" --jq '.[0] | "\(.number) \(.base.sha)"')
+          function find-archive() { find "${1}" -type f \( -name "async-profiler-*${2}.tar.gz" -o -name "async-profiler-*${2}.zip" \) ! -name '*-debug*' | head -n1; }
+          for p in linux-x64 linux-arm64 macos; do
+            gh run download -R "${REPO}" -n "async-profiler-${p}-${base_sha:0:7}" -D "base/${p}" 2>/dev/null || continue
+            gh run download -R "${REPO}" -n "async-profiler-${p}-${PR_SHA:0:7}"   -D "pr/${p}"   2>/dev/null || continue
+            bash compare.sh "$(find-archive "base/${p}" "${p}")" "$(find-archive "pr/${p}" "${p}")" >> report.md
+          done
+          gh pr comment "${pr}" -R "${REPO}" --body-file report.md --edit-last --create-if-none


Baraa-Hasheesh:
nit, this would always compare the current commit against the last public release 

Let's say I do a change that increases the binary size by 3% & that get's merged, after which I do a second merge that changes binary size by a further 3%, the above report will report the size change for the second commit as 6% 

This in turn would be misleading

benty-amzn:
Agreed, and that's part of the reason these are only informational comments and not any sort of blocking workflow.
I also didn't want to make this an extremely invasive change - we have more options available if we add an artifact to the build output, or spin up some infrastructure to track filesize <-> commit mappings.

As I see it, we have 3 relatively self-contained options:

- Compare against the latest release. This is what I've currently implemented, and has the flaws you call out in your comment.

- Compare against the latest nightly. This suffers from the same problems, just less - a recent change responsible for a size increase might not have triggered a nightly build yet. 

- Compare against the parent commit. The first problem with this is correctly identifying the parent, but the merge_base feels reasonable. This still isn't 100% proof against other changes getting merged before the size comparison is posted, and I don't think we can do it easily on Actions without adding additional build steps to get artifacts for the parent commit.

We could implement a combination as well, and post comparison against last release build and last nightly, which gives a more complete picture.

Baraa-Hasheesh:
> We could implement a combination as well, and post comparison against last release build and last nightly, which gives a more complete picture.

I think this would be nice

-------

Now this is my general opinion,
* it should be fine to remove LICENSE, README.md, include/asprof.h & bin/jfrconv from the comparisons

I believe the reason for this change was the concern that adding `c++ std` dependencies might inflate the binary size as those dependencies usually bring large amount of hidden code

LICENSE, README & include/asprof.h  are basically text files so generally speaking size won't increase that much for them & if it does it should be quite acceptable & fine

bin/jfrconv is basically a JAR with zero dependency on c++ std so it can be skipped 

In a matter of fact I would argue there is no need to check `bin/asprof` either as it's pure C with no dependencies on c++ libraries

* If the above holds true (Only check libasyncProfiler.so) I think the table can be simplified to something like this

| System | reference size | current size | diff % |
| ------- | ------- | ------- | -------- |
| aarch64 | X KB |   Y KB  | X % |
| x64 | X KB |   Y KB  | X % |
| macos | X KB |   Y KB  | X % |

lutkerd (approved):

lutkerd:
I agree with @Baraa-Hasheesh on a lot of the comments to include nightly builds in the comparison and to skip some of the text files in the tracking. I would still track asprof and the jar files, they can still grow or include additional dependencies at a later date, easy enough to track them now and it's the kind of thing we would likely forget to add if something changes within those files in the future.

I don't see anything blocking here and the improvements can be in separate PRs to keep this more contained and reviewable.

apangin:
It's all over-engineered, imo. 4 scripts that include and call each other in an unobvious way, along with a workflow that mixes inline bash with calls to these scripts, is too much for a relatively simple task. Try to keep it simple and straightforward. An illustrative example is where you parse .yml looking for hard-coded patterns to extract build commands - instead of just calling a single `make` command. Btw, the command from `build.yml` isn't good for local builds anway: for instance, you don't need to run coverage locally.

My overall stance to async-profiler development is - the less code, the better. Ideal solution is when a problem is solved with no code :)

So, I propose to give up "architecture" entirely and to re-implement the task in one script with minimum abstractions. I originally thought of just one workflow file, but if you want to run size comparison locally, I don't mind a extracting the common piece into a separate script. That said, I suggest not to overload this PR with stuff for local development - that can be added later.

apangin:
From a functional perspective, comparing size against the latest release is not very useful. Both other options you've mentioned ("nightly" build or a parent commit) make more sense. I would prefer the latter, but if it's more complicated, using nightly build is totally fine.

I also agree with previous comments about excluding text files, however, all of `libasyncProfiler.so`, `asprof`, `jfrconv` should be counted.

.github/workflows/binary-size-comparison.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
```
apangin: No `write` access in an untrusted context - definitely not when building an arbitrary third-party PR.
Fortunately, you can override permissions for a specific job, so grant `write` in the narrowest required scope.

scripts/compare-binary-sizes.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo ""
+    echo "Comparing \`${base_sha}\` against \`${treatment_sha}\`"
+    echo ""
+
+    local platform treatment_dir base_dir
+    for platform in linux-x64 linux-arm64 macos; do
+        treatment_dir="artifacts/async-profiler-${platform}-${treatment_sha}"
+        base_dir="artifacts/async-profiler-${platform}-${base_sha}"
+        bsdtar xf "${treatment_dir}"/* -C "${treatment_dir}"
+        bsdtar xf "${base_dir}"/* -C "${base_dir}"
```
apangin: No need to extract; it's enough to list contents of the archive

.github/workflows/binary-size-comparison.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  build-pr-linux-x64:
+    uses: ./.github/workflows/build.yml
```
apangin: It's not great to do so much redundant work: we build twice, and for each build we compile twice (another one for `coverage` target that we don't need here at all).
Is there a simple way to avoid that redundancy?

scripts/compare-binary-sizes.sh
```diff
@@ -0,0 +1,50 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+main() {
+    if [[ ${#} -ne 2 ]]; then
+        echo "Compare binary file sizes between 2 builds (treatment and base)"
+        echo "Prints a markdown report to stdout."
+        echo "Usage: compare-binary-sizes.sh <base-sha> <treatment-sha>"
+        echo "Expects build artifacts at: artifacts/async-profiler-{platform}-{sha7}/"
```
apangin: Would it be better to accept two filenames instead?
That will make the script more independent. I can then use it locally just for one platform I develop on and choose whatever baseline I want: major release, nightly, or another local build.
In this case, loop over platforms will be in yaml.

.github/workflows/binary-size-comparison.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+        with: { path: artifacts }
+      - name: Compare and comment
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          sudo apt-get -y install libarchive-tools
```
apangin: nit: I'd do this without bsdtar. For a small price of choosing between unzip/tar, this will slightly reduce build time and will allow running the script locally out-of-the-box.

.github/workflows/binary-size-comparison.yml
```diff
@@ -0,0 +1,19 @@
+name: Binary Size Comparison
+on: { workflow_run: { workflows: [CI], types: [completed] } }
+permissions: { contents: read, pull-requests: write, actions: read }
+env: { GH_TOKEN: "${{ github.token }}", REPO: "${{ github.repository }}", PR_SHA: "${{ github.event.workflow_run.head_sha }}" }
```
apangin: nit: I like concise code, but it's not about packing as mush as possible in one line )
In this case, it would be nicer to expand top-level maps.

lutkerd (approved):

.github/workflows/binary-size-comparison.yml
```diff
@@ -0,0 +1,19 @@
+name: Binary Size Comparison
```
apangin: Naming style differs from other workflows.
Also, I'd keep the workflow name consistent with the script name: `compare-binary-sizes`.

PR #1730, opened 2026-04-17
Support users setting output options for collapsed & flamegraph in the start command

Baraa-Hasheesh:
### Description
This change has two main goals:
* Enhance user experience when using async-profiler 
* Make the profiler behaviour more consistent for a given command 

**UX:** Currently certain flags are required at `stop` command to properly dump the output(collapsed & flamegraph) like `file` & `total`, this isn't that intuitive for the user 

With the new change the user can now fully front load everything in the `start` command for example `start,alloc,total,norm,file=output.html` & when wanting to stop the profiler, the user can simply call `stop` without the need to add additional flags here

**Consistency:** Currently if a user launches the profiler with the following command `start,alloc,file=output.collapsed`, the behaviour will be inconsistent depending on how the profiler is stopped for example:
- Profiler stopped via process death => File is dumped
- Profiler stopped via JfrRecording when using jfrsync => file is not dumped
- Profiler stopped via `stop` => file is not dumped

With this new change in all 3 cases the output file is dumped 

Note: This techincally solves a bug where the profiler won't dump the output in agent mode if the profiler was unintentionally stopped via JfrRecording (JfrSync_stopProfiler was called)

Note2: To avoid making the logic more complex than need be I decided on the following, only dump output specified in start command if & only if no output is given in the stop command, otherwise behaviour will default to previous behaviour of dumping output specified in stop command 

### Related issues
N/A

### Motivation and context
Make profiler behaviour more consistent & enhance UX by allowing front loading of dump options in start command 

### How has this been tested?
New test was added

---

### DIFF
--- a/src/flightRecorder.cpp
+++ b/src/flightRecorder.cpp
@@ -34,7 +34,7 @@
 INCLUDE_HELPER_CLASS(JFR_SYNC_NAME, JFR_SYNC_CLASS, "one/profiler/JfrSync")
 
 static void JNICALL JfrSync_stopProfiler(JNIEnv* env, jclass cls) {
-    Profiler::instance()->stop();
+    Profiler::instance()->expire(_global_args, false);
 }
 
 
--- a/src/javaApi.cpp
+++ b/src/javaApi.cpp
@@ -47,7 +47,7 @@ Java_one_profiler_AsyncProfiler_start0(JNIEnv* env, jobject unused, jstring even
 
 extern "C" DLLEXPORT void JNICALL
 Java_one_profiler_AsyncProfiler_stop0(JNIEnv* env, jobject unused) {
-    Error error = Profiler::instance()->stop();
+    Error error = Profiler::instance()->expire(_global_args, false);
 
     if (error) {
         throwNew(env, "java/lang/IllegalStateException", error.message());
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1541,17 +1541,19 @@ Error Profiler::runInternal(Arguments& args, Writer& out) {
             break;
         }
         case ACTION_STOP: {
-            Error error = stop();
-            if (args._output == OUTPUT_NONE) {
-                if (error) {
-                    return error;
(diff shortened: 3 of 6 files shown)


apangin:
Thank you for raising the issue.
I agree that the current behavior of the `stop` command may not be obvious to a new user. However, once you understand how collection and dumping work, everything becomes logical.

First of all, dumping results is a separate action from stopping the profiler. I don't want to merge them, because each has its own meaning and can be used independently. For example, you can dump results while the profiler is still running, or you can stop the profiler without dumping results. Of course, it is also possible to stop and dump at the same time, but it's not the `stop` command itself that writes the results. What triggers a dump is the format option (`collapsed`, `flamegraph`, etc.) or the `file` option, which automatically selects an output format.

Unlike other profilers, which usually dump results in just one format, async-profiler has supported multiple output formats from the beginning. In the past, I received many requests to allow dumping multiple views from a single profiling session. You can easily do this today: stop the profiler, then call `dump` with different options and filenames as many times as you want. In terms of consistency, it would actually seem less consistent to me to specify the first output file in the `start` command and all subsequent files in the `dump`/`stop` commands.

There are also scenarios in which a developer needs to pause and resume recording multiple times before dumping the final results, for example, to skip non-essential phases of job execution. Currently, `stop` without a filename achieves this goal; a subsequent `resume` command continues profiling without resetting the previously collected data. If you think of the `stop` command as a pause rather than a hard termination, everything falls into place. If there were a separate `pause` command, your proposal would make more sense, but we have what we have.

Finally, I want to mention JFR output as it stands out from the other formats. JFR is the only output that async-profiler writes on the fly. To use JFR format, the user _must_ specify the output file in the `start` command. In this case, there is no separate action to "dump" the results; `stop` simply finalizes the existing recording (so your point about `jfrsync` does not seem relevant). That means, if we set aside the other formats and consider JFR the only native output of async-profiler, then everything already works as you proposed. Users who are familiar with JDK Flight Recorder controls can enjoy a similar user experience with async-profiler.

PR #1733, opened 2026-04-28
Move otlp proto jar updates into a script

benty-amzn:
### Description
Move otlp updates into a script since they're not part of a normal build and don't belong in the Makefile

### How has this been tested?

Ran the script locally; passed and afterwards `git status` shows `modified:   test/gen/opentelemetry-gen-classes.jar`

---


apangin:
Thanks. I like the idea of scripting the download of the latest OTLP release.
However, creating pull requests automatically is an overkill, and it will not work anyway. If the schema changes, the code and the tests will require corresponding adjustments. I suggest not to make a workflow but to leave it as a separate script for running manually.

PR #1735, opened 2026-05-01
Add native lib include and exclude flag to limit CodeCache size

j-bahr:
Problem
=======
When async-profiler attaches to a JVM that has loaded hundreds of native libraries (e.g. FFI workloads with many .so files totaling several GB), Symbols::parseLibraries walks every mapped library and stores per-symbol metadata in CodeCache. In extreme scenarios it is possible that the RSS routinely grows by 5+ GB and asprof becomes unresponsive while holding the parse lock. In containerized environments this can also result in the process being killed by OOMD. The vast majority of those symbols are in code-generated or transitive-dependency libraries the user has no interest in profiling.

Goal
====
Give users a parse-time, library-level filter so they can scope symbol parsing to libraries they actually care about. Frames in filtered-out libraries already degrade gracefully to [unknown] in text/flame and "unknown" in JFR, so a missing library is safe. The only cost is resolution fidelity.

What changed
============
* New flags (default off, fully back-compat):

asprof CLI:  
```
--symbols-include GLOB   (repeatable)
--symbols-exclude GLOB   (repeatable)
```
  Agent:       
  ```
  syminclude=GLOB
  symexclude=GLOB
  ```
Patterns are fnmatch(3) globs matched against library basename.

* Precedence: an exclude match wins over an include match and over the auto-allow set. Otherwise, if any include is set the filter is a positive allowlist; with no include set the filter behaves as a denylist over a default-allow.

* Auto-allow: libjvm, the main executable, and any library located in the same directory as libjvm are always parsed (Java/JNI frames must always resolve). Excluding libjvm explicitly emits a Log::warn but is honored.

* Cache is additive across profile sessions: changing the filter on a later attach affects only newly discovered libraries; libs already in CodeCache continue to resolve.
```
Files
-----
src/arguments.{h,cpp}      add _symbols_include / _symbols_exclude
                           vectors + syminclude= / symexclude= cases
src/symbols.h              Symbols::setFilter / shouldParseLibrary API
src/symbols_linux.cpp      filter impl + JDK-root auto-allow + wiring
                           into parseLibraries; lift discovery cap
                           when filter is active so allowlisted libs
                           aren't lost behind unrelated entries
src/symbols_macos.cpp      same for macOS / dyld iteration
src/profiler.cpp           Symbols::setFilter() at top of start()
src/main/main.cpp          --symbols-include / --symbols-exclude CLI
                           flags + help text; description column
                           widened from 22 to 28 to accommodate
docs/ProfilerOptions.md    new rows in General options table
test/test/smoke/SmokeTests.java
                           coverage for include auto-allow,
                           exclude-libjvm no-op, post-attach exclude,
                           multi-pattern include, additive cache
                    	   across sessions, and rendering as bare
                           lib name under STYLE_LIB_NAMES
```

---


j-bahr:
@apangin to provide additional details, in production environments, we've seen growth of up to 15GB of RSS for processes with very large FFI work loads. We believe this is a combination of CodeCache size inefficiency as well as arena bloat from either glibc or jemalloc. We explored the idea of making the CodeCache more compact and efficient but we believe it'll still result in multi gigabyte regression in RSS. This leads to the process freezing on attach and in some cases the process being killed by OOMD. 

I am happy to close the PR if you have a better approach and am happy to work on implementing an alternative. Thanks for the consideration.

apangin:
Before we proceed with the review, could you please disclose usage of Generative AI tools used in preparation of any part of this PR, including the description, code, documentation, and tests.

j-bahr:
@apangin yes, this PR was AI assisted throughout with human supervision, review and validation.

apangin:
I suggest taking a step back and starting with the problem rather than a solution. This PR is not mergeable: it's too big and does not fit into the existing async-profiler architecture, more like a set of local ad-hoc patches. I assume it targets consequences of the problem, not the root cause. Moreover, the approach to skip some libraries entirely, even when user is not interested in function names, looks risky, because async-profiler parses shared objects not only to resolve symbols, but also to unwind stacks correctly. Skipping `.eh_frame` parsing may result in broken native stack traces.

You mentioned RSS growth and freezing on attach - these are two different issues with separate root causes; they require different treatment. I suggest to fix them in separate PRs. There was an attempt to fix freezing in #1423: that PR may need some polishing, but otherwise goes in the right direction.

As for the RSS issue, the description is not clear about the root cause. I'd like to see a deeper analysis. What structure takes memory exactly, and why? Can you show calculations? If you have more details, please share them in a dedicated issue, not as a comment in this PR, thanks.

j-bahr:
@apangin, Sounds good, I can provide additional details on the RSS problem in a new issues and we can discuss possible solutions there. For now I'll close this PR.

PR #1736, opened 2026-05-02
Address breaking changes in OTLP 1.10.0

ilyamor:
### Description
opentelemetry-proto v1.10 renumbers fields in the profiles `Sample` message. async-profiler currently emits the v1.9 layout, which is no longer wire-compatible with v1.10 readers.

| field                | v1.9 | v1.10 |
|----------------------|:----:|:-----:|
| stack_index          | 1    | 1     |
| values               | 2    | 4     |
| attribute_indices    | 3    | 2     |
| link_index           | 4    | 3     |
| timestamps_unix_nano | 5    | 5     |

### Related issues
n/a

### Motivation and context
We should pick up with the recent breaking changes in OTLP.

### How has this been tested?
Updated tests.

---


apangin:

For me to review the changes, could you please provide more context of the problem.
"We should pick up with the recent breaking changes in OTLP" does not sound like a strong justification. Who we? Why is it needed? What are consequences of picking these changes and what are consequences of not picking them? How was `opentelemetry-gen-classes.jar` generated? What does a new test verify exactly? Does it fail without your changes? Did you try the updated format with any of existing third-party software?

Also, if Generative AI tools were used in preparation of any part of this PR, please include proper attribution, e.g. via `Assisted-by:` field in the commit message. Thanks.

apangin:
@ilyamor Any comments to the above questions?

apangin:
Closing due to no response from the author.

arnabnandy7:
Hi @apangin, 

I ran into the compatibility issue described in this PR while adding OTLP profile export to `grafana/pyroscope-java`.

async-profiler 4.4 emits the Profiles 1.9 `Sample` layout, while Pyroscope 2.1.0 uses Profiles 1.10. The HTTP request is accepted, but `Sample.values` is interpreted as `attribute_indices`, so the resulting profile has no queryable samples. I reproduced this with a Java workload and confirmed that changing `Sample.values` from field 2 to field 4 fixes the wire-format mismatch.

Since the Profiles signal is still experimental and may continue to introduce breaking changes, which direction would you prefer?

1. Update async-profiler’s OTLP encoder to Profiles 1.10 now, with an end-to-end Pyroscope test.
2. Add an explicit OTLP schema/version option so users can select the backend-compatible layout.
3. Keep async-profiler on 1.9 until the Profiles schema stabilizes.

There is also no resource metadata in the generated payload today. Would you be open to an option or API for supplying attributes such as `service.name`, or should callers enrich the payload after `dumpOtlp()`?

I can prepare a new PR with the Pyroscope reproduction, regression test, and appropriate attribution once I know which compatibility approach you prefer.

cc @korniltsev-grafanista

apangin:
@arnabnandy7 [async-profiler 4.5](https://github.com/async-profiler/async-profiler/releases/tag/v4.5) was released yesterday. It is compatible with OpenTelemetry Profiles Alpha and verified to work with Grafana Pyroscope 2.1.1. That required more extensive [changes](https://github.com/async-profiler/async-profiler/pull/1768/changes) than proposed in this PR. Please try and report back if it works for your use-case. `service.name` is not currently included; it will be added in a separate change.

arnabnandy7:
> @arnabnandy7 [async-profiler 4.5](https://github.com/async-profiler/async-profiler/releases/tag/v4.5) was released yesterday. It is compatible with OpenTelemetry Profiles Alpha and verified to work with Grafana Pyroscope 2.1.1. That required more extensive [changes](https://github.com/async-profiler/async-profiler/pull/1768/changes) than proposed in this PR. Please try and report back if it works for your use-case. `service.name` is not currently included; it will be added in a separate change.

I can open a PR for `service.name` if that's not an issue, do let me know whether it's currently being worked on by someone else.

arnabnandy7:
Thanks for the pointer. I opened grafana/pyroscope-async-profiler#21 to bring upstream 4.5 into Grafana's fork while preserving its Grafana-specific context and trace ID changes.

Once the fork publishes v4.5.0.0, I will update pyroscope-java and run the Docker integration test against Pyroscope 2.1.1 to verify that the Fib profile can be queried back. I will report the result here as well.

For service.name, I will keep that as a separate follow-up as suggested.

PR #1737, opened 2026-05-06
Add support for HotCodeHeap

benty-amzn:
### Description
Adds support for HotCodeHeap by increasing the number of code caches from 3 to 4

### Motivation and context
https://bugs.openjdk.org/browse/JDK-8328186

### How has this been tested?
Profile an application with the hot code heap enabled. Before this change, entries will appear in the profile as
`[unknown_nmethod]` and `[no_Java_frame]`. After this change, they appear correctly. The included unit test recreates this and asserts on the behavior.

---

### DIFF
--- a/test/test/vmstructs/VmstructsTests.java
+++ b/test/test/vmstructs/VmstructsTests.java
@@ -6,6 +6,7 @@
 package test.vmstructs;
 
 import one.profiler.test.Assert;
+import one.profiler.test.Jvm;
 import one.profiler.test.Output;
 import one.profiler.test.Test;
 import one.profiler.test.TestProcess;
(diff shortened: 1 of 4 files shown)


test/test/vmstructs/VmstructsTests.java
```diff
@@ -33,4 +34,20 @@ public void checkJni(TestProcess p) throws Exception {
         p.waitForExit();
         Assert.isEqual(p.exitCode(), 0);
     }
+
+    @Test(
+        mainClass = DoBusyWork.class,
+        agentArgs = "start,event=cpu,collapsed,interval=1ms,file=%f",
+        jvmArgs = "-Xcomp -XX:-TieredCompilation -XX:+UnlockExperimentalVMOptions -XX:+HotCodeHeap -XX:+NMethodRelocation -XX:HotCodeIntervalSeconds=0 -XX:HotCodeSampleSeconds=10 -XX:HotCodeStablePercent=-1 -XX:HotCodeSamplePercent=100 -XX:HotCodeStartupDelaySeconds=0",
```
apangin: Why does the test work? It runs less than 10 seconds (`HotCodeSampleSeconds`), so no methods get into the HotCodeHeap during the test. Also, `-XX:-TieredCompilation` effectively disables "profiled code heap" segment.

To verify that we handle CodeCache segments properly, we want to have methods in all 4 segments.

PR #1740, opened 2026-05-11
Clear customEvents between chunks

alevymyers:
### Description

Add `customEvents.clear()` alongside `types.clear()` and `typesByName.clear()` when `JfrReader` moves to the next chunk. `customEvents` is keyed by chunk-scoped JFR type ID's (`typesByName()`), and without clearing it on new chunks, it can become stale. 

### Caveat 

This adds a burden to the caller for custom events. After this change, an entry added by an external caller is wiped at every chunk boundary, because `cacheEventTypes()` only re-registers the events hardcoded in `JfrReader`. Callers that register custom events in multi-chunk recordings will need to re-call `registerEvent` after each chunk. 

If that is acceptable, I think the patch solves the stale ID problem. If not, I can add a implementation to store, and re-register custom events if you think its the responsiblity of `JfrReader` to do it.

---

### DIFF
--- a/src/converter/one/jfr/JfrReader.java
+++ b/src/converter/one/jfr/JfrReader.java
@@ -65,6 +65,18 @@ public class JfrReader implements Closeable {
 
     private final Dictionary<Constructor<? extends Event>> customEvents = new Dictionary<>();
 
+    private final Map<String, Class<? extends Event>> registeredEvents = new HashMap<>();
+
+    // The following classes are instantiated reflectively.
+    // Make sure to list them in reachability-metadata.json.
+    {
+        registeredEvents.put("jdk.CPULoad", CPULoad.class);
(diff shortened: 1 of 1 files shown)


apangin:
Thank you for your proposal. I think the caveat you've mentioned is really important. Unfortunately, this change breaks a legit and most common use case when a JFR recording consists of homogeneous chunks.

alevymyers:
Ok, thank you for the response. See 4d033c5c49bda48f4234230f511aa4d9acfac4cb, where I've added a persistent Map storing the registered events, including the five built-ins. On every chunk transition, the map is iterated to re-build the `customEvents` `Dictionary` with the current chunk's type ID's.

apangin:
Thanks. I've merged a refactored version of your changes.

PR #1741, opened 2026-05-12
minor fix in the documentation

wind57:
It's a tiny fix in the documentation, on how many samples are taken in a particular case.

### DIFF
--- a/docs/FlamegraphInterpretation.md
+++ b/docs/FlamegraphInterpretation.md
@@ -47,7 +47,7 @@ Below are the sampling numbers:
 - `func4()`: 1 sample
 - `func1()->func5()`: 2 samples
 - `func2()->func8()`: 4 samples
-- `func2()->func6()`: 1 sample
+- `func2()->func6()`: 5 samples
 
 ## Sorting samples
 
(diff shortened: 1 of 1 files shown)


docs/FlamegraphInterpretation.md
```diff
@@ -47,7 +47,7 @@ Below are the sampling numbers:
 - `func4()`: 1 sample
 - `func1()->func5()`: 2 samples
 - `func2()->func8()`: 4 samples
-- `func2()->func6()`: 1 sample
+- `func2()->func6()`: 5 samples
```
apangin: There is no mistake here. There is exactly 1 sample with a stacktrace that starts with `func2` and ends with `func6` on the picture. At the same time, there are 4 samples starting with `func2` and ending with `func8`; it'll be probably less confusing if we write it as `func2()->func6()->func8()`.

wind57: right, sorry. that was my intention here. do you want me to write it in this manner?

wind57: changed.

apangin:
Good now. Thank you for the contribution.

PR #1742, opened 2026-05-12
feat(otlp): Populate Location.line with source line numbers

marco96casa:
## Summary

The OTLP serializer currently emits `Location` entries without line numbers — every frame maps 1:1 to a function with no `Line::line` field. This PR resolves JVM bytecode indices (BCI) to source line numbers using the existing `MethodInfo::getLineNumber()` path (same as the JFR writer), enabling line-level flamegraphs in OTLP-consuming backends like Grafana Pyroscope.

## Changes

- **`src/otlp.h`** — Add `Lookup*` member to `Otlp::Recorder`; introduce `LocationEntry` struct (`u32` function index + `jint` line, no padding) to track `(function_index, line_number)` pairs; use a `u64`-keyed `unordered_map` for zero-allocation dedup.
- **`src/otlp.cpp`** — New `getLocationIndex(const char*, ...)` method resolves BCI → line number for any frame with debug info via `Lookup::resolveMethod()`. Avoids heap allocation in the frame loop. The `location_table` writer now emits `Line::line` when the resolved line > 0.
- **`src/profiler.cpp`** — Construct a `Lookup` instance (with `OUTPUT_OTLP`) in `dumpOtlp()` and pass it to the `Recorder`. The `MethodMap` is populated lazily via JVMTI at dump time.
- **`test/test/otlp/OtlpTests.java`** — Add `lineNumbers` test asserting that `CpuBurner.burn` gets a resolved line number in OTLP output.

## Behavior

| Frame type | Before | After |
|---|---|---|
| Java interpreted | `line = 0` (omitted) | Resolved source line number |
| JIT-compiled with debug info | `line = 0` (omitted) | Resolved source line number |
| Native / JIT without debug info | `line = 0` | `line = 0` (unchanged) |

Line numbers are only emitted when available (`line > 0`), so the output remains valid for frames where resolution isn't possible.

## Testing

- Built and verified with `make` on Linux x86_64.
- Profiled a sample Java application and confirmed `Line::line` fields appear in the protobuf output for interpreted and JIT frames with debug info.
- Added automated test (`OtlpTests.lineNumbers`) that profiles `CpuBurner` and asserts line numbers are present.

### DIFF
--- a/src/otlp.cpp
+++ b/src/otlp.cpp
@@ -8,6 +8,33 @@
 
 namespace Otlp {
 
+size_t Recorder::getLocationIndex(const char* name, const ASGCT_CallFrame& frame) {
+    u32 func_idx = (u32)_functions.indexOf(name);
+    jint line = 0;
+
+    if (_lookup != nullptr && frame.bci > BCI_NATIVE_FRAME) {
+        // resolveMethod() caches MethodInfo per method_id in MethodMap (O(1) after first call).
+        // const_cast: resolveMethod() doesn't mutate the frame but its signature lacks const.
+        MethodInfo* mi = _lookup->resolveMethod(const_cast<ASGCT_CallFrame&>(frame));
+        if (mi != nullptr) {
+            jint bci = frame.bci;
+            bci = (bci & 0x10000) ? 0 : (bci & 0xffff);
+            line = mi->getLineNumber(bci);
+        }
+    }
+
+    // Dedup locations by (function, line) pair
+    u64 loc_key = ((u64)func_idx << 32) | (u32)line;
+    auto it = _location_keys.find(loc_key);
+    if (it != _location_keys.end()) {
+        return it->second;
+    }
+    size_t loc_idx = _location_entries.size();
+    _location_keys[loc_key] = loc_idx;
+    _location_entries.push_back({func_idx, line});
+    return loc_idx;
+}
+
 void Recorder::recordProfilesDictionary(const std::vector<CallTraceSample*>& call_trace_samples) {
     protobuf_mark_t dictionary_mark = _otlp_buffer.startMessage(ProfilesData::dictionary);
 
@@ -24,15 +51,16 @@ void Recorder::recordProfilesDictionary(const std::vector<CallTraceSample*>& cal
         _otlp_buffer.commitMessage(function_mark);
     });
 
-    // Write location_table
-    for (size_t function_idx = 0; function_idx < _functions.size(); ++function_idx) {
+    // Write location_table with line numbers
+    for (size_t i = 0; i < _location_entries.size(); ++i) {
+        const LocationEntry& entry = _location_entries[i];
         protobuf_mark_t location_mark = _otlp_buffer.startMessage(ProfilesDictionary::location_table, 1);
-        // TODO: set to the proper mapping when new mappings are added.
-        // For now we keep a dummy default mapping_index for all locations because some parsers
-        // would fail otherwise
         _otlp_buffer.field(Location::mapping_index, (u64)0);
         protobuf_mark_t line_mark = _otlp_buffer.startMessage(Location::lines, 1);
-        _otlp_buffer.field(Line::function_index, function_idx);
+        _otlp_buffer.field(Line::function_index, entry.function_index);
+        if (entry.line_number > 0) {
+            _otlp_buffer.field(Line::line, (u64)entry.line_number);
+        }
         _otlp_buffer.commitMessage(line_mark);
         _otlp_buffer.commitMessage(location_mark);
     }
@@ -78,7 +106,8 @@ void Recorder::recordStacks(const std::vector<CallTraceSample*>& call_trace_samp
                 continue;
             }
 
-            size_t location_idx = _functions.indexOf(_fn.name(trace->frames[j]));
+            const char* name = _fn.name(trace->frames[j]);
+            size_t location_idx = getLocationIndex(name, trace->frames[j]);
             _otlp_buffer.putVarInt(location_idx);
         }
         _otlp_buffer.commitMessage(location_indices_mark);
--- a/src/otlp.h
+++ b/src/otlp.h
@@ -9,8 +9,10 @@
 #include "engine.h"
 #include "frameName.h"
 #include "index.h"
+#include "lookup.h"
 #include "protobuf.h"
 #include "writer.h"
+#include <unordered_map>
 #include <vector>
 
 struct CallTraceSample;
@@ -78,6 +80,7 @@ namespace Function {
 
 namespace Line {
     const protobuf_index_t function_index = 1;
+    const protobuf_index_t line = 2;
 }
 
 namespace KeyValueAndUnit {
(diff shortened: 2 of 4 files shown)


src/otlp.cpp
```diff
@@ -8,6 +8,32 @@
 
 namespace Otlp {
 
+size_t Recorder::getLocationIndex(const std::string& name, const ASGCT_CallFrame& frame) {
+    size_t func_idx = _functions.indexOf(name);
+    jint line = 0;
+
+    if (_lookup != nullptr && frame.bci > BCI_NATIVE_FRAME) {
+        // const_cast: resolveMethod() doesn't mutate the frame but its signature lacks const
+        MethodInfo* mi = _lookup->resolveMethod(const_cast<ASGCT_CallFrame&>(frame));
```
alexanderdx, line 18: For large profiles (millions of samples × deep stacks), the JVMTI call here can become a performance bottleneck

Consider a cache keyed on `(method_id, bci)` before calling `resolveMethod` to avoid redundant JVMTI calls. The existing `_location_keys` dedup only helps after resolution

src/otlp.cpp
```diff
@@ -8,6 +8,32 @@
...
+        if (mi != nullptr) {
+            jint bci = frame.bci;
+            bci = (bci & 0x10000) ? 0 : (bci & 0xffff);
+            line = mi->getLineNumber(bci);
+        }
+    }
+
+    // Pack (func_idx, line) into a u64 key for efficient dedup
+    u64 loc_key = ((u64)func_idx << 32) | (u32)line;
```
alexanderdx, line 27: if the function index ever exceeds 2^32 the upper bits will be silently truncated, potentially causing key collisions? should we have a static assertion, or use explicit 32bit int for the func idx?

src/otlp.cpp
```diff
@@ -78,7 +105,8 @@ void Recorder::recordStacks(const std::vector<CallTraceSample*>& call_trace_samp
                 continue;
             }
 
-            size_t location_idx = _functions.indexOf(_fn.name(trace->frames[j]));
+            std::string name(_fn.name(trace->frames[j]));
+            size_t location_idx = getLocationIndex(name, trace->frames[j]);
```
alexanderdx: This allocates a `std::string` on every frame in the hot loop. If `_fn.name()` returns `const char*`, consider changing `getLocationIndex` to accept `const char*` or `std::string_view` to avoid the heap allocation — the string is only passed to `_functions.indexOf()`.

src/otlp.h
```diff
@@ -95,12 +98,20 @@ struct SampleInfo {
     size_t thread_name_index;
 };
 
+struct LocationEntry {
+    size_t function_index;
+    jint line_number;
+};
```
alexanderdx, line 104: `size_t function_index` (8 bytes) + `jint line_number` (4 bytes) => 4 bytes leftover for padding per entry. 

Not blocking, but `uint32_t` for `function_index` would halve the struct size if the function count will always fit in 32 bits (very likely in practice?)

src/profiler.cpp
```diff
@@ -1398,7 +1399,12 @@ void Profiler::dumpText(Writer& out, Arguments& args) {
 
 void Profiler::dumpOtlp(Writer& out, Arguments& args) {
     FrameName fn(args, args._style & ~STYLE_ANNOTATE, _epoch, _thread_names_lock, _thread_names);
-    Otlp::Recorder recorder(activeEngine(), fn, _start_time * 1000ULL, (OS::micros() - _start_time) * 1000ULL);
+    // MethodMap is populated lazily by Lookup::resolveMethod() via JVMTI queries.
+    // Safe to call here since dumpOtlp() runs on the caller thread, not in a signal handler.
+    MethodMap method_map;
+    Index packages, symbols;
+    Lookup lookup(&method_map, classMap(), &packages, &symbols, OUTPUT_OTLP);
+    Otlp::Recorder recorder(activeEngine(), fn, &lookup, _start_time * 1000ULL, (OS::micros() - _start_time) * 1000ULL);
```
alexanderdx, line 1407: non-blocking: `Recorder` stores a reference to stack-local `Lookup*` raw pointer. If the lifetime of `Recorder` is ever extended or heap-allocated, the `Lookup*` will become a dangling pointer once this local scope exits

src/otlp.cpp
marco96casa: Good point. `resolveMethod()` already caches the `MethodInfo` per `method_id` in the `MethodMap` (hash lookup, O(1) after first resolution), so the JVMTI call only happens once per unique method. The remaining cost is `getLineNumber()` which does a binary search over the method's line number table. These tables are typically small (tens of entries).

src/otlp.cpp
marco96casa: Fixed. Changed `LocationEntry::function_index` to `u32` and the packing now operates on `u32` values explicitly.

src/otlp.cpp
marco96casa: Fixed. `getLocationIndex` now accepts `const char*` directly. `_fn.name()` returns `const char*` and `Index::indexOf()` already has a `const char*` overload, so no intermediate `std::string` is constructed in the frame loop.

src/otlp.h
marco96casa: Fixed. `function_index` is now `u32`, so the struct is `{u32, jint}` = 8 bytes with no padding.

src/profiler.cpp
marco96casa: Both `Lookup` and `Recorder` are stack-local in `dumpOtlp()` with identical lifetimes, so this is safe as-is. If `Recorder` is ever heap-allocated or outlives the calling scope, this would need to change, but that would be a larger refactor of the dump path.

src/otlp.h
```diff
@@ -95,12 +98,20 @@ struct SampleInfo {
...
+struct LocationEntry {
+    u32 function_index;
+    jint line_number;
+};
+
 class Recorder {
   private:
     ProtoBuffer _otlp_buffer;
     FrameName& _fn;
+    Lookup* _lookup;
```
apangin, line 110: Both `FrameName` and `Lookup` serve the purpose of symbol resolution, but in different contexts.
`FrameName` is mostly for text formats, while `Lookup` is for JFR - binary dictionary-based format.
Mixing both together is a bad idea:
1. This is confusing (which one is used for what?)
2. This is redundant (a method is resolved in each structure independently, thus doubling cpu and memory consumption).
3. Each structure has its own lifecycle (e.g. `FrameName` has a static cache that is preserved across sessions, while `Lookup` does not).

See the [related comment](https://github.com/async-profiler/async-profiler/pull/1342#discussion_r2162317224) to other PR.

src/otlp.cpp
```diff
@@ -24,15 +51,16 @@ void Recorder::recordProfilesDictionary(const std::vector<CallTraceSample*>& cal
...
 
-    // Write location_table
-    for (size_t function_idx = 0; function_idx < _functions.size(); ++function_idx) {
+    // Write location_table with line numbers
+    for (size_t i = 0; i < _location_entries.size(); ++i) {
+        const LocationEntry& entry = _location_entries[i];
         protobuf_mark_t location_mark = _otlp_buffer.startMessage(ProfilesDictionary::location_table, 1);
-        // TODO: set to the proper mapping when new mappings are added.
-        // For now we keep a dummy default mapping_index for all locations because some parsers
-        // would fail otherwise
```
apangin, line 32: Why was it removed? I think there is an important note in this comment explaining why we emit zero `mapping_index` here.

src/otlp.cpp
```diff
@@ -8,6 +8,33 @@
...
+
+    // Dedup locations by (function, line) pair
+    u64 loc_key = ((u64)func_idx << 32) | (u32)line;
+    auto it = _location_keys.find(loc_key);
+    if (it != _location_keys.end()) {
+        return it->second;
+    }
+    size_t loc_idx = _location_entries.size();
+    _location_keys[loc_key] = loc_idx;
+    _location_entries.push_back({func_idx, line});
```
apangin, line 34: Isn't it what `Index` class does?

apangin:
@marco96casa Do you plan to continue working on this in the near future, or is it better to close the PR?

marco96casa:
Agree on closing it for now. Thanks!

PR #1746, opened 2026-05-19
Flush wall clock events on dump or jfr chunk rotation

### LINKED ISSUE #1249: Inaccurate Sample Count in WALL_BATCH Mode (v4.0)

benty-amzn:
### Description

Flush pending WallClock events when the user requests a dump or when the jfr chunk rotates, to prevent dropped samples described in the issue

### Related issues

Fixes #1249

### How has this been tested?

Included unit tests based on the reproducer posted to the related issue

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1114,6 +1114,7 @@ Error Profiler::flushJfr() {
     updateJavaThreadNames();
     updateNativeThreadNames();
 
+    if (_event_mask & EM_WALL) wall_clock.flush();
     lockAll();
     _jfr.flush();
     unlockAll();
@@ -1132,6 +1133,7 @@ Error Profiler::dump(Writer& out, Arguments& args) {
     if (_state == RUNNING) {
         updateJavaThreadNames();
         updateNativeThreadNames();
+        if (_event_mask & EM_WALL) wall_clock.flush();
     }
 
     switch (args._output) {
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -1113,8 +1113,8 @@ Error Profiler::flushJfr() {
 
     updateJavaThreadNames();
     updateNativeThreadNames();
-
     if (_event_mask & EM_WALL) wall_clock.flush();
+
     lockAll();
     _jfr.flush();
     unlockAll();
--- a/src/wallClock.cpp
+++ b/src/wallClock.cpp
@@ -3,10 +3,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <map>
 #include <string.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include "wallClock.h"
+#include "mutex.h"
 #include "profiler.h"
 #include "stackFrame.h"
 #include "tsc.h"
@@ -99,6 +101,8 @@ class ThreadCpuTimeBuffer {
 };
 
 static ThreadCpuTimeBuffer _thread_cpu_time_buf;
+static ThreadSleepMap _thread_sleep_state;
+static Mutex _thread_sleep_state_lock;
 
 
 long WallClock::_interval;
@@ -185,13 +189,24 @@ void WallClock::stop() {
     pthread_join(_thread, NULL);
 }
 
+void WallClock::flush() {
+    if (_mode != WALL_BATCH) return;
+
+    MutexLocker ml(_thread_sleep_state_lock);
+    for (ThreadSleepMap::iterator it = _thread_sleep_state.begin(); it != _thread_sleep_state.end(); ++it) {
+        if (it->second.counter != 0) {
+            recordWallClock(it->second, THREAD_SLEEPING, it->first);
+            it->second.counter = 0;
+        }
+    }
+}
+
 void WallClock::timerLoop() {
     int self = OS::threadId();
     ThreadFilter* thread_filter = Profiler::instance()->threadFilter();
     bool thread_filter_enabled = thread_filter->enabled();
     Mode mode = _mode;
 
-    ThreadSleepMap thread_sleep_state;
     ThreadList* thread_list = OS::listThreads();
     _thread_cpu_time_buf.reset();
     u64 cycle_start_time = OS::nanotime();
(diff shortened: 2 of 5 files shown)


src/os_linux.cpp
```diff
@@ -137,11 +137,12 @@ void OS::sleep(u64 nanos) {
     nanosleep(&ts, NULL);
 }
 
-void OS::uninterruptibleSleep(u64 nanos, volatile bool* flag) {
+void OS::uninterruptibleSleep(u64 nanos, volatile bool* flag, volatile bool* flush_pending) {
     // Workaround nanosleep bug: https://man7.org/linux/man-pages/man2/nanosleep.2.html#BUGS
     u64 deadline = OS::nanotime() + nanos;
     struct timespec ts = {(time_t)(deadline / 1000000000), (long)(deadline % 1000000000)};
-    while (*flag && clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, &ts, &ts) == EINTR);
+    while (*flag && (flush_pending == nullptr || !*flush_pending)
```
apangin: This is a veeery leaky abstraction. Why does general-purpose `uninterruptibleSleep` need to be aware of wall-clock engine flushing? It already looks at the flag provided from outside, it's then the caller's responsibility to manage this flag.
We can change `bool flag` to `int state` to allow multiple conditions for interruption. Or we can pass an arbitrary predicate instead.

src/wallClock.cpp
```diff
@@ -185,6 +186,17 @@ void WallClock::stop() {
...
 
+void WallClock::flush() {
+    if (!_running || _mode != WALL_BATCH) return;
+
+    MutexLocker ml(_flush_lock);
+    _flush_pending = true;
+    pthread_kill(_thread, WAKEUP_SIGNAL);
+    while (_flush_pending) {
+        _flush_lock.wait();
+    }
```
apangin: Waiting for other thread to complete flush is fragile, it's hard to verify that it's race-free. This also complicates the algorithm.
What if we do flush right in the dumping thread? This will require synchronizing access to `thread_sleep_state`, but this looks a smaller and safer change.

src/profiler.cpp
```diff
@@ -515,12 +515,7 @@ void Profiler::recordExternalSamples(u64 samples, u64 counter, int tid, u32 call
     _call_trace_storage.add(call_trace_id, samples, counter);
 
     u32 lock_index = getLockIndex(tid);
-    if (!_locks[lock_index].tryLock() &&
-        !_locks[lock_index = (lock_index + 1) % CONCURRENCY_LEVEL].tryLock() &&
-        !_locks[lock_index = (lock_index + 2) % CONCURRENCY_LEVEL].tryLock())
-    {
-        return;
-    }
+    _locks[lock_index].lock();
```
apangin: What is a reasoning behind this change, and what are the consequences?

src/wallClock.cpp
```diff
@@ -254,16 +271,18 @@ void WallClock::timerLoop() {
...
-        _thread_cpu_time_buf.drain(thread_sleep_state);
+        MutexLocker ml(_thread_sleep_state_lock);
+        _thread_cpu_time_buf.drain(_thread_sleep_state);
     }
 
     delete thread_list;
 
     // Flush remaining WallClock batches
-    for (ThreadSleepMap::const_iterator it = thread_sleep_state.begin(); it != thread_sleep_state.end(); ++it) {
+    for (ThreadSleepMap::const_iterator it = _thread_sleep_state.begin(); it != _thread_sleep_state.end(); ++it) {
```
apangin: Why isn't the last iteration protected by the lock?

src/wallClock.cpp
```diff
@@ -185,13 +189,25 @@ void WallClock::stop() {
     pthread_join(_thread, NULL);
 }
 
+void WallClock::flush() {
+    if (_mode != WALL_BATCH) return;
+
+    MutexLocker ml(_thread_sleep_state_lock);
+    _thread_cpu_time_buf.drain(_thread_sleep_state);
```
apangin: If we modify `_thread_cpu_time_buf` in different threads, then other `ThreadCpuTimeBuffer` operations (e.g. `reset`) should be under the lock, too. Except `add`, which _must_ be async-signal-safe.

However, I'm not sure that `drain` is needed here at all.

src/wallClock.cpp
```diff
@@ -254,16 +270,19 @@ void WallClock::timerLoop() {
...
+        MutexLocker ml(_thread_sleep_state_lock);
+        _thread_cpu_time_buf.drain(_thread_sleep_state);
     }
 
     delete thread_list;
 
     // Flush remaining WallClock batches
-    for (ThreadSleepMap::const_iterator it = thread_sleep_state.begin(); it != thread_sleep_state.end(); ++it) {
+    MutexLocker ml(_thread_sleep_state_lock);
+    for (ThreadSleepMap::const_iterator it = _thread_sleep_state.begin(); it != _thread_sleep_state.end(); ++it) {
```
apangin: Seems like code duplication. Can we just call `flush()` here?

apangin:
@benty-amzn Did you see the new test is [failing](https://github.com/async-profiler/async-profiler/actions/runs/26766863046/job/78895817121?pr=1746)?

macOS GHA runners are extremely overloaded; profiler cannot keep up with the sampling rate. Try increasing wall clock interval to 50ms.

PR #1753, opened 2026-05-29
Update the datagram socket test in LockTests so it no longer relies o…

benty-amzn:
…n network traffic; this is creating noise in GHA which makes the test flaky

### Description
`LockTests` sometimes fails in github actions because the time spent on networking, system calls, and firewall rules is much greater than on local developer systems. Rewrite the test to not depend on the network.

### How has this been tested?

---


apangin:
I decided to give this test one more chance and only removed the problematic assertion about the ratio - 656d96b57f1bcd7e54dee04d61985d340d6b93c3

This test was inspired by a real issue, and I used it in my async-profiler demos. Also, this test previously revealed an actual stack walking bug on Alpine Linux (not related to locking though). Anyway, if it keeps failing for a legitimate reason, I think we'll remove completely.

Issue #1567, opened 2025-10-28
The future of the Span API


pveentjer:
### Describe the feature

This branch contains the Span API:

https://github.com/async-profiler/async-profiler/tree/context-events

With the Span API there is an explicit API you can call to say that some action took a certain amount of time.

This is very useful for [Aeron](https://github.com/aeron-io/aeron). In Aeron we have [agents](https://github.com/aeron-io/agrona/blob/master/agrona/src/main/java/org/agrona/concurrent/Agent.java) (e.g. [Sender](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/Sender.java)/[Receivers](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/Receiver.java)/[Conductors](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java)) that run an eventloop. Every agent takes care of determining the current time and caching the time between loop iterations. After every iteration we signal the [DutyCycleTracker](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/DutyCycleTracker.java) that an iteration took a certain amount of time. 

If there is a stall on the agent, then currently it is difficult to find the cause of the latency outlier. That is why I started playing with the Span API. I wrote a bytecode-instrumenting agent that modifies the [DutyCycleTracker](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/DutyCycleTracker.java) to call the Span API. I made a slight modification in the API so that it doesn't call System.nanotime, but makes use of the nanotime already available in the agent.

My question is what the future is of the Span API. The work on the branch seems to have stalled. I would like to help out to get this new feature up and running.

### Use Case

Find the cause of latency outliers.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
Thank you for expressing interest in this feature. Your feedback helps us shape async-profiler roadmap accordingly.

To confirm: you are looking for a garbage-free API to emit spans that can be later correlated with CPU or Wall Clock profiling data, something like
```
Span.emit(long startTime, long endTime, String tag)
```
Is this correct?

[...]

Issue #1580, opened 2025-11-05
Callback for continuous profiling mode with `loop`


Baraa-Hasheesh:
### Describe the feature

As a user of async-profiler I want to do continuous profiling & between each two adjacent profiling session I want to run some logic to process/dump the current stacks within the profiler 

For example upload the JFR to an external service that does visualization on the collected stacks 

Currently this isn't really doable with vanilla agent mode,
The user will need to use the API mode & create some wrappers around the start/stop to handle the post processing for the generated profiles 

### Use Case

As discussed internally we wish to start doing optimization on the `loop` mode to allow faster startup of the profiling sessions, but that optimization won't be useful for some users due to the above mentioned

To make the future optimizations of loop more beneficial we should allow the user to implement profile processor even when using non API mode profiler 

### Proposed Solution

I propose a callback mechanism where the agent is able do callback the java process via a predefined class/method after each `stop` call, before the next `start`

`start,event=cpu,loop=5m,callback=pkg.MyClass.asprofCallback`

For example

```
package pkg;

public class MyClass {

     public static void asprofCallback(String profilePath) {
          ....
     }
}
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Baraa-Hasheesh:
Alternatively we can create a new light weight `stop` command, that we can call `pause` 

Following `pause` only `resume` or `stop` is allowed

apangin:
> I propose a callback mechanism where the agent is able do callback the java process via a predefined class/method

I don't like an idea of calling arbitrary user provided Java code from the profiler context. If done synchronously, this may cause disruption of the normal profiling operation. If asynchronous, this will be subject to all kinds of races.

Also, referencing a method by name is not quite clean API and is not friendly to custom ClassLoaders.

> Alternatively we can create a new light weight stop command, that we can call pause

This is better, but still far from ideal. First, because it complicates the contract between different commands. Second, because this encourages developers to process results while profiler is paused, that contradicts our intention to make continuous profiling smooth.

I think a better option would be to introduce a command that atomically returns profiling result and starts a new profiling session - something like `rotate`/`switch`/`flush`.

Baraa-Hasheesh:
> I think a better option would be to introduce a command that atomically returns profiling result and starts a new profiling session - something like `rotate`/`switch`/`flush`.

I think this is a reasonable approach

PR #1755, opened 2026-06-03
Span API

apangin:
### Description

Implement lightweight Span API for latency profiling.

A span is a `[start, end]` interval associated with the current thread, optionally labeled with a tag (an operation name, endpoint, etc.). Spans are recorded along with other profiling events, so the profile can later be filtered by a tag or duration.

**Design considerations:**

- API is all static and garbage-free.
- Timestamps are aligned with execution samples, whether they use TSC or CLOCK_MONOTONIC. In both cases, getting timestamp is one intrinsified Java method call.
- API itself is decoupled from the async-profiler native library. libasyncProfiler.so can be loaded separately: before Java classes, after Java classes, or not loaded at all - nothing will break.
- No overhead when profiler is not running.
- No JNI calls on the fast path, except for the one to record a span when absolutely needed.
- Two API flavors: one to always emit a span and another one to emit a span only if it encloses at least one profiling event.
- Works with virtual threads. However, in a virtual thread, "if-profiled" methods work the same way as "always-emit" methods.
- Spans accept an optional user-provided String tag. Tags are automatically deduplicated in the JFR recording.

### Related issues

#1567
#1289

### Motivation and context

There are two main use cases for the Span API:
1. Associate samples with a user-provided context (see #1289).
2. Troubleshoot latency outliers (https://github.com/async-profiler/async-profiler/pull/1201#issuecomment-2769236105).

### How has this been tested?

Added 3 new tests in the `span` folder.

---


src/api/one/profiler/Span.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Marks the start of a span on the current thread.
+     *
+     * @return a start timestamp to pass to {@link #end} / {@link #endIfProfiled},
+     *         or {@code 0} if async-profiler is not loaded
+     */
+    public static long start() {
+        if (Recording.state != Recording.RUNNING) {
+            return 0;
+        }
```
Baraa-Hasheesh, line 63: This condition is race prone in continuous profiling mode

It's quite possible that during the profiling session switch that this `start` API gets called & the API detects the `Recording.state` as `STOPPED`, which will result in the `startTime` being returned as 0,.

This in turn means it's quite possible that during continuous profiling sessions, a user will randomly lose/not emit spans.

----------

Note similar problem might happen on the `end` side where the profiler is `STOPPED` when the end call happens

apangin: Yes, this is by design - to keep Span API no-op when profiler is not running.
Restarting profiler already causes event loss - this is not unique to spans. To fix this systematically, a new approach to continuous profiling is needed that does not require stopping and starting profiler - #1580.

docs/IntegratingAsyncProfiler.md
```diff
@@ -57,6 +57,45 @@ can be found in [Profiler Options](ProfilerOptions.md).
...
+
+- `Span.start()` returns a start timestamp; pass it to `end` to record the span.
+- For very frequent operations, use `Span.endIfProfiled(span, tag)` instead of `end`. It records
+  the span only if at least one profiling sample was taken on the thread while it was open - spans
+  that enclose no sample are skipped.
+- `Span.emit(start, end, tag)` and `Span.emitIfProfiled(start, end, tag)` record a span with
+  explicit timestamps obtained from `Recording.timestamp()`, for cases where the interval is not
+  bounded by the current call.
+
+Spans appear as `profiler.Span` JFR events with `startTime`, `duration`, `eventThread` and `tag`
```
netudima, line 92: @apangin, thank you very much for working on this — it's a long-awaited feature!

Am I right that only one tag can be added for a span? For example, if I want to add more context, such as a username and an operation name, would I need to create two spans, or should I concatenate the values into a single tag?

apangin: Thank you for your feedback.

Tags are opaque by design. A single string is a simple yet flexible way to associate any application-specific context with profiling events. If you want to associate multiple attributes with one request, it's your choice how to serialize these attributes into one string.

[Contextual JFR events](https://docs.oracle.com/en/java/javase/25/docs/api/jdk.jfr/jdk/jfr/Contextual.html) are designed exactly for the use case you've described, but I don't see a good way to do the same for async-profiler while maintaining other requirements (e.g. being allocation-free). I'm happy to hear ideas.

src/api/one/profiler/Span.java
```diff
@@ -0,0 +1,113 @@
...
+    private static Class<?> getVirtualThreadClass() {
+        try {
+            return Class.forName("java.lang.BaseVirtualThread");
+        } catch (ClassNotFoundException e) {
+            return null;
+        }
+    }
+
+    // Carries timestamp of the last profiling event in the current thread
+    private static final ThreadLocal<ByteBuffer> LOCAL_BUF = new ThreadLocal<ByteBuffer>() {
```
franz1981, line 39: ThreadLocal.get() on a never-accessed thread allocates a new TL map entry, which matters for short-lived virtual threads

franz1981 (changes requested):

src/api/one/profiler/Span.java
apangin: Good point. An alternative would be to do a virtual thread check on every call to start/end. This is not too bad, but if this can be avoided for the most common use case, I'd prefer that way. After all, virtual threads are not for the maximum speed. Thoughts?

franz1981: That's another option indeed - in old Quarkus/Vertx versions we:
1. check if the platform version is >= 21
2. if it's >= 21 uses (static final) MethodHandle::invokeExact to check if is a virtual thread (this could be delivered without MH using multi-release jars based on your requirements); if inlined is indeed rather cheap and GC free(-ish), which looks like what the Span API aim to deliver

Sadly ScopedValue is not a good fit:
1. they can be inherited if not overridden (in scoped structured concurrency constructs)
2. it forces you to ask users to run their thread with the scoped value you provided (AFAIK)

The other reason against TL (not just for VT, tbh) is that without TL::remove you can cause ClassLoader(s) leaks - but TL::remove is pretty bad due to soft references, although IIRC @shipilev worked on intrinsifying it

apangin: Yes, this PR already has a cheap virtual thread check. But I'm saying I'd prefer to avoid it on the fast path altogether, if possible - that's what the current implementation does.

I'm not sure I understand your point about ClassLoader leak. What is the problem here exactly? Note that TL values are tiny ByteBuffers, not connected to async-profiler's classes.

franz1981: https://github.com/spring-attic/spring-cloud-sleuth/issues/27 this is one old example of the "leak" mentioned - not sure it applies to this, this chat just "unblocked a memory" from my past ^^

apangin: Thanks. Keys in the `ThreadLocalMap` are WeakReferences. What _can_ cause a real problem is `value` which is a strong reference ([example](https://github.com/open-telemetry/opentelemetry-java/issues/7083)). However, in my case, values are ByteBuffers (i.e. from the bootstrap ClassLoader), therefore, no leak here, even if a thread outlives async-profiler classes.

franz1981: Yep makes sense ^^

GC pressure from WeakReference processing is still not great, but you got my point on this alredy :heart:

apangin: OK, I actually benchmarked both versions and found no noticeable performance regressions with a virtual thread check on every call. I pushed the version that avoids creating thread locals for virtual threads.

pveentjer:
Will this implementation be able to handle the high rate of a busy spin loops? In Aeron we use various flavors of busy spinning depending on the agent, so conceptually something like this:

```
long nowNs = clock.nanoTime();
while (running) {
    final long span = Span.start();
    final int workCount = doWork();
    idleStrategy.idle(workCount);
    final long nextNs = clock.nanoTime();
    Span.endIfProfiled(span, "spin");
    dutyCycleTracker.measureAndUpdate(nowNs);
    nowNs = nextNs;
}
```

This means that endIfProfiled  can be called millions of times a second. The older span API was not able to deal with that, so I added a minimal time filter. In Aeron, we maintain the duration of single iteration in nanoseconds, so adding a filter on our side is trivial if that would be needed. 

```
long nowNs = clock.nanoTime();
while (running) {
    final long span = Span.start();
    final int workCount = doWork();
    idleStrategy.idle(workCount);
    final long nextNs = clock.nanoTime();
    if (nextNs - nowNs >= thresholdNs) {
        Span.endIfProfiled(span, "spin");
    }
    dutyCycleTracker.measureAndUpdate(nowNs);
    nowNs = nextNs;
}
```

But would be best if the Span API would be able to handle that.

apangin:
@pveentjer Yes, this should be fine. I've optimized `Span.endIfProfiled` even further to skip getting a timestamp when there are no profile samples. So, on the fast path, `Span.endIfProfiled` is a matter of checking one volatile field and one thread local.
For low-latency operations, the number of actually recorded spans will roughly match the sampling rate (e.g. 1000 if you profile with `interval=1ms`). You can add a custom logic above that to skip more spans; your example with `thresholdNs` is totally fine.

`Span.start()` gets a timestamp in the profiler's clock. This will be either `System.nanoTime` or `jdk.jfr.internal.JVM.counterTime` (on x86, where `rdtsc` is supported). Both are JVM intrinsics. If needed, you can force `System.nanoTime` clock with the `clock=monotonic` profiler option. If the application code already obtains timestamps that match profiler's clock, you can skip `Span.begin/end` entirely and emit spans with externally provided timestamps:
```
Span.emitIfProfiled(nowNs, nextNs, "spin");
```


To ensure no disk I/O happens in the application threads (when the recording buffer overflows), add the `mem` option. In this case, async-profiler will write events to an in-memory file then flush it on disk later when the profiling session ends. Alternatively, use `tmpfs` for the profiler recordings (`mem` option is not needed in this case).

petrbouda:
Already integrating Spans to Jeffrey Microscoper, it's just an initial support. It looks great so far! Do you plan any support for virtual thread, I know you want to keep async-profiler a little bit aside from virtual threads because of additional performance penalties. Currently, instead of virtual threads I see ForkJoinPool instead and very likely I can get a different one when the thread wakes up after blocking operation and I am not able correlate it.

However, it's great and think how to expand the support :) Thanks for that!
https://github.com/petrbouda/jeffrey/releases/download/v0.10.1-b10/microscope.jar

Screenshot from 2026-06-09 22-17-29
Screenshot from 2026-06-09 22-17-47
Screenshot from 2026-06-09 22-18-14
Screenshot from 2026-06-09 22-18-34
Screenshot from 2026-06-09 22-18-58
Screenshot from 2026-06-09 22-19-17
Screenshot from 2026-06-09 22-19-41

apangin:
@petrbouda Wow, that's impressive! Thanks!
Yes, virtual threads support is our next goal. This won't happen all at once, though. First goes unwinding of virtual frames for CPU profiling; spans in virtual threads will come later.

olivergillespie:
Thanks for this, it's working well for my needs.

One use-case to consider - many people already record the duration of their interesting segments to emit with metrics, and it's not always trivial/desirable to bundle another `long` in the context for tracking the async-profiler-specific start timestamp. It might be nice if there was a way to emit based on duration instead of timestamp (`Span.emit(long durationNanos, String tag)`). Since `timestamp()` doesn't have a specified unit, we can't just do `Span.emit(timestamp() - durationNanos, tag)`.

apangin:
Thank you for your feedback. I've [added](https://github.com/async-profiler/async-profiler/pull/1763) `Recording.clockFrequency()` method that enables conversions between user-provided timestamps and async-profiler's timestamps.

PR #1758, opened 2026-06-21
Parse DWARF unwinding info in .debug_frame section

apangin:
### Description

Reconstruct stack unwinding info from the `.debug_frame` section in two cases:
1. when parsing separate debuginfo file;
2. when a loaded shared object does not have `.eh_frame` section.

### Related issues

#1320

### Motivation and context

Async-profiler has been relying on the `.eh_frame` section for obtaining native stack traces when a library is compiled with `-fomit-frame-pointer` option. However, `.eh_frame` section is not always present, or it does not cover all the code. At the same time, DWARF unwinding information may exist in a separate debuginfo file in the `.debug_frame` section.

`.debug_frame` is different from `.eh_frame` in several aspects, but the underlying structure of the unwinding info is mostly the same: it's a list of Common Information Entry (CIE) records and Frame Description Entry (FDE) records, where FDEs contain instructions how to recover a stack frame (pc, sp, fp) at the given location.

### How has this been tested?

- Added a new integration test.
- Verified that both cases mentioned in #1320 work fine: 1) native stack traces are shown correctly on Alpine Linux when `musl-dbg` package is installed; 2) async-profiler can now profile GraalVM Native Images compiled with `-g` option.

---


Makefile
```diff
@@ -256,6 +256,11 @@ ifeq ($(OS_TAG),linux)
 
 	$(AS) -o $(TEST_LIB_DIR)/twiceatzero.o test/native/libs/twiceatzero.s
 	$(LD) -shared -o $(TEST_LIB_DIR)/libtwiceatzero.$(SOEXT) $(TEST_LIB_DIR)/twiceatzero.o --section-start=.seg1=0x4000 -z max-page-size=0x1000
+
+	$(CC) -shared -fPIC -O2 -g -fno-asynchronous-unwind-tables -fno-unwind-tables -fomit-frame-pointer -fno-optimize-sibling-calls $(INCLUDES) -Isrc -o $(TEST_LIB_DIR)/libdebugframe.$(SOEXT) test/native/libs/debugframe.c
```
benty-amzn, line 260: It doesn't look like the other test sources are built with `-O2`, is there a reason this one is?

src/symbols_linux.cpp
```diff
@@ -467,6 +473,19 @@ void ElfParser::parseDwarfInfo() {
     }
 }
 
+void ElfParser::parseDebugFrameSection() {
+    if (!DWARF_SUPPORTED || _cc->hasDwarfTable()) return;
+
+    ElfSection* debug_frame = findSection(SHT_PROGBITS, ".debug_frame");
+    if (debug_frame != NULL) {
```
benty-amzn, line 480: `nullptr`?

src/symbols_linux.cpp
```diff
@@ -327,6 +328,10 @@ bool ElfParser::parseFile(CodeCache* cc, const char* base, const char* file_name
         if (elf.validHeader()) {
             elf.calcVirtualLoadAddress();
             elf.loadSymbols(use_debug);
+            if (!use_debug || elf.findProgramHeader(PT_GNU_EH_FRAME) == NULL) {
```
benty-amzn, line 331: `nullptr`?

src/dwarf.cpp
```diff
@@ -122,13 +119,77 @@ void DwarfParser::parseFde() {
...
+
+    qsort(_table, _count, sizeof(FrameDesc), FrameDesc::comparator);
+}
+
+// CIE in .debug_frame differs from .eh_frame: it has no augmentation data,
+// and DWARF v4 adds address_size and segment_selector_size after the augmentation string.
+void DwarfParser::parseDebugCie() {
+    u32 cie_len = get32();
+    bool dwarf64 = cie_len == 0xffffffff;
+    _ptr += dwarf64 ? 8 + 8 : 4;
```
benty-amzn: I'm sure the compiler will optimize it out but the inline `8 + 8` scans a little weird, any reason not to use `dwarf64 ? 16 : 4`?

src/codeCache.h
```diff
@@ -177,6 +177,10 @@ class CodeCache {
         _debug_symbols = debug_symbols;
     }
 
+    bool hasDwarfTable() const {
+        return _dwarf_table != NULL;
```
benty-amzn, line 181: Should this be `nullptr` for consistency with the rest of the PR, or is it part of the C api?

Makefile
apangin: Yes, `-O2` (or `-O3`) complicates the shape of the compiled functions (from the stack walking perspective). Without optimizations, all of `Java_test_stackwalker_DebugFrameApp_run` and `level1-3` functions have a simple recognizable prologue that sets up a fixed-size frame. `-O2` breaks this pattern, especially for the leaf `level3` function: its frame is no longer fixed, `rbp` is used as a scratch register, `sqrt` function is inlined.
The particular optimizations don't matter; what matters is that `-O2` makes the frame layout less predictable, which is important for testing correctness of the stack unwinding logic in the absense of frame pointers.

apangin: In fact, this test also works without `-O2`, but the general consideration here is that if DWARF unwinding works for highly optimized code, it will likely work for non-optimized code too, but the opposite is not always true.

src/symbols_linux.cpp
apangin: I deliberately kept `NULL` in files that already have many `NULL`s but no `nullptr`s.
We'll eventually change everything to `nullptr`, but for now, I tried to keep consistency within each file.

src/dwarf.cpp
apangin: Changed to 16 and added a comment explaining these constants.

benty-amzn (approved):

PR #1769, opened 2026-07-12
#1406, #1756: Parse only essential libraries at load time

### LINKED ISSUE #1406: JVM application pauses when using asprof

### LINKED ISSUE #1756: Runtime attach fails on JVMs with many native libraries
### Describe the bug

On a process that has loaded more shared libraries than `MAX_NATIVE_LIBS` (2048), a runtime attach can fail with:

```
[ERROR] JVM does not support Tool Interface
```

The error is misleading, the real failure is that async profiler cannot locate the `libjvm.so` `CodeCache`, because it looks for it
inside the symbol cache array that is capped at `MAX_NATIVE_LIBS`. When `libjvm.so` happens to be mapped at a `/proc/self/maps` address past the first 2048 libraries, it is never parsed into that array, the lookup returns `NULL`, and attach aborts.

This reproduces using OpenJDK when running large FFI workloads with a large number of shared libraries ( 7,000+, with more than 2048 libraries mapped before `libjvm.so` )

### Root cause

`VM::init()` already obtains a pointer inside libjvm, then discards it and resolves libjvm by scanning the capped array instead:

- existing pointer inside libjvm.so
https://github.com/async-profiler/async-profiler/blob/019123b1b4d398c0eb51d57b64270b0ab750b2c1/src/vmEntry.cpp#L182
- Parses libs only up until the cap and scan the capped array
https://github.com/async-profiler/async-profiler/blob/019123b1b4d398c0eb51d57b64270b0ab750b2c1/src/vmEntry.cpp#L187-L190
- lib will be NULL causing it to return false, triggering the error 
https://github.com/async-profiler/async-profiler/blob/019123b1b4d398c0eb51d57b64270b0ab750b2c1/src/vmEntry.cpp#L194-L197

Which boils down to 
`VM::init()` 
-> `updateSymbols()` parses libraries but stops at the cap, so libjvm.so never gets parsed
-> `findLibraryByAddress(AsyncGetCallTrace)` therefore returns `NULL` 
-> `VMStructs::init(NULL)` does nothing, so `VMStructs::libjvm()` is left `NULL`
-> `findJvmLibrary("libj9vm")` hands back that same `NULL` 
-> `VM::init()` bails out with return `false` at the `if (lib == NULL)` check.
Agent_OnAttach() sees the false return and prints "JVM does not support Tool Interface"...

###  Expected:
  
  Expected: asprof <pid> attaches to the running JVM and begins profiling (or returns status), the same way it works when the process has fewer than `MAX_NATIVE_LIBS` native libraries, and the same way a `jcmd <pid> ...`  diagnostic attach succeeds on the very same process.

  ### Actual: 

asprof immediately fails with

  `[ERROR] JVM does not support Tool Interface`

  and exits non-zero, so no profile is collected. The message is incorrect, I believe that we should be able to produce a profile even when there are more shared libs than `MAX_NATIVE_LIBS`. 

### Reproduction Steps

Load more than `MAX_NATIVE_LIBS` shared objects from Java, with at least one loaded after libjvm so that libjvm's `/proc/self/maps` rank exceeds the cap, then attach:

```bash
# 1) build ~2600 tiny stub .so files
mkdir -p /tmp/manylibs
echo 'int s(void){return 0;}' | cc -shared -fPIC -xc - -o /tmp/manylibs/lib0.so
for i in $(seq 1 2600); do cp /tmp/manylibs/lib0.so /tmp/manylibs/lib$i.so; done

# 2) a JVM that dlopens them all, then idles
cat > /tmp/ManyLibs.java <<'EOF'
public class ManyLibs {
  public static void main(String[] a) throws Exception {
    for (int i = 0; i <= 2600; i++) {
      try { System.load("/tmp/manylibs/lib" + i + ".so"); } catch (Throwable t) {}
    }
    Thread.sleep(3_600_000);
  }
}
EOF
javac -d /tmp /tmp/ManyLibs.java
java -XX:+EnableDynamicAgentLoading -cp /tmp ManyLibs &

# 3) attach -> fails with "JVM does not support Tool Interface"
asprof status $!
```

I have a few thoughts on potential solutions to this. We could work around the cap by special casing the handling of libjvm, or excluding it from the cap. I believe this could be done by using the existing handle to `_asyncGetCallTrace` or by special casing and parsing libjvm first before the rest of the shard libs. 

This just covers up a slightly larger problem for us, the fact that we don't have a configurable cap for the maximum number of shared libraries. I am not sure if 2048 was chosen deliberately, but even if we solve the libjvm issue we will likely end up having missing symbols in our traces when the cap is enabled. 

Should we explore a solution that allows the cap to be configureable as compile time flags `-DMAX_NATIVE_LIBS=...`?

```cpp 
  #ifndef MAX_NATIVE_LIBS
  const int MAX_NATIVE_LIBS = 2048;
  #endif
```

Or could we explore adding a runtime flag `ASPROF_MAX_NATIVE_LIBS` that is read during agent load that constructs the array on the heap rather than being a fixed sized array? 

I'd be happy to work on a PR if you can provide guidance on how you would like to solve this directionally. Please let me know your thoughts.  
<hr>

**Additionally:** This is somewhat separate from this issue, but related to my current investigation. I would appreciate your thoughts on how to manage large FFI work loads like ours. We estimate that the code cache overhead for all shared libraries is ~3GB. This takes time and system resources. We worked around the loading time by attaching asprof as an agent at startup to pay parsing cost during warm up rather than at profile time. The memory overhead though lives for the life of the process and is never freed. Do you have any ideas on how we could improve both the time and memory overhead?

### Async-profiler version

v4.4

### Environment details

**OS/Arch:** Linux x86-64
**JVM:** OpenJDK (HotSpot), JDK 17 / 21 / 25
**Container**: Yes

apangin:
### Description

`Symbols::parseLibraries()` gets a new `essential_only` mode that parses only the libraries required at agent load time: `libjvm`, `libj9*`, and async-profiler itself.

### Related issues

- Fixes #1406
- Fixes #1756

### Motivation and context

`LateInitializer` runs as part of `dlopen()` when the agent is loaded. Parsing all shared libraries there has two problems:

- In a process with many/large libraries, `dlopen()` runs too long on a thread that cannot reach a safepoint, causing a long time-to-safepoint pause (#1406).
- When more than `MAX_NATIVE_LIBS` (2048) libraries are loaded, the scan may exhaust the limit before reaching `libjvm`, so the agent fails to attach with "JVM does not support Tool Interface" (#1756).

Parsing only the essential libraries makes agent load fast and guarantees `libjvm` is always parsed. Libraries parsed at load time are remembered, so `Profiler::start()` does not parse them again. `libj9*` is included so that OpenJ9 symbols remain available in `VM::init()` as before.

### How has this been tested?

- CI
- Manual: dynamic attach and `-agentpath` at startup; no dynamic-agent warning on JDK 21+
- Reproduced #1756 with a test app that loads 2100 extra libraries: the old build fails to attach, the new build attaches instantly and profiles correctly

---

### DIFF
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -740,7 +740,17 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
     fclose(f);
 }
 
-static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count) {
+static bool isEssentialLibrary(const char* file, const char* map_start, const char* map_end) {
+    const void* this_lib = (const void*)isEssentialLibrary;
+    if (this_lib >= map_start && this_lib < map_end) {
+        return true;  // async-profiler's own library
+    }
+
+    const char* base_name = strrchr(file, '/');
+    return base_name != NULL && (strncmp(base_name + 1, "libjvm.", 7) == 0 || strncmp(base_name + 1, "libj9", 5) == 0);
+}
+
+static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count, bool essential_only) {
     FILE* f = fopen("/proc/self/maps", "r");
     if (f == NULL) {
         return;
@@ -776,6 +786,9 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
         }
 
         if (map.isExecutable()) {
+            if (essential_only && !isEssentialLibrary(map.file(), map_start, map_end)) {
+                continue;
+            }
             SharedLibrary& lib = libs[inode];
             if (lib.file == nullptr) {
                 lib.file = strdup(map.file());
@@ -794,7 +807,7 @@ static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs,
     fclose(f);
 }
 
-void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
+void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols, bool essential_only) {
     MutexLocker ml(_parse_lock);
 
     if (_in_parse_libraries || array->count() >= MAX_NATIVE_LIBS) {
@@ -815,7 +828,7 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
     }
 
     std::unordered_map<u64, SharedLibrary> libs;
-    collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count());
+    collectSharedLibraries(libs, MAX_NATIVE_LIBS - array->count(), essential_only);
 
     for (auto& it : libs) {
         u64 inode = it.first;
--- a/src/symbols_linux.cpp
+++ b/src/symbols_linux.cpp
@@ -747,7 +747,11 @@ static bool isEssentialLibrary(const char* file, const char* map_start, const ch
     }
 
     const char* base_name = strrchr(file, '/');
-    return base_name != NULL && (strncmp(base_name + 1, "libjvm.", 7) == 0 || strncmp(base_name + 1, "libj9", 5) == 0);
+    return base_name != NULL && (
+               strncmp(base_name + 1, "libjvm.", 7) == 0 ||
+               strncmp(base_name + 1, "libj9", 5) == 0 ||
+               strncmp(base_name + 1, "libazsys", 8) == 0
+           );
 }
 
 static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count, bool essential_only) {
--- a/src/symbols.h
+++ b/src/symbols.h
@@ -18,7 +18,7 @@ class Symbols {
 
   public:
     static void parseKernelSymbols(CodeCache* cc);
-    static void parseLibraries(CodeCacheArray* array, bool kernel_symbols);
+    static void parseLibraries(CodeCacheArray* array, bool kernel_symbols, bool essential_only = false);
 
     static bool haveKernelSymbols() {
         return _have_kernel_symbols;
--- a/src/symbols_macos.cpp
+++ b/src/symbols_macos.cpp
@@ -191,7 +191,11 @@ static bool isEssentialLibrary(const char* path, const mach_header* image_base) 
     }
 
     const char* base_name = path != NULL ? strrchr(path, '/') : NULL;
-    return base_name != NULL && (strncmp(base_name + 1, "libjvm.", 7) == 0 || strncmp(base_name + 1, "libj9", 5) == 0);
+    return base_name != NULL && (
+               strncmp(base_name + 1, "libjvm.", 7) == 0 ||
+               strncmp(base_name + 1, "libj9", 5) == 0 ||
+               strncmp(base_name + 1, "libazsys", 8) == 0
+           );
 }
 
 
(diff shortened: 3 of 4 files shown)


src/symbols_linux.cpp
```diff
@@ -740,7 +740,17 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
...
 
-static void collectSharedLibraries(std::unordered_map<u64, SharedLibrary>& libs, int max_count) {
+static bool isEssentialLibrary(const char* file, const char* map_start, const char* map_end) {
+    const void* this_lib = (const void*)isEssentialLibrary;
+    if (this_lib >= map_start && this_lib < map_end) {
+        return true;  // async-profiler's own library
+    }
+
+    const char* base_name = strrchr(file, '/');
+    return base_name != NULL && (strncmp(base_name + 1, "libjvm.", 7) == 0 || strncmp(base_name + 1, "libj9", 5) == 0);
```
Baraa-Hasheesh: I would vote to also give priority to some zing libs as well 

https://github.com/async-profiler/async-profiler/blob/master/src/cpuEngine.cpp#L70

If the profiler is run with Zing it possible to hit the "**Could not set pthread hook**" error due to the fact that "libazsys" was never parsed

Baraa-Hasheesh (commented):
change looks good to me, one small suggestion related to zing support

src/symbols_linux.cpp
apangin: Thank you for the review. I remember about Zing libraries, but they are not a problem as far as I can see, since `CpuEngine::setupThreadHook()` is called from `Profiler::start` -> `_engine->start()` after the full library scan is made in `Profiler::start()`. Or am I missing something?

Baraa-Hasheesh: The problem is that due to the 2048 lib limit it's possible that this library is never parsed & skipped due to the mentioned limit

So https://github.com/async-profiler/async-profiler/issues/1756 won't be fixed

Baraa-Hasheesh: What I mean that in the full scan, this lib is at the end of the list so it might be skipped

apangin: Ah, that's true. I've added `libazsys` to the early parsing phase too. Thanks for the right call!

apangin: The change in `symbols_macos.cpp` is not really needed (no Zing on macOS), but it's harmless - I added it to keep files in sync.

Baraa-Hasheesh: > The change in symbols_macos.cpp is not really needed (no Zing on macOS), but it's harmless - I added it to keep files in sync.

Was going to raise that, but yes I agree keeping the files in sync looks better & easier to understand the overall purpose

Issue #1743, opened 2026-05-14
OpenJDK fastdebug crashes at assert(Thread::current() == ThreadLocalStorage::thread()) failed: TLS mismatch!


apangin:
### Describe the bug

The issue was originally reported against OpenJDK:
https://bugs.openjdk.org/browse/JDK-8382795

However, it turns out to be a glibc defect triggered by async-profiler.
We can work around the bug in async-profiler.

### Reproduction Steps

1. A new thread starts and calls `Thread::initialize_thread_current()`.
2. `ThreadLocalStorage::set_thread(this)` calls `pthread_setspecific`, which populates this thread's slot for the given key: https://codebrowser.dev/glibc/glibc/nptl/pthread_setspecific.c.html#86-87
Even though the C code initializes `seq` field first and `data` field after, the order of stores in the compiled code is reversed (nothing prevents gcc from doing this "optimization").
    ```
   <+88>: mov %r12d,%ebx
   <+91>: mov %rbp,0x8(%rax)  # level2->data = value
   <+95>: xor %r8d,%r8d
   <+98>: mov %rbx,(%rax)     # level2->seq = seq
    ```

4. A profiling signal arrives between two stores, e.g. at offset <+95>.
5. async-profiler detects if it is sampling a native thread or a Java thread by calling `pthread_getspecific` with the same key.
6. `pthread_getspecific` reads non-null `data` field, but discovers that `seq` is zero. It therefore falls to the "unlikely" branch that treats the value stale and resets it to NULL: https://codebrowser.dev/glibc/glibc/nptl/pthread_getspecific.c.html#60
7. Subsequent `ThreadLocalStorage::thread()` reads NULL and triggers the assertion.

### Async-profiler version

4.4

### Environment details

x86_64 / glibc 2.35 / OpenJDK 25 fastdebug

Baraa-Hasheesh:
Hello @apangin 

I was taking a look at this problem to attempt to create a solution (I have a rough idea on how to do that using `dl` standard functions to find `pthread_setspecific ` address boundaries)

However while I was analyzing the code base I noticed 2 things:
- This issue also may happen to `asprof_thread_local_data` 

This happens as the initial get call will call `pthread_setspecific` [here](https://github.com/async-profiler/async-profiler/blob/master/src/threadLocalData.cpp#L33) 

During recording of sample we call `getIfPresent` which calls ` pthread_getspecific(_profiler_data_key)`

-> We have a pair of set/get using same pthread key

[...]

PR #1771, opened 2026-07-18
Fix glibc TLS corruption

Baraa-Hasheesh:
### Description
In this change I'm addressing the glibc race that causes the tls entry to be unset, this is done by simply calling the pthread_setspecific two times back to back for the problematic keys

To achieve this with minimum changes the pthread_setspecific hook is now always enabled even if no CPU engine is selected, the CPU engine now only is responsible for setting itself to allow creation of timers/perf-events

Note: 
I did consider a more complex solution [here](https://github.com/async-profiler/async-profiler/compare/master...Baraa-Hasheesh:async-profiler:refs/heads/address-boundary-fix) that checks if `pthread_getspecific` is safe to call during the stackwalking, however after comparing the pros & cons of both, this solution wins for me

Pros:
- Less complex & lower change footprint
- Happens outside the of the StackWalking (hotpath), so will have a lower performance overhead overall 

Cons (minor):
- Problem is only glibc specific so we're calling pthread_setspecific twice for macOS & MUSL when it's not needed

### Related issues
#1743

### Motivation and context
Fix problem in glibc that causes the TLS entry to be unset

### How has this been tested?
I wasn't able to create a viable reproducer, but in theory the change should work

---

### DIFF
--- a/src/cpuEngine.cpp
+++ b/src/cpuEngine.cpp
@@ -32,6 +32,9 @@ static int pthread_setspecific_hook(pthread_key_t key, const void* value) {
     }
 
     if (value != NULL) {
+        // workaround for https://github.com/async-profiler/async-profiler/issues/1743
+        // having 2 different set functions guarantees the TLS will be correct after both are executed
+        pthread_setspecific(key, value);
         int result = pthread_setspecific(key, value);
         CpuEngine::onThreadStart();
         return result;
@@ -80,11 +83,17 @@ bool CpuEngine::setupThreadHook() {
 
 void CpuEngine::enableThreadHook() {
     *_pthread_entry = (void*)pthread_setspecific_hook;
+}
+
+void CpuEngine::enableEngine() {
     storeRelease(_current, this);
 }
 
 void CpuEngine::disableThreadHook() {
     *_pthread_entry = (void*)pthread_setspecific;
+}
+
+void CpuEngine::disableEngine() {
     storeRelease(_current, nullptr);
 }
 
--- a/src/cpuEngine.cpp
+++ b/src/cpuEngine.cpp
@@ -32,10 +32,9 @@ static int pthread_setspecific_hook(pthread_key_t key, const void* value) {
     }
 
     if (value != NULL) {
-        // workaround for https://github.com/async-profiler/async-profiler/issues/1743
-        // having 2 different set functions guarantees the TLS will be correct after both are executed
-        pthread_setspecific(key, value);
         int result = pthread_setspecific(key, value);
+        // Workaround for #1743: the second call repairs TLS if it was corrupted by the nested pthread_getspecific
+        pthread_setspecific(key, value);
         CpuEngine::onThreadStart();
         return result;
     } else {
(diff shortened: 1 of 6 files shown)


apangin:
Thanks for looking into this.
I agree the proposed solution looks better than checking `pthread_getspecific` bounds in the stack walking code.
However, I'm thinking of an even smaller change that avoids calling `pthread_getspecific` entirely in the common case:
```
VMThread* VMThread::current() {
    if (_tls_slot_offset >= 0) {
        return ((VMThread**)pthread_self())[_tls_slot_offset];
    }
    // Fallback to pthread_getspecific
    return _tls_index >= 0 ? (VMThread*)pthread_getspecific((pthread_key_t)_tls_index) : NULL;
}
```
We can discover `_tls_slot_offset` with a similar trick that we do for `_tls_index`.

Baraa-Hasheesh:
> We can discover `_tls_slot_offset` with a similar trick that we do for `_tls_index`

Are you suggesting doing a manual scan of the  internal `pthread` structure when we detect a `glibc` running in search for an exact offset for the TLS that stores the `VMThread` structure?

Happy path (`key < PTHREAD_KEY_2NDLEVEL_SIZE`):
- First we need to find the [specific_1stblock](https://codebrowser.dev/glibc/glibc/nptl/descr.h.html#pthread::specific_1stblock)
- After which we can do `specific_1stblock[key]`

-> Above suggested trick can be used, as the specific_1stblock is flat within the structure

Unlikely path  (`key >= PTHREAD_KEY_2NDLEVEL_SIZE`)::
- First we need to find the [specific](https://codebrowser.dev/glibc/glibc/nptl/descr.h.html#pthread::specific) for that key
- After which we can access the key

`pthread_self()->specific[key / PTHREAD_KEY_2NDLEVEL_SIZE][key % PTHREAD_KEY_2NDLEVEL_SIZE]`

-> Above suggested trick won't work, as we have 2 levels of indirection,

**Note both code paths depend on  `PTHREAD_KEY_2NDLEVEL_SIZE` & `PTHREAD_KEY_1STLEVEL_SIZE` that was used to compile the glibc (Thankfully these should be relatively constant)**

apangin:
First-level TLS block (covering 32 thread locals) is inlined in the pthread structure. We don't need to know the specific offset where the array starts - we can just probe the offset dynamically like we do in `VMStructs::initTLS`. No need to care about 2nd-level block: a genuine Java process will always have pthread_key within the first 32 slots. The only case when it does not hold is when a native applications allocates 32 or more pthread_keys and then creates an embedded JVM. This is a very uncommon scenario, but even in this case, nothing breaks catastrophically - we just fallback to the current behavior.

This is not what I've finally decided - I'm exploring alternatives.

apangin:
OK, after more thoughts, I'm leaning towards the approach with `pthread_setspecific` hook as it's less hacky than the offset probing. The hook is already installed for CPU profiling, which a dominant mode, so installing the hook unconditionally will be totally fine. However, after this change, `pthread_setspecific` hook will no longer be exclusive to CpuEngine, so I propose a small refactoring to move `setup/enable/disableThreadHook` functions out of CpuEngine to Profiler.

apangin:
OK, I've merged this and will do a follow-up refactoring shortly. I think pthread_setspecific hook will line up nicely with the existing `dlopen_hook` in profiler.cpp. We can probably merge them in one function to install/uninstall both hooks together.

Baraa-Hasheesh:
Thanks, I was planning to do the needed refactoring/changes over the weekend

apangin:
No problem, I can leave it to you then.

PR #1775, opened 2026-07-24
Configurable rate limiter for JFR events

apangin:
### Description

New `--ratelimit` option that limits the number of JFR events emitted per second, individually configurable per event category.

Supported categories: : `cpu`, `alloc`, `lock`, `wall`, `nativemem`, `nativelock`, `trace`, `span`.

Example: `--ratelimit cpu:1000,alloc:200,span:100`

Implemented as a per-category token budget refilled every second by the profiler's Timer thread. Events exceeding the budget are discarded early, before the stack walk, so throttled samples cost almost nothing.

Unused budget carries over to the next second (up to 100%) to tolerate momentary bursts. Effective limits are recorded as `jdk.ActiveSetting` events of the corresponding event types. The feature applies to JFR output only and adds no overhead when not configured.

### Motivation and context

In continuous profiling, an event flood can blow up recording size and profiling overhead. It's not always possible to select the optimal profiling interval ahead of time to avoid this. Rate limiting gives users a predictable per-second bound.

### How has this been tested?

- New native tests for parsing `ratelimit` argument.
- New integration test `JfrTests.rateLimit`.
- Manually verified emitted event counts match the exact budget math and `jfr print` shows the `ratelimit` settings.

---


src/profiler.cpp
```diff
@@ -393,6 +394,14 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
     atomicInc(_total_samples);
 
     int tid = OS::threadId();
+    if (!RateLimit::allow(event_type)) {
+        if (event_type == PERF_SAMPLE) {
+            // Need to reset PerfEvents ring buffer, even though we discard the sample
+            PerfEvents::resetBuffer(tid);
```
Baraa-Hasheesh: should we increment a counter to indicate a skipped sample as we do below?

`atomicInc(_failures[-ticks_skipped]);`

src/profiler.cpp
```diff
@@ -512,6 +525,10 @@ void Profiler::recordExternalSample(u64 counter, int tid, EventType event_type,
 }
 
 void Profiler::recordExternalSamples(u64 samples, u64 counter, int tid, u32 call_trace_id, EventType event_type, Event* event) {
+    if (!RateLimit::allow(event_type)) {
```
Baraa-Hasheesh, line 525: This handles
- LIVE objects -> happens when profiler is stopped
- WALL sample flush -> profiler stop, or on demand by user 

For case 1 the check is not needed

For case 2, applying the rate limit looks iffy for me.

What do you think about not applying the limits at all at this code path?

Baraa-Hasheesh (commented):
**Suggestion**

This might complicate the solution, but it's possible to enhance this feature for WALL engine by checking the limits before interrupting the thread

-> It doesn't need to be super accurate you can create a `peek` function that just checks without decrementing the budget & keeping the actual decrement in the recordSample path

src/arguments.h
```diff
@@ -280,7 +284,9 @@ class Arguments {
         _title(NULL),
         _minwidth(0),
         _reverse(false),
-        _inverted(false) {
+        _inverted(false)
+    {
+        for (int i = 0; i < EC_CATEGORIES; i++) _rate_limit[i] = -1;
```
benty-amzn, line 289: `_rate_limit{-1}` in the initializer list would be more consistent with existing code

apangin: That will initialize only the first element with -1 but all the rest with 0.
I could write `_rate_limit{-1, -1, -1, -1, -1, -1, -1, -1}`, but it will be easy to miss when the number of categories grows.

src/profiler.cpp
apangin: Thanks for flagging this. I initially thought that `ticks_skipped` is only useful for text output, but it is also used for calculating stack walking stats, so incrementing the counter for rate-limited samples would be right.
I've pushed a [fix](https://github.com/async-profiler/async-profiler/pull/1775/changes/a490082607d1b5c2961212af6cb7b66fa0ab0979), along with a small refactoring to reduce code duplication.

src/profiler.cpp
apangin: `LiveObject` events are not rate-limited; the check has no effect on them.
`WallClock` events are throttled, so limits should apply wherever events come from.

Frankly speaking, rate-limiting WallClock events with the existing mechanism is not a good idea, since it may skew the profile significantly. I have a better idea how to reduce the number of WallClock events by without affecting profile accuracy - this can be done by a better coalescing of repeated events that have exactly the same stack trace. But that's a separate story.

To your original point, `recordExternalSamples` is a general-purpose function, I want to keep the check here to make sure rate limits apply no matter what events come this way now or in the future.

Issue #1668, opened 2026-01-28
Wallclock is not filtered by `include=`


maciejkwidzinski:
### Describe the bug

Can't filter wallclock profiles via AP agent `include=` option.

### Expectation
The JFR contains only a profile of our tests. There should only be one thread called `Test worker`. All stack traces should contain `org.gradle.api.internal.tasks.testing`.
Both for "Method Profiling Sample"s and "Wall-Clock Sleeping" samples are filtered.

### Reality
The profile is not filtered:
<img width="1632" height="424" alt="Image" src="https://github.com/user-attachments/assets/253c9fb2-650d-4336-a32c-fb5c771f019b" />

Method Profiling Sample events are filtered properly:
<img width="1341" height="426" alt="Image" src="https://github.com/user-attachments/assets/0eead748-23a2-447d-be5b-6b30478ed961" />

But "Wall-Clock Sleeping" events contain all stacktraces from hundreds of threads:
<img width="1566" height="579" alt="Image" src="https://github.com/user-attachments/assets/d908f622-d80b-4d73-a7db-e642fed6fe72" />

### Reproduction Steps

1. Run via IntelliJ (agent) profiling a Gradle test task.
2. Use IJ-bundled agent (v4.1) or provide v4.3 manually (doesn't make a difference).
3. Agent options: `event=wall,interval=2ms,include=org.gradle.api.internal.tasks.testing.*`.
4. Wait for tests to finish and IntelliJ profiler GUI to open.

De-facto settings:
<img width="1606" height="728" alt="Image" src="https://github.com/user-attachments/assets/43cf217c-e14f-4282-ab6b-a0a42115e1c6" />

#### Just filter via GUI, select the `Test worker` thread
It's ok for initial analysis, but:
* can't diff in GUI, because IJ diff no longer provides the thread picker
    * it would have to join by threads from separate profiles, but there's no generic way to do this: thread ids change, names are not unique and can by dynamic
* why send 300 MB JFR files, when a much smaller one would suffice

#### Use `filter=`
Luckily, the `Test worker` always has id=1 in IJ profiler GUI. But `filter=1` doesn't work.
It's because thread filtering is by native thread id, which is not predictable ahead of time.

#### Filter it via `jfrconv`
That works somewhat, but:
* I cannot output JFR with it
* HTML is cool and all, but IJ can read JFR profiles and display profiling info in code, it's tremendously useful
* can't diff in IJ
[...]

PR #1777, opened 2026-07-24
feat: Filter wall samples by thread name instead of thread id

benty-amzn:
### Description
Changes the `filter` parameter from a series of numeric thread ID ranges to a pattern-based filter on thread name
Parameter can be supplied multiple times for multiple patterns like include/exclude

### Related issues
#1668

### Motivation and context
See [discussion](https://github.com/async-profiler/async-profiler/issues/1668#issuecomment-3818895351) on the related issue

### How has this been tested?
Included unit tests, profiled a gradle daemon with a filter and confirmed daemon threads were not included in the profile

---


src/threadNameHook.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+static void setEntry0(JNIEnv* env, jclass cls, jlong entry) {
+    const JNINativeMethod method = {
+        (char*)"setNativeName", (char*)"(Ljava/lang/String;)V", (void*)(uintptr_t)entry
+    };
+    hook_registered = env->RegisterNatives(thread_class, &method, 1) == 0;
+}
+
+Error installThreadNameHook() {
```
apangin: This all looks overcomplicated. I know it was inspired by the similar code in `LockTracer`, but there is a substantial difference. `Unsafe_park` is a private function in `libjvm.so`, it may not even have a corresponding symbol if `libjvm.so` is stripped - hence those complications. On the contrary, `JVM_SetNativeThreadName` is a public exported symbol appearing in all JVM versions, so we know ahead of time the original function to call. Furthermore, `JVM_SetNativeThreadName` is defined in `libjvm.so` but called from `libjava.so` via PLT/GOT. To hook this function, we may skip `RegisterNatives` completely and setup a native-level hook like we do for `malloc` and `pthread_*` functions. As a bonus, this should work correctly even with multiple instances of async-profiler loaded simultaneously, unlike the `RegisterNatives` approach where only one instance can hook `JVM_SetNativeThreadName` at a time.

src/profiler.cpp
```diff
@@ -684,16 +695,48 @@ void Profiler::setThreadInfo(int tid, const char* name, jlong java_thread_id) {
...
-    if (_update_thread_names) {
-        JitWriteProtection jit(true);  // workaround for JDK-8262896
-        jvmtiThreadInfo thread_info;
-        int native_thread_id = VMThread::nativeThreadId(jni, thread);
-        if (native_thread_id >= 0 && jvmti->GetThreadInfo(thread, &thread_info) == 0) {
-            jlong java_thread_id = VMThread::javaThreadId(jni, thread);
-            setThreadInfo(native_thread_id, thread_info.name, java_thread_id);
-            jvmti->Deallocate((unsigned char*)thread_info.name);
+void Profiler::onThreadRename(int tid, const char* name) {
+    MutexLocker state_lock(_state_lock);
```
apangin: Why is this two-level lock?
`onThreadRename` is called in the context of a runnable Java thread. Profiler thread may hold `_state_lock` for an extended period, e.g., while dumping a large profile. During this time, an innocent Java thread will remain stalled.

src/profiler.cpp
```diff
@@ -130,18 +132,27 @@ void Profiler::addRuntimeStub(const void* address, int length, const char* name)
     CodeHeap::updateBounds(address, (const char*)address + length);
 }
 
-void Profiler::onThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
-    if (_thread_filter.enabled()) {
-        _thread_filter.remove(OS::threadId());
+bool Profiler::matchesThreadFilter(const char* name) {
+    return name != NULL &&
+           std::any_of(_thread_matchers.begin(), _thread_matchers.end(),
```
apangin: Try to avoid bringing in new STL classes/functions, unless there is a good reason to.

src/threadNameHook.cpp
apangin: **Edit:** `JVM_SetNativeThreadName` is _not_ called via PLT; GOT patching won't work.
`RegisterNatives` is still needed, but the previous comment about getting address of `JVM_SetNativeThreadName` directly still holds.

docs/ProfilerOptions.md
```diff
@@ -59,7 +59,7 @@ The below options are `action`s for async-profiler and common for both `asprof`
...
-| `--filter FILTER`    | `filter=FILTER`    | In the wall-clock profiling mode, profile only threads with the specified ids.<br>Example: `asprof -e wall -d 30 --filter 120-127,132,134 Computey`                                                                                                                                                                                                                                                                                                                                                                                         |
+| `--filter PATTERN`   | `filter[=PATTERN]` | In the wall-clock profiling mode, profile only threads whose names match the given pattern(s). `--filter` can be specified multiple times. A pattern may begin or end with a star `*` that denotes any (possibly empty) sequence of characters. Threads are matched as they start and whenever they are renamed. Without a pattern, filtering is enabled with an empty set of threads, to be populated through `AsyncProfiler.addThread()`.<br>Example: `asprof -e wall -d 30 --filter 'pool-1-thread-*' --filter main Computey`            |
```
Baraa-Hasheesh: are we concerned about breaking backward compatibility?

src/arguments.h
```diff
@@ -183,7 +183,8 @@ class Arguments {
     const char* _log;
     const char* _loglevel;
     const char* _unknown_arg;
-    const char* _filter;
+    bool _filter_enabled;
```
Baraa-Hasheesh, line 187: `_filter_enabled` is not needed, you can directly check `_filter.empty()`

src/profiler.cpp
```diff
@@ -934,6 +994,28 @@ Error Profiler::start(Arguments& args, bool reset) {
...
+        }
+
+        _thread_filter.clear();
+        {
+            MutexLocker ml(_thread_names_lock);
+            _thread_matchers.assign(args._filter.begin(), args._filter.end());
+        }
+
+        _update_thread_names = true;
+        updateJavaThreadNames();
```
Baraa-Hasheesh: Is `updateJavaThreadNames` needed here?

If I'm not wrong it's sufficient to only call `updateNativeThreadNames`

src/profiler.cpp
```diff
@@ -934,6 +994,28 @@ Error Profiler::start(Arguments& args, bool reset) {
...
 
+    bool filter_threads = args._filter_enabled && (_engine == &wall_clock || hasEvent(EC_WALL));
+    _thread_filter.setEnabled(filter_threads);
+
+    if (filter_threads && !args._filter.empty()) {
+        if (VM::loaded() && (error = installThreadNameHook())) {
+            return error;
+        }
+
+        _thread_filter.clear();
```
Baraa-Hasheesh: `_thread_filter.clear()`  should be done outside the check

apangin:
I see several issues with the way how current instrumentation of `Thread.setNativeName` works:
- Hook is never uninstalled, causing an overhead even when profiler is not running. Uninstalling a hook by re-registering native method would lead to [another issue](https://bugs.openjdk.org/browse/JDK-8369219).
- Running the second instance of async-profiler will permanently disable the hook of the first instance (e.g., running an ad-hoc profiling session when the continuous profiler is active)
- Filter does not apply to non-Java threads started in the middle of a profiling session. This is also relevant for JVM Compiler and GC threads.

Given shortcomings of the current implementation and extra complexity it introduces, I think it would be better to split the change into separate PRs. Or maybe don't support thread renaming at all in the first round.

src/profiler.cpp
benty-amzn: The native thread names are truncated to 15 characters, at least on my local linux machine. Depending on the exact length of the thread name and the pattern the user supplies, not checking the full Java thread names could be inaccurate.

src/arguments.h
benty-amzn: Filtering enabled is distinct from "one or more patterns exist" - it supports cases like this one from the tests:

```
        profiler.execute("start,event=wall,filter");
        profiler.addThread(included);
```

src/profiler.cpp
apangin: Full thread snapshots are now made twice: in the beginning and in the end of a profiling session. The last one is likely redundant if we track all thread creations and terminations.

src/profiler.cpp
```diff
@@ -765,6 +766,41 @@ void Profiler::updateNativeThreadNames() {
...
+    jvmtiThreadInfo thread_info;
+    int tid = VMThread::nativeThreadId(jni, thread);
+    if (tid >= 0 && jvmti->GetThreadInfo(thread, &thread_info) == 0) {
+        _thread_filter.update(tid, thread_info.name);
+        jvmti->Deallocate((unsigned char*)thread_info.name);
+    }
+}
+
+void Profiler::applyThreadFilter() {
+    ThreadList* thread_list = OS::listThreads();
```
apangin, line 780: This list includes all threads, Java and native.
For Java threads, you first apply filter based on their truncated native names, then apply it once again based on Java names.

apangin: This method duplicates the most expensive part of `updateNativeThreadNames` and  `updateJavaThreadNames`, but without updating `_thread_names`. Since the work is done anyway, would it make sense to update `_thread_names` map and then use it for filtering? This can also help with the above comment (to filter Java threads once).

src/profiler.cpp
```diff
@@ -765,6 +766,41 @@ void Profiler::updateNativeThreadNames() {
...
+            _thread_filter.update(tid, name_buf);
+        }
+    }
+    delete thread_list;
+
+    if (VM::loaded()) {
+        jvmtiEnv* jvmti = VM::jvmti();
+        jint thread_count;
+        jthread* thread_objects;
+        if (jvmti->GetAllThreads(&thread_count, &thread_objects) == 0) {
```
apangin, line 794: While reviewing this, I noticed a leak in the existing async-profiler code - `updateJavaThreadNames`. This code allocates a bunch of JNI handles which may stay alive for a long period of time. `PushFrame`/`PopFrame` may be needed to avoid this problem.

src/threadFilter.h
```diff
@@ -25,6 +27,8 @@ class ThreadFilter {
     u32* _bitmap[MAX_BITMAPS];
     bool _enabled;
     volatile int _size;
+    std::vector<Matcher> _include;
+    std::vector<Matcher> _exclude;
```
apangin, line 31: Hmm. `ThreadFilter` is probably not the best name for this structure - it is more like a low-level bitset suitable for async-signal context. It is also used in other places (`Recording::_thread_set`).

`std::vector<Matcher>` makes it a higher-level construct. It suggests splitting this into two classes, e.g. `ThreadBitSet` and `ThreadFilter`.

src/profiler.cpp
```diff
@@ -134,6 +134,7 @@ void Profiler::addRuntimeStub(const void* address, int length, const char* name)
 void Profiler::onThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
     if (_thread_filter.enabled()) {
         _thread_filter.remove(OS::threadId());
+        filterThread(jvmti, jni, thread);
```
apangin, line 137: The first line always removes the thread while the second one may add it again.
Maybe we can check the filter first and then either add or remove the thread to the set?

src/main/main.cpp
```diff
@@ -494,7 +495,7 @@ int main(int argc, const char** argv) {
             format << ",features=" << String(args.next()).replace(',', "+");
 
         } else if (arg == "--filter") {
-            format << ",filter=" << String(args.next()).replace(',', ";");
+            params << ",filter=" << args.next();
```
apangin: You can merge it with the generic case that covers `all`, `live`, etc.

src/arguments.cpp
```diff
@@ -557,6 +560,15 @@ int Arguments::parseTimeout(const char* str) {
     return 0xff000000 | hh << 16 | mm << 8 | ss;
 }
 
+bool Arguments::parseThreadFilter(const char* pattern) {
+    if ((pattern[0] != '+' && pattern[0] != '-') || pattern[1] == 0) {
```
apangin: I don't quite like the fact that _include_ pattern needs a `+` prefix. Perhaps, we can make it optional?
To me, it would be more intuitive to have `includePattern` and `!excludePattern` (prefixed with `!`), wdyt?

src/profiler.cpp
```diff
@@ -132,17 +132,14 @@ void Profiler::addRuntimeStub(const void* address, int length, const char* name)
 }
 
 void Profiler::onThreadStart(jvmtiEnv* jvmti, JNIEnv* jni, jthread thread) {
-    if (_thread_filter.enabled()) {
-        _thread_filter.remove(OS::threadId());
-    }
```
apangin, line 138: Why did you remove it in `onThreadStart` but not in `onThreadEnd`?

src/profiler.cpp
```diff
@@ -973,6 +987,13 @@ Error Profiler::start(Arguments& args, bool reset) {
         return Error("Running method tracing and Java method sampling in parallel is not supported");
     }
 
+    bool filter_threads = args._filter_enabled && (_engine == &wall_clock || hasEvent(EC_WALL));
+    _thread_filter.init(filter_threads, args._threadfilter_include, args._threadfilter_exclude);
+    if (filter_threads && !(args._threadfilter_include.empty() && args._threadfilter_exclude.empty())) {
+        _update_thread_names = true;
+        applyThreadFilter();
+    }
```
apangin: nit: I'd move this down in the code, after cheap checks

src/arguments.cpp
benty-amzn: While it's an odd choice, `!` is a valid first character for a thread name. AFAIK, any ascii character we might use for include/exclude is, and so making it optional introduces ambiguity between `!thread*` as a thread name and `!thread*` meaning "exclude `thread*`"

src/profiler.cpp
benty-amzn: for thread start it moved to `pthread_setspecific_hook`

src/arguments.cpp
apangin: Yeah I know. Have never seen such weird thread names, though. Even for that unlikely case, a workaround is to put `*` in the beginning: `*!thread*`.
IMO, `!` is visually better. Compare with `--filter -pool-1-*` for instance (thread names often contain `-`).
`--filter -thread-name` also reads as a separate flag.

A real alternative is to have two different options. I just didn't know good names for these options so ruled them out in the beginning )

src/profiler.cpp
apangin: `pthread_setspecific_hook` is called both on start and termination.

apangin:
Please avoid force-pushing, unless you do a major rewrite of everything.
This makes harder to review incremental changes.

src/arguments.cpp
Baraa-Hasheesh: > A real alternative is to have two different options. I just didn't know good names for these options so ruled them out in the beginning )

I would vote for this as it would maintain backwards compatibility (thread IDs on old option)

What about

- include_thread
- exclude_thread

apangin: A bit too long; underscore is ugly in connection with `--` prefix; overlap with existing --include/--exclude options.
I thought about `--only-thread`/`--skip-thread`, but can't say I like it very much either.

benty-amzn: 2 flags certainly seems like the cleanest solution to me, but I don't have any better names to suggest at the moment.

docs/ProfilerOptions.md
```diff
@@ -128,3 +130,31 @@ By default, async-profiler merges stack traces starting from the outermost (e.g.
 - `otlp` - dump events in OpenTelemetry format.
 
 It is possible to specify multiple dump options at the same time.
+
+### Pattern Syntax
+
+A pattern matches the name of a thread or stack frame. Patterns may be prefixed or
```
apangin, line 136: Worth mentioning that patterns are case-sensitive.

src/main/main.cpp
```diff
@@ -79,6 +79,9 @@ static const char USAGE_STRING[] =
     "  --nativelock time   pthread mutex/rwlock profiling threshold in nanoseconds\n"
     "  --wall interval     wall clock profiling interval\n"
     "  --nobatch           legacy wall clock sampling without batch events\n"
+    "  --filter            in wall clock mode, profile only threads registered via the API\n"
```
apangin, line 82: I think this option shall remain API-only, i.e. without `asprof` exposure.

src/arguments.cpp
```diff
@@ -294,7 +294,26 @@ Error Arguments::parse(const char* args) {
...
             CASE("filter")
-                _filter = value == NULL ? "" : value;
+                _filter_enabled = true;
+                if (value != NULL && value[0] != 0) {
+                    msg = "filter does not take a value; use ithread/xthread";
+                }
+
+            CASE("ithread")
+                _filter_enabled = true;
+                if (value == NULL || value[0] == 0) {
```
apangin, line 304: Empty string is a valid pattern.

src/profiler.cpp
```diff
@@ -758,14 +762,24 @@ void Profiler::updateNativeThreadNames() {
             if (it == _thread_names.end() || it->first != tid) {
                 if (OS::threadName(tid, name_buf, sizeof(name_buf))) {
                     _thread_names.insert(it, std::map<int, std::string>::value_type(tid, name_buf));
+                    if (_thread_filter.enabled()) {
+                        _thread_filter.update(tid, name_buf);
+                    }
                 }
+            } else if (_thread_filter.enabled()) {
+                _thread_filter.update(tid, it->second.c_str());
```
apangin, line 770: Why is that? What cases does this update cover?

test/test/api/ThreadFilterApi.java
```diff
@@ -0,0 +1,41 @@
...
+        Thread included = new Thread(ThreadFilterApi::included);
+        Thread excluded = new Thread(ThreadFilterApi::excluded);
+        included.start();
+        excluded.start();
+
+        AsyncProfiler profiler = AsyncProfiler.getInstance();
+        profiler.execute("start,event=wall,filter");
+        profiler.addThread(included);
+        profiler.addThread(excluded);
+        profiler.removeThread(excluded);
```
apangin, line 24: How does this work? What prevents a sample to land between `addThread` and `removeThread`?

src/profiler.cpp
```diff
@@ -711,7 +710,13 @@ void Profiler::setupSignalHandlers() {
...
 void Profiler::setThreadInfo(int tid, const char* name, jlong java_thread_id) {
     MutexLocker ml(_thread_names_lock);
-    _thread_names[tid] = name;
+    std::map<int, std::string>::iterator it = _thread_names.lower_bound(tid);
+    if (it == _thread_names.end() || it->first != tid) {
+        _thread_names.insert(it, std::map<int, std::string>::value_type(tid, name));
+        _thread_filter.update(tid, name);
+    } else {
+        it->second = name;
+    }
```
apangin, line 719: What was a reason for replacing `_thread_names[tid] = name;` with a more complicated snippet?
I guess, to skip updating filter when a thread name is already known. But that is actually a good thing, even though we agreed that it's not required to handle all renamings.

apangin: I'd leave `_thread_filter.update` in `updateThreadName()` as in your previous revision.

Note that `updateNativeThreadNames()` does not overwrite a known thread name on purpose, because native thread names are limited to 15 characters only. So, we first assign thread names to all Java threads, then scan through all platform threads to name the remaining (non-Java) threads.

src/profiler.cpp
```diff
@@ -995,6 +1000,21 @@ Error Profiler::start(Arguments& args, bool reset) {
...
+
+    _thread_filter.init(filter_threads, args._threadfilter_include, args._threadfilter_exclude);
+    if (filter_by_name) {
+        updateJavaThreadNames();
+        updateNativeThreadNames();
+
+        MutexLocker ml(_thread_names_lock);
+        for (std::map<int, std::string>::iterator it = _thread_names.begin(); it != _thread_names.end(); ++it) {
+            _thread_filter.update(it->first, it->second.c_str());
+        }
```
apangin, line 1015: If `updateJavaThreadNames`+`updateNativeThreadNames` always update the filter as they should, this loop will be redundant.

src/arguments.h
```diff
@@ -182,7 +182,9 @@ class Arguments {
...
+    std::vector<const char*> _threadfilter_include;
+    std::vector<const char*> _threadfilter_exclude;
```
apangin, line 187: nit: `_thread_filter_include` (with underscore) would be better style-wise, although a bit too long.
maybe just `_thread_include` (`_include_thread`, `_threads_include`)?

src/profiler.cpp
```diff
@@ -758,7 +760,10 @@ void Profiler::updateNativeThreadNames() {
             if (it == _thread_names.end() || it->first != tid) {
                 if (OS::threadName(tid, name_buf, sizeof(name_buf))) {
                     _thread_names.insert(it, std::map<int, std::string>::value_type(tid, name_buf));
+                    _thread_filter.update(tid, name_buf);
                 }
+            } else {
+                _thread_filter.update(tid, it->second.c_str());
```
apangin: Since you've brought this line back, my previous question remains: what is this line for? It just re-applies filter to all threads that have been already filtered. If this is only for `resume` command, it is not relevant: `resume` should be called with the same arguments as `start`, otherwise its behavior is undefined. Even if we wanted to support this case (probably not), we'd do that differently.

docs/ProfilerOptions.md
```diff
@@ -128,3 +130,32 @@ By default, async-profiler merges stack traces starting from the outermost (e.g.
...
+### Thread Filter
+
+`--ithread PATTERN` and `--xthread PATTERN` restrict wall clock sampling to a subset of
+threads. Thread names are matched against a [Pattern](#pattern-syntax): `--ithread` includes
+matching threads, `--xthread` excludes them. Patterns are matched at profiling start, when a
+Java thread starts, and on each dump. The `filter` option, available through the API only,
+enables filtering without patterns: no threads are sampled until added with
+`AsyncProfiler.addThread()`.
+These options have no effect unless wall clock profiling is enabled.
+Note: combining `ithread`/`xthread` with `addThread()`/`removeThread()` is not supported.
```
apangin: Can you explain this again please? I don't see why it can't, at least, during one profiling session without pauses and restarts. We don't necessarily need to support this, but if it comes for free, why not?

For example, what if I always want to profile main thread (which I specify in the `ithread` option) then add other worker threads dynamically as they execute tasks I'm interested in.

PR #1779, opened 2026-08-01
Move pthread_setspecific hook to profiler space

Baraa-Hasheesh:
### Description
Follow up to #1771, after that change the `pthread_setspecific` hook is more generic & no longer fits under the CpuEngine class/file 

Moving the hooking function into the profiler class/file similar to what's already done for the dlopen hook

### Related issues
NA

### Motivation and context
Code & structure consistency 

### How has this been tested?
`make test`

---

### DIFF
--- a/src/cpuEngine.cpp
+++ b/src/cpuEngine.cpp
@@ -4,7 +4,6 @@
  */
 
 #include <errno.h>
-#include <pthread.h>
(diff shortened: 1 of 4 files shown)


apangin:
Useful cleanup. Thanks.

PR #1783, opened 2026-08-05
Reduce flakiness in parseMultiModeRecording on JDK 8

benty-amzn:
### Description
The JVM starts a thread's blocked-time clock before it posts the JVMTI contended-enter event that async-profiler timestamps. Posting that event transitions the thread to native, which blocks if a safepoint is pending. Time spent waiting there counts toward ThreadMXBean blocked time but falls outside the profiler's window, lowering the ratio the test asserts. Enabling UseCountedLoopSafepoints brings behavior in line with newer versions.

### Motivation and context
`parseMultiModeRecording` is flaky on JDK 8 on highly-contended GHA runners

### How has this been tested?
Local repro in a docker container based on the GHA runner
The additional flag IgnoreUnrecognizedVMOptions is set because LoopStripMiningIter is not available on JDK 8. It is set on other versions to keep the value the [same as default](https://github.com/openjdk/jdk25u-dev/blob/63176eeac932d12cee18461f7dd84474359e3c5e/src/hotspot/share/gc/g1/g1Arguments.cpp#L234) on newer JDKs

---

### DIFF
--- a/test/test/jfr/JfrTests.java
+++ b/test/test/jfr/JfrTests.java
@@ -84,7 +84,8 @@ public class JfrTests {
      * @param p The test process to profile with.
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true)
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmVer = {9, Integer.MAX_VALUE})
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmVer = {8, 8}, jvmArgs = "-XX:+UseG1GC -XX:+UseCountedLoopSafepoints")
     public void parseMultiModeRecording(TestProcess p) throws Exception {
         Output output = p.waitForExit(TestProcess.STDOUT);
         assert p.exitCode() == 0;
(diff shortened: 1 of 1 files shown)


test/test/jfr/JfrTests.java
```diff
@@ -84,7 +84,7 @@ public void parseRecording(TestProcess p) throws Exception {
      * @param p The test process to profile with.
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true)
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmArgs = "-XX:+IgnoreUnrecognizedVMOptions -XX:+UseG1GC -XX:+UseCountedLoopSafepoints -XX:LoopStripMiningIter=1000")
```
apangin: `-XX:+UseG1GC -XX:+UseCountedLoopSafepoints -XX:LoopStripMiningIter=1000` are the default settings in JDK 11+, and there is no `-XX:LoopStripMiningIter` option in JDK 8.

I think it will be less confusing to keep the test without custom JVM flags on JDK 11+, but add a separate `@Test` specifically for JDK 8. `@Test` annotation can be repeated multiple times for the same method.

test/test/jfr/JfrTests.java
```diff
@@ -84,7 +84,8 @@ public void parseRecording(TestProcess p) throws Exception {
      * @param p The test process to profile with.
      * @throws Exception Any exception thrown during profiling JFR output parsing.
      */
-    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true)
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmVer = {9, Integer.MAX_VALUE})
+    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmVer = {8, 8}, jvmArgs = "-XX:+UseG1GC -XX:+UseCountedLoopSafepoints")
```
apangin:
```suggestion
    @Test(mainClass = JfrMultiModeProfiling.class, agentArgs = "start,event=cpu,alloc,lock=0,quiet,jfr,file=%f", output = true, jvmVer = 8, jvmArgs = "-XX:+UseG1GC -XX:+UseCountedLoopSafepoints")
```

PR #1790, opened 2026-08-22
Align sampling behaviour across output formats

Baraa-Hasheesh:
### Description
I noticed that for the `recordExternalSample`  & `recordExternalSamples` we save the stack trace before a lock is acquired.

This means the output generated by the profiler might differ depending on the format selected by the user
- If JFR is selected then the events/samples will not be reported
- If Collapsed/Flamegraph is selected then the samples will be reported 

This change simply moves the lock check to be done first:
- The generated output will be the same no matter the output format
- Avoids recording unneeded traces when using JFR output formats

A secondary issue this change fixes, is the fact that `recordExternalSamples` did not update the `_total_samples` & the `_failures[-ticks_skipped]` counters

### Related issues
N/A

### Motivation and context
- Make the profiler output more consistent across the various output formats
- Account for `recordExternalSamples` in the sampling stats

### How has this been tested?
`make test`

---

### DIFF
--- a/src/profiler.cpp
+++ b/src/profiler.cpp
@@ -494,8 +494,9 @@ u64 Profiler::recordSample(void* ucontext, u64 counter, EventType event_type, Ev
 
 void Profiler::recordExternalSample(u64 counter, int tid, EventType event_type, Event* event, int num_frames, ASGCT_CallFrame* frames) {
     atomicInc(_total_samples);
+    int lock_index;
 
-    if (!RateLimit::allow(event_type)) {
+    if (!RateLimit::allow(event_type) || (lock_index = tryLock(tid)) < 0) {
         atomicInc(_failures[-ticks_skipped]);
         return;
     }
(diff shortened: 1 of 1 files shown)


apangin:
As far as I understand, this PR intends to solve inconsistency between JFR and FlameGraph output in case of failed lock acquisitions. But why is it a problem in the first place?

It is uncommon to dump the same profile in both formats. People choose one or another. With JFR output, recordings can be later converted to any other format without need to dump twice.
OK, even if someone dumps `collapsed` after `jfr`, they may get a _more accurate_ profile that includes samples missed in JFR. What your PR proposes is to make `collapsed`/`flamegraph` output _worse_, so that it misses samples similarly to JFR (even when JFR recording is not active).

Other variants of `recordSample` acquire lock in the beginning because they need a buffer for stack frames. In `external` variants, buffer is not needed, because stack frames are supplied from outside. `CallTraceStorage` is a concurrent lock-free structure; we can benefit from this property and call `put` right away without acquiring locks, thus reducing lock contention. While `recordExternalSample(s)` always succeed in the current implementation (except for JFR recordings), this PR adds potential failures due to lock contention.

The secondary fix does not look right either.
`_total_samples` and `ticks_skipped` are used to calculate average stack walking time, see `Profiler::writeMetrics` and `Profiler::logStats`. We don't walk stack in `recordExternalSamples`, so incrementing counters there would distort the metrics. Also, `recordExternalSamples` is called for live objects, which are already counted by `recordSample`.

PR #1791, opened 2026-08-23
Send profiles directly to an OTel-compatible endpoint

apangin:
### Description

async-profiler can now send profiles directly to an OpenTelemetry-compatible endpoint. When the output file name is an `http://` or `https://` URL, the profile is submitted with an HTTP POST request instead of being written to a local file. The default output format for a remote endpoint is `otlp`.
```
asprof --loop 60s -f http://localhost:4318/v1development/profiles <pid>
```

Together with the `loop` option, this turns async-profiler into an agent for continuous profiling.

The feature works with all profiler interfaces: `asprof`, `-agentpath`, Java API, and C API.
Sending requires `libcurl` to be available. It is loaded dynamically on first use.

### Motivation and context

OTLP-compatible backends (OpenTelemetry Collector, Grafana Pyroscope) accept profiles over HTTP, but until now users had to manage delivery themselves: dump profiles to files and upload them with external tooling. This PR allows continuous profiling out-of-the-box with just one configuration option.

### How has this been tested?

Tested locally with Pyroscope.
Automated tests will be added in a follow-up.

---


src/httpClient.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <algorithm>
+#include <dlfcn.h>
+#include <stdio.h>
+#include <string.h>
+#include "httpClient.h"
+#include "log.h"
+#include "mutex.h"
+#include "writer.h"
+
+constexpr long HTTP_TIMEOUT_MS = 10000;
```
Baraa-Hasheesh, line 15: IMO this should be configurable.

src/httpClient.cpp
```diff
@@ -0,0 +1,166 @@
...
+        default:
+            return "Content-Type: text/plain";
+    }
+}
+
+Error HttpClient::send(const char* url, const char* data, size_t len, Output format) {
+    using namespace Curl;
+
+    if (!initialize()) {
+        Log::info("Failed to send profile to %s: libcurl not found", url);
```
Baraa-Hasheesh: Question

The logs here look to be more of an error rather than an info, why not use `Log::error` here & in other places in this file?

src/httpClient.cpp
```diff
@@ -0,0 +1,166 @@
...
+        return Error("Could not load libcurl");
+    }
+
+    CURL* curl = easy_init();
+    if (curl == nullptr) {
+        Log::info("Failed to send profile to %s: curl_easy_init failed", url);
+        return Error("Failed to initialize libcurl");
+    }
+
+    // Append default OTLP profiles path if URL does not specify a path
```
Baraa-Hasheesh: should this behaviour be added to the documentation?

src/profiler.cpp
```diff
@@ -1665,28 +1671,45 @@ Error Profiler::run(Arguments& args) {
...
+        MutexLocker ml(_state_lock);
 
-    if (args._file != NULL && args._output != OUTPUT_NONE && args._output != OUTPUT_JFR) {
-        FileWriter out(args.file());
-        if (!out.is_open()) {
-            return Error("Could not open output file");
-        }
-        error = dump(out, args);
+        Error error = stop(restart);
         if (error) {
```
Baraa-Hasheesh, line 1681: Unrelated to change but I noticed a bug here related to `args._fdtransfer` 

If the file dump fails the fd server connection is never closed

src/profiler.cpp
```diff
@@ -1665,28 +1671,45 @@ Error Profiler::run(Arguments& args) {
...
+                    return Error("Could not open output file");
+                }
+                error = dump(out, args);
+            }
+            if (error) {
+                return error;
+            }
+        }
+
+        if (restart) {
```
Baraa-Hasheesh, line 1700: unrelated to change, but it might be a good idea to add a validation to prevent loop from running if the user does not provide an output file

-> Profiler is mostly wasting CPU time collecting samples that no one will read

-> It might be a good idea to enforce the file to be either a URL endpoint or a dynamic file (has % syntax such as %n)

src/profiler.cpp
```diff
@@ -1665,28 +1671,45 @@ Error Profiler::run(Arguments& args) {
...
+        }
     }
 
-    if (restart) {
-        args._fdtransfer = false;  // keep the previous connection
-        args._file_num++;
-        return start(args, true);
+    if (http_out.size() > 0) {
+        // Send profiles outside stop-start transaction to minimize the gap between interations
+        return HttpClient::send(file, http_out.buf(), http_out.size(), args._output);
```
Baraa-Hasheesh, line 1712: Question here if the upload fails the profiler will simply keep running in the new session, should we stop the profiler in that case?

The reason I'm raising this is that with this new behaviour we have a divergence in behaviour between the output formats 

For on desk dump if the sample dump fails then the profiler will stop 

For URL dump if the sample dump fails then the profiler will continue 

-> I personally would vote to not stop the profiler, but would also vote to bring the other dump formats to this same alignment, What do you think?

src/httpClient.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+#ifdef __APPLE__
+    static const char* const LIB_NAMES[] = {"libcurl.4.dylib", "libcurl.dylib"};
+#else
+    static const char* const LIB_NAMES[] = {"libcurl.so.4", "libcurl-gnutls.so.4", "libcurl-nss.so.4", "libcurl.so"};
+#endif
+
+    static bool loadLibrary() {
+        for (int i = 0; i < sizeof(LIB_NAMES) / sizeof(LIB_NAMES[0]); i++) {
+            _libcurl = dlopen(LIB_NAMES[i], RTLD_NOW);
```
Baraa-Hasheesh, line 63: I would personally add `RTLD_NODELETE` here

src/httpClient.cpp
apangin: Maybe. I just don't know yet what this configuration may look like. In practice, I think 10 seconds is a meaningful timeout in almost all cases.

src/httpClient.cpp
apangin: `Log::error` is for fatal errors. A failed upload can be a transient error, it does not interrupt continuous profiling.
Whether it should be `warn` or `info` is arguable, but certainly not an `error`. OK, it should be probably `warn` for library loading issues, but `info` for uploading errors (since the latter may not even be a client issue - it may be a problem on the server side).

src/profiler.cpp
Baraa-Hasheesh: ~~Never mind this will never happen~~

apangin: Yeah, good observation. Let's fix this separately. Worth mentioning in #1652 so it's not forgotten.

src/profiler.cpp
apangin: What if it is a profiler stress test? :)
I mean, it's fine, we don't try to guess user's intentions.

src/profiler.cpp
Baraa-Hasheesh: nevermind, I think we should remove the error checks for `dump` call here

As it stands the call will never fails inside the `expire` function 

Having the error check simply makes it harder to do proper tracing of what actually happens 

We only need to check for the `out.is_open` before calling dump

apangin: Current behavior is the right one, I think.
Uploading to a remote endpoint is fundamentally different from writing to a local file, there is no requirement (or expectation) for failures on both paths to have the same effect.

Upload failure can be caused by an issue on a remote server out of operator's control. OTLP specification explicitly requires clients to tolerate temporary connection errors.
File write failure is more likely to be fatal (e.g. out of disk space). And it's the same system where async-profiler runs.

src/profiler.cpp
Baraa-Hasheesh: https://github.com/async-profiler/async-profiler/pull/1652#issuecomment-5458740683

src/httpClient.cpp
apangin: But why? There is no `dlclose` anyway.

Baraa-Hasheesh: I treat it as the intention, 

This library is loaded & never meant to be unloaded

So I would personally add `RTLD_NODELETE` to make it clearer

I'm fine with not doing it as what you mentioned stands 
>no dlclose

src/httpClient.cpp
```diff
@@ -0,0 +1,166 @@
...
+            return "Content-Type: text/html";
+        default:
+            return "Content-Type: text/plain";
+    }
+}
+
+Error HttpClient::send(const char* url, const char* data, size_t len, Output format) {
+    using namespace Curl;
+
+    if (!initialize()) {
```
Baraa-Hasheesh, line 110: I would move the initialize to a different function & call that during the start of the profiler if Arguments::isUrl is true 

-> This would allow us to give an early indication to the user that the upload will fail without the need to wait for the full session/loop to conclude

src/httpClient.cpp
Baraa-Hasheesh: While I agree that 10 seconds is meaningful timeout in almost all cases.
I'm thinking of this from the view point of the current API mode setup.

- start session using `start,event=cpu`
- I stop the session & dump the session using `stop,otlp,file=https://mysite.com/myendpoint`

The thread where I call the stop/dump will potentially hang for 10seconds (Server is having a problem)
-> This in turn can delay the shutdown of the JVM process by those aforementioned 10 seconds, as the JVM will need to wait on that thread in most cases before exiting.


In loop mode the user wouldn't notice this hang/delay as it's happening on the background deamon timer thread.

apangin: > The thread where I call the stop/dump will potentially hang for 10seconds

Yes, it will, but what are alternatives? Even a normal on-disk dump may take several seconds.
This does not pause application threads anyway.

src/httpClient.cpp
apangin: Yeah, that was my first intention. But this comes with some caveats.
`Arguments::isUrl` check is valid only after resolving placeholders, because I may specify something like
`filename=%{OTEL_EXPORTER_OTLP_ENDPOINT}`.
Initializing libcurl _only_ in `start` is not enough, because I may start profiling without a filename, but specify a URL later in `stop` command.
Of course, these aren't big problems, I just did not want to complicate code for an unlikely case when libcurl cannot be loaded. It's cleaner to keep initialization close to the place where libcurl is used.

PR #1795, opened 2026-08-27
Validate jmethodID of compiled frames before storing

### LINKED ISSUE #1794: Dragonwell JDK: SIGSEGV in isStaleMethodId() at JFR dump
[jvmdump_trimmed.log](https://github.com/user-attachments/files/31563466/jvmdump_trimmed.log)
 Attached is the hs_err crash log from the reproduction above, sanitized and trimmed to the crash-relevant parts (crash header, faulting thread, registers, faulting 
   instructions, and the libasyncProfiler mapping). If you need any additional information from the full log, please let me know and I'll provide it.

**Description**

If nmethods are reclaimed (CodeCache sweep) while a profiling session is active, `walkVM` may silently store a garbage jmethodID for a compiled frame. The value is never dereferenced during sampling, so the setjmp/longjmp crash protection never fires. The first dereference happens at dump time — `*(VMMethod**)id` in `isStaleMethodId()` — on the dumping thread, outside the protected region, and the JVM dies with SIGSEGV (SI_KERNEL, #GP) on a non-canonical address.

My first report had no reproduction; I now have a reliable one, plus a core-dump analysis of the exact failure path. Sorry for the quality of the first version.

**Environment**

AJDK 11.0.27.26 (any HotSpot <= 25 where jmethodID is directly dereferenceable), x86-64, official async-profiler 4.5 release, `cstack=vm`.

**Steps to reproduce**

1. `asprof start <pid>`
2. Cause CodeCache reclamation while the session is active. I use a churn script against the target process: repeatedly exclude and re-allow a few hot methods via `jcmd Compiler.directives_add/clear`, invalidating their nmethods and forcing deopt/recompile, plus periodic `GC.run`, whose class unloading triggers a CodeHeap sweep (this is the step that produced the observed usage drop):
Image

3. `asprof stop <pid>` → crash during the dump phase.
Image

The timeline lines up: the crash occurred at 00:46:37, 9 seconds after the sweep window above. The loop (start, wait for a sweep, stop) reproduces reliably; I can share the churn script if useful. I also collected 6 earlier independent production crashes over several months — all 7 crashes stored the identical value `0x8b4c000000e8840f`.

Crash log from `asprof stop` on the official 4.5 (abridged; full file available):

```
SIGSEGV (0xb), si_code: 128 (SI_KERNEL), TRAPNO 0xd (#GP)
RBX=0x8b4c000000e8840f is an unknown value

C  [libasyncProfiler.so+0x39777] FrameName::name(ASGCT_CallFrame&, bool)+0xd87
C  [libasyncProfiler.so+0x41095] Profiler::dumpText(Writer&, Arguments&)+0x5f5
C  [libasyncProfiler.so+0x4d046] Profiler::dump(Writer&, Arguments&)+0x3f6
C  [libasyncProfiler.so+0x4d1be] Profiler::runInternal(Arguments&, Writer&)+0xae
C  [libasyncProfiler.so+0x50410] Agent_OnAttach+0x300        <- asprof stop
V  [libjvm.so] JvmtiExport::load_agent_library
```

**Analysis (from a core dump of one of the production crashes)**

The edge condition is a stale nmethod: the code block has been swept, but the metadata chain the walk dereferences (Method → ConstMethod → ConstantPool → holder → jmethodIDs) is still mapped and readable, because the freed memory has been reused for new code. In `VMMethod::id()`, only the first load is protected by SafeAccess; the following dereferences are guarded by `goodPtr()`, which checks `>= 0x1000` and alignment, but not address validity. The chain completes without faulting, and `return ids[num + 1]` reads a garbage, non-canonical jmethodID from reused code memory. Since no fault occurs, the longjmp protection never gets a chance; the value is stored as-is.

The trace decoded from the sample storage in the core:

```
[... GC frames ...]
[frame type=C1, bci=4, method_id=0x8b4c000000e8840f]   <- dirty frame
[break_not_walkable]
```

The dirty frame precedes `break_not_walkable`: it was written first, then the walk faulted further up and longjmp truncated the trace — but the dirty frame had already been stored (the same value was also found in a second, independent storage buffer).

**Possible fix**

The three compiled-frame store sites could go through `getMethodId()`, as the interpreter path already does:

```cpp
fillFrame(frames[depth++], type, 0, getMethodId(nm->method()));
fillFrame(frames[depth++], type, scope.bci(), getMethodId(scope.method()));
jmethodID method_id = getMethodId(method);   // compiler-task pseudo frame
```

This way `validatedId()`'s self-check (`*(VMMethod**)id == this`) runs at sampling time, inside `walkVM`'s setjmp protection window: a garbage value is either rejected by the alignment/self-reference checks, or faults inside the protected region, where longjmp catches it and the sample is truncated as `break_not_walkable`. The error is then handled within the existing crash protection, instead of being deferred to dump time and becoming a JVM crash.

guyinyou:
Fixes #1794

The compiled-frame and inlined-scope paths in `walkVM` store `VMMethod::id()` without ever dereferencing it. In certain edge conditions `id()` returns a garbage non-canonical value (see the linked issue for the crash evidence: R13 = 8 bytes of machine code, SI_KERNEL/#GP). The value is silently persisted and only dereferenced later by `isStaleMethodId()` at dump time — outside `walkVM`'s setjmp/longjmp crash protection.

This change routes all three sites through `getMethodId()`, which dereferences the id inside the protected region: a garbage id faults at sampling time, is caught by the existing longjmp protection, and the sample is dropped as `break_not_walkable` — nothing invalid ever reaches the call-trace storage. NULL ids are already rendered as "unknown" by `Lookup`, same as the interpreter path today.

Verified: full local build passes with no warnings.

### DIFF
--- a/src/stackWalker.cpp
+++ b/src/stackWalker.cpp
@@ -348,7 +348,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
             } else if (nm->isNMethod()) {
                 int level = nm->level();
                 FrameTypeId type = details && level >= 1 && level <= 3 ? FRAME_C1_COMPILED : FRAME_JIT_COMPILED;
-                fillFrame(frames[depth++], type, 0, nm->method()->id());
+                fillFrame(frames[depth++], type, 0, getMethodId(nm->method()));
 
                 if (nm->isFrameCompleteAt(pc)) {
                     if (depth == 1 && frame.unwindEpilogue(nm, (uintptr_t&)pc, sp, fp)) {
@@ -365,7 +365,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
                                 type = scope_offset > 0 ? FRAME_INLINED :
                                        level >= 1 && level <= 3 ? FRAME_C1_COMPILED : FRAME_JIT_COMPILED;
                             }
-                            fillFrame(frames[depth++], type, scope.bci(), scope.method()->id());
+                            fillFrame(frames[depth++], type, scope.bci(), getMethodId(scope.method()));
                         } while (scope_offset > 0 && depth < max_depth);
                     }
 
@@ -442,7 +442,7 @@ int StackWalker::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth, 
                 } else if (mark == MARK_COMPILER_ENTRY && features.comp_task && vm_thread != NULL) {
                     // Insert current compile task as a pseudo Java frame
                     VMMethod* method = vm_thread->compiledMethod();
-                    jmethodID method_id = method != NULL ? method->id() : NULL;
+                    jmethodID method_id = getMethodId(method);
                     if (method_id != NULL) {
                         fillFrame(frames[depth++], FRAME_JIT_COMPILED, 0, method_id);
                     }


apangin:
I'm afraid this is a wrong solution to a wrong problem.
At best, it only hides subsequences of a different bug.

Let's discuss the related issue first.

PR #1796, opened 2026-08-28
Support JFR event settings in jfrsync

### LINKED ISSUE #1793: Support JFR event settings in --jfrsync
### Describe the feature

`--jfrsync` supports enabling individual JFR events:

```text
--jfrsync +jdk.JavaExceptionThrow
```

However, it cannot configure event settings such as stack traces.

Could it support the standard JFR setting syntax?

```text
--jfrsync '+jdk.JavaExceptionThrow#stackTrace=true'
```

This is related to #750.

### Use Case

Enable exception events with stack traces without creating and distributing a custom `.jfc` file.

### Proposed Solution

Support `#stackTrace=true` for explicitly enabled events. This could correspond to:

```java
recording.enable(eventName).withStackTrace();
```

The existing `+event` syntax would remain unchanged.

Would this fit the intended design of `jfrsync`?

### Acknowledgements

MrChang0:
### Description

Support JFR event settings in explicit `jfrsync` event lists:

```text
--jfrsync '+jdk.JavaExceptionThrow#stackTrace=true'
```

The existing `+event` syntax remains unchanged.

### Related issues

Closes #1793

### Motivation and context

This allows configuring settings such as `stackTrace` and `threshold` without creating and distributing a custom `.jfc` file.

### How has this been tested?

Added an integration test that verifies stack traces are recorded for `jdk.JavaExceptionThrow`. Tested on JDK 11 and JDK 21.

---


docs/ProfilerOptions.md
```diff
@@ -73,7 +73,7 @@ The below options are `action`s for async-profiler and common for both `asprof`
...
+| `--jfrsync CONFIG`   | `jfrsync[=CONFIG]` | Start Java Flight Recording with the given configuration synchronously with the profiler. The output .jfr file will include all regular JFR events, except that execution samples will be obtained from async-profiler. This option implies `-o jfr`.<br>`CONFIG` is a predefined JFR profile, a JFR configuration file (.jfc), or a list of JFR events started with `+`. Event settings may be specified as `+event#setting=value`.<br>Example: `asprof -e cpu --jfrsync '+jdk.JavaExceptionThrow#stackTrace=true' -f combined.jfr 8983`                                                                                                                          |
```
apangin: Does it make sense to enable stackTrace for all custom events by default?

src/helper/one/profiler/JfrSync.java
```diff
@@ -76,6 +76,24 @@ public static void start(String fileName, String settings, int eventMask) throws
...
 
+    private static void enableEvent(Recording recording, String event) throws ParseException {
+        int separator = event.indexOf('#');
+        if (separator < 0) {
+            recording.enable(event);
+            return;
+        }
+
+        int equals = event.indexOf('=', separator + 1);
+        if (separator == 0 || equals <= separator + 1 || equals == event.length() - 1) {
```
apangin: This is too strict. Empty string is a perfectly valid value for certain settings, e.g. `MethodTrace#filter`.
In general, over-defensive programming is considered bad style in this repository. Even JDK standard API accepts
```
rec.enable("jdk.JavaExceptionThrow").with("", "")
```
so why should async-profiler reject it?

test/test/jfr/ExceptionThrow.java
```diff
@@ -0,0 +1,25 @@
...
+    private static void throwException() {
+        try {
+            throw new Exception(MESSAGE);
+        } catch (Exception ignored) {
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        long startTime = System.currentTimeMillis();
+        while (System.currentTimeMillis() - startTime < 2000) {
```
apangin: Why wait for 2 seconds? Does it matter for this test if we throw 1 or 100000 exceptions?

docs/ProfilerOptions.md
MrChang0: Adjusted the description :)

test/test/jfr/ExceptionThrow.java
MrChang0: Agree. Simplified the test to throw only one exception.

src/helper/one/profiler/JfrSync.java
MrChang0: Agreed. I removed the validation and now treat a missing = as an empty value, so +jdk.MethodTrace#filter is accepted. I also added it to the test.

src/helper/one/profiler/JfrSync.java
```diff
@@ -76,6 +76,25 @@ public static void start(String fileName, String settings, int eventMask) throws
...
+    private static void enableEvent(Recording recording, String event) {
+        int separator = event.indexOf('#');
+        if (separator < 0) {
+            recording.enable(event);
+            return;
+        }
+
+        int equals = event.indexOf('=', separator + 1);
+        if (equals < 0) {
+            event += '=';
```
apangin, line 88: This is long and inefficient.
A simpler way to do the same would be something like
```
int eq = event.indexOf('=', separator + 1);
String setting = eq >= 0 ? event.substring(separator + 1, eq) : event.substring(separator + 1);
String value = eq >= 0 ? event.substring(eq + 1) : "";
```

src/helper/one/profiler/JfrSync.java
```diff
@@ -76,6 +76,25 @@ public static void start(String fileName, String settings, int eventMask) throws
         recording.start();
     }
 
+    private static void enableEvent(Recording recording, String event) {
+        int separator = event.indexOf('#');
+        if (separator < 0) {
+            recording.enable(event);
```
apangin, line 82: I mean, why don't enable stackTraces by default for all custom events as they are already enabled in standard JFR profiles? So that users don't have to add `#stackTrace=true` every time.

MrChang0: Agreed. I've prepared a minimal local update using `withStackTrace()` when enabling each event, followed by the requested setting. It deliberately does not preserve `#stackTrace=false` across later entries for the same event; those entries can re-enable stack traces. This keeps the change small without additional state tracking.

apangin: This turned out to be a bad idea. Not all events have a meaningful stack trace (e.g. `jdk.ExceptionStatistics`). Also, if multiple settings were specified and `stackTrace` was set to `false` in the first one, the second setting would reset `stackTrace` back to `true`.

I reverted the change.

apangin:
Fix merged. Thank you for the contribution.

Issue #1786, opened 2026-08-19
JEP 401 value classes not tracked with `alloc` event


galderz:
### Describe the bug

FYI @franz1981

The reproducer (AllocReproducer.java) allocates equal numbers of identity class (`IdentityPoint`) and value class (`ValuePoint`) instances. When profiled with async-profiler's alloc event both v4.5 release and latest source), only identity class allocations appear. Value class allocations are completely missing:

```
$ javac --enable-preview --source 28 AllocReproducer.java

$ java --enable-preview -agentpath:$HOME/opt/async-profiler/lib/libasyncProfiler.dylib=start,event=alloc,file=alloc_profile.jfr AllocReproducer
Profiling started
Starting allocation reproducer …
  Identity class: class AllocReproducer$IdentityPoint
  Value    class: class AllocReproducer$ValuePoint
  Iterations per round: 50000000

Round 0
Round 1
Round 2
Round 3
Round 4

$ jfr print --events jdk.ObjectAllocationInNewTLAB alloc_profile.jfr | grep 'objectClass = ' | sed 's/.*objectClass = //' | sort | uniq -c | sort -rn
7680 AllocReproducer$IdentityPoint (classLoader = null)
```

JVMTI now requires support value objects to be requested separately, otherwise no events are emitted and currently AsyncProfiler does not request it:

```
void JvmtiExport::post_sampled_object_alloc(JavaThread *thread, oop object) {
...
  const bool is_inline = object->is_inline();
  if (is_inline && !JvmtiExport::can_support_value_objects()) {
    return;
  }
```

One way to fix it would be to signal it in `vmEntry.cpp`:

```
diff --git a/src/vmEntry.cpp b/src/vmEntry.cpp
index e8d66f0..5359a9a 100644
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -250,6 +250,14 @@ bool VM::init(JavaVM* vm, bool attach) {
     capabilities.can_tag_objects = 1;
     _jvmti->AddCapabilities(&capabilities);

+    // Request value object support if available (JDK 28+ / Valhalla).
+    // This is needed so that SampledObjectAlloc events are fired for value class instances.
+    // AddCapabilities silently ignores unsupported capabilities on older JDKs.
+    jvmtiCapabilities value_caps = {0};
+    // can_support_value_objects is bit 45 (0-indexed) in jvmtiCapabilities
[...]

PR #1800, opened 2026-09-08
Support tracking allocations of JEP 401 value classes

galderz:
### Description

Adds tracking of allocations for JEP 401 value classes.

### Related issues

It fixes issue #1786

### Motivation and context

Without the fix, no allocations of JEP 401 value classes are being tracked. You need to track this to understand how effective reference scalarization/flatterning is within the context of the given application.

### How has this been tested?

Verified that the reproducer in issue #1786 passes. Run `make test` successfully on darwin/aarch64.

---

### DIFF
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -260,7 +260,7 @@ bool VM::init(JavaVM* vm, bool attach) {
     // this is just ignored and no harm is done.
     jvmtiCapabilities value_caps = {0};
     // can_support_value_objects is bit 45 (0-indexed) in jvmtiCapabilities
-    ((unsigned int*)&value_caps)[1] |= (1u << 13);
+    ((unsigned int*)&value_caps)[1] = (1u << 13);
     _jvmti->AddCapabilities(&value_caps);
 
     jvmtiEventCallbacks callbacks = {0};
--- a/src/vmEntry.cpp
+++ b/src/vmEntry.cpp
@@ -250,6 +250,14 @@ bool VM::init(JavaVM* vm, bool attach) {
     capabilities.can_tag_objects = 1;
     _jvmti->AddCapabilities(&capabilities);
 
+    // Request value object support if available (JDK 28+ / Valhalla).
+    // This is needed so that SampledObjectAlloc events are fired for value class instances.
+    // AddCapabilities silently ignores unsupported capabilities on older JDKs.
+    jvmtiCapabilities value_caps = {0};
+    // can_support_value_objects is bit 45 (0-indexed) in jvmtiCapabilities
+    ((unsigned int*)&value_caps)[1] |= (1u << 13);
+    _jvmti->AddCapabilities(&value_caps);
+
     jvmtiEventCallbacks callbacks = {0};
     callbacks.VMInit = VMInit;
     callbacks.VMDeath = VMDeath;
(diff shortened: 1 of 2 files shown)


src/vmEntry.cpp
```diff
@@ -250,6 +250,14 @@ bool VM::init(JavaVM* vm, bool attach) {
     capabilities.can_tag_objects = 1;
     _jvmti->AddCapabilities(&capabilities);
 
+    // Request value object support if available (JDK 28+ / Valhalla).
+    // This is needed so that SampledObjectAlloc events are fired for value class instances.
+    // AddCapabilities silently ignores unsupported capabilities on older JDKs.
```
apangin: Are you sure? If this were true, you could add this bit in a single call to `AddCapabilities` above.

src/vmEntry.cpp
```diff
@@ -250,6 +250,14 @@ bool VM::init(JavaVM* vm, bool attach) {
...
+    jvmtiCapabilities value_caps = {0};
+    // can_support_value_objects is bit 45 (0-indexed) in jvmtiCapabilities
+    ((unsigned int*)&value_caps)[1] |= (1u << 13);
```
apangin: nit: `|=` is redundant as the original value is known to be `0`.

src/vmEntry.cpp
```diff
@@ -250,6 +250,14 @@ bool VM::init(JavaVM* vm, bool attach) {
     capabilities.can_tag_objects = 1;
     _jvmti->AddCapabilities(&capabilities);
 
+    // Request value object support if available (JDK 28+ / Valhalla).
+    // This is needed so that SampledObjectAlloc events are fired for value class instances.
```
apangin, line 254: Note that the object reference passed to the callback will be `NULL` for value objects.
Async-profiler uses these references for live object tracing. Even though this should work correctly (`NewWeakGlobalRef` accepts NULLs), I would suggest making the intention clear and do not call `live_refs.add` at all when the object is NULL.

src/vmEntry.cpp
galderz: `AddCapabilities` returns `JVMTI_ERROR_NOT_AVAILABLE` on older versions when this value is set.

The reason to keep the call that sets `AddCapabilities` for value classes separate is that any errors thrown there don't affect the rest of capabilities set in `_jvmti->AddCapabilities(&capabilities);` above. If the JDK supports value classes, the capability is set, otherwise nothing happens.

So, the reason to call `AddCapabilities` a second time is deliberate. The value returned by `AddCapabilities` is not used so there is not harm done because the capabilities set in JVMTI are additive.

Can you suggest other ways to do this?

src/vmEntry.cpp
galderz: Applied in update

src/vmEntry.cpp
galderz: Applied in update

src/vmEntry.cpp
galderz: I've changed the comment to make the intention more explicit.

apangin: Sorry, I meant the phrase "AddCapabilities silently ignores unsupported capabilities on older JDKs" was misleading; you could just remove it. No need for half-page comments explaining trivial things.

apangin: Never mind, I'll amend it myself.

apangin:

galderz:
Thanks @apangin for the review!

PR #1805, opened 2026-09-14
Support regex matching for JFR span tag filters

### LINKED ISSUE #1804: Support regex span tag filtering in JFR converter
### Describe the feature

In [Apache Spark PR #58770](https://github.com/apache/spark/pull/58770), I proposed adding task attempt spans to executor profiling output. Each span has a composite tag containing task metadata, for example:

```text
SparkTask:stageId=0,stageAttemptNumber=0,partitionId=0,taskAttemptId=0,attemptNumber=0,status=succeeded
```

I would like to filter profiling samples by `taskAttemptId` alone. Currently, `jfrconv --tag` requires an exact match of the entire tag, so I also need to know the stage, partition, attempt number, and completion status.

We can add support for Java regular expressions to `--tag`, using whole-tag matching consistent with `--include` and `--exclude`. 

### Use Case

This would allow selecting a task attempt with:

```bash
jfrconv --wall \
  --tag 'SparkTask:.*,taskAttemptId=123,.*' \
  profile-exec-1.jfr task-123.html
```

Or selecting multiple task attempts:

```bash
jfrconv --wall \
  --tag 'SparkTask:.*,taskAttemptId=(123|456),.*' \
  profile-exec-1.jfr selected-tasks.html
```

### Proposed Solution

_No response_

### Acknowledgements

wForget:
### Description

Change `--tag` to accept Java regular expressions with whole-tag matching, consistent with `--include` and `--exclude`. Update help text and documentation with matching semantics and literal escaping examples.

### Related issues

Closes #1804

### Motivation and context

One use case for spans is to associate samples with a user-provided context. This context may contain multiple pieces of information, and we may want to filter samples using only part of it. Adding regular expression support to `--tag` makes this possible.

For example, in [Apache Spark PR #58770](https://github.com/apache/spark/pull/58770), I proposed recording task attempt spans with tags containing stage, partition, task attempt, and completion information. Regex matching allows selecting samples by `taskAttemptId` without specifying the entire tag:

```bash
jfrconv --wall \
  --tag 'SparkTask:.*,taskAttemptId=123,.*' \
  profile-exec-1.jfr task-123.html
```

### How has this been tested?

Added JFR converter tests.


docs/ConverterUsage.md
```diff
@@ -102,6 +103,21 @@ See the [profiler options documentation](ProfilerOptions.md#options-applicable-t
 `jfrconv` utility is provided in `bin` directory of the async-profiler package.
 It requires JRE to be installed on the system.
 
+### Filter by span tag
```
apangin: This addition begins the "jfrconv examples" section. However, filtering by span tags is an advanced use case, so keep the basic examples first, and move your change to the end of the section.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +82,51 @@ public void visit(Event event, long samples, long value) {
...
+    private static void checkTagFilter(String file, Arguments args, String... expectedTags) throws IOException {
+        try (JfrReader jfr = new JfrReader(file)) {
+            int[] checked = new int[1];
+            JfrConverter converter = new JfrConverter(jfr, args) {
+                @Override
+                protected void collectEvents(Filter filter) throws IOException {
+                    for (SpanEvent span; (span = jfr.readEvent(SpanEvent.class)) != null; ) {
+                        if (span.duration == 0) continue;
+                        long time = jfr.eventTimeToNanos(span.time + span.duration / 2);
+                        boolean expected = Arrays.asList(expectedTags).contains(span.tag());
```
apangin: I understand this is not a performance-critical piece of code, but in general, creating a new list object and doing linear search on each iteration hurt my eyes. This could be a HashSet created before the loop.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +82,51 @@ public void visit(Event event, long samples, long value) {
...
+        String file = p.getFilePath("%f");
+        checkTagFilter(file, new Arguments("--tag", "busyRequest"), "busyRequest");
+        checkTagFilter(file, new Arguments("--tag", "busy.*"), "busyRequest", "busyOptional");
+        checkTagFilter(file, new Arguments("--tag", "busyRequest|idleNormal"), "busyRequest", "idleNormal");
+        checkTagFilter(file, new Arguments("--tag", "\\QbusyRequest\\E"), "busyRequest");
+        checkTagFilter(file, new Arguments("--tag", "busy"));
+        checkTagFilter(file, new Arguments("--tag", "missing"));
+        checkTagFilter(file, new Arguments("--tag", ".*"),
+                "busyRequest", "idleRequest", "busyOptional", "idleOptional", "idleNormal");
+        checkTagFilter(file, new Arguments("--tag", "busy.*", "--latency", "200"), "busyRequest");
```
apangin: What is this code supposed to test? Regular expression syntax? JCK already covers that.

async-profiler tests are more integration tests than unit tests. Choose a couple of scenarios (no more than that) and test them end-to-end.

apangin (commented):
Thank you for the contribution. The change itself is useful, but the docs and the tests need some edits.

wForget:
> Thank you for the contribution. The change itself is useful, but the docs and the tests need some edits.

@apangin Thanks for your review. I updated documentation and rewrote the `tagFilter` test case by referencing the `latencyFilter` test case.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +79,53 @@ public void visit(Event event, long samples, long value) {
         }
     }
 
+    @Test(mainClass = Tagger.class, agentArgs = "start,jfr,wall,file=%f", runIsolated = true)
```
apangin: Why `runIsolated`? What makes the test sensitive to background noise?

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +79,53 @@ public void visit(Event event, long samples, long value) {
...
+        checkTagFilter(file, new Arguments("--tag", "showcase.*"), 0, 1, 2);
+        checkTagFilter(file, new Arguments("--tag", "missing"));
+    }
+
+    private static void checkTagFilter(String file, Arguments args, int... expectedMethods) throws IOException {
+        try (JfrReader jfr = new JfrReader(file)) {
+            boolean[] found = new boolean[3];
+            JfrConverter converter = new JfrConverter(jfr, args) {
+                @Override
+                protected void convertChunk() {
```
apangin: This breaks encapsulation and complicates the code in general.

The purpose of the test is to verify that the `--tag` filter works correctly, right? So why not just convert to .jfr to .collapsed and check presence of strings like other tests do?

test/test/jfrconverter/JfrconverterTests.java
wForget: `runIsolated` is unnecessary, removed

test/test/jfrconverter/JfrconverterTests.java
wForget: > The purpose of the test is to verify that the `--tag` filter works correctly, right? So why not just convert to .jfr to .collapsed and check presence of strings like other tests do?

Thanks for the guidance. I've updated the test case along those lines.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +80,34 @@ public void visit(Event event, long samples, long value) {
...
+            JfrToFlame converter = new JfrToFlame(jfr, args);
+            converter.convert();
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            converter.dump(out);
+            String text = out.toString("UTF-8");
+            if (expectedMethods.length == 0) {
+                assert text.isEmpty() : "Expected no output, but got: " + text;
+            } else {
+                for (String method : expectedMethods) {
+                    assert text.contains(method) : "Expected method " + method + " to be present in the output";
```
apangin: This only checks presence of the expected methods, but not absence of others.

test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -79,6 +80,34 @@ public void visit(Event event, long samples, long value) {
...
+        checkTagFilter(file, new Arguments("--output", "collapsed", "--tag", "missing"));
+    }
+
+    private static void checkTagFilter(String file, Arguments args, String... expectedMethods) throws IOException {
+        try (JfrReader jfr = new JfrReader(file)) {
+            JfrToFlame converter = new JfrToFlame(jfr, args);
+            converter.convert();
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            converter.dump(out);
+            String text = out.toString("UTF-8");
```
apangin: This entire method is redundant; conversion can be done with a one-liner:
```
Output out = Output.convertJfrToCollapsed(file, "--tag", tag);
```

test/test/jfrconverter/JfrconverterTests.java
wForget: Thanks, I’ve updated it to use exact matching.

apangin:
OK, I simplified the test and merged the change. Thank you for the contribution.

Issue #1803, opened 2026-09-14
Allow selective deletion of specific frames/information when using jfrconv --diff


Baraa-Hasheesh:
### Describe the feature

Recently I was using the async-profiler to try to find the execution overhead difference between two different versions of some application using the `--diff` option 

However in my use case I was recording flamegraphs with CPU & Thread information, which caused a lot of pain when checking the differences (different thread names/IDs & different CPU cores) 

Here I'm proposing adding an option to simply allow me to remove some frames when dumping the stacks, for example the following stack `[CPU-1][thread1 tid={threadID}];method1;method2` would become `method1;method2`

This would make it more straight forward when comparing the differences using the existing `--diff` option

### Use Case

Allow user to ignore CPU & Thread ID information when doing a `--diff`

### Proposed Solution

**Approach#1:** a simple `--xframe {pattern}` option that simply removes the frames from the stacks but keep the stacks themselves 

**Apporach#2:** Extend `--norm` to effect that information 
- For example don't include thread id when `--norm` is selected

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

apangin:
For the described use-case of stripping CPU and Thread frames, there is already a working option: `--skip 2`.
E.g. `jfrconv --diff --skip 2 new.jfr old.jfr` will omit two bottom frames when building a flamegraph.

I just noticed that `--skip` option works for JFR-to-flame or collapsed-to-flame conversion, but not for `flame-to-flame` conversion, unfortunately, but this should be easy to fix.

apangin:
Alright, the fix is trivial indeed: #1806

Baraa-Hasheesh:
The provided solution should be good enough for my use case, closing this ticket

PR #1806, opened 2026-09-14
#1803: jfrconv --skip option does not work for flame-to-flame conversion

apangin:
### Description

`jfrconv --skip` option does not work when the source file is an HTML FlameGraph.
This PR fixes the issue by forcing FlameGraph rebuild through the `FlameGraph.addSample` code path - this is where the `skip` option is checked.

### Related issues

#1803

### How has this been tested?

Recorded two .html profiles with `threads,record-cpu` options
and ran `jfrconv --diff --skip 2 old.html new.html`.

---


Baraa-Hasheesh (approved):
LGTM

Issue #1807, opened 2026-09-17
SIGSEGV crash in Dictionary::lookup when profiling with Pyroscope


dongwang123:
Hello：
 
We are experiencing intermittent JVM crashes (SIGSEGV) on a small number of container instances when using async-profiler via Pyroscope. The crash occurs in the native `Dictionary::lookup` function.
  Most containers run fine — only a few pods are affected, which suggests a possible race condition or memory corruption issue.

How Pyroscope is started:
We enable **CPU, ALLOC, and LOCK** profiling simultaneously via the Pyroscope Java agent:
  java -javaagent:pyroscope.jar
    -Dpyroscope.application.name=***
    -Dpyroscope.server.address=***
    -Dpyroscope.profiler.event=CPU
    -Dpyroscope.profiler.alloc=4096k
    -Dpyroscope.profiler.lock=20ms
    -jar app.jar

Environment:
**OS**: Linux (amd64), containerized (Kubernetes)
  - **JRE**: Java(TM) SE Runtime Environment 17.0.10+11-LTS-240
  - **JVM**: Java HotSpot(TM) 64-Bit Server VM (mixed mode, sharing, tiered, compressed oops, compressed class ptrs, G1 GC)
  - **async-profiler version**: 4.5.0.0
  - **Pyroscope version**: 2.9.1
  - **Container base image**: <e.g., oracle/graalvm-ce:17, eclipse-temurin:17, etc.>

Crash details:
A fatal error has been detected by the Java Runtime Environment:
SIGSEGV (0xb) at pc=0x00007f3bf4e08965, pid=1, tid=596
JRE version: Java(TM) SE Runtime Environment (17.0.10+11) (build 17.0.10+11-LTS-240)
Java VM: Java HotSpot(TM) 64-Bit Server VM (17.0.10+11-LTS-240, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
Problematic frame:
C  [libasyncProfiler-linux-x64-94af16d95b27e7e3346a51e8d43e744de92a62a3.so+0xd965]  Dictionary::lookup(char const*, unsigned long)+0x65
An error report file with more information is saved as:

If you would like to submit a bug report, please visit:
 https://bugreport.java.com/bugreport/crash.jsp
The crash happened outside the Java Virtual Machine in native code.
See problematic frame for where to report the bug.

[...]

PR #1809, opened 2026-09-20
Obtain class IDs under the sample lock

apangin:
### Description

Fixes use-after-free in `Dictionary::lookup` that happens when a profiling session is restarted while lock or allocation events are still being recorded.

This PR moves class name resolution under `Profiler::_locks` so that it cannot coincide with `_class_map.clear()` in `Profiler::start`.

### Related issues

#1807

### How has this been tested?

A stress test that reproduces the crash: 32 threads taking ReentrantLocks and allocating instances of 20000 classes, while the main thread restarts the profiler every 50 ms with `start,event=cpu,alloc=1,lock=1,jfr`.

---


src/lockTracer.cpp
```diff
@@ -215,9 +215,12 @@ void JNICALL LockTracer::UnsafeParkHook(JNIEnv* env, jobject instance, jboolean
         _orig_unsafe_park(env, instance, isAbsolute, time);
         u64 park_end_time = TSC::ticks();
 
-        const u64 duration = park_end_time - park_start_time;
-        if (updateCounter(_total_duration, duration, _interval)) {
-            recordContendedLock(PARK_SAMPLE, park_start_time, park_end_time, lock_name, park_blocker, time);
+        // Profiling session might have been stopped or restarted while the thread was parked
+        if (_enabled && park_start_time >= _start_time) {
```
Baraa-Hasheesh: Wouldn't this still potentially race?

`_enabled` is set to `true` as part of `installTraps` which happens way earlier than the set of `_start_time`

Baraa-Hasheesh: also as a note `_start_time` is a normal static value, we should probably change it to `volatile` or maybe use atomics?

src/event.h
```diff
@@ -36,9 +36,21 @@ class Event {
...
+
   public:
-    u32 _class_id;
+    EventWithClass() : _class_name(nullptr) {}
+
+    void setClassName(const char* class_name, u32 len);
+    void setClassSignature(const char* class_sig);
+
+    // Must be called under the sample lock (Profiler::_locks).
+    // May allocate; do not call in asynchronous signal handlers.
```
Baraa-Hasheesh, line 51: This is called for `recordAllocationOutsideTLAB` &  `recordAllocationInNewTLAB` which can happen in asynchronous signal handlers (`allocaTracer`)

Wouldn't that cause an issue?

apangin: AllocTracer's handlers are _synchronous_ - they are only called at the specific locations when a thread is known not to be inside `malloc`.

src/lockTracer.cpp
apangin: `volatile` or atomics are redundant: compiler can't cache values across callbacks, and there is no happens-before requirement here. Opaque read is enough.

There is no race with respect to where `_enabled` is set, but I noticed a different bug: since `park` traps are no longer uninstalled, lock samples will continue being reported during the next profiling session started without `lock`. I fixed that by resetting `_start_time` in `LockTracer::stop`.

Baraa-Hasheesh (approved):
LGTM

benty-amzn (approved):

PR #1810, opened 2026-09-21
Simplified `latencyFilter` test and made `tagFilter` test more reliable

apangin:
### Description

Simplified `latencyFilter` test and made `tagFilter` test more reliable.

### Motivation and context

Recently added `tagFilter` failed sometimes, because it assumed only 3 stack traces in the output file. That was too strong assumtion; there may be other valid stack traces when a thread is inside the JVM runtime, e.g.
```
StringTable::do_intern
ConstantPool::string_at_impl
ConstantPool::resolve_constant_at_impl
Bytecode_loadconstant::resolve_constant
InterpreterRuntime::resolve_ldc
test/jfrconverter/Tagger.main
```
Additionally, I refactored `latencyFilter` test using the same pattern as in `tagFilter`.

### How has this been tested?

`tagFilter` and `latencyFilter` no longer fail locally.

---


test/test/jfrconverter/JfrconverterTests.java
```diff
@@ -25,72 +21,44 @@ public class JfrconverterTests {
 
     @Test(mainClass = CpuBurner.class, agentArgs = "start,jfr,all,file=%f")
     public void heatmapConversion(TestProcess p) throws Exception {
-        Output out = p.waitForExit("%f");
+        p.waitForExit("%f");
```
Baraa-Hasheesh: as we're ignoring the return value here, why not use the `p.waitForExit()` here & other places

apangin: You are right; I missed that.
Updated.

Baraa-Hasheesh (approved):
LGTM

